PR #16, opened 2018-04-11
add debian support to get-deps

necrobuffalo:
Dependency names on Ubuntu and Debian are identical (only tested on stretch.)


wez:
Sorry, I got swamped and totally missed this PR!  Thanks!

PR #21, opened 2019-02-23
Initial xkbcommon support

### LINKED ISSUE #18: Not being able to insert special characters
## Is it a build problem?
No, build happens happens happily!
I use `--release`

## Did something not work the way you expected?
Yes, I cannot type any of this characters:
```
~ ç Ç á è Ï õ Ũ ñ
```
As well as anything that uses "Alt Gr" mod, as `/?`(I'm on a notebook) `¹²³£¢¬{[}§ªº€`, also, nothing on numpad works `/*--+789456123`.
Tabs hotkey (Super-T, Super-[, Super-]) don't works either....

### What did you try?
Do what I always do, from bash, fish, tmux, vim...
Can't even `cd ~/Projects`

Trace:
https://drive.google.com/open?id=1rQVuDGW1Gjf-KNWJkQFzXAQ7yAQ8YIUJ

### What did you expect to happen?
The same behavious as anywhere

### What actually happened?
Any other thing I typed worked, It's way faster than konsole, ligatures, colors and text effects (bold) are fine too.

### Others note:
I use i3, on archlinux, with latest NVIDIA drivers and libglvnd. Only setting is `font = { fontconfig_pattern = "Fira Code" }`

There is also this error, visible with RUST_BACKTRACE:
https://gist.github.com/PedroHLC/0e8b46a77f6925136842110bab4cdded

Lompik:
This adds basic support for X11's dead keys & compositions unicode input. Most keys are passed to ```key_down``` via their char representation.

What doesn't work yet: 
   - multi codepoint unicode such as ```J́``` . Around 257 of those are reachable via compose.
   - remove consumed modifiers 

Fixes #18


src/xwindows/keyboard.rs
```diff
@@ -0,0 +1,590 @@
...
+                0,
+                events as u16,
+                map_parts as u16,
+                map_parts as u16,
+                None,
+            );
+
+            cookie
+                .request_check()
+                .expect("failed to select notify events from xcb xkb");
```
wez: `?` operator here too

src/xwindows/keyboard.rs
```diff
@@ -0,0 +1,590 @@
...
+    }
+
+    pub fn process_key_event(
+        &self,
+        xcb_ev: &xcb::KeyPressEvent,
+    ) -> Option<(KeyCode, KeyModifiers)> {
+        let xcode = xcb_ev.detail() as xkb::Keycode;
+        let xsym = self.state.borrow().key_get_one_sym(xcode);
+        let pressed = (xcb_ev.response_type() & !0x80) == xcb::KEY_PRESS;
+        if pressed {
```
wez: I tend to prefer early-return over `if-then-else` as it results in fewer things for human brains to parse:

```
if !pressed {
   return None;
}
self.compose_state.borrow_mut().feed(xsym);
```

src/xwindows/keyboard.rs
```diff
@@ -0,0 +1,590 @@
...
+        }
+        if self.mod_is_active(xkb::MOD_NAME_ALT) {
+            // Mod1
+            res |= KeyModifiers::ALT;
+        }
+        if self.mod_is_active(xkb::MOD_NAME_LOGO) {
+            // Mod4
+            res |= KeyModifiers::SUPER;
+        }
+        if self.mod_is_active("Mod3") {
```
wez: Just curious: is this something that we'd want to make configurable in the future? (doesn't need to be part of this PR)

src/xwindows/keyboard.rs
```diff
@@ -0,0 +1,590 @@
...
+        self.keymap.replace(new_keymap);
+        eprintln!("updated keymap");
+    }
+
+    pub fn get_device_id(&self) -> i32 {
+        self.device_id
+    }
+}
+
+#[cfg_attr(rustfmt, rustfmt_skip)]
```
wez: would it be the end of the world to let `rustfmt` run on this?
I value greatly the fact that a code formatter will mercilessly apply styling to code so that humans don't need to nitpick over the details in code reviews, even if it doesn't quite format to the taste of any one person :-)

src/xwindows/keyboard.rs
```diff
@@ -0,0 +1,590 @@
...
+            xkb::KEY_Tab =>                      KeyCode::Tab,
+          //xkb::KEY_ISO_Left_Tab =>             KeyCode::LeftTab,
+            xkb::KEY_BackSpace =>                KeyCode::Backspace,
+            xkb::KEY_Return =>                   KeyCode::Char(0xdu8 as char),
+            xkb::KEY_Insert =>                   KeyCode::Insert,
+            xkb::KEY_Delete =>                   KeyCode::Delete,
+            xkb::KEY_Clear =>                    KeyCode::Delete,
+            xkb::KEY_Pause =>                    KeyCode::Pause,
+            xkb::KEY_Print =>                    KeyCode::Print,
+          // 0x1005FF60 =>                       KeyCode::SysRq,         // hardcoded Sun SysReq
```
wez: There are a lot of commented sections in here.  Could you add a comment on the top of the function that explains why they are here?  For instance, is it something we might want to uncomment in the future, and if so, what we'd need to do first.  Alternatively, if we wouldn't want to uncomment them, it would be good to explain that also.   A third option is to delete the commented lines.

src/xwindows/keyboard.rs
```diff
@@ -0,0 +1,590 @@
...
+
+struct XkbGenericEvent {
+    base: xcb::Event<xcb_xkb_generic_event_t>,
+}
+
+impl XkbGenericEvent {
+    pub fn response_type(&self) -> u8 {
+        unsafe { (*self.base.ptr).response_type }
+    }
+    #[allow(non_snake_case)]
```
wez: I think you can remove this attribute and the one on `device_id`

src/guiloop/x11.rs
```diff
@@ -440,6 +440,11 @@ impl GuiEventLoop {
     fn process_xcb_event(&self, event: &xcb::GenericEvent) -> Result<(), Error> {
         if let Some(window_id) = Self::window_id_from_event(event) {
             self.process_window_event(window_id, event)?;
+        } else { // [FIXME] can't identify window from xkb keyboard event
```
wez: travis will complain about this not being rust formatted; make sure that you `cargo fmt` so that you get a green build status

src/xwindows/keyboard.rs
```diff
@@ -0,0 +1,590 @@
...
+         // xkb::KEY_XF86LaunchA =>              KeyCode::LaunchA,
+         // xkb::KEY_XF86LaunchB =>              KeyCode::LaunchB,
+         // xkb::KEY_XF86LaunchC =>              KeyCode::LaunchC,
+         // xkb::KEY_XF86LaunchD =>              KeyCode::LaunchD,
+         // xkb::KEY_XF86LaunchE =>              KeyCode::LaunchE,
+         // xkb::KEY_XF86LaunchF =>              KeyCode::LaunchF,
+            _ => {return None;}
+        };
+    Some(res)
+}
```
wez: this is a really long function (especially with all the commented out bits), so it would be good to move it into its own file (perhaps back into `xkeysyms.rs`?)

Lompik:
this need ```xkbcommon >= 0.5.0``` release 5years ago. It seems that trusty has ```0.4.1```

src/guiloop/x11.rs
```diff
@@ -440,6 +440,11 @@ impl GuiEventLoop {
...
+            let mut windows = self.windows.borrow_mut();
+            for w in windows.by_id.values_mut() {
+                w.dispatch_event(event)?;
```
wez: I think this is perhaps the only potential technical issue in this PR.
Is there a possibility that we'd dispatch the same events to multiple windows?
In a quick test with opening two windows at once, my normal non-IME text seems to do the right thing, so maybe it's ok and it just feels slightly bad because it is `O(number-of-windows)`?

wez:
> this need `xkbcommon >= 0.5.0` release 5years ago. It seems that trusty has `0.4.1`

You could add:

```
dist: xenial
```

to the travis config; I'm fine with raising that requirement

src/xwindows/keyboard.rs
Lompik: yes but keep in mind that it's already configurable in X11 (I rebind caps lock to ```Control``` and ```left_control``` to ```hyper/mod4```).
this is how I see this would work:
  - add a ```HYPER``` variant in ```KeyModifiers```.  
  - keep 1:1 X11 modifiers map to ```KeyModifiers```
  - Allow users to configure based on ```KeyModifiers```

Lompik:
probably just need to force rebuild ```harfbuzz-sys```

wez:
Thanks!
The build error is weird.  Try removing the `cache: cargo` line from the travis config and see if that helps to blow the cache

Lompik:
You might need to clear cache manually,

https://docs.travis-ci.com/user/caching/#clearing-caches

I tried to remove harfbuzz building in 30e2de9 while moving to xenial so rust ```harfbuzz``` is probably built with system's HB. sorry about that.

wez:
> You might need to clear cache manually, Builds without cache passes.
> 
> https://docs.travis-ci.com/user/caching/#clearing-caches

I cleared the caches and restarted the build

> I tried to remove harfbuzz building in [30e2de9](https://github.com/wez/wezterm/commit/30e2de994e602a7a99644a054353691cd87a1fc2) while moving to xenial so rust `harfbuzz` is probably built with system's HB. sorry about that.

If xenial has a new enough harfbuzz, I’m also fine with using that over building the bundled version; I’d love to unbundle that code!  I can sort that out after we get a green build here

Lompik:
> If xenial has a new enough harfbuzz, I’m also fine with using that over building the bundled version; I’d love to unbundle that code! I can sort that out after we get a green build here

This has been tried here https://travis-ci.org/wez/wezterm/builds/497429630 . HB in xenial is not new enough.

src/xwindows/keyboard.rs
wez: this functions are snake_case already, so you should be able to remove the attributes :)

wez:
> > If xenial has a new enough harfbuzz, I’m also fine with using that over building the bundled version; I’d love to unbundle that code! I can sort that out after we get a green build here
> 
> This has been tried here https://travis-ci.org/wez/wezterm/builds/497429630 . HB in xenial is not new enough, we need >= `1.0.5` , but xenial has `1.0.1` (release in 2015). Xenial was release by Ubuntu in Apr 2016 & travis Oct 2018 !
> Maybe try a ppa ? https://docs.travis-ci.com/user/reference/xenial/#third-party-apt-repositories-removed

Since I've already got a [fork of harfbuzz-sys for some mac stuff](https://github.com/servo/rust-harfbuzz/pull/128), I think I'll add a feature flag to turn on freetype support and then we can just leave it to that crate to build a new enough version.

wez:
OK, we have a green build, thanks!

In the description at the top of this PR there are a couple of TODO's; I think you may have resolved one of them, so can you review and update those based on the current state of this PR?

wez:
Rebased and pushed; many thanks!

PR #22, opened 2019-02-23
Test to see whether we need our bundled harfbuzz library any more

wez:


wez:
We do, as the one built by the crate disables freetype support

PR #24, opened 2019-03-02
Complete Control key interpretation

Lompik:
example key currently lacking:
  - ```Ctrl+[``` sends escape
  - ```Ctrl+Space``` /```Ctrl+2``` sends null

As of now, both xkbcommon/glutin do not interpret those (they could).

### DIFF
--- a/term/src/terminalstate.rs
+++ b/term/src/terminalstate.rs
@@ -765,23 +765,31 @@ impl TerminalState {
             (Insert, _, _, SHIFT, _) => paste!(),
             (Char('v'), ..) if mods == KeyModifiers::SUPER => paste!(),
 
-            (Char(c), CTRL, _, SHIFT, _) if c <= 0xff as char && c > 0x40 as char => {
-                // If shift is held we have C == 0x43 and want to translate
-                // that into 0x03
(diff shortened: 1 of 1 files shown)


wez:
Thanks for continuing to contribute!
I think this X11 specific translation should happen in the xwindows layer because I'm concerned that this change might break translation in the glutin flavors of things.
Also: I've been thinking a little bit about making a user-configurable key binding configuration and processing layer that will need to sit at the layer above terminalstate for reasons alluded to in the changes I've recently pushed regarding the `mux` component.

I'd like your input on that but need a little bit more time to form some ideas; I'll open an issue with those thoughts when I have some time to write them down so that you'll have an opportunity to weigh in before code gets written for that.

In the meantime though, would you mind moving these translations up to `xwin.rs`?
Hopefully we can preserve the mappings as-is in terminalstate; ideally we won't unintentionally change the behavior of the other platforms/glutin with this PR.

Lompik:
My intent here is not to push X11 specific stuff everywhere. I do believe those keys needs translation regardless of the platform. I reckon I don't have a windows/macos to test this out though. 
I'll close this for now, we can revisit later as needed.

PR #26, opened 2019-05-05
Load a basic menu on macOS

fanzeyi:
This will allow copying texts and system shortcuts on macOS. (~~Pasting text doesn't work right now~~ nvm it's some bug in oh-my-zsh.) 

For example, ⌘W will be able to close current terminal window, ﻿﻿⌘M will be able to hide the current window.

image

image

### DIFF
--- a/assets/macos/WezTerm.app/Contents/Resources/Base.lproj/menu.xib
+++ b/assets/macos/WezTerm.app/Contents/Resources/Base.lproj/menu.xib
@@ -1,0 +1,628 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<document type="com.apple.InterfaceBuilder3.Cocoa.XIB" version="3.0" toolsVersion="14490.70" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES" customObjectInstantitationMethod="direct">
+    <dependencies>
+        <plugIn identifier="com.apple.InterfaceBuilder.CocoaPlugin" version="14490.70"/>
+    </dependencies>
+    <objects>
+        <customObject id="-2" userLabel="File's Owner" customClass="NSApplication"/>
+        <customObject id="-1" userLabel="First Responder" customClass="FirstResponder"/>
+        <customObject id="-3" userLabel="Application" customClass="NSObject"/>
+        <customObject id="YLy-65-1bz" customClass="NSFontManager"/>
+        <menu title="Main Menu" systemMenu="main" id="AYu-sK-qS6">
+            <items>
+                <menuItem title="WezTerm" id="1Xt-HY-uBw">
+                    <modifierMask key="keyEquivalentModifierMask"/>
+                    <menu key="submenu" title="WezTerm" systemMenu="apple" id="uQy-DD-JDr">
+                        <items>
+                            <menuItem title="About WezTerm" id="5kV-Vb-QxS">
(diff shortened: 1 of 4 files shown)


fanzeyi:
Looking at the giant xib file now I'm thinking maybe I should create the menus programmatically.. Although that would mean translating a few hundred lines of [Swift](https://gist.github.com/fanzeyi/b7e339695856d5258b421142bf96209b) into Rust calling Objective-C. I'm not sure which solution is better now :/

install.sh
```diff
@@ -3,12 +3,17 @@
...
-      cp target/release/wezterm $APP && \
-        echo "Installed to $APP"
+
+    # compile xib
+    BASE_LPROJ="$APP/Contents/Resources/Base.lproj/"
+
+    rm -rf "$APP"
+    cp -r assets/macos/WezTerm.app "$APP" && \
+      cp target/release/wezterm "$APP" && \
+        ibtool --compile "$BASE_LPROJ/menu.nib" "$BASE_LPROJ/menu.xib" && \
```
wez: I had to run `sudo xcode-select -s /Applications/Xcode_10.2.app` to get ibtool to run; it doesn't run with the CLI only flavor of the developer tools.  I think that is probably fine, but its worth checking that it will run in the `ci/deploy.sh` script which packages up the .zip file that is produced when I push a tag.

I think we need to add this same step to that script, and should probably change the `.travis.yml` to always run `ci/deploy.sh` in the `script` section, rather than in the `before_deploy` section so that it gets tested in PRs too.

assets/macos/WezTerm.app/Contents/Resources/Base.lproj/menu.xib
```diff
@@ -0,0 +1,628 @@
...
+                                                <action selector="stopSpeaking:" target="-1" id="dX8-6p-jy9"/>
+                                            </connections>
+                                        </menuItem>
+                                    </items>
+                                </menu>
+                            </menuItem>
+                        </items>
+                    </menu>
+                </menuItem>
+                <menuItem title="Format" id="jxT-CU-nIS">
```
wez: I think this menuItem should be removed; it's not applicable to the terminal and it blocks some of the existing hotkeys eg: CMD-T should create a new tab, but instead pops up a font dialog that isn't connected to anything

assets/macos/WezTerm.app/Contents/Resources/Base.lproj/menu.xib
```diff
@@ -0,0 +1,628 @@
...
+                            <menuItem isSeparatorItem="YES" id="m54-Is-iLE"/>
+                            <menuItem title="Close" keyEquivalent="w" id="DVo-aG-piG">
+                                <connections>
+                                    <action selector="performClose:" target="-1" id="HmO-Ls-i7Q"/>
+                                </connections>
+                            </menuItem>
+                        </items>
+                    </menu>
+                </menuItem>
+                <menuItem title="Edit" id="5QF-Oa-p0T">
```
wez: I think most the sub items in this menu don't apply and can be removed

wez:
a lot has changed internally since this PR was made.  The idea I have for this in the future is to introduce a layer in the code that abstracts the concept of a window menu bar so that we can populate it for both mac and windows (linux is notably lacking a standard for this, and will require more effort to eventually implement).

I'm going to close this PR because it won't land with its current approach!

PR #29, opened 2019-06-02
Fix and clean-up for attributes and colors

markbt:
Fix a bug in `flush_pending_attr` where if non-color attributes (e.g. bold) change, but the color does not, the color is lost.  This is because changing the attributes is actually implemented by resetting all attributes (`\e[m`) and then setting the new ones.  This has the side-effect of clearing all colors, so we need to set the colors again.

Since changing the attributes is actually done by resetting all the attributes and setting new ones again, the fallback implementation can be made simpler by doing the same.

### DIFF
--- a/termwiz/src/render/terminfo.rs
+++ b/termwiz/src/render/terminfo.rs
@@ -47,31 +47,17 @@ impl TerminfoRenderer {
         });
     }
 
-    #[cfg_attr(
-        feature = "cargo-clippy",
-        allow(clippy::cyclomatic_complexity, clippy::cognitive_complexity)
-    )]
     fn flush_pending_attr<W: UnixTty + Write>(&mut self, out: &mut W) -> Fallible<()> {
-        macro_rules! attr_on_off {
-            ($cap_on:ident, $cap_off:ident, $attr:expr, $accesor:ident, $sgr:ident) => {
-                let value = $attr.$accesor();
-                if value != self.current_attr.$accesor() {
-                    let out = out.by_ref();
-                    let on: bool = value.into();
-                    if on {
-                        if let Some(attr) = self.get_capability::<cap::$cap_on>() {
-                            attr.expand().to(out)?;
-                        } else {
-                            write!(out, "{}", CSI::Sgr(Sgr::$sgr(value)))?;
-                        }
-                    } else {
-                        if let Some(attr) = self.get_capability::<cap::$cap_off>() {
-                            attr.expand().to(out)?;
-                        } else {
-                            write!(out, "{}", CSI::Sgr(Sgr::$sgr(value)))?;
-                        }
-                    }
+        macro_rules! attr_on {
+            ($cap:ident, $sgr:expr) => {
+                if let Some(attr) = self.get_capability::<cap::$cap>() {
+                    attr.expand().to(out.by_ref())?;
+                } else {
+                    write!(out, "{}", CSI::Sgr($sgr))?;
                 }
+            };
+            ($sgr:expr) => {
+                write!(out, "{}", CSI::Sgr($sgr))?;
             };
         }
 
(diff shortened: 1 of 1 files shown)


termwiz/src/render/terminfo.rs
```diff
@@ -52,32 +52,37 @@ impl TerminfoRenderer {
...
+            let mut current_foreground = self.current_attr.foreground;
+            let mut current_background = self.current_attr.background;
+
             if !attr.attribute_bits_equal(&self.current_attr) {
-                if let Some(sgr) = self.get_capability::<cap::SetAttributes>() {
+                // Updating the attribute bits also resets the colors.
+                current_foreground = ColorAttribute::Default;
+                current_background = ColorAttribute::Default;
+
+                // The SetAttributes capability can only handle single underline and slow blink.
```
wez: This looks like it has some unintentionally broad impact; the intent of the prior code was to use the terminfo-provided SetAttributes sequences for as much as possible, falling back on our manual mostly ECMA-48 CSI sequences.

With this change, if we have single-underline or slow-blink then we won't use the terminfo at all.
While in most cases this will be a wash, I wouldn't be surprised if there was some terminal out there that wasn't totally ECMA-48 and this would be a tricky issue to run down.

I think that we should probably have this logic:

* If cap::SetAttributes is available, use it to set the attributes that it supports, then manually do the remaining attributes "by hand"
* Otherwise do it "by hand"

termwiz/src/render/terminfo.rs
```diff
@@ -52,32 +52,37 @@ impl TerminfoRenderer {
         allow(clippy::cyclomatic_complexity, clippy::cognitive_complexity)
```
wez: I'd be curious to see whether `cargo clippy` is significantly happier about this with your simplification.  I doubt it, but am curious nonetheless!

termwiz/src/render/terminfo.rs
```diff
@@ -52,32 +52,37 @@ impl TerminfoRenderer {
...
+                    if attr.underline() != Underline::Double && attr.blink() != Blink::Rapid {
+                        self.get_capability::<cap::SetAttributes>()
+                    } else {
+                        None
+                    };
+
+                if let Some(sgr) = sgr_cap {
                     sgr.expand()
                         .bold(attr.intensity() == Intensity::Bold)
                         .dim(attr.intensity() == Intensity::Half)
```
wez: Now that I look at this, the old code was treating double underline as equivalent to single here, and any blink as blink.  I'm ok with that as a first approximation, and then we can output the explicit double-underline and fast-blink as followups if needed.

termwiz/src/render/terminfo.rs
```diff
@@ -1157,6 +1127,46 @@ mod test {
         );
     }
 
+    #[test]
+    fn color_after_attribute_change() {
```
wez: thanks for adding a test for this!

termwiz/src/render/terminfo.rs
markbt: If we have single-underline or slow-blink we should still use the terminfo-based SetAttributes.  It's just the existence of double-underline or rapid-blink that makes us fall back to the old behaviour.

termwiz/src/render/terminfo.rs
markbt: Hmm.. we could drop them by testing for `attr.underline() == Underline::Single` (so we only set underline if it is single underline), and then later on do `if attr.underline() == Underline::Double { attr_on!(Sgr::Underline(Underline::Double))` (like we do for italics).  How does that sound?

termwiz/src/render/terminfo.rs
markbt: `cargo clippy` doesn't complain about this code if I remove this.  I'll remove it.  (You have other clippy warnings and errors, by the way).

wez:
Thanks!

markbt:
Oh wait, I just realised this is wrong now (it removes single underline and slow blink handling from the fallback path).  I'll prep a PR to fix that.

PR #30, opened 2019-06-02
set underline and blink in fallback path

markbt:
pull/29 inadvertently removed the fallback path for setting single
underline and slow blink.  Restore it.

### DIFF
--- a/termwiz/src/render/terminfo.rs
+++ b/termwiz/src/render/terminfo.rs
@@ -47,6 +47,7 @@ impl TerminfoRenderer {
         });
     }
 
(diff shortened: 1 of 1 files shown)


termwiz/src/render/terminfo.rs
```diff
@@ -47,6 +47,7 @@ impl TerminfoRenderer {
         });
     }
 
+    #[cfg_attr(feature = "cargo-clippy", allow(clippy::cognitive_complexity))]
```
wez: ah well, it was good while it lasted :)

markbt: Apparently 27 is too much.  Need `cargo clippy --i-am-a-genius`.

PR #39, opened 2019-06-25
add parse_first_as_vec

markbt:
Add a convenience function to the escape parser that, like `parse_first`,
matches only the first escape sequence, but instead collects all matching
actions.


markbt:
See https://github.com/markbt/streampager/commit/eb1853fe948f86ba2b3eeb81bbfe7df8981f5d33 for an example of this being used to avoid a second parse.

wez:
Thanks!

Issue #46, opened 2019-10-12
Option to link against system libraries of freetype, libpng and zlib


lilydjwg:
## Is it a build problem?

Yes.

```
$ rustup show
command not found.
```
I do not use rustup on my Arch Linux.

## Did something not work the way you expected?

The build system can't be instructed to use system libraries of freetype, libpng and zlib, which is unexpected.

By making use of system libraries, I have the following benefits:

* much less download time (I'm in a sad place where the speed of cloning from GitHub is about 15KB/s)
* less build time
* receives updates from the distribution

### What did you try?

Read the source to see if I can set any environment variables to achieve this.

wez:
You mentioned `rustup` above: whether you use rustup or not shouldn't have any bearing on how freetype is linked in.  You need to have a reasonably recent version of rust + cargo installed to build wezterm.

freetype and harfbuzz are co-dependent libraries and most distributions do not have the right combination of these available and built with the correct build options, which is why we bundle and build them.  These libraries do not get updated very frequently by distributions (which is another reason for bundling) and weakens the argument that you will receive updates from your distribution.

I spent a lot of time getting the build in its current reproducible state so, while I have empathy for the slow downloads you're experiencing, I'm sorry to say that I don't think the amount of effort required to get the build working that way is worth the time.

If you, or someone else, wanted to submit a PR that makes this possible I'm open to it, but I would want it to not be the default mode of operation.

As an alternative proposal, consider making a PR for the azure pipelines configuration that captures the built artifacts using the same mechanism we use for the other targets; then we would have a binary tarball or package download for Arch.  This feels like a more impactful use of time!

wez:
Thanks for the pull requests!  I'm currently traveling so it may take a few days before I can fully review these!

wez:
I decided against allowing this; see https://github.com/wez/wezterm/pull/48#issuecomment-568784247 for the rationale

Logarithmus:
@wez why don't you use https://lib.rs/crates/harfbuzz_rs & https://lib.rs/crates/freetype?
Do they lack any features?

wez:
When I started this, they didn't exist and the only existing bindings linked only against system libraries, but I require static linkage of known versions so that I have consistent behavior across windows, mac and linux.  In particular, freetype, harfbuzz, zlib and png are problematic on windows; all are required for proper color emoji support.

If those other crates now meet those requirements then I'm open to a PR that targets them.

PR #48, opened 2019-10-14
link against system libraries when WEZRERM_SYSDEPS envvar is set to 1

lilydjwg:
Following #46, this patch makes it possible to use system libraries instead.

I don't know how well this will work on different distributions but it works on Arch, and freetype2 (and related) API should be stable enough.

### DIFF
--- a/deps/freetype/build.rs
+++ b/deps/freetype/build.rs
@@ -74,9 +74,16 @@ fn freetype() {
 }
 
 fn main() {
-    zlib();
-    libpng();
-    freetype();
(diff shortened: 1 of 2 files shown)


wez:
Thanks for this.  I've been going back and forth on whether to take this PR, but ultimately: it is important for wezterm to control the freetype and harfbuzz builds because there is a lot of variance in the versions available and their build options.

https://github.com/wez/wezterm/commit/11b79827dee45f8f914a61bc6810af68aa34e366 resolves what I think was the main pain point in your original issue by adding the submodule sources into a source tarball and including that on the releases page.

PR #49, opened 2019-10-15
Add line drawing

markbt:
Add support to termwiz for DEC line drawing mode.  This is recorded as a cell attribute.


markbt:
I've rebased onto latest master to check this all still works.  It would be nice to get this merged as I have pending commits for adding DEC line drawing support to streampager.  I can't remember what I ran that prompted me to add that support, but something must still be using it.

markbt:
Ok, that's fair enough.  I'll put the translation layer in streampager.

PR #60, opened 2019-11-05
Add color palette to RenderableState

fanzeyi:
`ClientTab` was returning the default palette since it doesn't really have `ColorPalette` stored anywhere. We can actually get a copy of `ColorPalette` when we create `ClientTab`s from `Mux`, and we can store it in the `RenderableState`. (As `LocalTab` stores it in `TerminalState`, I didn't look into why but I guess it is a state because we can change the color palette on the fly.)

----

As a result, now running WezTerm with `connect sb` can show me the correct palette configured in my local WezTerm configuration.
﻿
image

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -927,6 +927,13 @@ impl Config {
 
         Ok(cmd)
     }
+
+    pub fn palette(&self) -> term::color::ColorPalette {
+        self.colors
+            .as_ref()
+            .map(Into::into)
+            .unwrap_or_else(Default::default)
+    }
 }
 
 #[derive(Debug, Deserialize, Clone)]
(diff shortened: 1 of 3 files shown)


src/config.rs
```diff
@@ -978,3 +985,9 @@ impl From<Palette> for term::color::ColorPalette {
         p
     }
 }
+
+impl From<Palette> for term::color::ColorPalette {
+    fn from(cfg: Palette) -> term::color::ColorPalette {
+        (&cfg).into()
+    }
+}
```
fanzeyi: I'm kinda surprised this is not implemented automatically.

wez: The trick for this turned out to be:

```
  self.colors
            .as_ref()
            .cloned() // <--- !!
            .map(Into::into)
            .unwrap_or_else(Default::default)
```

PR #61, opened 2019-11-05
Support GNUReadline-style Alt-Backspace

Imxset21:
Alt-Backspace is the GNUReadline-style shortcut for kill-previous-word.
In wezterm, this currently doesn't work, since this gets trapped by the generic `Backspace` match case.
This diff adds a more specific case for when Backspace is combined with Alt to produce the correct sequence.

This is definitely a hack and there should probably be a more general solution for GNUReadline combinations, but this solves the immediate problem.

### DIFF
--- a/term/src/terminalstate.rs
+++ b/term/src/terminalstate.rs
@@ -846,6 +846,7 @@ impl TerminalState {
         let to_send = match (key, ctrl, alt, shift, self.application_cursor_keys) {
             (Tab, ..) => "\t",
             (Enter, ..) => "\r",
+            (Backspace, _, ALT, ..) => "\x1b\x08",
             (Backspace, ..) => "\x08",
             (Escape, ..) => "\x1b",
             // Delete


wez:
Thanks! I'll follow up and test various other key combinations to see which others need to be allow with alt

Imxset21:
@wez for reference, the full list of the readline key bindings (both for vi and Emacs modes) can be found at https://linux.die.net/man/3/readline

Default Key Bindings

The following is a list of the default emacs and vi bindings. Characters with the eighth bit set are written as M-<character>, and are referred to as metafied characters. The printable ASCII characters not mentioned in the list of emacs standard bindings are bound to the self-insert function, which just inserts the given character into the input line. In vi insertion mode, all characters not specifically mentioned are bound to self-insert. Characters assigned to signal generation by stty(1) or the terminal driver, such as C-Z or C-C, retain that function. Upper and lower case metafied characters are bound to the same function in the emacs mode meta keymap. The remaining characters are unbound, which causes readline to ring the bell (subject to the setting of the bell-style variable).

PR #67, opened 2019-11-21
mux: smooth rate limiter

ony:
Clumsy example of making rate limiter be more smooth.

See comments for ac028da1b649378e5d9fb62de7a528d0fb720233

### DIFF
--- a/src/mux/mod.rs
+++ b/src/mux/mod.rs
@@ -11,6 +11,7 @@ use log::{debug, error};
 use portable_pty::ExitStatus;
 use promise::Future;
 use std::cell::{Ref, RefCell, RefMut};
+use std::cmp::min;
 use std::collections::HashMap;
 use std::io::Read;
 use std::rc::Rc;
@@ -46,6 +47,7 @@ pub struct Mux {
 }
 
 fn read_from_tab_pty(config: Arc<Config>, tab_id: TabId, mut reader: Box<dyn std::io::Read>) {
+    const MIN_CHUNKSIZE: u32 = 200;
     const BUFSIZE: usize = 32 * 1024;
     let mut buf = [0; BUFSIZE];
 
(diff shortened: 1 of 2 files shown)


wez:
Thanks for looking at this!

I've been idly looking at this too; I came up with this helper for the `RateLimiter`; it's not perfect, but it does end up admitting eg: 1 byte per ~10ms with your 100 byte/s example.

I have this in a commit that I'm about ready to push, but I'd appreciate your feedback:

```rust
    /// Attempt to admit up to `amount` number of items.
    /// On success, returns the amount that were actually admitted,
    /// which may be less than the requested amount.
    /// If no items can be admitted immediately, returns a duration
    /// of time after which the caller should retry to admit.
    pub fn admit_check(&mut self, mut amount: u32) -> Result<u32, Duration> {
        loop {
            match self.lim.check_n(amount) {
                Ok(_) => return Ok(amount),
                Err(NegativeMultiDecision::BatchNonConforming(_, over)) if amount == 1 => {
                    return Err(over.wait_time_from(Instant::now()));
                }
                _ => {}
            };

            // try again with half the size.
            // This isn't a perfectly efficient approach, especially
            // with a very large input buffer size, but it is reasonable;
            // we use a 32k buffer which means that in the worst case
            // (where the buffer is 100% full), we'll take ~15 iterations
            // to reach a decision of a single byte or a sleep delay.
            amount = amount / 2;
        }
    }
```

wez:
(I'll push my commit to make it easier for you to see and critique and we can go from there)

ony:
@wez, thank you for quick response. I left my comments on your commit. I'm not sure if there is much difference between binary search or linear estimation based on duration as for temporary work-around while looking for a better dependency.

I'm not that familiar with Rust yet. I think that your API reduces a lot of boilerplate and allows easier migration to async implementation of pumping PTY to tab.

PR #72, opened 2019-12-06
Various cleanups

jsgf:
1. Use plain bools for italic/bold font settings
2. Use info for various chitchat messages
3. Implement Debug for fontconfig `Pattern` and `FontSet`

Should be no functional change.


wez:
Thanks!

PR #90, opened 2019-12-28
Only blink cursor when it's not moving

jsgf:
The cursor should only start blinking when its been sitting in once place;
a moving cursor should always be visible. Rather than making the cursor blink
relative to the creation time of the window, track its position, and use the
last time it moved as the blinking timebase.

Fixes https://github.com/wez/wezterm/issues/83


wez:
Looks good, thanks!

PR #95, opened 2019-12-29
Focus cursor

jsgf:
Put the cursor into a "not focused" state when the window doesn't have input focus. This means:
1. not blinking, and
2. if its a block cursor, show an outline

This addresses https://github.com/wez/wezterm/issues/93. It is based on https://github.com/wez/wezterm/pull/90.

This doesn't handle focus events for Windows or macOS yet.

### DIFF
--- a/src/frontend/gui/termwindow.rs
+++ b/src/frontend/gui/termwindow.rs
@@ -95,6 +95,7 @@ impl PrevCursorPos {
 
 pub struct TermWindow {
     window: Option<Window>,
+    focused: bool,
     fonts: Rc<FontConfiguration>,
     /// Window dimensions and dpi
     dimensions: Dimensions,
@@ -177,6 +178,11 @@ impl WindowCallbacks for TermWindow {
 
     fn as_any(&mut self) -> &mut dyn Any {
         self
+    }
+
+    fn focus_change(&mut self, focused: bool) {
+        log::debug!("Setting focus to {:?}", focused);
+        self.focused = focused;
     }
 
     fn mouse_event(&mut self, event: &MouseEvent, context: &dyn WindowOps) {
@@ -612,6 +618,7 @@ impl TermWindow {
             dimensions.pixel_height,
             Box::new(Self {
                 window: None,
+                focused: false,
                 mux_window_id,
                 fonts: Rc::clone(fontconfig),
                 render_metrics,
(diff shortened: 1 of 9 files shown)


termwiz/src/surface/mod.rs
```diff
@@ -33,6 +33,7 @@ pub enum Position {
 pub enum CursorShape {
     Hidden,
     Default,
+    BlockOutline,
```
wez: since the `termwiz` crate also allows targeting other terminals than just `wezterm` (it's sort of a bit like a curses API replacement in here, with some core shared types), this enum should in theory be constrained to cursor shapes that are settable as attributes.

With that in mind, I'm not sure that this new variant strictly belongs here, although I'll admit that it feels like a convenient spot for it.

Considering that we're just using the same box sprite for this, what do you think about removing this enum variant and dealing with this at render time?

window/src/lib.rs
```diff
@@ -114,6 +114,9 @@ pub trait WindowCallbacks: Any {
     /// Called when the window is resized, or when the dpi has changed
     fn resize(&mut self, dimensions: Dimensions) {}
 
+    /// Call when window gains/loses focus
```
wez: nit: s/Call/Called/

src/frontend/gui/termwindow.rs
```diff
@@ -2084,25 +2133,31 @@ impl TermWindow {
...
+                    CursorShape::Hidden
                 }
             } else {
                 shape
+            };
+            // Get the cursor shape if the window is defocused.
+            if self.focused {
+                shape
+            } else {
+                shape.defocused()
```
wez: GH won't let me expand this, but the block just below manages the background color; I think you'll want to augment that for the `!focused + Block` case so that it uses the regular cell attributes and then I think we don't need a new shape variant

termwiz/src/surface/mod.rs
jsgf: I'll have a look.

jsgf:
I added a guess at window focus handling for Windows, but I couldn't work out how to do it for macOS.

wez:
Thanks!  I can verify the Windows focus behavior and take a crack at macOS and Wayland

jsgf:
I did Wayland, but also didn't test it.

PR #107, opened 2020-01-04
Remove unused dependency

quark-zju:
Detected by `cargo-udeps`.

I was wondering why @markbt/streampager requires `jpeg-decoder`.


wez:
Thanks :-)

PR #114, opened 2020-01-11
Use current tab directory as default for new tabs.

jsgf:
Seems to work locally, haven't tested remotely/different domains yet.

Fixes issue https://github.com/wez/wezterm/issues/113

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -373,7 +373,7 @@ async fn async_run_ssh(opts: SshCommand, params: SshParameters) -> anyhow::Resul
     domain.attach().await?;
 
     let window_id = mux.new_empty_window();
-    let tab = domain.spawn(PtySize::default(), cmd, window_id)?;
+    let tab = domain.spawn(PtySize::default(), cmd, None, window_id)?;
     let fontconfig = Rc::new(FontConfiguration::new());
     gui.spawn_new_window(&fontconfig, &tab, window_id)?;
 
@@ -429,7 +429,7 @@ fn run_serial(config: config::ConfigHandle, opts: &SerialCommand) -> anyhow::Res
     domain.attach().wait()?;
 
     let window_id = mux.new_empty_window();
-    let tab = domain.spawn(PtySize::default(), None, window_id)?;
+    let tab = domain.spawn(PtySize::default(), None, None, window_id)?;
     gui.spawn_new_window(&fontconfig, &tab, window_id)?;
 
     gui.run_forever()
@@ -506,7 +506,7 @@ async fn spawn_tab_in_default_domain_if_mux_is_empty(
 
     let tab = mux
         .default_domain()
-        .spawn(PtySize::default(), cmd, window_id)?;
+        .spawn(PtySize::default(), cmd, None, window_id)?;
     let fontconfig = Rc::new(FontConfiguration::new());
     front_end()
         .unwrap()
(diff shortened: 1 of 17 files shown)


src/main.rs
```diff
@@ -373,7 +373,7 @@ async fn async_run_ssh(opts: SshCommand, params: SshParameters) -> anyhow::Resul
     domain.attach().await?;
 
     let window_id = mux.new_empty_window();
-    let tab = domain.spawn(PtySize::default(), cmd, window_id)?;
+    let tab = domain.spawn(PtySize::default(), cmd, None, window_id)?;
```
wez: It would be nice to add a config option to specify the default cwd: on the surface of things it would simply be:

```
    /// or much better name
   default_working_dir:  Option<PathBuf>,
```

but that should probably be domain specific rather than global.
Totally fine to punt on this.

src/server/codec.rs
```diff
@@ -368,6 +368,7 @@ pub struct Spawn {
     /// If None, create a new window for this new tab
     pub window_id: Option<WindowId>,
     pub command: Option<CommandBuilder>,
+    pub command_dir: Option<String>,
```
wez: github won't let me comment on lines that aren't in the diff, so I'm picking this one :-p

I think you'll also want to add the dir to `WindowAndTabEntry` above so that we can record it when the domain is attached, and to `GetTabRenderChangesResponse` so that we can notice and update when the remote cwd is updated.

(I think you'll want to look for any place that we're dealing with the `title` and add similar processing for the cwd)

src/ssh.rs
```diff
@@ -338,12 +338,16 @@ impl Domain for RemoteSshDomain {
         &self,
         size: PtySize,
         command: Option<CommandBuilder>,
+        command_dir: Option<String>,
```
wez: FWIW, the ssh protocol doesn't define a way to specify the cwd, so it would be acceptable to just ignore this parameter here.
In theory you could augment the program that we run to turn it into `cd CWD && CMD`, but we'd need to be confident that the remote system was running a compatible shell, and that isn't guaranteed these days now that Windows can run openssh and spawn `cmd.exe` on it.

src/main.rs
jsgf: When would this default be used? I see one of:
- override tab cwd
- override $HOME
- used by default if one or both of the above don't exist/isn't accessible

wez: Heh, and this is where the complexity comes in :)

My initial thought was that the config default would be used:
* when spawning the initial tab in a window (I think all of the call sites in `main.rs` are these)
* For the `SpawnTabDomain::Domain` and `SpawnTabDomain::DomainName` cases where we can't pull the cwd from an existing tab
* Where the tab cwd `is_none`

I can see that someone might potentially want to choose how to resolve the cwd if we have one from the current tab and one in the config file, so perhaps there should be an option to specify whether the cwd from the tab should take precedence.

jsgf: Does it make sense for the default be a property of a domain?

wez: Yeah, because a domain can be remote and may have a different filesystem layout.   So I think we'd end up with a "global" default that applies to the default domain (in the `Config` struct), and then the explicitly defined domains would have that value too (`TlsDomainClient`, `UnixDomain`, but not `SshDomain`).

It's probably not very convenient to plumb that config through, so you can feel empowered to punt on this unless you really want to :-)

jsgf: Think I'll pass on it for now.

If you can specify the command, then you can effectively override the current dir anyway, so maybe it isn't all that necessary to have specific config for it. This mechanism is useful when the current dir is a dynamic property of the session.

jsgf:
OK, directory tracking is working in the mux case, at least with unix domain sockets.

wez:
Thanks!

PR #132, opened 2020-01-25
second attempt at building for older macs

wez:
refs: https://github.com/wez/wezterm/issues/128


wez:
Pushed as bd179fa56c614a173c5afbc9b104270b904c0bc1

PR #144, opened 2020-02-04
Make some termwiz deps optional

quark-zju:
The main issue is `palette`, which does not build with buck. I also replaced `derive_builder` with an ad-hoc macro.

### DIFF
--- a/termwiz/src/caps/mod.rs
+++ b/termwiz/src/caps/mod.rs
@@ -52,74 +52,74 @@
 //! With all this in mind, this module presents a `Capabilities` struct
 //! that holds information about a terminal.   The `new_from_env` method
 //! implements some heuristics (a fancy word for guessing) to compute
-//! the terminal capabilities, but also offers a `ProbeHintsBuilder`
+//! the terminal capabilities, but also offers a `ProbeHints`
 //! that can be used by the embedding application to override those choices.
+use crate::builder;
 use anyhow::Error;
-use derive_builder::*;
 use semver::Version;
 use std::env::var;
 use terminfo::{self, capability as cap};
 
-/// Use the `ProbeHintsBuilder` to configure an instance of
-/// the `ProbeHints` struct.  `ProbeHints` are passed to the `Capabilities`
-/// constructor to influence the effective set of terminal capabilities.
-#[derive(Debug, Default, Builder, Clone)]
-#[builder(default)]
-pub struct ProbeHints {
-    /// The contents of the TERM environment variable
-    term: Option<String>,
+builder! {
+    /// Use the `ProbeHints` to configure an instance of
+    /// the `ProbeHints` struct.  `ProbeHints` are passed to the `Capabilities`
+    /// constructor to influence the effective set of terminal capabilities.
+    #[derive(Debug, Default, Clone)]
+    pub struct ProbeHints {
+        /// The contents of the TERM environment variable
+        term: Option<String>,
 
-    /// The contents of the COLORTERM environment variable.
-    /// <http://invisible-island.net/ncurses/ncurses-slang.html#env_COLORTERM>
-    colorterm: Option<String>,
+        /// The contents of the COLORTERM environment variable.
+        /// <http://invisible-island.net/ncurses/ncurses-slang.html#env_COLORTERM>
+        colorterm: Option<String>,
 
-    /// The contents of the TERM_PROGRAM environment variable
-    term_program: Option<String>,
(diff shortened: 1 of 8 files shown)


termwiz/src/color.rs
```diff
@@ -71,31 +69,34 @@ impl RgbColor {
...
-
     pub fn to_tuple_rgba(self) -> RgbaTuple {
-        Srgba::<u8>::new(self.red, self.green, self.blue, 0xff)
-            .into_format()
-            .into_components()
-    }
-
-    pub fn to_linear_tuple_rgba(self) -> RgbaTuple {
-        self.to_linear().into_components()
+        (
```
wez: This is the one thing I'm unsure about in this PR; there are some subtle differences between SRGB and RGB to do with the non-linear scaling in SRGB.  I think this change is good, but it's worth making sure that the old implementation was just doing simple divide by 255; the consequence of this not being right is that the colors will have the wrong brightness.  This is more obvious with brighter color schemes!

termwiz/src/color.rs
```diff
@@ -71,31 +69,34 @@ impl RgbColor {
...
+    /// `color-names` feature is disabled.
+    ///
     /// The list of names can be found here:
     /// <https://ogeon.github.io/docs/palette/master/palette/named/index.html>
     pub fn from_named(name: &str) -> Option<RgbColor> {
-        palette::named::from_str(&name.to_ascii_lowercase()).map(|color| {
-            let color = Srgb::<u8>::from_format(color);
-            Self::new(color.red, color.green, color.blue)
-        })
+        #[cfg(feature = "color-names")]
```
wez: I think I'd like to replace this with a lazy-static hashmap initialized from rgb.txt.
That file comes originally from X11, but vim also has a decent copy:

```
$ wc -l /usr/share/vim/vim82/rgb.txt
782 /usr/share/vim/vim82/rgb.txt
$ head /usr/share/vim/vim82/rgb.txt
255 250 250             snow
248 248 255             ghost white
248 248 255             GhostWhite
245 245 245             white smoke
245 245 245             WhiteSmoke
220 220 220             gainsboro
255 250 240             floral white
255 250 240             FloralWhite
253 245 230             old lace
253 245 230             OldLace
```

it should hopefully be pretty trivial to `include_str!` that file and build up a map of the lowercased name -> rgb values for use here

termwiz/src/caps/mod.rs
```diff
@@ -52,74 +52,74 @@
...
-/// Use the `ProbeHintsBuilder` to configure an instance of
-/// the `ProbeHints` struct.  `ProbeHints` are passed to the `Capabilities`
-/// constructor to influence the effective set of terminal capabilities.
-#[derive(Debug, Default, Builder, Clone)]
-#[builder(default)]
-pub struct ProbeHints {
-    /// The contents of the TERM environment variable
-    term: Option<String>,
+builder! {
+    /// Use the `ProbeHints` to configure an instance of
```
wez: Will Rust let us do something like:

```
#[deprecated("Use ProbeHints directly")]
pub type ProbeHintsBuilder = ProbeHints;
```

to help guide anyone that might be using the hint builder?   I don't think this is critical and it may only be streampager that was using this anyway; if we can't deprecate the use of a type alias then we don't need to worry about it.

termwiz/src/color.rs
wez: I added an rgb.txt based alternative to the palette crate for this case, so we shouldn't need the color names feature in this PR

termwiz/src/color.rs
quark-zju: I kept both functions, although this to_linear one does not seem to be used.

termwiz/src/caps/mod.rs
quark-zju: I looked at the generated code and didn't quite like the API:

    pub struct ProbeHintsBuilder {
        #[doc = " The contents of the TERM environment variable"]
        term: ::std::option::Option<Option<String>>,
        ...
    }

    impl ProbeHintsBuilder {
        #[doc = " The contents of the TERM environment variable"]
        #[allow(unused_mut)]
        pub fn term(&mut self, value: Option<String>) -> &mut Self {
            let mut new = self;
            new.term = ::std::option::Option::Some(value);
            new
        }

        pub fn build(&self)
         -> ::std::result::Result<ProbeHints, ::std::string::String> {
             ...
        }
    }

I'd like to get rid of the double `Option`, and make `build` infalliable (or not needed). So the API is incompatible now.

If we want compatibility I can restore the old API.

PR #145, opened 2020-02-05
vtparse: avoid build.rs codegen

quark-zju:
Change build.rs codegen to const_fns. This makes vtparse more friendly for buck
build.

Note const_fn functions still have limitation on the current stable (1.41)
rustc (ex. native "match" or "if" cannot be used in const_fn). So I used some
tricks to get it compile.

(I have manually checked the content of `TRANSITIONS` array - it didn't change)


vtparse/src/transitions.rs
```diff
@@ -0,0 +1,343 @@
...
+/// Match `u8` using given patterns. Return `OptionPack`.
+/// The patterns should not overlap.
+macro_rules! match_action_state {
+    ( $name:ident => {
+        $( $a:tt $( ..= $b:tt )? => ($action:ident, $state:ident), )*
+    }) => {
+        OptionPack(
+            $(
+                ({
+                    const B: u8 = [$a $(,$b)?][[$a $(,$b)?].len() - 1];
```
wez: This is pretty hard work to understand at a glance; could you add some comments to explain it?
I *think* I can follow it, but it's been a solid 15 minutes of reading and thinking :)

wez: (this actually looks like perl code!)

quark-zju:
I added some comments and links in the code.

PR #160, opened 2020-02-18
Cleanup: let structopt parse ssh user/host/port instead of doing it by hand

bcully:
Implement FromStr and Display for SshParameters, so that they can be
parsed and displayed in the standard way.

The motivation for this is that I want to add gateway hosts to the ssh command, and
would like to be able to add something like `gateway: Vec<SshParamters>` to SshCommand
to do it.


wez:
Looks good, thanks!

PR #165, opened 2020-04-06
Preserve clipboard contents on single left click

butterflysky:
Previously, left clicking into a wezterm window would empty the clipboard, even if the mouse wasn't dragged.

This PR changes that behavior; now the clipboard contents are preserved unless the selection is non-empty.

This required two changes; one on mouse press and one on mouse release. I'm calling this out because I'm not sure why the clipboard contents were set to a new String on left mouse press before.

I tested this change by running in lldb with a breakpoint on [this line](https://github.com/wez/wezterm/blob/6d5a7ad143872374b0fc480d529ea619f6d850af/src/frontend/gui/termwindow.rs#L2748), clicking into the terminal, and stepping through, validating the clipboard contents before and after.

### DIFF
--- a/src/frontend/gui/termwindow.rs
+++ b/src/frontend/gui/termwindow.rs
@@ -2792,7 +2792,6 @@ impl TermWindow {
                         // Initiate a selection
                         self.selection(tab.tab_id())
                             .begin(SelectionCoordinate { x, y: stable_row });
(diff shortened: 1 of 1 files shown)


wez:
I patched in and a tried a few ad-hoc tests; looks good!
Thanks for this!

PR #174, opened 2020-05-03
Generate builds for more platforms

wez:


wez:
Squashed and pushed as ba804589

PR #197, opened 2020-05-29
maybe enable building for arm64 ubuntu

wez:


wez:
(this needs better support from GH actions; abandoning for now)

PR #212, opened 2020-06-09
termwiz: add CursorVisibility

markbt:
Some applications want to control the cursor visibility, without affecting
cursor shape.  Provide the `CursorVisibility` change type for this purpose.

Use this when dropping `UnixTerminal` to ensure the cursor is visible
without affecting the user's cursor shape.

### DIFF
--- a/termwiz/src/surface/mod.rs
+++ b/termwiz/src/surface/mod.rs
@@ -29,6 +29,18 @@ pub enum Position {
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+pub enum CursorVisibility {
+    Hidden,
+    Visible,
+}
+
+impl Default for CursorVisibility {
+    fn default() -> CursorVisibility {
+        CursorVisibility::Visible
+    }
+}
+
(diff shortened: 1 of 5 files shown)


termwiz/src/surface/mod.rs
```diff
@@ -268,6 +289,7 @@ impl Surface {
             Change::ClearToEndOfScreen(color) => self.clear_eos(*color),
             Change::CursorColor(color) => self.cursor_color = *color,
             Change::CursorShape(shape) => self.cursor_shape = *shape,
+            Change::CursorVisibility(visibility) => self.cursor_shape = (*visibility).into(),
```
wez, line 292: I think this is the only thing that I have questions about; should we change the shape here when we've logically only changed the visibility?  Should the shape internally separate visibility from configured shape?  That raises some other questions though: how do we know the current shape when we start?  Should the internal shape state be something like:

```
struct CursorShape {
    shape: enum {
       Unknown,
       SteadyBlock,
       .. other shapes but not invisible
   }
   visibility: enum {
      Unknown,
      Visible,
      NotVisible
  }
}
```

so that we know we we should manipulate it?

wez: I pushed a commit based on top of this that separates visibility from shape; thanks for kicking this off!

PR #226, opened 2020-06-20
Update to allsorts-0.4

luben:
Only the bare minimum changes


wez:
Thanks for looking at this; it's been on my mind!

I gave this a quick test using `cargo run -- start --font-shaper allsorts` and things still seem to work!

PR #245, opened 2020-07-17
Fixed bug where setting the current directory was impossible on UNIX …

john01dav:
…for non-default terminal commands


wez:
Ah, good catch, thanks!

PR #258, opened 2020-09-06
portable_pty - support for x86_64-unknown-freebsd target

TimeToogo:
Compiling the portable-pty crate for `x86_64-unknown-freebsd` currently generates a build error:

```
error[E0308]: mismatched types
   --> /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/portable-pty-0.3.0/src/unix.rs:122:53
    |
122 |         if unsafe { libc::ioctl(self.0.as_raw_fd(), libc::TIOCGWINSZ, &mut size as *mut _) } != 0 {
    |                                                     ^^^^^^^^^^^^^^^^
    |                                                     |
    |                                                     expected `u64`, found `u32`
    |                                                     help: you can convert an `u32` to `u64`: `libc::TIOCGWINSZ.into()`
```

Seems `libc::TIOCGWINSZ` is defined as u32 on some targets.

### DIFF
--- a/pty/src/unix.rs
+++ b/pty/src/unix.rs
@@ -107,7 +107,14 @@ impl PtyFd {
             ws_ypixel: size.pixel_height,
         };
 
-        if unsafe { libc::ioctl(self.0.as_raw_fd(), libc::TIOCSWINSZ, &ws_size as *const _) } != 0 {
+        if unsafe {
+            libc::ioctl(
+                self.0.as_raw_fd(),
+                libc::TIOCSWINSZ as _,
+                &ws_size as *const _,
+            )
+        } != 0
+        {
             bail!(
                 "failed to ioctl(TIOCSWINSZ): {:?}",
                 io::Error::last_os_error()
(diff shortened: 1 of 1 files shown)


pty/src/unix.rs
```diff
@@ -119,7 +119,14 @@ impl PtyFd {
 
     fn get_size(&self) -> Result<PtySize, Error> {
         let mut size: winsize = unsafe { mem::zeroed() };
-        if unsafe { libc::ioctl(self.0.as_raw_fd(), libc::TIOCGWINSZ, &mut size as *mut _) } != 0 {
+        if unsafe {
+            libc::ioctl(
+                self.0.as_raw_fd(),
+                libc::TIOCGWINSZ.into(),
```
wez, line 125: for this sort of thing I would typically do: `libc::TIOCGWINSZ as _` to magically infer the right type, but I'm also fine with `.into()`.

I'm curious why line 110 above doesn't need this same treatment; was it only this line that had a compile failure?

TimeToogo: Honestly I didn't know about `_` as a placeholder for type inference, very handy, updated to match your style.

In regards to `libc::TIOCSWINSZ` that didn't cause a compile error on this target but upon checking [some targets](https://github.com/rust-lang/libc/search?q=TIOCSWINSZ&unscoped_q=TIOCSWINSZ) also define this as `u32` so i've applied the same cast for that also.

PR #311, opened 2020-10-29
lineedit: add a way to render previews

quark-zju:
Make it possible to implement rendering previews in new lines, seen in
nodejs [1]. The preview is cleared when the line is accepted (Enter)
or canceled (Ctrl+C).

[1]: https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V12.md#repl-previews

### DIFF
--- a/termwiz/src/lineedit/host.rs
+++ b/termwiz/src/lineedit/host.rs
@@ -49,6 +49,13 @@ pub trait LineEditorHost {
         vec![OutputElement::Text(prompt.to_owned())]
     }
 
+    /// Given a reference to the current line being edited, render a preview
+    /// of its outcome. The preview is cleared when the input is accepted,
+    /// or canceled.
+    fn render_preview(&self, _line: &str) -> Vec<OutputElement> {
+        Vec::new()
+    }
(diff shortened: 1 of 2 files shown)


jankatins:
DO I understand that right that when I enter `echo whatever` the preview will show `whatever`? If so: what happens if I type `rm -rf /<cursor>` and wait a bit? Will that execute already?

wez:
> DO I understand that right that when I enter `echo whatever` the preview will show `whatever`? If so: what happens if I type `rm -rf /<cursor>` and wait a bit? Will that execute already?

This is a change to the line editor library in the Terminal Wizardry (termwiz) crate that resides in this repo, rather than a change to the default behavior of wezterm the terminal, so it won't cause your normal shell session to speculatively execute half-baked shell commands!

This PR allows an embedding application to supply an optional preview line as part of processing during line editing; the nature and semantics of that preview are entirely specific to the embedding application, and it needs to actively write code that uses this line editor and enables the preview.

wez (approved):
Nice and simple!

PR #356, opened 2020-11-26
Fixes building on aarch64

evs-ch:
It seems like 827d94a broke building on aarch64. The fix is pretty
much just adapted from bf962c8.
I know little about rust, so I might've missed some obvious issues with
this PR - it seems to work so far, though.

### DIFF
--- a/wezterm-font/src/fcwrap.rs
+++ b/wezterm-font/src/fcwrap.rs
@@ -190,7 +190,7 @@ impl Pattern {
     pub fn add_charset(&mut self, charset: &CharSet) -> anyhow::Result<()> {
         unsafe {
             ensure!(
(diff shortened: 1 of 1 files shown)


wez:
Thanks for helping out!  The CI is upset because it wants this to be `cargo fmt`'d; could you add that to this PR?

evs-ch:
Sure thing :) 

Btw: thanks for actually making wezterm in the first place – it took me quite some time to find a terminal emulator that's Wayland-native, knows how to deal with modern fonts and font features *and* runs on an systems only supporting older OpenGL/GLES-versions (for reasons of, well, closed source driver blobs and a still ongoing efforts to get things re-implemented open source)).

PR #378, opened 2020-12-12
Update matrix room url

bew:
It is using the new room name integration between matrix & gitter since they merged (see: https://matrix.org/blog/2020/09/30/welcoming-gitter-to-matrix)


wez:
There are a couple of other places that need updating too; see: https://github.com/wez/wezterm/commit/d99b025e24aa8f75d5cd9a50ecfd6ad92bca2eff

bew:
> There are a couple of other places that need updating too; see: [d99b025](https://github.com/wez/wezterm/commit/d99b025e24aa8f75d5cd9a50ecfd6ad92bca2eff)

Thanks for the pointer, will change them all when I get back to a computer

PR #384, opened 2020-12-19
Update snapshot tests to use k9@0.10.0

aaronabramov:
I'm gradually improving snapshot testing macro devx in k9 and preparing
to ship v1. Before i do this i'm changing the inline snapshot macro to be
just `snapshot!()` that takes `Debug` trait an an arg and figures out
serialization of it.

this project has just a few snapshot tests and is prefect to test the migration (if that's ok)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1786,9 +1786,9 @@ dependencies = [
 
 [[package]]
 name = "k9"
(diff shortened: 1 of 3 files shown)


wez:
Looks good; thanks!

PR #396, opened 2020-12-26
docs: update default keybinding info for CloseCurrentTab w/ its boolean param

cxiao:
CloseCurrentTab now takes a boolean parameter (confirm), but the default keybinding table in the docs hasn't been updated to reflect that.


wez:
Thanks!

Issue #406, opened 2020-12-29
Switching termwiz to thiserror instead of anyhow


passcod:
Would you be open to switching the error type of the termwiz crate from Anyhow to some custom type (e.g. with thiserror)? I'm happy to do the work for it.

As it's a library crate, it makes it hard to integrate with anything that's not using anyhow itself.

wez:
I'll gladly take a PR that does that; thanks!

PR #407, opened 2020-12-30
Convert termwiz from anyhow to its own custom error types

passcod:
Ref #406 

Only remaining uses of anyhow in the termwiz crate:
- Tests and examples
- Wrapping the FileDescriptor errors

This change does bring some fairly minor breaking changes on the API especially regarding the public traits (only insofar as the error types in Results change).

I've run tests locally on linux but not on windows.

I can convert the FileDescriptor errors in a follow up PR if you want.


termwiz/src/escape/osc.rs
```diff
@@ -1,14 +1,79 @@
...
+
 use bitflags::bitflags;
 use num_derive::*;
 use num_traits::FromPrimitive;
 use ordered_float::NotNan;
-use std::collections::HashMap;
-use std::fmt::{Display, Error as FmtError, Formatter};
-use std::str;
+use std::{
+    collections::HashMap,
```
wez, line 9: I prefer the import style on the LHS; it feels easier to grok and diff

mux/src/termwiztermtab.rs
```diff
@@ -272,33 +275,45 @@ impl std::io::Write for TermWizTerminalRenderTty {
 }
 
 impl termwiz::render::RenderTty for TermWizTerminalRenderTty {
-    fn get_size_in_cells(&mut self) -> anyhow::Result<(usize, usize)> {
+    fn get_size_in_cells(&mut self) -> Result<(usize, usize), RenderError> {
```
wez, line 278: FWIW, the mux crate doesn't need to use `thiserror`; it's not a published library crate, just an application helper crate.

termwiz/Cargo.toml
```diff
@@ -27,6 +27,7 @@ regex = "1"
 semver = "0.9"
 serde = {version="1.0", features = ["rc", "derive"], optional=true}
 terminfo = "0.7"
+thiserror = "1.0.23"
```
wez, line 30: I prefer `thiserror = "1.0"` rather than specifying the minor number here.
The `Cargo.lock` file tracks the resolved versions.

termwiz/src/error.rs
```diff
@@ -0,0 +1,37 @@
...
+use thiserror::Error;
+
+/// Convenient return type for functions.
+pub type Result<T> = std::result::Result<T, Error>;
+
+/// Main error type.
+#[derive(Debug, Error)]
+pub enum Error {
+    /// Generic I/O error.
+    #[error("i/o: {0}")]
```
wez, line 11: I think it would save some bikeshedding if these all (except for `ImpossibleState`) just passed through to the underlying `Display` impl;

```
#[error("{0}")]
```

for each variant; that way we don't need to worry about the additional phrasing for the different variants.

wez: Sort of on the same line of thinking: which of these errors are you thinking of handling in code in consumers of `termwiz`?  There's kind of a lot of effort to produce all these different error variants, and in turn, that makes it more difficult to iterate on termwiz in the future.

I can understand not wanting to bleed `anyhow` into consuming code, but I'm also concerned about the fan-out of enum variants and the "fan-in" impact within the crate.

What do you think about reducing the scope of the error variants here so that there are fewer to manage both inside and outside of the `termwiz` crate?

The most basic way of looking at that is that this `Error` type here is largely opaque outside of the crate (just implements the error related traits).

Most of the error conditions raised during escape sequence parsing aren't really discrete conditions that the calling code can do anything about, so I don't think that having so many error variants is especially helpful.  Also: I do rather enjoy the convenience of `bail!`, `ensure!` and `anyhow!` macros for constructing errors.  It is certainly lazy, but it seems like "good lazy" given that distinguishing most (or all?) of these cases generally doesn't matter.

What do you think?

termwiz/src/render/terminfo.rs
```diff
@@ -807,27 +793,27 @@ mod test {
     }
 
     impl Terminal for FakeTerm {
-        fn set_raw_mode(&mut self) -> anyhow::Result<()> {
-            bail!("not implemented");
+        fn set_raw_mode(&mut self) -> StdResult<(), TerminalError> {
+            unimplemented!()
```
wez, line 797: this change will make the program panic with no chance to handle the error, whereas previously the caller could decide to handle this

wez (commented):
Thanks for taking this on!
What do you think about reducing the number of error variants?  See my inline comment for more of my thinking around that

mux/src/termwiztermtab.rs
passcod: Yeah, but anyhow::Error doesn't implement std::error::Error so it can't be used in this. I used thiserror just to make the inner error type, not to expose it

termwiz/src/render/terminfo.rs
passcod: yeah, but this is in test code?

termwiz/src/error.rs
passcod: Didn't really have a reason, I only mechanically converted the existing `bail!` etc messages with the goal to not discard any information, but more from a perspective of _developer_ debuggability (seemed annoying to delete all the bail! etc stringed info) rather than consumers necessarily caring about, say, the particular ioctl that was called. Generally, though:

 - getting io::Error information that I don't have to parse out of a string is useful, because often it can be "no permissions" which can be handled differently from "file doesn't exist" which is also different from "no space left on device" etc.
 - Fmt, Regex, Utf8, these all can collapse into a generic opaque type. Same for where they appear in sub error types.
 - The wrapped sub error types are mostly to distinguish these two broad categories to make it easier for cases that don't care about the more precise errors to glob those out without having a single "every error is a string" type.

Going at an extreme (not serious), we could well have a top-level

```rust
#[derive(Debug, Error)]
pub enum Error {
	#[error("{0}")]
	Wrapped(anyhow::Error),
}

impl From<anyhow::Error> for Error {
	fn from(err: anyhow::Error) -> Self {
		Self::Wrapped(err)
	}
}
```

and be mostly done, keeping everything with anyhow internally.

More seriously though, and without redoing everything, I could add a `Generic(anyhow::Error)` with a `From` impl as above to the existing enums, and you'd be able to use anyhow macros internally with abandon, then specialise to a variant if needed/requested. (That's how I did the conversion piece by piece in the first place.) Then I'd undo all of the specialised error conversions that were a bit overzealous (like all the Tty variants).

termwiz/src/render/terminfo.rs
passcod: Undone

passcod:
Okay, did a partial reduction (only done for the main and terminal errors), for discussion purposes.

termwiz/src/error.rs
markbt: Rather than `#[error("{0}")]` you should instead use `#[error(transparent)]` for errors that you want to pass through directly.

Similarly, errors fields marked as `#[source]` (or `#[from]` which implies source) shouldn't be included in the formatting.  The user of the error can walk the error chain if they want to know more, and if you include the source error in the formatting then users that *do* walk the chain will get the error printed twice, which looks ugly.

I made simliar changes in streampager [here](https://github.com/markbt/streampager/commit/44cae6d89b983903186f4c602d9c35fe97c503e5).

passcod: Ah, good point. I'll change things around (tomorrow, it's late here).

wez:
Thanks for doing this; I wound up with 5d360ae3657473ba45c84fec0fbcdce82a80eec8 which conflicts with this PR.
While I didn't take your PR as-is, it was valuable to help me see and think through what I wanted to see happen here.
I opted to just push what I thought rather than have a protracted slow motion discussion here, especially as it's taken me a few days to get back to you.

I opted to keep `Error` opaque to a similar degree as it was with `anyhow::Error`.
You can use eg: `Error::source` to extract underlying errors if necessary, and I'm open to hearing concrete situations where the ergonomics could be improved!

passcod:
Absolutely fine, thank you for this anyway! ---- On Fri, 08 Jan 2021 21:41:11 +1300  notifications@github.com  wrote ----
Thanks for doing this; I wound up with 5d360ae which conflicts with this PR.
While I didn't take your PR as-is, it was valuable to help me see and think through what I wanted to see happen here.
I opted to just push what I thought rather than have a protracted slow motion discussion here, especially as it's taken me a few days to get back to you.
I opted to keep Error opaque to a similar degree as it was with anyhow::Error.
You can use eg: Error::source to extract underlying errors if necessary, and I'm open to hearing concrete situations where the ergonomics could be improved!

PR #409, opened 2020-12-30
Instruct docs.rs to build with widgets and serde features

passcod:
As seen in https://github.com/async-rs/async-std/blob/master/Cargo.toml#L20-L22

That way, docs.rs will have the `Widget` struct documentation available.


wez:
I published termwiz 0.11 today and the docs now show widgets: https://docs.rs/termwiz/0.11.0/termwiz/widgets/index.html
Thanks!

PR #412, opened 2021-01-01
added hyperfile functionality

zeusintuivo:
= Branch created: Fri Jan  1 20:15:48 CET 2021
  +---- This task: Fri Jan  1 20:15:48 CET 2021
 
 
 Make it possible to click on the filenames with line nummbers 
 I am not a rust programmer but I am basically cloning the hyperlink 
 functionality to make it possible to click a file and open it with 
 local editor. 
 
 Any ideas from real rust programmers or better ways of doing it are welcomed. 
 
 
 
20:15pm ---


wez:
Thanks for contributing!
However, I don't think you need to add a whole new concept to achieve what you're looking for here.

I'd suggest doing something like the following:

* Define a hyperlink rule to match the filename patterns of interest
* Define an `open-uri` callback in your `wezterm.lua`; docs are here: https://wezfurlong.org/wezterm/config/lua/wezterm/on.html
* In you callback, if the URI looks like your filename URI, you can use [pane:get_current_working_dir](https://wezfurlong.org/wezterm/config/lua/pane/get_current_working_dir.html) to resolve the current directory of the pane to produce an absolute path
* Then you can spawn eg: `xdg-open` or `start` to open the newly computed file URL, or you can directly spawn your preferred editor with the parameters you need.

zeusintuivo:
> Thanks for contributing!
> However, I don't think you need to add a whole new concept to achieve what you're looking for here.
> 
> I'd suggest doing something like the following:
> 
>     * Define a hyperlink rule to match the filename patterns of interest
> 
>     * Define an `open-uri` callback in your `wezterm.lua`; docs are here: https://wezfurlong.org/wezterm/config/lua/wezterm/on.html
> 
>     * In you callback, if the URI looks like your filename URI, you can use [pane:get_current_working_dir](https://wezfurlong.org/wezterm/config/lua/pane/get_current_working_dir.html) to resolve the current directory of the pane to produce an absolute path
> 
>     * Then you can spawn eg: `xdg-open` or `start` to open the newly computed file URL, or you can directly spawn your preferred editor with the parameters you need.

So the suggestion here, is to have the file be opened by the browser and then browser redirect it a file. It is far more complex to even try to figure out to configure xdg-open, and even more complex to make it work.  Plus not a very efficient technique, making the browser open the file. 

Also, when I look into the code, this approach that you tell me, it seems it might not work, becuase if I understand it correctly, the code now checks if the "url exists" so technically a filepath will always be return false. 

As far as "hightlight" it works. I can get it to highlight files, but the trigger, that detects the urls, will always return false, and the detected filenames will never open a browser. 

If you have some receipe that works, I will like to know, becuase I searched in the documentation and there is notthing that regard.

wez:
With this config file:

```lua

wezterm.on("open-uri", function(window, pane, uri)
  local start, match_end = uri:find("hyperfile:");
  if start == 1 then
    local name = uri:sub(match_end+1);
    local start, match_end = name:find(":");
    local lineno = name:sub(match_end+1)
    name = name:sub(1, match_end-1)

    local action = wezterm.action{SpawnCommandInNewWindow={
         args={"vim", "+"..lineno, name}
      }};
     window:perform_action(action, pane);

    -- prevent the default action from opening in a browser
    return false
  end
end)

return {
  hyperlink_rules = {
    {
      regex = "^\\s*[a-zA-Z0-9/_\\-\\. ]+\\.?[a-zA-Z0-9]+:[0-9]+",
      format = "hyperfile:$0"
    }
  }
}
```

Then to get the output that you want to match:

```bash
$ cd wezterm
$ grep -Rn message wezterm-gui/src/gui/**/*.rs
wezterm-gui/src/gui/mod.rs:78:                        Connection::get().unwrap().terminate_message_loop();
wezterm-gui/src/gui/mod.rs:83:        self.connection.run_message_loop()
wezterm-gui/src/gui/overlay/confirm_close_pane.rs:12:fn run_confirmation_app(message: &str, term: &mut TermWizTerminal) -> anyhow::Result<bool> {
wezterm-gui/src/gui/overlay/confirm_close_pane.rs:38:    let wrapped = textwrap::fill(message, text_width);
wezterm-gui/src/gui/overlay/confirm_close_pane.rs:40:    let message_rows = wrapped.split("\n").count();
wezterm-gui/src/gui/overlay/confirm_close_pane.rs:44:    let top_row = (size.rows - (message_rows + 2)) / 2;
wezterm-gui/src/gui/overlay/confirm_close_pane.rs:46:    let button_row = top_row + message_rows + 1;
wezterm-gui/src/gui/overlay/confirm_close_pane.rs:225:            con.terminate_message_loop();
wezterm-gui/src/gui/termwindow.rs:1962:                        con.terminate_message_loop();
```

Clicking on `wezterm-gui/src/gui/termwindow.rs:1962` will cause a new window to spawn and run `vim +1962 wezterm-gui/src/gui/termwindow.rs` inside it.

zeusintuivo:
> ```lua
> wezterm.on("open-uri", function(window, pane, uri)
>   local start, match_end = uri:find("hyperfile:");
>   if start == 1 then
>     local name = uri:sub(match_end+1);
>     local start, match_end = name:find(":");
>     local lineno = name:sub(match_end+1)
>     name = name:sub(1, match_end-1)
> 
>     local action = wezterm.action{SpawnCommandInNewWindow={
>          args={"vim", "+"..lineno, name}
>       }};
>      window:perform_action(action, pane);
> 
>     -- prevent the default action from opening in a browser
>     return false
>   end
> end)
> 
> return {
>   hyperlink_rules = {
>     {
>       regex = "^\\s*[a-zA-Z0-9/_\\-\\. ]+\\.?[a-zA-Z0-9]+:[0-9]+",
>       format = "hyperfile:$0"
>     }
>   }
> }
> ```

Nice, thanks, this works! This is exactly what I was looking for. But out of curiosity. You can see that I took several hours to complete this PR. I would like to see compile in my laptop. I fixed the tests. But, it is now basically a clone of hyperlinks with a hyperfile functionally. Just out of curiosity, how would you complete this PR, so hyperfile works out of the box in this code structure. Just for the sake of learning.

termwiz/src/escape/osc.rs
```diff
@@ -384,6 +387,7 @@ osc_entries!(
     SetCurrentWorkingDirectory = "7",
     /// See https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda
     SetHyperlink = "8",
+    SetHyperfile = "88",
```
wez, line 390: this uses OSC 88, but other places still use and refer to OSC 8

termwiz/src/caps/mod.rs
```diff
@@ -302,6 +314,12 @@ impl Capabilities {
         self.hyperlinks
     }
 
+    /// Does the terminal support hyperfiles?
+    /// See <https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda>
```
wez, line 318: this spec doesn't apply to hyperfiles

termwiz/README.md
```diff
@@ -12,6 +12,7 @@ Included functionality:
 * `Surface` models a terminal display and its component `Cell`s
 * Terminal attributes are aware of modern features such as
   True Color, [Hyperlinks](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda)
+  True Color, [Hyperfiles](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda)
```
wez, line 15: `True Color` is a separate feature already mentioned above

term/src/test/mod.rs
```diff
@@ -141,10 +141,19 @@ impl TestTerm {
...
+        let osc = OperatingSystemCommand::SetHyperfile(Some(file.as_ref().clone()));
+        self.print(format!("{}", osc));
+    }
+
     fn hyperlink_off(&mut self) {
         self.print("\x1b]8;;\x1b\\");
     }
 
+    fn hyperfile_off(&mut self) {
+        self.print("\x1b]8;;\x1b\\");
```
wez, line 154: 8 vs 88

docs/hyperfiles.markdown
```diff
@@ -0,0 +1,44 @@
...
+      format = "$0",
+    },
+  }
+}
+```
+
+### Explicit Hyperfiles
+
+wezterm supports the relatively new [Hyperfiles in Terminal
+Emulators](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda)
```
wez, line 31: not the hyperfile spec

docs/hyperfiles.markdown
```diff
@@ -0,0 +1,44 @@
...
+specification that allows emitting text that can be clicked and resolve to a
+specific URL, without the URL being part of the display text.  This allows
+for a cleaner presentation.
+
+The gist of it is that running the following bash one-liner:
+
+```bash
+printf '\e]8;;example.json\e\\This is a file\e]8;;\e\\\n'
```
wez, line 39: 8 vs 88

termwiz/src/surface/line.rs
```diff
@@ -22,6 +23,12 @@ bitflags! {
         const SCANNED_IMPLICIT_HYPERLINKS = 1<<2;
         /// true if we found implicit hyperlinks in the last scan
         const HAS_IMPLICIT_HYPERLINKS = 1<<3;
+        /// The line contains 1+ cells with explicit hyperlinks set
+        const HAS_HYPERFILE = 1<<1;
```
wez, line 27: these bit assignments collide with those they're copying above; they should be `1<<4`, `1<<5` and so on

wez:
> But out of curiosity. You can see that I took several hours to complete this PR. 

I appreciate you taking the time to dive in!

> I would like to see compile in my laptop. I fixed the tests. But, it is now basically a clone of hyperlinks with a hyperfile functionally. Just out of curiosity, how would you complete this PR, so hyperfile works out of the box in this code structure. Just for the sake of learning.

Right now the CI is red because the PR needs to be formatted using `cargo fmt` so I can't see precisely what is failing, but I've put in a couple of quick comments in the review.

The main reason *not* to pursue this PR is that it adds a new escape sequence and a concept that no other terminal has, and one that overlaps the existing Hyperlinks spec; the implementation right now appears to overload OSC 8 for that purpose, which seems like a conflicting OSC code assignment (ah, in some places 88, others 8).

Adding a new terminal escape requires not only support from the emulator but also client code; the `termwiz` crate does double-duty as both part of the terminal emulator model in `wezterm` but also as a client library for console applications, and needs to know how to determine if a terminal supports a given feature.  For something that is really "emulator UI side" like this, we shouldn't need to change either the internal terminal model, introduce a new escape sequence or change any other client programs.

The second reason not to take this PR as-is, is that there is a significant amount of code duplication between Hyperlinks and Hyperfiles which means that the same issues will need to resolved in both copies of the code, making overall maintenance more difficult.

So overall, I would be inclined to avoid making so many changes to the rust code; if we were to bake in this kind of file clicking feature I would want to find a way to keep the link portions as hyperlinks and try to find a way to make it easy to enable something like the lua config snippet for common editors.  At the moment that is something we can do with better docs, but I'm also open to the idea of expanding the lua functions we have available to make this simpler for folks in the future!

termwiz/src/escape/osc.rs
zeusintuivo: Here I used 8 at first. But I noticed that this specific line when having hyperfile as 8 and hyperlink as 8, the tests will fail. Because the hyperlink would receive a hyperfile, when the hyperfile=8 is assigned secondly, or if I switch places. And if I place hyperlink = 8 secondly then this would make the hyperfile test fail. So I added a random number 88, I don't know now what OSC 8 is, but I observed that 88 number fixed the interaction.

termwiz/src/caps/mod.rs
zeusintuivo: Right, I thought I made up the term hyperfile given the hyperlink name. I could delete this line

termwiz/README.md
zeusintuivo: Here I can delete this too. I

termwiz/src/escape/osc.rs
zeusintuivo: Do you recon we could use another number besides 8 ?

term/src/test/mod.rs
zeusintuivo: Good catch if I am using 88 then it should be 88, right. Unless you know of a better number?

docs/hyperfiles.markdown
zeusintuivo: Ok, I will delete this line

docs/hyperfiles.markdown
zeusintuivo: Good catch again

termwiz/src/surface/line.rs
zeusintuivo: Ok, I do knot know what "1<<" stands for, but I am going to make change...mmm. done.

zeusintuivo:
I am going to make this format chances and suggestions and run cargo fmt.

termwiz/src/caps/mod.rs
zeusintuivo: done

termwiz/README.md
zeusintuivo: done

docs/hyperfiles.markdown
zeusintuivo: done

zeusintuivo:
About the lua config file. 
What if I want to also read 

Diff in /Users/benutzer/_/itch/wezterm/termwiz/src/lib.rs at line 46:

Then I am testing an extra step like this. 

```lua
local function isEmpty(s)
	-- Sample use
	-- if isempty(foo) then
 	--  	foo = "default value"
	-- end
  return s == nil or s == ''
end


wezterm.log_error("Hello!");
wezterm.on("open-uri", function(window, pane, uri)
  	wezterm.log_error("0 uri " .. uri);
  local m_fr, m_to = uri:find("hyperfile:");
  	wezterm.log_error("0 from " .. m_fr);
  	wezterm.log_error("0   to " .. m_to);
  if m_fr == 1 then
    local name = uri:sub(m_to+1);
	  if isEmpty(name) then
	  	wezterm.log_error("name is nil");
	  end
    	wezterm.log_error("1 name " .. name);
    local m_fr = name:find(":");
	  if isEmpty(m_fr) then
	  	wezterm.log_error("m_fr is nil : not found");
			local m_fr = 0
			local m_to = 0
			local m_fr = name:find("Diff in");
		  if isEmpty(m_fr) then
			  wezterm.log_error("m_fr is nil 'Diff in ' not found");
		  else
	    	local m_fr, m_to = name:find("Diff in ");
	    	wezterm.log_error("2 from " .. m_fr);
		  	wezterm.log_error("2   to " .. m_to);
		  	local name = name:sub(m_to+1);
	    	wezterm.log_error("2 name " .. name);
				local m_fr = name:find(" at line ");
			  if isEmpty(m_fr) then
				  wezterm.log_error("m_fr is nil ' at line ' not found");
				else
		    	local m_fr, m_to = name:find(" at line ");
		    	wezterm.log_error("3 from " .. m_fr);
			  	wezterm.log_error("3   to " .. m_to);
			  	local lineno = name:sub(m_to+1)
			  	wezterm.log_error("3 line " .. lineno);
		      name = name:sub(1, m_fr-1)
		    	wezterm.log_error("3 name " .. name);
		    	local action = wezterm.action{SpawnCommandInNewWindow={
		         args={"nano", "+"..lineno, name}
		      }};
		     window:perform_action(action, pane);
			  end
		  end
		else
    	local m_fr, m_to = name:find(":");
    	wezterm.log_error("1 from " .. m_fr);
	  	wezterm.log_error("1   to " .. m_to);
	    local lineno = name:sub(m_to+1);
    	wezterm.log_error("1 line " .. lineno);
      name = name:sub(1, m_to-1)
    	wezterm.log_error("1 name " .. name);
    	local action = wezterm.action{SpawnCommandInNewWindow={
         args={"nano", "+"..lineno, name}
      }};
     window:perform_action(action, pane);
	  end
    return false
  end
end)
return {
  hyperlink_rules = {
    {
      regex = "^\\s*[a-zA-Z0-9/_\\-\\. ]+\\.?[a-zA-Z0-9]+:[0-9]+",
      format = "hyperfile:$0"
    },
    {
       regex = "Diff in [a-zA-Z0-9/_\\-\\. ]+\\.?[a-zA-Z0-9]+",
       format = "hyperfile:$0"
    }
  }
}
```

termwiz/src/surface/line.rs
zeusintuivo: done

term/src/test/mod.rs
zeusintuivo: done

zeusintuivo:
I think the only thing left pending for me to do, is to add another ""open-file" service, because now it uses the "open-uri" clone funtionality.  Unless you know a better way.

wez:
> What if I want to also read
> `Diff in /Users/benutzer/_/itch/wezterm/termwiz/src/lib.rs at line 46:`

I'd use captures in the hyperlink rule; the parentheses define captured subexpressions.
The `format` can then reference them using $1 and $2; $0 is the entire matched string,
$1 is the first capture, $2 the second and so on.

So if you match that whole pattern, you can put parens around the filename and line number portions,
and then `format` can stick them together so that you always have `hyperfile:FILENAME:LINENO`,
and that helps keep the open-uri event simpler.

```lua
     {
        regex = "Diff in ([a-zA-Z0-9/_\\-\\. ]+\\.?[a-zA-Z0-9]+) at line ([0-9]+):",
        format = "hyperfile:$1:$2"
     }
```

termwiz/src/escape/osc.rs
wez: I'd suggest going with 88 as a thought experiment to see this working, but overall: I don't think we should be adding a new OSC code for this when we can use hyperlink rules with the existing hyperlinks OSC 8 :)

zeusintuivo:
Ok, so I added the "open-file" extra event. It seems all tests pass. I can build it. But it seems that the highlight does not work when running the produced build. It also seems the hyperfile extra functionality is not getting triggered.  

Now, comes a learning bit. Which I might need the masters' assistance, if you may, tell me what is wrong or missing in the highlight. Could it be that hyperlink and hyperfile are canceling each other like you mentioned that now there is extra scanning work going on. Or that it is just missing more moving parts? 

I also tried the new format of reading which works fine, with the normal build; to what I have also placed it in the hyperfile code in the markdown. 
```lua
local function isEmpty(s)
  -- Sample use
  -- if isempty(foo) then
  --    foo = "default value"
  -- end
  return s == nil or s == ''
end
local function runAction(window, pane, name, lineno)
  -- Sample use
  -- runAction(window, pane, name, lineno)
  local action = wezterm.action{SpawnCommandInNewWindow={
    -- args={"vim", "+"..lineno, name}
    -- args={"nano", "+"..lineno, name}
    -- args={"pstorm", name..":"..lineno}
    -- args={"mine", name..":"..lineno}
    -- args={"webstorm", name..":"..lineno}
    -- args={"charm", name..":"..lineno}
    -- args={"subl", name..":"..lineno}
    -- args={"brackets", name}
    -- args={"code", "-g", name..":"..lineno}
    args={"code-insiders", "-g", name..":"..lineno}
  }};
  window:perform_action(action, pane);
end

wezterm.on("open-file", function(window, pane, uri)
  local m_fr, m_to = uri:find("hyperfile:");
  if m_fr == 1 then
    local name = uri:sub(m_to+1);
    local m_fr = name:find(":");
    if isEmpty(m_fr) then
      local m_fr = 0
      local m_to = 0
      local m_fr = name:find("Diff in");
      if false == isEmpty(m_fr) then
        local m_fr, m_to = name:find("Diff in ");
        local name = name:sub(m_to+1);
        local m_fr = name:find(" at line ");
        if false == isEmpty(m_fr) then
          local m_fr, m_to = name:find(" at line ");
          local lineno = name:sub(m_to+1)
          name = name:sub(1, m_fr-1)
          runAction(window, pane, name, lineno);
        end
      end
    else
      local m_fr, m_to = name:find(":");
      local lineno = name:sub(m_to+1);
      name = name:sub(1, m_to-1)
      runAction(window, pane, name, lineno);
    end
    -- prevent the default action from opening in a browser
    return false
  end
  -- otherwise, by not specifying a return value, we allow later
  -- handlers and ultimately the default action to caused the
  -- URI to be opened in the browser
end)
return {
  hyperfile_rules = {
    {
      regex = "^\\s*[a-zA-Z0-9/_\\-\\. ]+\\.?[a-zA-Z0-9]+:[0-9]+",
      format = "hyperfile:$0"
    },
    {
       regex = "Diff in [a-zA-Z0-9/_\\-\\. ]+\\.?[a-zA-Z0-9]+",
       format = "hyperfile:$1:$2"
    }
  }
}
```
Only differences here would be that right now with the production build  I am implementing `wezterm.on("open-uri", ` instead of `wezterm.on("open-file", ` and the regex name on production is  `  hyperlink_rules = {` instead of `  hyperfile_rules = {`

wez:
As mentioned above, I'm not going to merge this functionality.

PR #426, opened 2021-01-13
ci: generate universal binaries for macOS

wez:
This should persuade the CI to install both the arm and intel flavors
of the rust toolchain on macOS, and the deploy script to generate
a universal binary.


wez:
Manually pushed this to master in 42b95cd4134af898f43624426d1b39b39fa7fc74

PR #435, opened 2021-01-22
Reduce compile memory required by vtparse

bugadani:
Instead of 3800MB, let's top out at ~220 :) Probably raises MSRV, though. Also, untested.

### DIFF
--- a/vtparse/src/transitions.rs
+++ b/vtparse/src/transitions.rs
@@ -49,40 +49,19 @@ macro_rules! define_function {
 
 /// Apply all u8 values to `fn(u8) -> u8`, return `[u8; 256]`.
 macro_rules! define_table {
-    ( $func:tt ) => {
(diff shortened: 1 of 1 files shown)


vtparse/src/transitions.rs
```diff
@@ -49,40 +49,22 @@ macro_rules! define_function {
...
-            iter=($($iter)*+16)
-            remaining=[$($remaining)*]
-        )
-    };
+    ( $func:tt ) => {{
+        const fn gen() -> [u8; 256] {
+            let mut arr = [0; 256];
+
+            let mut i = 0u8;
+            loop {
```
wez, line 57: can this be the more idiomatic `for i in 0..256 {` style loop?

wez:
Nice!  Thanks for looking at this!  I have a minor question/comment inline above!

vtparse/src/transitions.rs
bugadani: No it doesn't compile yet I think.

PR #437, opened 2021-01-24
Add Sync marker to child process created through pty spawn

CompuIves:
This will also allow sharing references of the process across different threads. I need this to share state between different threads.

I don't think that this will break any existing behaviour, from what I've seen (and from what the compiler tells) every data structure in there is `Sync` compatible.

### DIFF
--- a/pty/src/win/mod.rs
+++ b/pty/src/win/mod.rs
@@ -3,6 +3,7 @@ use anyhow::Context as _;
 use std::io::{Error as IoError, Result as IoResult};
 use std::os::windows::io::{AsRawHandle, RawHandle};
 use std::pin::Pin;
+use std::sync::Mutex;
 use std::task::{Context, Poll};
 use winapi::shared::minwindef::DWORD;
 use winapi::um::minwinbase::STILL_ACTIVE;
@@ -18,13 +19,14 @@ use filedescriptor::OwnedHandle;
 
 #[derive(Debug)]
 pub struct WinChild {
-    proc: OwnedHandle,
+    proc: Mutex<OwnedHandle>,
 }
 
 impl WinChild {
     fn is_complete(&mut self) -> IoResult<Option<ExitStatus>> {
         let mut status: DWORD = 0;
-        let res = unsafe { GetExitCodeProcess(self.proc.as_raw_handle(), &mut status) };
+        let proc = self.proc.lock().unwrap();
+        let res = unsafe { GetExitCodeProcess(proc.as_raw_handle(), &mut status) };
         if res != 0 {
             if status == STILL_ACTIVE {
                 Ok(None)
@@ -37,7 +39,8 @@ impl WinChild {
     }
 
     fn do_kill(&mut self) -> IoResult<()> {
-        let res = unsafe { TerminateProcess(self.proc.as_raw_handle(), 1) };
+        let proc = self.proc.lock().unwrap();
+        let res = unsafe { TerminateProcess(proc.as_raw_handle(), 1) };
         let err = IoError::last_os_error();
         if res != 0 {
             Err(err)
@@ -62,11 +65,12 @@ impl Child for WinChild {
         if let Ok(Some(status)) = self.try_wait() {
             return Ok(status);
         }
+        let proc = self.proc.lock().unwrap();
         unsafe {
-            WaitForSingleObject(self.proc.as_raw_handle(), INFINITE);
+            WaitForSingleObject(proc.as_raw_handle(), INFINITE);
         }
         let mut status: DWORD = 0;
-        let res = unsafe { GetExitCodeProcess(self.proc.as_raw_handle(), &mut status) };
+        let res = unsafe { GetExitCodeProcess(proc.as_raw_handle(), &mut status) };
         if res != 0 {
             Ok(ExitStatus::with_exit_code(status))
         } else {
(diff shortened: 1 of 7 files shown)


wez:
Thanks for this!
Looks like something needs to be tweaked for the Windows build to be happy:

```
Error[E0053]: method `spawn_command` has an incompatible type for trait
   --> pty\src\win\conpty.rs:115:5
    |
115 |     fn spawn_command(&self, cmd: CommandBuilder) -> anyhow::Result<Box<dyn Child + Send>> {
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected trait `Child + Send + Sync`, found trait `Child + Send`
    | 
   ::: pty\src\lib.rs:132:5
    |
132 |     fn spawn_command(&self, cmd: CommandBuilder) -> Result<Box<dyn Child + Send + Sync>, Error>;
    |     -------------------------------------------------------------------------------------------- type in trait
    |
    = note: expected fn pointer `fn(&ConPtySlavePty, CommandBuilder) -> std::result::Result<Box<(dyn Child + Send + Sync + 'static)>, _>`
               found fn pointer `fn(&ConPtySlavePty, CommandBuilder) -> std::result::Result<Box<(dyn Child + Send + 'static)>, _>`
```

May just be able to get away with adding the marker on the impl there also?

CompuIves:
Ah yep, you're completely right! Forgot to compile for Windows. Commit incoming...

wez (changes requested):
a bit of a stumbling block still:

```
error[E0277]: `*mut c_void` cannot be shared between threads safely
361
   --> pty\src\win\conpty.rs:118:12
362
    |
363
118 |         Ok(Box::new(child))
364
    |            ^^^^^^^^^^^^^^^ `*mut c_void` cannot be shared between threads safely
365
    |
366
    = help: within `WinChild`, the trait `Sync` is not implemented for `*mut c_void`
367
    = note: required because it appears within the type `OwnedHandle`
368
    = note: required because it appears within the type `WinChild`
369
    = note: required for the cast to the object type `dyn Child + Send + Sync`
```

We can't simply mark `OwnedHandle` as `Sync` because it wouldn't be true.

I think you'll need to do something to `WinChild` in `pty/src/win/mod.rs` to make it `Sync` safe; perhaps wrapping the `proc` in a `Mutex` would do the job?

https://github.com/wez/wezterm/blob/master/pty/src/win/mod.rs#L19-L22

CompuIves:
Thanks a lot for the help on this! I can't seem to get the Windows version of the build to run, so these suggestions are really helpful. I've just committed a fix (hopefully) 🤞 .

CompuIves:
Ah, I just realized that I'm locking a Mutex here:

https://github.com/wez/wezterm/pull/437/files#diff-9e47a53d152496778518f3b10fe3d466cd2fd43d7e3e0b0de24100451399de13R93

In a future poll. That's an interesting one, I could make this Mutex an async version, and then use a blocking lock where it makes sense?

pty/src/win/mod.rs
```diff
@@ -62,11 +65,12 @@ impl Child for WinChild {
         if let Ok(Some(status)) = self.try_wait() {
             return Ok(status);
         }
+        let proc = self.proc.lock().unwrap();
         unsafe {
-            WaitForSingleObject(self.proc.as_raw_handle(), INFINITE);
+            WaitForSingleObject(proc.as_raw_handle(), INFINITE);
```
wez, line 70: Waiting while a mutex is held is a potential source of trouble.
There are at least two possible ways to resolve this:

* Grab the raw handle under the lock, but release it prior to the wait:

```rust
let handle = self.proc.lock().unwrap().as_raw_handle();
unsafe {
    WaitForSingleObject(handle, INFINITE);
}
```

This feels like it might be potentially possible for eg: the handle to be closed outside of the scope of the Mutex guarding it, but in practice, since `wait` requires a mutable reference it should be theoretically impossible for there to be a concurrent actor.

* Make a copy of the handle:

See https://docs.rs/filedescriptor/0.7.3/filedescriptor/struct.OwnedHandle.html#method.try_clone

```rust
let handle = self.proc.lock().unwrap().try_clone()?;
unsafe {
   WaitForSingleObject(handle.as_raw_handle(), INFINITE);
}
```

This is safer, but a little more heavy weight, and the `try_clone` could fail due to kernel resource shortage (although: the system would likely be badly borked if that were true so something else
would be likely to fail as well).

pty/src/win/mod.rs
```diff
@@ -85,7 +89,9 @@ impl std::future::Future for WinChild {
             Ok(None) => {
                 struct PassRawHandleToWaiterThread(pub RawHandle);
                 unsafe impl Send for PassRawHandleToWaiterThread {}
-                let handle = PassRawHandleToWaiterThread(self.proc.as_raw_handle());
+
+                let proc = self.proc.lock().unwrap();
```
wez, line 93: It'd probably be better (wrt. unexpected deadlock/blocking) to constrain the scope of this lock, similar to the first suggestion I made above.

This `PassRawHandleToWaiterThread` construct is a (pre-existing) little wonky thing.  Similar to above it is safer to `try_clone`, but note that `poll` could be called relatively more frequently, so it is potentially more likely to fail to clone the handle, but also: this code is already spawning a new thread for each call through this code path, which is already a bit gross, but unavoidable because each call has its own `waker`, so I don't know if it really makes anything worse to add a `try_clone` in here.

CompuIves: That makes sense to me, and like you said, if the `try_clone` only happens in case of lack of resources then there are probably other things to worry about in this case.

CompuIves:
Yes, those are really good points. I also got a Windows machine now to run and see the clippy results! I made the changes, thanks a ton for the elaborate feedback, I'm learning a lot from it so I really appreciate it!

wez:
Thanks!

PR #438, opened 2021-01-24
Add config option to specify default current working directory

dfrankland:
Specifies the default current working directory for the initially spawned program.

### DIFF
--- a/config/src/lib.rs
+++ b/config/src/lib.rs
@@ -526,6 +526,10 @@ pub struct Config {
     /// as the positional arguments to that command.
     pub default_prog: Option<Vec<String>>,
 
+    /// Specifies the default current working directory for the initially
+    /// spawned program.
+    pub default_cwd: Option<PathBuf>,
+
     /// Specifies a map of environment variables that should be set
     /// when spawning commands in the local domain.
     /// This is not used when working with remote domains.
@@ -1241,6 +1245,10 @@ impl Config {
     }
 
     pub fn apply_cmd_defaults(&self, cmd: &mut CommandBuilder) {
+        if let Some(ref cwd) = self.default_cwd {
+            cmd.cwd(cwd);
+        }
+
         for (k, v) in &self.set_environment_variables {
             cmd.env(k, v);
         }
--- a/config/src/lib.rs
+++ b/config/src/lib.rs
@@ -1245,7 +1245,9 @@ impl Config {
     }
 
     pub fn apply_cmd_defaults(&self, cmd: &mut CommandBuilder) {
-        if let Some(ref cwd) = self.default_cwd {
+        // Apply `default_cwd` only if `cwd` is not already set, allows `--cwd`
+        // option to take precedence
+        if let (None, Some(cwd)) = (cmd.get_cwd(), &self.default_cwd) {
             cmd.cwd(cwd);
         }
 
--- a/config/src/lib.rs
+++ b/config/src/lib.rs
@@ -526,8 +526,9 @@ pub struct Config {
     /// as the positional arguments to that command.
     pub default_prog: Option<Vec<String>>,
 
-    /// Specifies the default current working directory for the initially
-    /// spawned program.
+    /// Specifies the default current working directory if none is specified
+    /// through configuration or OSC 7 (see docs for `default_cwd` for more
+    /// info!)
     pub default_cwd: Option<PathBuf>,
 
     /// Specifies a map of environment variables that should be set
--- a/docs/config/launch.markdown
+++ b/docs/config/launch.markdown
@@ -49,6 +49,48 @@ to edit your wezterm configuration:
 wezterm start -- vim ~/.wezterm.lua
 ```
 
+## Specifying the current working directory
+
+If you'd like `wezterm` to start running a program in a specific working
+directory you can do so via the config, CLI, and when using
+[`SpawnCommand`](lua/SpawnCommand.md):
+
+* Setting the [`default_cwd`](lua/config/default_cwd.md) via the config:
+
+  ```lua
+  return {
+    default_cwd = "/some/path",
+  }
+  ```
+
+* One off program in a specific working directory via the CLI:
+
+  ```bash
+  wezterm start --cwd /some/path
+  ```
+
+* The [`SpawnCommandInNewTab`](lua/keyassignment/SpawnCommandInNewTab.md),
+  [`SpawnCommandInNewWindow`](lua/keyassignment/SpawnCommandInNewWindow.md),
+  [`SpawnTab`](lua/keyassignment/SpawnTab.md) key assignments, and the
+  [Launcher Menu](#the-launcher-menu) described below all accept a
+  [`SpawnCommand`](lua/SpawnCommand.md) object that accepts an optional `cwd` field:
+
+  ```lua
+  {
+    label = "List files in /some/path",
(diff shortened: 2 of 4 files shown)


config/src/lib.rs
```diff
@@ -1241,6 +1245,10 @@ impl Config {
     }
 
     pub fn apply_cmd_defaults(&self, cmd: &mut CommandBuilder) {
+        if let Some(ref cwd) = self.default_cwd {
```
wez, line 1248: I'm a little "worried" about how this interacts with other things that want to set the cwd.
There are a couple of places that come to mind:

 * The logic that applies the OSC 7 specified cwd to the command; `LocalDomain::spawn` and `LocalDomain::split_pane`.  Those apply the cwd *after* calling `apply_cmd_defaults` so those seem ok
 * `wezterm start --cwd /some/where` which pre-populates the cwd before `spawn_tab_in_default_domain_if_mux_is_empty` calls `spawn` which calls down to `apply_cmd_defaults` here.
 * Specifying a `cwd` setting in a launch menu item (see: https://wezfurlong.org/wezterm/config/launch.html#the-launcher-menu).  I didn't check to see if this would behave correctly or not; would be great if you could verify that!

I think the `--cwd` option should override the value specified in the configuration file, so what I think needs to be done is to introduce:

```rust
   pub fn get_cwd(&self) -> Option<&OsString> {
       self.cwd.as_ref()
   }   
```

to the `CommandBuilder` in `pty/src/cmdbuilder.rs`

so that the logic here in this function can be something like:

```rust
    if let Some(ref cwd) = self.default_cwd {
       // Allow `--cwd` option to take precedence
       if cmd.get_cwd().is_none() {
          cmd.cwd(cwd);
       }
    }
```

docs/config/lua/config/default_cwd.md
```diff
@@ -0,0 +1,4 @@
+# `default_cwd`
+
+Specifies the default current working directory for the initially spawned
+program.
```
wez, line 4: Would be great to expand on this because the outcome is a bit more complicated/nuanced; it's probably a good idea to explain this by adding a `Current Working Directory` section to https://wezfurlong.org/wezterm/config/launch.html#launching-programs, as well as covering the same information here:

If `wezterm start --cwd /some/path` is used to specify the current working directory, that will take precedence for the initially created window.

Otherwise, the initial window will use the value specified by the `default_cwd` configuration option.

Commands launched using `SpawnCommand` (link to https://wezfurlong.org/wezterm/config/lua/SpawnCommand.html) will use the `cwd` specified in the `SpawnCommand` (if any).

Otherwise: Panes/Tabs/Windows created after the first will generally try to resolve the current working directory of the current Pane, preferring a value set by OSC 7 (link to Shell Integration: https://wezfurlong.org/wezterm/shell-integration.html) and falling back to attempting to lookup the cwd of the current process group leader attached to a local Pane.  If no cwd can be resolved, then the `default_cwd` will be used.  If `default_cwd` is not specified, then the home directory of the user will be used.

config/src/lib.rs
```diff
@@ -526,6 +526,10 @@ pub struct Config {
     /// as the positional arguments to that command.
     pub default_prog: Option<Vec<String>>,
 
+    /// Specifies the default current working directory for the initially
```
wez, line 529: Might be better to go for a more succinct:

> Specifies the default current working directory if none is specified through configuration or OSC 7 (see docs for `default_cwd` for more info!)

wez (changes requested):
Thanks also for this!

A minor but important change needs to be made, and I've suggested some extra prose for the docs!

config/src/lib.rs
dfrankland: For the last bit of logic there you suggest, do you have any feelings about me writing it like this?

```rust
// Apply `default_cwd` only if `cwd` is not already set, allows `--cwd`
// option to take precedence
if let (None, Some(cwd)) = (cmd.get_cwd(), &self.default_cwd) {
    cmd.cwd(cwd);
}
```

wez: Go for it!

docs/config/lua/config/default_cwd.md
```diff
@@ -0,0 +1,47 @@
...
+                               |                        Use OSC 7
+                        Is `default_cwd`                  value
+                            defined?
+                     __________|__________
+                    |                     |
+                   Yes                    No
+                    |                     |
+                   Use                 Use home
+              `default_cwd`            directory
+```
```
dfrankland, line 47: I am not sure if this is too verbose for you or not. I felt it rendered pretty well:

Screenshot from 2021-01-25 16-58-36

docs/config/lua/config/default_cwd.md
```diff
@@ -0,0 +1,47 @@
...
+              |                                             |
+       Opened with CLI                  Opened with a `SpawnCommand` that includes
+      and `--cwd` flag?                  `cwd` or is a new pane, tab, or window?
+    __________|__________                         __________|__________
+   |                     |                       |                     |
+  Yes                    No                      No                   Yes
+   |                     |                       |                     |
+  Use                    |                Is there a value    Use `cwd` specified
+`--cwd`                  |                 set by OSC 7?       by `SpawnCommand`
+                         |             __________|__________
```
wez, line 34: That looks great!   Did you render that by hand?

There's a bit missing; If there is no OSC 7 value, then there is a "can the cwd of the process group leader be resolved?" if yes, use that, if no --> "if default_cwd defined".

A footnote describing that could be something like:

> On macOS and Linux, wezterm can attempt to resolve the process group leader and then attempt to resolve its current working directory.  This is not guaranteed to succeed, and there are a number of potential edge cases (which is another reason for configuring your shell to use OSC 7 sequences).

docs/config/launch.markdown
```diff
@@ -49,6 +49,44 @@ to edit your wezterm configuration:
...
+  }
+  ```
+
+* One off program in a specific working directory via the CLI:
+
+  ```bash
+  wezterm start --cwd /some/path
+  ```
+
+* Using `SpawnCommand` in a specific working directory:
```
wez, line 72: I'd tweak this to link to where/how the user might do this:

The [SpawnCommandInNewTab](https://wezfurlong.org/wezterm/config/lua/keyassignment/SpawnCommandInNewTab.html), [SpawnCommandInNewWindow](https://wezfurlong.org/wezterm/config/lua/keyassignment/SpawnCommandInNewWindow.html), [SpawnTab](https://wezfurlong.org/wezterm/config/lua/keyassignment/SpawnTab.html) key assignments, and the [Launcher Menu](https://wezfurlong.org/wezterm/config/launch.html#the-launcher-menu) described below all accept a [SpawnCommand](https://wezfurlong.org/wezterm/config/lua/SpawnCommand.html) object that accepts an optional `cwd` field

dfrankland:
I've amended the changes to the last commit. This is how it looks rendered now:

Screenshot from 2021-01-26 09-35-24
Screenshot from 2021-01-26 09-34-41

docs/config/lua/config/default_cwd.md
dfrankland: Yes, I've made this flow chart by hand. I make a lot of charts like this for work, but not in ascii art fortunately :blush:

wez:
This is great, thank you!

PR #439, opened 2021-01-25
Add erase mode to ClearScrollback action

dfrankland:
The new erase mode `ScrollbackAndViewport` clears all lines, both visible and those scrolled off the top of the viewport, making the prompt line the new first line and resetting the scrollbar thumb to the full height of the window.

This is the behavior that Hyper / xterm has for clearing the terminal.

https://github.com/xtermjs/xterm.js/blob/28bcd88fb0f7d32c686461dc4594996d1877920e/src/browser/Terminal.ts#L1169-L1187

### DIFF
--- a/mux/src/pane.rs
+++ b/mux/src/pane.rs
@@ -117,6 +117,7 @@ pub trait Pane: Downcast {
     fn palette(&self) -> ColorPalette;
     fn domain_id(&self) -> DomainId;
 
+    fn erase_buffer(&self) {}
     fn erase_scrollback(&self) {}
 
     /// Called to advise on whether this tab has focus
--- a/mux/src/pane.rs
+++ b/mux/src/pane.rs
@@ -2,6 +2,7 @@ use crate::domain::DomainId;
 use crate::renderable::*;
 use crate::Mux;
 use async_trait::async_trait;
+use config::keyassignment::ScrollbackEraseMode;
 use downcast_rs::{impl_downcast, Downcast};
 use portable_pty::PtySize;
 use rangeset::RangeSet;
@@ -117,8 +118,7 @@ pub trait Pane: Downcast {
     fn palette(&self) -> ColorPalette;
     fn domain_id(&self) -> DomainId;
 
-    fn erase_buffer(&self) {}
-    fn erase_scrollback(&self) {}
+    fn erase_scrollback(&self, _erase_mode: ScrollbackEraseMode) {}
 
     /// Called to advise on whether this tab has focus
     fn focus_changed(&self, _focused: bool) {}
--- a/term/src/terminalstate.rs
+++ b/term/src/terminalstate.rs
@@ -674,7 +674,7 @@ impl TerminalState {
         }
     }
 
-    pub fn erase_buffer(&mut self) {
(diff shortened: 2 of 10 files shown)


wez (changes requested):
Thanks for diving in to this!
I think the implementation looks ok (although I haven't looked too closely at the actual buffer manipulation yet!), I just have a couple of organizational changes that should help make this also work with multiplexer sessions :-)

term/src/terminalstate.rs
```diff
@@ -674,6 +674,19 @@ impl TerminalState {
...
+
+        let row_index = self.screen.phys_row(self.cursor.y);
+        let row = self.screen.lines[row_index].clone();
+
+        self.erase_in_display(EraseInDisplay::EraseDisplay);
+
+        self.screen.lines[0] = row;
+
+        self.cursor.y = 0;
+    }
```
dfrankland, line 688: Thanks for the review! This is a great idea, I didn't realize that actions could be configured for enum variants that have associated data. I have one question about the implementation of the enum for the mode. It seems that the enum needs to be (de)serializable, which is why I defined it in the `config::keyassignment` module, but the `term` crate isn't associated with the `config` crate since it is supposed to be standalone. Would you like for me to keep `erase_buffer` as a function or should I define another enum in this `terminalstate` module which the enum in `cofig::keyassignment` would map to?

wez: I'd be fine with keeping this new function, although I think we could make its name more closely match the enum variant (perhaps: `erase_scrollback_and_viewport`?).

The other alternatives are OK, but feel like a lot of effort relative to the reward.

mux/src/pane.rs
```diff
@@ -118,6 +118,7 @@ pub trait Pane: Downcast {
     fn palette(&self) -> ColorPalette;
     fn domain_id(&self) -> DomainId;
 
+    #[allow(unused)]
     fn erase_scrollback(&self, erase_mode: ScrollbackEraseMode) {}
```
dfrankland: I hope this alright. I figured it's better to allow `erase_mode` to be unused than prefix it with `_`

wez: I'd prefer `_erase_mode` here; attributes are more to type and can diverge from the code that led to the attribute being added.

The `_` prefix is understood to mean that it is unused in a given instance, even though it feels a little weird for a trait definition to have unused parameters.  This is a little bit of a wart in Rust TBH.

dfrankland: No problem. Would you like / be okay with me to rebase as well? I can change `erase_mode` to `_erase_mode` in the initial commit I added it. I don't want you to lose your PR comments unless this is the last thing before you are ready to merge.

wez: I'm totally fine with you rebasing/squashing/force pushing.  My next steps here when I've got a few minutes in the right headspace later this evening are:

* Actually look closely at the screen/line buffer manipulation (I've only glossed over it so far)
* Assuming it's good (nothing obviously bad in my skimming so far), squash and merge
* Add an entry in the changelog; briefly summarizing the change; you're welcome to preempt me on that!

Issue #310, opened 2020-10-28
ghostly text artifacts after clearing screen on macos with transparency enabled


erf:
## Describe the bug

When `window_background_opacity` is less than 1.0, e.g. `0.3`, and i list some text in the terminal, then resize the window, sometimes the text (buffer?) is not cleared. I also noticed this some times when using the [Kitty](https://github.com/kovidgoyal/kitty) terminal some time ago. This might be an error with some macOS spesific code.

I noticed when i open Wezterm from the terminal i get this ERROR message, not sure if it has to do with anything but i thought i mention it.

```
% wezterm
 2020-10-28T09:25:59.204Z ERROR wezterm_gui::gui::termwindow > OpenGL initialized! ANGLE (Metal Renderer: Intel(R) Iris(TM) Plus Graphics 640) OpenGL ES 3.0.0 (ANGLE 2.1.0.8230df39a500) is_context_loss_possible=false
^C
```

## Environment (please complete the following information):

 - OS: macOS
 - Version: 20200909-002054-4c9af461-240-g3808cf3f

## To Reproduce

1. enable transparency by setting `window_background_opacity=0.3`
2. `ls` some text in the terminal window
3. resize the window, make it larger
4. now clear the window with `Ctrl-L`
5. notice the text is remained in the window, although with slight less opacity
6. resize the window again to make the text go away

## Configuration

Only set `window_background_opacity=0.3`

## Expected behavior

I would expect the text to always clear on `Ctrl-L`

## Screenshots

![Screenshot 2020-10-28 at 10 36 32](https://user-images.githubusercontent.com/1562523/97420279-67b23900-190b-11eb-94d2-07f95a9a60fe.png)
![Screenshot 2020-10-28 at 10 25 13](https://user-images.githubusercontent.com/1562523/97420288-68e36600-190b-11eb-8618-70b695db33fc.png)

wez:
Seems to be specific to macOS.  Neither Windows nor Linux have a similar issue.

This also appears whether Metal (via EGL) is used or the older CGL is used (you can use `prefer_egl=false` to force CGL if you want to compare this for yourself, but it appears the same).

erf:
Did you manage to recreate it? Do you think it's worth looking into?

wez:
Yeah, I can reproduce it.  I'm not sure what can be done about it; wezterm does clear the sufaces (because otherwise we'd have the same problem on all systems, with or without opacity), so I'm not sure how to make macos properly respect that.
[...]

PR #445, opened 2021-01-30
Only disable drop shadow on macOS when window background opacity is transparent

fanzeyi:
as title.

see #310 

https://user-images.githubusercontent.com/409951/106346474-50694b80-626c-11eb-9614-fcdbf88bd9c3.mov

### DIFF
--- a/window/src/os/macos/window.rs
+++ b/window/src/os/macos/window.rs
@@ -396,11 +396,6 @@ impl Window {
             let _: () = msg_send![*window, setTabbingMode:2 /* NSWindowTabbingModeDisallowed */];
 
             window.setReleasedWhenClosed_(NO);
-            window.setOpaque_(NO);
-            // Turn off the window shadow, because when the background is transparent
-            // having the shadow enabled seems to correlate with ghostly remnants
-            // see: https://github.com/wez/wezterm/issues/310
-            window.setHasShadow_(NO);
             let ns_color: id = msg_send![Class::get("NSColor").unwrap(), alloc];
             window.setBackgroundColor_(cocoa::appkit::NSColor::clearColor(ns_color));
 
@@ -453,6 +448,7 @@ impl Window {
                 .insert(window_id, Rc::clone(&window_inner));
 
             let window = Window(window_id);
+            window.config_did_change();
 
             inner.borrow_mut().enable_opengl()?;
             // Synthesize a resize event immediately; this allows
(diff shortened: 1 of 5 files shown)


window/src/os/macos/window.rs
```diff
@@ -397,10 +398,7 @@ impl Window {
 
             window.setReleasedWhenClosed_(NO);
             window.setOpaque_(NO);
```
wez, line 400: I think the artifacts in the other issue came about because this opacity boolean is NO and shadow is YES.  Just enabling the shadows when the alpha level is opaque may not prevent artifacts from showing up (not sure).

wez (commented):
I'm OK with the spirit of this change but I'm not sure about the implementation.
There are two main things about it that don't sit well with me:

* The added method to control the opacity doesn't change the opacity at all (that's a function of the alpha in the renderer)
* There's some fan-out through to all of the different platforms from the change in the construction of the window

What do you think about exposing the opacity configuration through the relatively recently introduced `WindowConfiguration` trait for the window layer over here:

https://github.com/wez/wezterm/blob/master/window/src/configuration.rs#L20

That way we don't need to add more parameters to the constructor.
I think the problem then becomes with letting the window layer know that the config changed.

So, perhaps a method like `config_did_change` or `advise_or_config_change` or some better name could be added to `WindowOps` and `WindowOpsMut` would do the job?

In the macos implementation of that method you'd use the `WindowConfiguration` trait to see what the opacity is set to and then `window.setHasShadow_` and `window.setOpaque_` accordingly.

window/src/os/macos/window.rs
fanzeyi: Ah, so if we change opaque correctly here we can have avoid that issue even with drop shadow on?

wez: We have to have opaque = NO for alpha < 1.0 to work.  But when the window is not opaque, drop shadows result in artifacts.  I don't know if the artifacts are caused by the alpha not being 1.0 or if it is just because opaque=NO (or both). 

I think to be safest, we should set both opaque and drop shadow flags to the appropriate values based on the configured opacity level.

fanzeyi: Ah, I only realized now that my understanding of "opaque" is the opposite of what it means (hence the direction I set in the last commit is incorrect). 😬

wez:
Thanks!

PR #446, opened 2021-01-30
Add `dual` fonts

leiserfg:
Dual spaced fonts (despite the name) are also monospaced fonts. This PR will allow 
to use fonts like Iosevka Term (That's mainly while I'm doing it 😄).


wez:
Thanks!  I tweaked it a little in b49c7a7a72920a8e63fc1a1f4725e95b4863d673 to avoid the iter/map/chain, and to iterate over MONO and DUAL more directly; it felt a bit more natural.  I also added the same concept to the logic that searches fallback; I don't think it changes much in practice, but it makes those two code paths more consistent with each other!

PR #459, opened 2021-02-05
Add cli option --config-file

bew:
It works! :tada: 

I didn't find a documentation for global flags/options, is it currently missing?

I'm not fond of the duplication between wezterm/wezterm-gui/wezterm-mux-server.. But this is a larger problem I think, it shouldn't be solved in this PR.

For the implementation
- I extracted the existing config load logic to a function that takes a Path, and make it used when the flag is given.

- I changed the signature of `reload` (to avoid having `reload()` and `reload_from_path(path)` functions), to take a `ConfigFileSelection` (not sure about the naming), to choose between searching the config file or using a given one.

### DIFF
--- a/config/src/lib.rs
+++ b/config/src/lib.rs
@@ -1034,8 +1034,8 @@ impl Config {
         let lua = make_lua_context(path)?;
         let config: mlua::Value = smol::block_on(
             lua.load(&s)
-            .set_name(path.to_string_lossy().as_bytes())?
-            .eval_async(),
+                .set_name(path.to_string_lossy().as_bytes())?
+                .eval_async(),
         )?;
         cfg = luahelper::from_lua_value(config).with_context(|| {
             format!(
--- a/config/src/lib.rs
+++ b/config/src/lib.rs
@@ -974,6 +974,19 @@ pub struct LoadedConfig {
     lua: Option<mlua::Lua>,
 }
 
+struct PathPossibility {
+    path: PathBuf,
+    is_required: bool,
+}
+impl PathPossibility {
+    pub fn required(path: PathBuf) -> PathPossibility {
+        PathPossibility { path, is_required: true }
+    }
+    pub fn optional(path: PathBuf) -> PathPossibility {
+        PathPossibility { path, is_required: false }
+    }
+}
+
 impl Config {
     pub fn load() -> Result<LoadedConfig, Error> {
         // Note that the directories crate has methods for locating project
@@ -982,8 +995,8 @@ impl Config {
         // so we do this bit "by-hand"
 
         let mut paths = vec![
-            CONFIG_DIR.join("wezterm.lua"),
-            HOME_DIR.join(".wezterm.lua"),
+            PathPossibility::optional(CONFIG_DIR.join("wezterm.lua")),
+            PathPossibility::optional(HOME_DIR.join(".wezterm.lua")),
         ];
         if cfg!(windows) {
             // On Windows, a common use case is to maintain a thumb drive
@@ -996,26 +1009,27 @@ impl Config {
             // dir as the executable that will take precedence.
             if let Ok(exe_name) = std::env::current_exe() {
                 if let Some(exe_dir) = exe_name.parent() {
-                    paths.insert(0, exe_dir.join("wezterm.lua"));
+                    paths.insert(0, PathPossibility::optional(exe_dir.join("wezterm.lua")));
                 }
             }
         }
         if let Some(path) = std::env::var_os("WEZTERM_CONFIG_FILE") {
             log::trace!("Note: WEZTERM_CONFIG_FILE is set in the environment");
-            paths.insert(0, path.into());
+            paths.insert(0, PathPossibility::required(path.into()));
         }
 
         if let Some(path) = CONFIG_FILE_OVERRIDE.lock().unwrap().as_ref() {
             log::trace!("Note: config file override is set");
-            paths.insert(0, path.clone());
+            paths.insert(0, PathPossibility::required(path.clone()));
         }
 
-        for p in &paths {
+        for path_item in &paths {
+            let p = path_item.path.as_path();
             log::trace!("consider config: {}", p.display());
             let mut file = match fs::File::open(p) {
                 Ok(file) => file,
                 Err(err) => match err.kind() {
-                    std::io::ErrorKind::NotFound => continue,
+                    std::io::ErrorKind::NotFound if !path_item.is_required => continue,
                     _ => bail!("Error opening {}: {}", p.display(), err),
                 },
             };
--- a/config/src/lib.rs
+++ b/config/src/lib.rs
@@ -59,6 +59,7 @@ lazy_static! {
     pub static ref CONFIG_DIR: PathBuf = xdg_config_home();
     pub static ref RUNTIME_DIR: PathBuf = compute_runtime_dir().unwrap();
     static ref CONFIG: Configuration = Configuration::new();
+    static ref CONFIG_FILE_OVERRIDE: Mutex<Option<PathBuf>> = Mutex::new(None);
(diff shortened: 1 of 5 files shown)


wezterm/src/main.rs
```diff
@@ -29,6 +29,10 @@ struct Opt {
     #[structopt(short = "n")]
     skip_config: bool,
 
+    /// Specify the configuration file to use, overrides WEZTERM_CONFIG_FILE
```
wez, line 32: I'd rather say `overrides the normal configuration file resolution` as that is a bit more precise

wezterm/src/main.rs
```diff
@@ -29,6 +29,10 @@ struct Opt {
...
+    #[structopt(long = "config-file", parse(from_os_str))]
```
wez, line 33: you can add something like `conflicts_with = "skip_config"` in here to have clap/structopt error out if both `-n` and `--config-file` are used together.

I mention this because I have been bitten by `WEZTERM_CONFIG_FILE=/tmp/issueXXX.lua wezterm -n` not doing what I expect a few times while investigating issues!

wezterm/src/main.rs
```diff
@@ -228,7 +232,12 @@ fn run() -> anyhow::Result<()> {
 
     let opts = Opt::from_args();
     if !opts.skip_config {
-        config::reload();
+        if let Some(ref config_file) = opts.config_file {
```
wez, line 235: I would have tackled this a bit differently:

* Added a `static ref CONFIG_FILE_OVERRIDE: Mutex<Option<PathBuf>> = Mutex::new(None)` `config/src/lib.rs`, and then add a function to set it:

```rust
pub fn set_config_file_override(path: &Path) {
   CONFIG_FILE_OVERRIDE.lock().unwrap().replace(path.to_path_buf());
}
```

Then here in main.rs the logic would be:

```rust
if let Some(config_file) = opts.config_file.as_ref() {
   config::set_config_file_override(config_file);
}
config::reload();
```

then see my comment in `config/src/lib.rs` for where I'd hook that up

config/src/lib.rs
```diff
@@ -997,47 +1006,11 @@ impl Config {
             paths.insert(0, path.into());
         }
 
-        for p in &paths {
```
wez, line 1000: (see my comment down in main.rs before reading this one)

then, this logic could be kept over here with a minor addition being inserted here (github won't let me comment on lines that are not part of a PR; this logically belongs right after the insertion logic for the env var above so that this insert takes precedence at the start of the vec):

```rust
   if let Some(path) = CONFIG_FILE_OVERRIDE.lock().unwrap().as_ref() {
      paths.insert(0, path.clone());
   }
```

config/src/lib.rs
```diff
@@ -251,8 +251,13 @@ pub fn configuration() -> ConfigHandle {
     CONFIG.get()
 }
 
-pub fn reload() {
-    CONFIG.reload();
+pub enum ConfigFileSelection<'a> {
```
wez, line 254: This is a nice idea, but there are a couple of cases in the implementation below where I think it would misbehave.  My suggestion would be to not do this and use the CONFIG_FILE_OVERRIDE global that I suggest elsewhere in this review.

However, an alternative way to express that idea is for `CONFIG_FILE_OVERRIDE` to be  `Mutex<ConfigFileSelection>` rather than an `Mutex<Option<PathBuf>>`.  For that to work out, you'd need to make the `FromPath` variant own its data: `FromPath(PathBuf)`.

I'm not sure that it adds much value over `Option<PathBuf>`; it adds more lines and a locally defined type that is equivalent to `Option`, so I think the maintenance cost is slightly higher in the long run, and the `Option` version just wins out for me.

config/src/lib.rs
```diff
@@ -317,7 +322,7 @@ impl ConfigInner {
                     log::trace!("event:{:?}", event);
                     if let Some(path) = extract_path(event) {
                         log::debug!("path {} changed, reload config", path.display());
-                        reload();
+                        reload(ConfigFileSelection::Search);
```
wez, line 325: I mentioned misbehavior in a comment above; here's a problem scenario:

You set `--config-file` on the command line.  If you update that file, this code triggers and instead of reloading the original file it now loads using the default config file resolution instead of directly loading the file that you specified via the command line option.

For this to be right, we need to remember that the cli override was used.

(It may just happen to squeak by because we also set the environment variable to match what we loaded, but that's more of a lucky coincidence and is the kind of subtle behavior that could get overlooked and broken over time if it isn't explicitly handled)

wez (commented):
Thanks for taking a crack at this!  This is something that I'd use, so I definitely appreciate it!

I think this could be a smaller PR with fewer changes if you follow my suggestion of adding a global to remember the config file path.

wezterm/src/main.rs
bew: I tried to add `conflicts_with` on `config_file` structopt, but when testing it doesn't seem to work at all, I can do `./target/debug/wezterm -n --config-file conf.lua` without errors :confused:

wezterm/src/main.rs
bew: done!

config/src/lib.rs
bew: > It may just happen to squeak by because we also set the environment variable to match what we loaded, but that's more of a lucky coincidence and is the kind of subtle behavior that could get overlooked and broken over time if it isn't explicitly handled

Good point you're right, it's better when the behavior is explicit!

config/src/lib.rs
bew: I used `Mutex<Option<PathBuf>>` as suggested, and with the config file override terminology it makes a lot of sense :+1:

config/src/lib.rs
bew: I did this, but I think it's a bit wrong:
If the file given with `--config-file foo/bar` does NOT exist, the next possibility will be considered.
:point_right:  I would expect it to fail to load (and fallback to default config?) if the given file is unreadable, no?

bew:
I don't know how you prefer to do code reviews, so I didn't marked your comments as resolved, so you can see my answers and mark then as resolved if it is ok for you.

config/src/lib.rs
wez: Good catch.  I think this a good place to introduce a little enum:

```rust
enum Item {
   Required(PathBuf),
   Optional(PathBuf),
}
```

Then the members of the `paths` vec here are `Item::Required` for entries produced from the `CONFIG_FILE_OVERRIDE` or the `WEZTERM_CONFIG_FILE` environment, and `Item::Optional` for all of the others.

Then in the loop below the `NotFound => continue` line would bail for `Required` and continue for `Optional`.

wezterm/src/main.rs
wez: I wonder if it wants `skip-config` instead of `skip_config`; `structopt` presumably translates the name to use hyphens to pass down to the underlying `clap` crate.

bew: that was it!

bew:
Replying to https://github.com/wez/wezterm/pull/459#discussion_r571450100

> > I would expect it to fail to load (and fallback to default config?) if the given file is unreadable, no?
>
> Good catch. I think this a good place to introduce a little enum:
> ```rs
> enum Item {
>    Required(PathBuf),
>    Optional(PathBuf),
> }
> ```

It should be done now! tested locally, it works great!
I didn't implemented with the enum you suggested though, I found it was quite hard to use in the for loop over paths items.
I'm not fond of the naming of `PathPossiblity` but don't know what to name it to.. (Item feels way too generic, or PathItem ? but it's weird, looks like a path component.. anyway..)

I hope you like it :smiley:

config/src/lib.rs
bew: see reply below https://github.com/wez/wezterm/pull/459#issuecomment-774679104

wez:
Thanks!

bew:
> Thanks!

Youhou, thank you 😊

Issue #461, opened 2021-02-06
Undocumented options


bew:
Here are the list of undocumented options, based on this version of the config: [config/src/lib.rs@b6d85f94](https://github.com/wez/wezterm/blob/b6d85f941db2f5df389d007fb34fa18e0139a25a/config/src/lib.rs)

- [x] adjust_window_size_when_changing_font_size
- [x] allow_square_glyphs_to_overflow_width #462 
- [x] automatically_reload_config #462 
- [x] check_for_updates #462 
- [x] check_for_updates_interval_seconds #462
- [x] daemon_options
- [x] debug_key_events
- [x] enable_csi_u_key_encoding
- [ ] experimental_shape_post_processing (won't be documented I guess)
- [x] native_macos_fullscreen_mode
- [x] ~~periodic_stat_logging~~ - internal developer use
- [x] prefer_egl
- [ ] ratelimit_mux_line_prefetches_per_second
- [x] ~~resolved_palette~~ - internal cache, not user configurable
- [x] scroll_to_bottom_on_input #462
- [x] selection_word_boundary #462
- [x] show_tab_index_in_tab_bar
- [x] show_update_window #462
- [x] ssh_domains
- [x] swap_backspace_and_delete
- [x] tab_and_split_indices_are_zero_based
- [x] tls_clients
- [x] tls_servers
- [x] ~~use_local_build_for_proxy~~ - internal developer use

---
For the record, here is how I made the list:
1. copy all lines in `Config` struct ([example](https://github.com/wez/wezterm/blob/b6d85f941db2f5df389d007fb34fa18e0139a25a/config/src/lib.rs#L459-L858)) in an empty file
2. (vim) Remove all docs with `:v/pub/d`
3. (vim) Cleanup the lines with `:%norm dndwf:D`
4. (vim) Sort all options `:%sort`
5. save file to `all_options`
6. find missing docs with:

   ```sh
   while read opt; do
     rg $opt docs/config/ > /dev/null || echo $opt
   done <all_options
   ```

bew:
The last one from that list is `ratelimit_mux_line_prefetches_per_second` just want to note that I have no idea how to document that.

I re-ran the steps, and the undocumented options are:
```
experimental_shape_post_processing
periodic_stat_logging
ratelimit_mux_line_prefetches_per_second
resolved_palette
use_local_build_for_proxy
```
I updated the first post to add them (:

wez:
[...]

PR #462, opened 2021-02-06
Add some missing option doc

bew:
- automatically_reload_config
- check_for_updates
- check_for_updates_interval_seconds
- show_update_window
- allow_square_glyphs_to_overflow_width
- scroll_to_bottom_on_input
- selection_word_boundary

Part of #461

### DIFF
--- a/docs/config/lua/config/scroll_to_bottom_on_input.md
+++ b/docs/config/lua/config/scroll_to_bottom_on_input.md
@@ -1,3 +1,5 @@
 # `scroll_to_bottom_on_input`
 
-When true (the default) scrolls to the bottom of the scrollback on input.
+When `true` (the default), the viewport will automatically scroll to the
+bottom of the scrollback when there is input to the terminal so that you
+can see what you are typing.
--- a/docs/config/lua/config/scroll_to_bottom_on_input.md
+++ b/docs/config/lua/config/scroll_to_bottom_on_input.md
@@ -1,0 +1,3 @@
+# `scroll_to_bottom_on_input`
+
+When true (the default) scrolls to the bottom of the scrollback on input.
--- a/docs/config/lua/config/show_update_window.md
+++ b/docs/config/lua/config/show_update_window.md
@@ -1,6 +1,6 @@
 # `show_update_window`
 
-When the automatic update checks detects a new version of Wezterm, this
+When Wezterm checks for an update and detects a new version, this
 option controls whether a window is shown with information about the new
 available version and links to download/install it.
 
--- a/docs/config/lua/config/show_update_window.md
+++ b/docs/config/lua/config/show_update_window.md
@@ -1,0 +1,14 @@
+# `show_update_window`
+
+When the automatic update checks detects a new version of Wezterm, this
+option controls whether a window is shown with information about the new
+available version and links to download/install it.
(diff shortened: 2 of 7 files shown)


docs/config/misc.md
```diff
@@ -65,6 +65,32 @@ return {
   -- Set this to "NeverPrompt" if you don't like confirming closing
   -- windows every time
   window_close_confirmation = "AlwaysPrompt",
+
+  -- When true (the default), watch the config file and reload it
+  -- automatically when it is detected as changing.
+  -- When false, you will need to manually trigger a config reload.
+  automatically_reload_config = true,
```
wez, line 72: Thanks for picking this up!

My plan was to break this `misc.md` file up so that there is a file per option over in https://github.com/wez/wezterm/tree/main/docs/config/lua/config
That way each option can be linked to from other parts of the docs and it is a little bit easier to search for things.

Would you mind putting these newly documented items into files in that directory instead?

docs/config/misc.md
```diff
@@ -65,6 +65,32 @@ return {
...
+  -- with links to download/install it. It doesn't automatically download
+  -- the release. No data are collected for the wezterm project as part
+  -- of this.
+  -- By default it is checked once every 24 hours.
+  -- Set `check_for_updates` to `false` to disable this completely or set
+  -- `check_for_updates_interval_seconds` for an alternative update interval.
+  -- You can also disable the update window with `show_update_window`.
+  check_for_updates = true,
+  check_for_updates_interval_seconds = 86400,
+  show_update_window = true,
```
bew, line 85: For these 3, do I make a single file for all of them? (like update.md somewhere) or make a file for each? (duplicating some info)

wez: If it's clearer to duplicate (perhaps it's just a couple of lines in two places) then I'd lean towards that, but otherwise, I'd suggest putting the main explanation into one of them and having the other related options briefly describe their purpose and link to the more fleshed out information.

docs/config/lua/config/scroll_to_bottom_on_input.md
```diff
@@ -0,0 +1,3 @@
+# `scroll_to_bottom_on_input`
+
+When true (the default) scrolls to the bottom of the scrollback on input.
```
wez, line 3: I'd say:

> When `true` (the default), the viewport will automatically scroll to the bottom of the scrollback when there is input to the terminal so that you can see what you are typing.

docs/config/misc.md
bew: done!

bew:
Should be good to go for these options

docs/config/lua/config/show_update_window.md
```diff
@@ -0,0 +1,14 @@
+# `show_update_window`
+
+When the automatic update checks detects a new version of Wezterm, this
```
MuhammedZakir, line 3:
```suggestion
When a new version of Wezterm is detected, this
```

or

```
When Wezterm checks for an update and detects a new version, this
```


Because of "checks detects", I had to re-read this part to make sure it isn't a typo. Besides, in the next para, "check_for_updates" is mentioned. So I think we don't need to mention "automatic update checks" here.

PR #465, opened 2021-02-08
Add Ctrl-u in search mode to clear the current search pattern

bew:
This is a small addition to the search mode (until keys for this mode are configurable :eyes:).

Ctrl-u is a common key in shells, vim, readline to clear the input buffer, and I use it often.


wez:
Sounds good; could you add that to the docs too?
https://wezfurlong.org/wezterm/scrollback.html#searching-the-scrollback

wez:
Thanks!

PR #490, opened 2021-02-17
unix: Correctly signal child process on exit

bew:
@p00f reported on Matrix an issue where zsh would not save its ~/.zsh_history file when closing wezterm (which needs to 'close' the shell) but it works when closing the shell (which will then close wezterm).

The reason seems to be because currently wezterm tries to kill the child process when closing, instead of sending a signal to the process to tell it its terminal is closing.

NOTE: For some reason I couldn't reproduce the issue on my Linux, but @p00f confirmed that it worked.

---
On Matrix @wez, you suggested to implement it bit differently:
> I suppose that sending SIGHUP makes sense, however, **that isn't guaranteed to kill an arbitrary program** because it may be ignoring HUP (eg: nohup). The usual approach for gracefully killing things is first to deliver SIGTERM, give it a few moments to handle it and if it hasn't gone away, follow up with SIGKILL. I think in the context of a pty we can do SIGHUP, SIGTERM and then SIGKILL.

Highlighted:
> that isn't guaranteed to kill an arbitrary program

The thing is, from what I read the terminal should NOT kill the program (cf wikipedia extract on SIGUP below) but must simply send the SIGHUP signal to the process. By default that signal kills the process unless it configured a signal handler.

Extract from https://en.wikipedia.org/wiki/SIGHUP#Modern_usage:
> With the decline of access via serial line, the meaning of SIGHUP has changed somewhat on modern systems, often meaning a controlling pseudo or virtual terminal has been closed. If a command is executed inside a terminal window and the terminal window is closed while the command process is still running, it receives SIGHUP.

### DIFF
--- a/pty/src/lib.rs
+++ b/pty/src/lib.rs
@@ -190,6 +190,11 @@ impl Child for std::process::Child {
     fn kill(&mut self) -> IoResult<()> {
         #[cfg(unix)]
         {
+            // On unix, we send the SIGHUP signal instead of trying to kill
+            // the process. The default behavior of a process receiving this
+            // signal is to be killed unless it configured a signal handler.
+            // In this case the process is responsible of itself, releasing
+            // resources and terminating itself, or become a daemon.
             unsafe { libc::kill(self.id() as i32, libc::SIGHUP) };
             Ok(())
         }
--- a/pty/src/lib.rs
+++ b/pty/src/lib.rs
@@ -44,10 +44,10 @@
 //! session with an implementation of `PtySystem`, allowing
 //! you to use the same pty interface with remote ptys.
 use anyhow::Error;
+use libc;
 #[cfg(feature = "serde_support")]
 use serde_derive::*;
 use std::io::Result as IoResult;
-use libc;
 
 pub mod cmdbuilder;
 pub use cmdbuilder::CommandBuilder;
--- a/pty/src/lib.rs
+++ b/pty/src/lib.rs
@@ -47,6 +47,7 @@ use anyhow::Error;
 #[cfg(feature = "serde_support")]
 use serde_derive::*;
 use std::io::Result as IoResult;
+use libc;
 
 pub mod cmdbuilder;
 pub use cmdbuilder::CommandBuilder;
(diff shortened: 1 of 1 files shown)


pty/src/lib.rs
```diff
@@ -187,6 +188,17 @@ impl Child for std::process::Child {
...
 
     fn kill(&mut self) -> IoResult<()> {
+        #[cfg(unix)]
+        {
+            // On unix, we send the SIGHUP signal instead of trying to kill
+            // the process. The default behavior of a process receiving this
+            // signal is to be killed unless it configured a signal handler.
+            // In this case the process is responsible of itself, releasing
+            // resources and terminating itself, or become a daemon.
+            unsafe { libc::kill(self.id() as i32, libc::SIGHUP) };
```
wez, line 198:
```suggestion
            let result = unsafe { libc::kill(self.id() as i32, libc::SIGHUP) };
            if result == 0 {
               Ok(())
            } else {
               Err(IoError::last_os_error())
            }
```

pty/src/lib.rs
```diff
@@ -187,6 +188,17 @@ impl Child for std::process::Child {
     }
 
     fn kill(&mut self) -> IoResult<()> {
+        #[cfg(unix)]
+        {
+            // On unix, we send the SIGHUP signal instead of trying to kill
```
wez, line 193: While sending HUP first makes sense, the semantics of this method is that if it is successful, then it is assumed that the process has been killed.
This is important because calling `.wait()` will block until the process exits; if that process is ignoring HUP then it can potentially wedge the terminal GUI thread:

https://github.com/wez/wezterm/blob/main/mux/src/localpane.rs#L509-L510

The suggestion I made about allowing a grace period before upgrading to a regular kill allows both a chance for the shell to handle SIGHUP and for wezterm to not be held ransom by processes that won't die when we need them to!

I can't think of a good reason for the process directly attached to the pty here to ignore SIGHUP.  *Handling* it makes sense, but ignoring only makes sense for eg: a child process that is effectively a daemon process (spawned via `nohup`, in which case, it won't ever be associated directly with the pty, because `nohup` spawns a child process that detaches from the pty).

PR #512, opened 2021-03-07
windows: Add option to treat left ctrl-alt as altgr

### LINKED ISSUE #472: Windows: All Ctrl-Alt-key incorrectly sends Alt-key
## Describe the bug

TSIA

## Environment

 - OS: Windows (double checked, Linux is ok)
 - Version: 20210203-095643-70a364eb-26-g9bf419eb~

## Configuration

```lua
return {
  -- tested with and without
  use_dead_keys = false,
}
```

## To Reproduce

1. Open an ssh connection to a (linux) server
    _(local shells are also broken w.r.t keys, but I think it's another topic since other terminals have the same behavior)_
2. Open cat
3. Press Ctrl-Alt-z (or any alpha keys)

## Expected behavior

Observe the output `^[z` (Alt-z) instead of `^[^Z` (Alt-Ctrl-z)

bew:
The previous behavior was to always treat ctrl-alt as altgr on Windows,
this has been done to better support altgr through a VNC session,
but this is very unintuitive when you don't need this behavior.

Closes #472

### DIFF
--- a/docs/config/keys.md
+++ b/docs/config/keys.md
@@ -100,6 +100,24 @@ return {
 }
 ```
 
+### Microsoft Windows and Ctrl-Alt <-> AltGr
+
+*since: nightly*
+
+If you are using a layout with an *AltGr* key, you may experience issues
+when running inside a VNC session, because VNC emulates the AltGr keypresses
+by sending plain *Ctrl-Alt* keys, which won't be understood as AltGr.
+
+To fix this behavior you can tell WezTerm to treat left *Ctrl-Alt* keys as
+*AltGr* with the option `treat_left_ctrlalt_as_altgr`. Note that the key bindings using separate Ctrl and Alt won't be triggered anymore.
+
+```lua
+return {
+  treat_left_ctrlalt_as_altgr = true
+}
+```
+
+
 ### Defining Assignments for key combinations that may be composed
 
 When a key combination produces a composed key result, wezterm will look up
--- a/docs/config/keys.md
+++ b/docs/config/keys.md
@@ -102,20 +102,7 @@ return {
 
 ### Microsoft Windows and Ctrl-Alt <-> AltGr
 
-*since: nightly*
-
-If you are using a layout with an *AltGr* key, you may experience issues
-when running inside a VNC session, because VNC emulates the AltGr keypresses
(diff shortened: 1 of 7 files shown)


window/src/os/windows/window.rs
```diff
@@ -1579,14 +1579,20 @@ unsafe fn key(hwnd: HWND, msg: UINT, wparam: WPARAM, lparam: LPARAM) -> Option<L
             // We set RIGHT_ALT as a hint to ourselves that AltGr is in
             // use (we use regular ALT otherwise) so that our dead key
             // resolution can do the right thing.
-            //
+            modifiers |= Modifiers::RIGHT_ALT;
+        } else if config().treat_left_ctrlalt_as_altgr()
```
wez, line 1583: Should also constrain this to layouts with AltGr mappings:

```suggestion
        } else if inner.keyboard_info.has_alt_gr()
 && config().treat_left_ctrlalt_as_altgr()
```

docs/config/keys.md
```diff
@@ -100,6 +100,24 @@ return {
 }
 ```
 
+### Microsoft Windows and Ctrl-Alt <-> AltGr
+
+*since: nightly*
+
+If you are using a layout with an *AltGr* key, you may experience issues
```
wez, line 107: Please also add a `docs/config/lua/config/treat_left_ctrlalt_as_altgr.md` with this information inside it.

It's unfortunate to duplicate the content in two places, and this particularly page is already quite long, so I'd suggest moving this block into that separate page, and then here in this doc you can have a shorter paragraph that says:

> If you are using VNC and a keyboard layout with dead keys, then you may wish to enable \[treat_left_ctrlalt_as_altgr\](lua/config/treat_left_atrlalt_as_altgr.md).

Please also add a line to `docs/changelog.md` summarizing this PR and linking to that new doc page!

wez (changes requested):
Looks good! A couple of minor things to tweak before we merge it!

window/src/os/windows/window.rs
bew: Done

docs/config/keys.md
bew: Done!

PR #576, opened 2021-03-25
update k9 to 0.11.0 to make it work with rustc@1.51.0

aaronabramov:
1.51.0 no longer takes non string literals in `panic()`. k9@0.11.0 fixes that


wez:
Thanks!

Issue #584, opened 2021-03-29
Gaps around block drawing glyphs


JeanMertz:
Using the [vertical bar](https://en.wikipedia.org/wiki/Vertical_bar) (`|`) together with `line_height` other than `1.0` causes spacing between two bars positioned above each other.

Technically, this makes perfect sense. But aesthetically, the use-case for a vertical bar is often to mimic a continuous line from top to bottom in a terminal. For example, in Vim, it's often used to draw a line between two splits.

I used Kitty before, and it renders vertical bars as "expected", even with increased line spacing:

<img width="214" alt="Screenshot 2021-03-29 at 22 20 17" src="https://user-images.githubusercontent.com/383250/112895117-002c2f80-90dd-11eb-8361-2da6f936d2d3.png">

While Wezterm does not:

<img width="265" alt="Screenshot 2021-03-29 at 22 20 29" src="https://user-images.githubusercontent.com/383250/112895126-03bfb680-90dd-11eb-9b8a-57cc10a2e002.png">

(there's also a difference in vertical text positioning between the two, but I filed #582 to track that)

I can see this being closed as "working as intended" because it really is, but I hope that you agree that this could potentially warrant a special case to cover the most common use-case for a character like this.

wez:
Kitty has [special case logic for a variety of box drawing glyphs](https://github.com/kovidgoyal/kitty/blob/master/kitty/fonts/box_drawing.py#L768).  So far, wezterm only does that for the "Block Elements" unicode range (https://github.com/wez/wezterm/issues/433), but I think it is reasonable to expand the scope to some or all of https://unicode-table.com/en/blocks/box-drawing/

JeanMertz:
> but I think it is reasonable to expand the scope to some or all of [unicode-table.com/en/blocks/box-drawing](https://unicode-table.com/en/blocks/box-drawing/)

Thanks for the pointer.

It appears the character in my use-case is the  [Box Drawings Light Vertical](https://unicode-table.com/en/2502/), so that would indeed be captured by special-casing the "box drawing" characters.

GladOSkar:
The same applies to powerline characters btw.

Config for both `wezterm` and `kitty`:
```
font_size = 9
line_height = 1.4
```

`wezterm`:
![image](https://user-images.githubusercontent.com/9497832/113486373-a8157480-94b2-11eb-96f7-b90597653ea5.png)

`kitty`:
[...]

PR #588, opened 2021-03-30
custom Unicode box drawing glyphs

JeanMertz:
This is an initial push towards solving #584.

It is unlikely we'll want to merge the PR as is, since it:

- misses a bunch of box drawing characters (currently only `light vertical` is implemented)
- duplicates a lot of "block drawing" logic that we should probably merge
- doesn't include any tests

I figured I'd push this up regardless, as it solved an immediate need for my usage, and I am running this fork for now, until there is more guidance on the preferred way forward with this feature.

Here's an example with custom `line_height` and the vertical line going all the way to the edges of the line:

Screenshot 2021-03-30 at 10 25 24

### DIFF
--- a/wezterm-gui/src/glyphcache.rs
+++ b/wezterm-gui/src/glyphcache.rs
@@ -219,6 +219,36 @@ impl BlockKey {
     }
 }
 
+/// Represents a Box Drawing glyph, decoded from
+/// <https://en.wikipedia.org/wiki/Box_Drawing_(Unicode_block)>
+/// <https://www.unicode.org/charts/PDF/U2500.pdf>
+#[derive(Copy, Clone, PartialEq, Eq, Hash, Debug)]
+pub enum BoxDrawingKey {
+    LightVertical,
+}
+
+impl BoxDrawingKey {
+    pub fn from_char(c: char) -> Option<Self> {
+        use BoxDrawingKey::*;
+
+        let c = c as u32;
+        Some(match c {
+            0x2502 => LightVertical,
+            _ => return None,
+        })
+    }
+
+    pub fn from_cell(cell: &termwiz::cell::Cell) -> Option<Self> {
+        let mut chars = cell.str().chars();
+        let first_char = chars.next()?;
+        if chars.next().is_some() {
+            None
+        } else {
+            Self::from_char(first_char)
+        }
+    }
+}
+
 #[derive(Debug)]
 pub struct ImageFrame {
     duration: Duration,
@@ -324,6 +354,7 @@ pub struct GlyphCache<T: Texture2d> {
     frame_cache: HashMap<(usize, usize), Sprite<T>>,
     line_glyphs: HashMap<LineKey, Sprite<T>>,
     block_glyphs: HashMap<BlockKey, Sprite<T>>,
+    box_drawing_glyphs: HashMap<BoxDrawingKey, Sprite<T>>,
     metrics: RenderMetrics,
 }
 
@@ -346,6 +377,7 @@ impl GlyphCache<ImageTexture> {
             metrics: metrics.clone(),
             line_glyphs: HashMap::new(),
             block_glyphs: HashMap::new(),
+            box_drawing_glyphs: HashMap::new(),
         })
     }
 }
@@ -375,6 +407,7 @@ impl GlyphCache<SrgbTexture2d> {
             metrics: metrics.clone(),
             line_glyphs: HashMap::new(),
             block_glyphs: HashMap::new(),
+            box_drawing_glyphs: HashMap::new(),
         })
     }
 
@@ -385,6 +418,7 @@ impl GlyphCache<SrgbTexture2d> {
         self.glyph_cache.clear();
         self.line_glyphs.clear();
         self.block_glyphs.clear();
+        self.box_drawing_glyphs.clear();
     }
 }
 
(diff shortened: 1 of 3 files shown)


wezterm-gui/src/glyphcache.rs
```diff
@@ -745,6 +779,46 @@ impl<T: Texture2d> GlyphCache<T> {
         self.block_sprite(block)
     }
 
+    fn box_drawing_sprite(&mut self, box_drawing: BoxDrawingKey) -> anyhow::Result<Sprite<T>> {
```
JeanMertz, line 782: This is a stripped version of `fn block_sprite`.

I can see two ways forward to merge the two:

1. Nest `BlockKey` into `BoxDrawingKey` as they serve the same purpose and use the same logic. Having said that, it does appear that box drawing and block elements are distinct from each other in the Unicode specification, so this might not be what we'd want.
2. Add a top-level `CustomDrawingGlyph` (or differently named) enum that contains both `BoxDrawing` and `Block`, which would allow us to have a single `fn custom_drawing_sprite` to avoid the duplication.

There are other ways to do this, but I didn't want to go in any specific direction yet without knowing your preference.

wezterm-gui/src/glyphcache.rs
```diff
@@ -745,6 +779,46 @@ impl<T: Texture2d> GlyphCache<T> {
...
+                    cell_rect.origin.x + x as isize,
+                    cell_rect.origin.y + self.metrics.cell_size.height,
+                ),
+                white,
+            );
+        };
+
+        use BoxDrawingKey::*;
+        match box_drawing {
+            LightVertical => draw_vertical(&mut buffer, 10),
```
JeanMertz, line 807: `10` here was arbitrarily chosen as a test-case. I wonder if there's (semi-)official documentation anywhere on the exact details on how to draw the different characters?

wezterm-gui/src/termwindow/render.rs
```diff
@@ -1144,6 +1163,51 @@ impl super::TermWindow {
...
+        cursor_shape: Option<CursorShape>,
+        glyph_color: LinearRgba,
+        underline_color: LinearRgba,
+        bg_color: LinearRgba,
+        white_space: TextureRect,
+    ) -> anyhow::Result<()> {
+        let sprite = gl_state
+            .glyph_cache
+            .borrow_mut()
+            .cached_box_drawing(box_drawing)?
```
JeanMertz, line 1183: This function is 99% duplicated from `populate_block_quad`, with the exception being this call here, so there's plenty of room to DRY up this logic.

wezterm-gui/src/glyphcache.rs
wez: I'd vote for adding renaming `BlockKey` `CustomGlyphKey` and then adding `LightVertical` to that enum.

wezterm-gui/src/glyphcache.rs
wez: There don't appear to be guidelines on exactly how thick the "light" variant is vs. the "heavy" variant.

The font metrics do tell us the desired height of an underline (take a look in the `line_sprite` function below: `self.metrics.underline_height`).

We also know the aspect ratio of the glyph (width-to-height ratio), so I'd suggest using the `underline_height` to compute a vertical bar thickness for the "light" variant, then eg: double that for the heavy variant.

wez (commented):
Thanks for taking a run at implementing this!
I think merging these changes into the existing block drawing code will satisfy the DRY ideal!

wezterm-gui/src/glyphcache.rs
wez: (and then center that line in the box)

JeanMertz:
@wez I pushed some changes. Happy to hear what you think.

Here's how it currently looks:

Screenshot 2021-03-31 at 07 58 55

The test data could use an extra pass to make it more useful (i.e. render multiple subsequent horizontal lines), but I figured this was a good enough start.

wezterm-gui/src/glyphcache.rs
```diff
@@ -733,16 +784,88 @@ impl<T: Texture2d> GlyphCache<T> {
...
+            for x in rect.min.x..rect.max.x {
+                for y in rect.min.y..rect.max.y {
+                    let pixel = buffer.pixel_mut(x, y);
+                    *pixel = white.as_srgba32();
+                }
+            }
+        };
+
+        let center = cell_rect.center();
+        let light_thickness = self.metrics.underline_height as usize;
```
wez, line 829: I think that it may make sense to aspect correct this; most monospace fonts are taller than they are wide, so a height metric should probably be multiplied by `width / height` in floating point to convert to a width metric, and then have a min size of 1 pixel.  Eg: `(underline_height * width / height).min(1)`.  For the double thickness, multiply by 2 before converting to integer

wez (commented):
Just a quick comment; I'm looking at this from my iPad before calling it a night, so this isn't a thorough review!

wezterm-gui/src/glyphcache.rs
JeanMertz: > most monospace fonts are taller than they are wide

That's true, but are you convinced that applies in this case? The intent of these characters is to _draw boxes_ on the screen, it would seem weird to me to have a vertical line be wider than a horizontal one, just because its length is also taller?

That is to say, if I were to draw a box using:

```
┌─┐
│ │
└─┘
```

I'd expect the width of the line to not differ between the vertical and horizontal lines (it would also mean the corners would have to draw their horizontal and vertical parts differently, which makes them not conform to the Unicode standard).

Unless your point is to have _all_ lines be drawn using the aspect ratio, which we can do, but then a horizontal line will be less thick than an underline, not sure if that matters?

JeanMertz: (Unless drawing actual underlines also already takes the aspect ratio in account?)

wez: if the font specifies that the underline height is 2px, and the cell metrics are 8px wide and 16px tall: if you make the vertical lines 2px wide then they will appear too wide.  Aspect correcting them will make them have the correct proportions, which in this case should be 1px wide (`2px * 8px / 16px`).

If you set the font size to something large (eg: 144 or even double that) then this difference will be easier to see.

JeanMertz:
Quick update that I've been using this PR (and rebasing it locally on `main`) for a while now, I just haven't found the time yet to finish it up to be merge-able.

wez:
Nightly builds now synthesize the full box drawing range of unicode

PR #599, opened 2021-03-31
Capture middle mouse events on macos

guswynn:
I noticed that no matter what I did, even trying multiple mouses, I couldn't get middle-mouse-paste to work on my macos version of wezterm

Looking through the code, I found it was missing bindings to `otherMouse*` events. I added them, and lo and behold, it worked!

One question: I dont really understand the difference between the press type in `MouseEventKind` and the `buttons` in `MouseEvent`.
This code causes the middle mouse click to happen when I click ANY of the non left/right buttons on my mouse. I am not sure if thats what we want (until/if we add button bindings would change this), or if not, how I should structure the code to avoid that. I suppose we'd have to compare the decoded mouse_buttons with the MouseEventKind in `mouse_common`?

### DIFF
--- a/window/src/os/macos/window.rs
+++ b/window/src/os/macos/window.rs
@@ -1501,9 +1501,8 @@ impl WindowView {
         unsafe {
             let button_number = NSEvent::buttonNumber(nsevent);
             // Button 2 is the middle mouse button (scroll wheel)
-            // At least on the tested mouse, the dedicated middle mouse button
-            // is button 4. This should probably be made configurable at some point
-            if [2, 4].contains(&button_number) {
+            // but is the dedicated middle mouse button on 4 button mouses
+            if button_number == 2 {
                 Self::mouse_common(this, nsevent, MouseEventKind::Release(MousePress::Middle));
             }
         }
@@ -1581,7 +1580,7 @@ impl WindowView {
         unsafe {
             let button_number = NSEvent::buttonNumber(nsevent);
             // See `other_mouse_up`
-            if [2, 4].contains(&button_number) {
+            if button_number == 2 {
                 Self::mouse_common(this, nsevent, MouseEventKind::Press(MousePress::Middle));
             }
         }
--- a/window/src/os/macos/window.rs
+++ b/window/src/os/macos/window.rs
@@ -1577,7 +1577,14 @@ impl WindowView {
     }
 
     extern "C" fn other_mouse_down(this: &mut Object, _sel: Sel, nsevent: id) {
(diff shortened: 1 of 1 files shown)


window/src/os/macos/window.rs
```diff
@@ -1496,6 +1496,10 @@ impl WindowView {
         Self::mouse_common(this, nsevent, MouseEventKind::Release(MousePress::Right));
     }
 
+    extern "C" fn other_mouse_up(this: &mut Object, _sel: Sel, nsevent: id) {
+        Self::mouse_common(this, nsevent, MouseEventKind::Release(MousePress::Middle));
```
wez, line 1500: regarding not triggering for non-middle buttons, I think you want something like:

```rust
let button_number = NSEvent::buttonNumber(nsevent);
// if the above doesn't work, do it the hard way: let button_number: isize = msg_send![nsevent, buttonNumber];
if button_number == 1 {
   Self::mouse_common(this, nsevent, MouseEventKind::Release(MousePress::Middle));
}
```

I'm not sure if the middle mouse button is 1 or not, but you can `log::info!("button {}", button_number);` to see what numbers show up.

wez (commented):
>  I dont really understand the difference between the press type in MouseEventKind and the buttons in MouseEvent.

Ah, yeah, it's a bit clunky.  `MouseButtons` is a mask showing which buttons are down (like a modifier mask), but `MousePress` is used to track a discrete individual button press event as part of the streak (double, triple, quadruple etc.) click counting.

macOS itself has a similar distinction; `NSEvent::buttonNumber` vs. `NSEvent::pressedMouseButtons`

I suppose that `MousePress` should really also have `X1` and `X2` enum variants if we want to support non-middle other buttons in configurable mouse bindings.

guswynn:
On my mouse, button 2 is the scroll wheel click, button 4 is the dedicated middle mouse button. On iterm2 only button 4 triggers the default middle mouse button binding.
I am inclined to leave this as a catch all for "middle mouse" buttons, and then in a later pr add fine-grained configuration for which button numbers map to which buttons (possibly when adding support for X1 etc. buttons) but generalizing that across platforms may be non-trivial

guswynn:
Wait never mind, I must have misread something, button == 2 should work as it matches the behavior in other programs (chrome, iterm2)

guswynn:
On my 4 button mouse, the scroll wheel is one of the X buttons I think (button 4), i think the only reasonable thing to do here is just record button 2 (until we have X buttons, though at that point we may want to have more than 2 X buttons), sorry for the noise, the pr should be ready!

PR #603, opened 2021-04-01
wezterm-gui: allow title prefixes to be hidden

LawnGnome:
Another option here would be to implement a simple templating system for the window title, but selfishly, this does what I'm looking for in terms of just disabling the `[count/total]` prefix.

### DIFF
--- a/config/src/lib.rs
+++ b/config/src/lib.rs
@@ -879,6 +879,10 @@ pub struct Config {
     #[serde(default)]
     pub hide_tab_bar_if_only_one_tab: bool,
 
+    /// If true, hide all prefixes in the window title.
+    #[serde(default)]
+    pub hide_window_title_prefixes: bool,
+
     #[serde(default)]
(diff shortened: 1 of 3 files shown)


wez:
Thanks for this!  I'm a little on the fence about it: I have been thinking of providing a way to configure what shows up in the title bar to give more control, but have been punting until someone asked for it.  Let me sleep on this a bit: if it will take too much effort to do the more configurable approach in a reasonable time then I'm ok with this change, but otherwise, I'd like to avoid adding a new option only to remove it very soon afterwards!

wez:
I'll be giving this some thought, together with https://github.com/wez/wezterm/issues/647 this weekend

wez:
Nightly builds now have a [format-window-title](https://wezfurlong.org/wezterm/config/lua/window-events/format-window-title.html) event that gives you a lot more control over this!

PR #610, opened 2021-04-01
Add ActivateLastTab command

alexgartrell:
This replicates `last-window` in tmux. To pull this off, I
deliberately store the last tab whenever I'm activating a new one or
spawning a new one. I had to do this explicitly rather than hooking
set_active, because we end up setting the active tab briefly for some
common operations like moving a tab.

### DIFF
--- a/docs/config/lua/keyassignment/ActivateLastTab.md
+++ b/docs/config/lua/keyassignment/ActivateLastTab.md
@@ -1,0 +1,14 @@
+# ActivateLastTab
+
+Activate the last active tab. If there is none, it will do nothing.
+
+
+```lua
+return {
+  keys = {
+    {key="o", mods="LEADER|CTRL", action="ActivateLastTab"},
+  }
+}
+```
+
+
--- a/mux/src/window.rs
+++ b/mux/src/window.rs
@@ -10,6 +10,7 @@ pub struct Window {
     id: WindowId,
     tabs: Vec<Rc<Tab>>,
     active: usize,
+    last_active: Option<TabId>,
     clipboard: Option<Arc<dyn Clipboard>>,
     invalidated: bool,
 }
@@ -20,6 +21,7 @@ impl Window {
             id: WIN_ID.fetch_add(1, ::std::sync::atomic::Ordering::Relaxed),
             tabs: vec![],
             active: 0,
+            last_active: None,
             clipboard: None,
             invalidated: false,
         }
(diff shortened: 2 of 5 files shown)


bew:
I'd like to suggest `ActivateLastTab`, because Previous looks like to opposite of Next,which is not right.

alexgartrell:
@bew  great point. It should have occurred to me when I reminded myself of how tmux did it :)

docs/config/lua/keyassignment/ActivateLastTab.md
```diff
@@ -0,0 +1,16 @@
+# ActivateLastTab
+
+Activate the last active tab index. If there is none, it will do
+nothing. If the tabs get moved around, it will move to the last
+tab index rather than the last tab itself.
```
wez, line 5: is that the desired behavior, or just simple to implement behavior?
If desired, then I'd suggest adding a brief example of what that means if the user reorganizes.

However, I think it isn't too much extra complexity to switch to the prior tab using its tabid with the inline suggestions!

mux/src/window.rs
```diff
@@ -10,6 +10,7 @@ pub struct Window {
     id: WindowId,
     tabs: Vec<Rc<Tab>>,
     active: usize,
+    last_active: Option<usize>,
```
wez, line 13:
```suggestion
    last_active: Option<TabId>,
```

mux/src/window.rs
```diff
@@ -124,9 +126,15 @@ impl Window {
...
 
+    #[inline]
+    pub fn get_last_active_idx(&self) -> Option<usize> {
+        self.last_active
+    }
+
     pub fn set_active(&mut self, idx: usize) {
         assert!(idx < self.tabs.len());
         self.invalidated = true;
+        self.last_active = Some(self.active);
```
wez, line 137:
```suggestion
        self.last_active = self.get_by_idx(self.active).map(|tab| tab.tab_id());
 ```

File: mux/src/window.rs [hunk 4]
```diff
@@ -124,9 +126,15 @@ impl Window {
         self.active
     }
 
+    #[inline]
+    pub fn get_last_active_idx(&self) -> Option<usize> {
+        self.last_active
```
wez, line 131:
```suggestion
        self.last_active.map(|tab_id| self.idx_by_id(tab_id))
```

wez (commented):
Thanks for this!
Some suggestions inline!

alexgartrell:
The issue is that move_tab removes and reinserts, so set_active gets called multiple times. Plus if you activate the same tab twice, it'll be its own "last_active." I'll see if I can figure out a way to get the ideal behavior that's not overly gross though.

wez:
Thanks for this!  I tweaked it a little because it felt like I could easily forget to save the tab if I changed some code later, and I found a single omission in the [Tab Navigator](https://wezfurlong.org/wezterm/config/lua/keyassignment/ShowTabNavigator.html).

Thanks again!

PR #620, opened 2021-04-02
move sudo into script to not use sudo for rust checking

xunnanxu:
Rust is typically installed per user rather than globally and hence when sudo is used typically either the executable is not found or a wrong version is used instead.

Update the docs

Test:
Ran `./get-deps` and verified rust version check passed and the package manager was executed with sudo privilige

### DIFF
--- a/get-deps
+++ b/get-deps
@@ -51,12 +51,16 @@ if ! check_rust_version ; then
   exit 1
 fi
 
+if test -z "${SUDO+x}" && hash sudo 2>/dev/null; then
+  SUDO="sudo"
+fi
+
 # Centos may not have lsb_release installed
 if test -e /etc/centos-release || test -e /etc/fedora-release; then
   if test -x /bin/dnf ; then
-    YUM=dnf
+    YUM="$SUDO dnf"
   else
-    YUM=yum
+    YUM="$SUDO yum"
   fi
   # Fedora 33 moved some perl bits around
   $YUM install -y perl-FindBin perl-File-Compare || true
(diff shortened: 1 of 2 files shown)


wez:
Thanks for contributing!
This won't work as written because a number of the CI environments either don't have sudo at all, or won't have the ability to run sudo.

Can you share more about the issue you want to resolve with this change?

xunnanxu:
Hi @wez so the script was updated in 8a40241 to run `rustc` to check the version. However this would break in systems where user has a different version of rust than the system one (which may be lower like 1.48) or system doesn't have anything at all as the PATH is per user only via rustup. In that case, rustc should be invoked in original user (as that's the one used in later actual compilation) rather than root.

xunnanxu:
And actually after that commit trunk is broken (you see the error in CI build).

get-deps
```diff
@@ -51,12 +51,14 @@ if ! check_rust_version ; then
   exit 1
 fi
 
+SUDO=sudo
```
wez, line 54: I'd suggest doing something like this, which should allow the caller to specify what `SUDO` expands to, otherwise, only sets it if sudo is present in the path.  That might be sufficient to fix the human use case and the CI use case

```
if test -z "${SUDO+x}" ; then
   if hash sudo 2>/dev/null ; then
      SUDO="sudo"
   fi
fi
```

wez (changes requested):
back to you for now!

get-deps
xunnanxu: Thanks. I've updated the PR. Please take a look.

wez:
Thanks!

PR #663, opened 2021-04-08
Support unicode charaters in `selection_word_boundary`

guswynn:
I want to support https://unicodelookup.com/#%E2%94%82/1 (this is used as the tmux pane separator)
This is marginally slower, and it only supports "Unicode Scalar Value"s (rust char) and not graphemes, but it should be a start!

### DIFF
--- a/wezterm-gui/src/selection.rs
+++ b/wezterm-gui/src/selection.rs
@@ -52,7 +52,7 @@ pub struct SelectionRange {
 }
 
 fn is_double_click_word(s: &str) -> bool {
-    match s.len() {
+    match s.chars().count() {
         1 => !config::configuration().selection_word_boundary.contains(s),
         0 => false,
         _ => true,


wezterm-gui/src/selection.rs
```diff
@@ -52,7 +52,7 @@ pub struct SelectionRange {
 }
 
 fn is_double_click_word(s: &str) -> bool {
-    match s.len() {
+    match s.chars().count() {
```
wez, line 55: if you change line 8 to:

```rust
use wezterm_term::{SemanticZone, StableRowIndex, unicode_column_width};
```

then I think this can be changed and support graphemes
```suggestion
    match unicode_column_width(s) {
```

wez (commented):
Thanks for this!

Do you want to try making this work with graphemes?  Suggested changes inline!

wezterm-gui/src/selection.rs
wez: although my suggested changes might not be sufficient: not sure what it does for eg: double-width graphemes or longer combining sequences.
If it doesn't obviously work, then I'll take you original PR :)

guswynn:
@wez I tried your suggestion with நி, which appeared to be 2 columns, and even without it in my word boundary, selection appeared to only select half of it. Therefore, I think getting large graphemes to work with selection in general may be more difficult, so we can probably stick with `chars` until we have a better way of handling graphemes for selection in general (I think https://www.unicode.org/reports/tr29/tr29-37.html is the relevant paper)

wez:
Thanks!

PR #665, opened 2021-04-09
termwiz: do not use terminfo for unsupported 256-colors

quark-zju:
I noticed edenscm log -p (using streampager) cannot render 256 colors
if TERM is not "*-256color", despite forcing enabling true colors in
streampager (https://github.com/markbt/streampager/pull/28).

I tracked it down here. The problem is that we ask terminfo for colors
it does not claim to support. Fix it by using fallback CSI rendering
for colors exceeding the terminfo max color.


wez:
Looks good, thanks!

PR #676, opened 2021-04-11
Fix: copy PrimarySelection to Clipboard on Windows

zoritle:
Windows dose not have `PrimarySelection`, copy anything selected will mass up clipboard.


wez:
Would you mind explaining a bit more about the issue you're trying to solve?
It is intentional that selecting text populates the clipboard on Windows.

zoritle:
> 
> 
> Would you mind explaining a bit more about the issue you're trying to solve?
> It is intentional that selecting text populates the clipboard on Windows.

This is not a common behavior on Windows. Automatically copying selected text to clipboard is usually not a Windows user intended. If you copy some thing, and then subconsciously select some text, your clipboard got overwritten, since there are no concept of `PrimarySelection`. And if you are using a clipboard history manager(like ditto or Win+V),it will mess up the history.

Most Windows user are habituated to copy/paste explicitly, since it's all other software's behavior. Or copy by right mouse click after selection, which is the behavior of default Windows console.

window/src/os/windows/window.rs
```diff
@@ -683,10 +683,15 @@ impl WindowOps for Window {
         )
     }
 
-    fn set_clipboard(&self, _clipboard: Clipboard, text: String) -> Future<()> {
-        Future::result(
-            clipboard_win::set_clipboard_string(&text).context("Error setting clipboard"),
-        )
+    fn set_clipboard(&self, clipboard: Clipboard, text: String) -> Future<()> {
```
wez, line 686: How about introducing a new configuration option named something like `treat_primary_selection_as_clipboard_on_windows` that controls this?
I want the default to be be `true`, but it would allow you to set it to false
to get the behavior that you want.

You can add the option in a similar way to this:
https://github.com/wez/wezterm/blob/main/config/src/lib.rs#L666-L667

Then to access the config, something like this:

```rust
fn set_clipboard(&self, clipboard: Clipboard, text: String) -> Future<()> {
    Connection::with_window_inner(self.0, move |inner| {
        if matches!(clipboard, Clipboard::Clipboard) || inner.config.treat_primary_selection_as_clipboard_on_windows {
         clipboard_win::set_clipboard_string(&text).context("Error setting clipboard"),
       } else {
         Ok(())
       }
    })
}

               
```

zoritle: I think this is redundant. There's already a property https://wezfurlong.org/wezterm/config/lua/keyassignment/CopyTo.html to set the copy destination, if someone would like selection automatically copied they can just set the destination to be `Clipboard` or `ClipboardAndPrimarySelection`. And if you want this behavior to be default, you can make the mousebinding default to these value as well.
Since the `PrimarySelection` is defined and only supported on the X platform (probably wayland i'm not sure), it's technically correct to ignore this operation on unsupported platform, and, if copy to clipboard is expected, set it (or default) to clipboard is more intuitive.

bew:
What about adding an `InternalClipboard` or `InternalSelection`, that would mimic the behavior of the `PrimarySelection` from Linux, except it only works inside wezterm?
This way selecting and pasting in wezterm would work, without overwriting the clipboard, unless you explicitly use `CopyTo="Clipboard"`.

The default would still be to copy to clipboard on selection, to match the other Windows terminals (as far as I remember), but this would allow the people that don't want that behavior to still have an internal selection accessible when on wezterm, and the ability to copy explicitly in the clipboard on demand, if they so wish

stappersg:
Breaking now **four years of silence** for requesting to close this stalled merge request.

bew:
It's an old one for sure.
If this is still something someone wants, Ithink it'd be better to make an issue first to discuss about it.

PR #720, opened 2021-04-23
conditionally include process divination for quit-confirm check on an…

dmfay:
…y win/linux/osx instead of all

I've always gotten the "Really kill this window...?" confirmation even with nothing open, and on looking in the docs and finding `skip_close_confirmation_for_processes_named` nothing I tried worked. I think it's just using the wrong predicate in `divine_process_list`; this behaves as expected on Linux.


wez:
Good catch... it's sort of hilarious that this ended up as `all` rather than `any` :-p

PR #733, opened 2021-04-25
Better function for undercurl

ModProg:
This improves the rendering of the undercurl.

Before:
image

After:
image

All that is changed is the wave generation function.

What I noticed is, that the antialiasing is broken, therefor I removed
the second pixel draw, as that made the curl twice as thick as the
other lines.

### DIFF
--- a/wezterm-gui/src/glyphcache.rs
+++ b/wezterm-gui/src/glyphcache.rs
@@ -899,7 +899,7 @@ impl<T: Texture2d> GlyphCache<T> {
 
                 for row in 0..self.metrics.underline_height as usize {
                     let value = (255. * (vertical - v).abs()) as u8;
-                    add(x, row + y + v as usize, 0x10, max_y, buffer);
+                    add(x, row + y + v as usize, 0x88, max_y, buffer);
                 }
             }
             println!("{:?}", *buffer);
--- a/wezterm-gui/src/glyphcache.rs
+++ b/wezterm-gui/src/glyphcache.rs
@@ -889,7 +889,7 @@ impl<T: Texture2d> GlyphCache<T> {
                 let pixel = buffer.pixel_mut(x, y);
                 let (current, _, _, _) = SrgbaPixel::with_srgba_u32(*pixel).as_rgba();
                 let value = current.saturating_add(val);
-                *pixel = SrgbaPixel::rgba(value, value, value, value).as_srgba32();
+                *pixel = SrgbaPixel::rgba(0xFF, 0xFF, 0xFF, val).as_srgba32();
             }
 
             for x in 0..self.metrics.cell_size.width as usize {
--- a/wezterm-gui/src/glyphcache.rs
+++ b/wezterm-gui/src/glyphcache.rs
@@ -893,14 +893,13 @@ impl<T: Texture2d> GlyphCache<T> {
             }
 
             for x in 0..self.metrics.cell_size.width as usize {
-                let vertical = wave_height as f32 * (x as f32 * x_factor).cos();
-                let v1 = vertical.floor();
-                let v2 = vertical.ceil();
+                let vertical = wave_height as f32 / -3. * (x as f32 * x_factor).sin()
+                    + wave_height as f32 / 2.;
+                let v = vertical.round();
 
                 for row in 0..self.metrics.underline_height as usize {
-                    let value = (255. * (vertical - v1).abs()) as u8;
-                    add(x, row + y + v1 as usize, 255 - value, max_y, buffer);
-                    add(x, row + y + v2 as usize, value, max_y, buffer);
+                    let value = (255. * (vertical - v).abs()) as u8;
+                    add(x, row + y + v as usize, 255 - value, max_y, buffer);
                 }
             }
         };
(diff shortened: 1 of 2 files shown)


wez:
Thanks for diving in to this!

image

For my font size, the old rendering (top) looks a bit better because it looks more wavy; the revised code seems to be a bit too shallow which makes the wave a bit too subtle IMO; it's hard to distinguish between curl and dashed underlines.

Would you mind tweaking this a bit more? I think the amplitude needs a bit more exaggeration.

ModProg:
will do, any idea on how to get the Antialiasing working? that could help get curvyness across.

Maybe the vertical scale could also be a little dependent on the space (like the double line that gets closer to the letters).

wez:
For anti-aliasing, the `val` parameter to that inline `add` function is intended to do that job; you should be able to set it to any greyscale value from `0` to `0xff`.  So it should be just a matter of computing a reasonable grey value for the anti-aliasing.  There's probably some literature that will give a good suggestion on how to do that properly--I'm not really a computer graphics expert and I may have mis-transcribed the original code from some example I found on the internet.

It may make sense to also set the alpha value to `val`; currently it looks like it is hard coded to `0xff`.

ModProg:
It does not seam to make a difference setting the alpha to a lower number (see commit)
0xFF:
grafik
0x10:
grafik
0x10 without setting color:
grafik

ModProg:
I found the resposible lines of code (see last commit). Looking at them I wondered if that value should be a float instead of a color, as it containes the same value 4 times
grafik

wezterm-gui/src/glyphcache.rs
```diff
@@ -889,20 +889,21 @@ impl<T: Texture2d> GlyphCache<T> {
                 let pixel = buffer.pixel_mut(x, y);
                 let (current, _, _, _) = SrgbaPixel::with_srgba_u32(*pixel).as_rgba();
                 let value = current.saturating_add(val);
-                *pixel = SrgbaPixel::rgba(value, value, value, 0xff).as_srgba32();
+                *pixel = SrgbaPixel::rgba(0xFF, 0xFF, 0xFF, val).as_srgba32();
```
wez, line 892:
```suggestion
                *pixel = SrgbaPixel::rgba(value, value, value, value).as_srgba32();
```

I think we probably want this, which is closer to the original logic, but also sets the alpha channel value as a pre-multiplied value

wezterm-gui/src/glyphcache.rs
```diff
@@ -889,20 +889,21 @@ impl<T: Texture2d> GlyphCache<T> {
...
+                let vertical = wave_height as f32 / -3. * (x as f32 * x_factor).sin()
+                    + wave_height as f32 / 2.;
+                let v = vertical.round();
 
                 for row in 0..self.metrics.underline_height as usize {
-                    let value = (255. * (vertical - v1).abs()) as u8;
-                    add(x, row + y + v1 as usize, 255 - value, max_y, buffer);
-                    add(x, row + y + v2 as usize, value, max_y, buffer);
+                    let value = (255. * (vertical - v).abs()) as u8;
+                    add(x, row + y + v as usize, 0x88, max_y, buffer);
```
wez, line 902:
```suggestion
                    add(x, row + y + v as usize, value, max_y, buffer);
```

passing value should give some amount a grey-level control and an opportunity for AA.

I think the original secondary `add` call might be good to restore now that you've found that the alpha value wasn't respected in the shader.  That second `add` with `255 - value` appears to be the thing that computes the anti-aliased pixel value.

wezterm-gui/src/line-frag.glsl
```diff
@@ -25,7 +25,7 @@ void main() {
     // if the underline glyph isn't transparent in this position then
     // we take the underline color, otherwise we'll leave the color
     // at the background color.
-    color = o_underline_color;
+    color = under_color * o_underline_color;
```
wez, line 28: Ah, that's a good find!

wezterm-gui/src/line-frag.glsl
```diff
@@ -23,9 +23,9 @@ void main() {
   vec4 under_color = sample_texture(atlas_nearest_sampler, o_underline);
   if (under_color.a != 0.0) {
     // if the underline glyph isn't transparent in this position then
-    // we take the underline color, otherwise we'll leave the color
+    // we overlay the underline color, otherwise we'll leave the color
     // at the background color.
-    color = o_underline_color;
+    color = color * (1.0 - under_color) + under_color * o_underline_color;
```
wez, line 28: My original intent for this was to draw the underline color over the top of the text if the underline was set, but the implementation was a bit wonky as you've discovered.

Can you share a screenshot of how this looks?  It may not be strictly required; for example, an alternative to this is `color = under_color` to take all of the rgba information from your computed underline glyph.

ModProg: The reasoning was, that if the Underline is drawn over the cursor, the half transparent pixels should be rendered over the cursor. I'm not sure if this actually  is what this results in. Will double check later, and send you a screenshot.

ModProg: grafik

The main thing that need's to be improved now is just thickness, through a better rasterization. 

Would it make sense to look into the fontrenderer, and maybe delegate this to it as well?

wez: We can't pass this sort of data into freetype, at least, not without defining a glyph in a font file.
I think this looks good.

I'm going to merge this, but adjust this line to be `color = under_color * o_underline_color;` as I don't think the more complex math is required; the prior color is the background color and it's fine to just stamp the underline color over the top and make the shader a bit cheaper to execute.

wez:
Thanks!

PR #754, opened 2021-05-01
Bump `libz-sys` to 1.1.3 to make wezterm build on systems which use musl libc

Logarithmus:
See https://github.com/rust-lang/libz-sys/pull/78 for more context.


wez:
Sounds good; thanks!

PR #759, opened 2021-05-04
Allow VoidSymbol keycode to be seen by wezterm

digitallyserviced:
When using `setxkbmap` to disable the functionality of CapsLock, it shows as `0x00FFFFFF VoidSymbol` keycode. That was not being picked up by wezterm.

These changes allow it to be picked up as VoidSymbol so the CapsLock key can be used without any unintended side-effects. 

This is how `xev` and other applications see it once disabled and could even be used when disabling other keys to have them be used similarly.

Makes for a great LEADER independent from anything else without going shift crazy.

### DIFF
--- a/termwiz/src/input.rs
+++ b/termwiz/src/input.rs
@@ -86,6 +86,12 @@ pub struct KeyEvent {
 #[cfg_attr(feature = "use_serde", derive(Serialize, Deserialize))]
 #[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
 pub enum KeyCode {
+    // other keys when disabling its original use will show up as this keycode
+    // e.g. setxkbmap -option caps:none 
+    // prevents the key from operating as originally intended and
+    // makes it possible to utilize it as a leader/modifier with no oddities
+    VoidSymbol,
+
     /// The decoded unicode character
     Char(char),
 
@@ -274,6 +280,9 @@ mod windows {
                 Some(unicode) if unicode > '\x00' => {
                     let mut buf = [0u8; 4];
                     self.buf
+                        .extend_with(unicode.encode_utf8(&mut buf).as_bytes());
+                    self.process_bytes(callback, true);
+                    return;
                         .extend_with(unicode.encode_utf8(&mut buf).as_bytes());
                     self.process_bytes(callback, true);
                     return;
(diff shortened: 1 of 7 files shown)


termwiz/src/input.rs
```diff
@@ -86,6 +86,12 @@ pub struct KeyEvent {
 #[cfg_attr(feature = "use_serde", derive(Serialize, Deserialize))]
 #[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
 pub enum KeyCode {
+    // other keys when disabling its original use will show up as this keycode
+    // e.g. setxkbmap -option caps:none 
+    // prevents the key from operating as originally intended and
+    // makes it possible to utilize it as a leader/modifier with no oddities
+    VoidSymbol,
```
wez, line 93: I don't think there is a way to represent this for the terminal, so I'd be inclined to remove it from this particular enum and the corresponding matches in the termwiz/term crates

wezterm-gui/src/termwindow/keyevent.rs
```diff
@@ -300,6 +300,7 @@ impl super::TermWindow {
             WK::RightAlt => KC::RightAlt,
             WK::Pause => KC::Pause,
             WK::CapsLock => KC::CapsLock,
+            WK::VoidSymbol => KC::VoidSymbol,
```
wez, line 303:
```suggestion
            WK::VoidSymbol => return Key::None,
```

wez (commented):
Thanks for this!  What do you think about not including void symbol at the terminal layer?  There's no way to encode that concept in terminal input/output, so it doesn't seem necessary.

wez:
Please also `cargo fmt` and update the PR to make the CI happy!

termwiz/src/input.rs
digitallyserviced: It would seem my understanding of the code is still a bit green. I actually had wondered why are we defining a semi-duplicate `enum` of `KeyCodes`? I apparently did not look too deeply into the implementations of them and just naively plastered it everywhere I saw `CapsLock`.  **I will look into this further.**

digitallyserviced:
I think I cleaned up the unnecessary KeyCodes so that only where it needs to be added for key events and bindings are in there. Also `cargo fmt` and re-tested.

wez:
Thanks!

PR #773, opened 2021-05-09
docs: Correct Lua example for PasteFrom action.

### LINKED ISSUE #772: The doc example for PasteFrom is incorrect
The Lua example has:

```lua
 {key="V", mods="CTRL", action=PasteFrom="Clipboard"},
```

This is invalid Lua. I believe it should be:

```lua
 {key="V", mods="CTRL", action=wezterm.action{PasteFrom="Clipboard"}},
```

Both key definitions in that example have the same issue. 

By the way, I searched the `config/lua` directory for other cases of the same problem and didn't find any, so I believe this particular error is isolated to `PasteFrom`

***

Please forgive my not using the issue template - 95% of it was not applicable to this problem.

KerfuffleV2:
Please note that I know Lua but not Wezterm. I'm including this in the hope that it's useful but it's possible that it is not the correct fix.

Closes #772

### DIFF
--- a/docs/config/lua/keyassignment/PasteFrom.md
+++ b/docs/config/lua/keyassignment/PasteFrom.md
@@ -16,10 +16,10 @@ local wezterm = require 'wezterm';
 return {
   keys = {
     -- paste from the clipboard
-    {key="V", mods="CTRL", action=PasteFrom="Clipboard"},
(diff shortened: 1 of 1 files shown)


bew (approved):

wez:
Looks good, thanks!

PR #788, opened 2021-05-12
Adding ability to redirect stdio to filedescriptor

robinbernon:
Adding ability to redirect stdio to filedescriptor

### DIFF
--- a/filedescriptor/src/lib.rs
+++ b/filedescriptor/src/lib.rs
@@ -210,6 +210,13 @@ pub struct FileDescriptor {
     handle: OwnedHandle,
 }
 
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+pub enum StdioDescriptor {
+    Stdin,
+    Stdout,
+    Stderr,
+}
+
 impl FileDescriptor {
     /// Create a new descriptor from some object that is convertible into
     /// the system `RawFileDescriptor` type.  This consumes the parameter
@@ -256,6 +263,22 @@ impl FileDescriptor {
     /// that can be successfully made non-blocking.
     pub fn set_non_blocking(&mut self, non_blocking: bool) -> anyhow::Result<()> {
         self.set_non_blocking_impl(non_blocking)
+    }
+
+    /// Attempt to redirect stdio to the underlying handle and return
+    /// a `FileDescriptor` wrapped around the original stdio source.
+    /// Since the redirection requires kernel resources that may not be
+    /// available, this is a potentially fallible operation.
+    /// Supports StdIn, StdOut, and StdErr redirections.
+    pub fn redirect_stdio<F: AsRawFileDescriptor>(
+        f: &F,
+        stdio: StdioDescriptor,
+    ) -> anyhow::Result<Self> {
+        match stdio {
+            StdioDescriptor::Stdin => Self::redirect_stdin(f),
+            StdioDescriptor::Stdout => Self::redirect_stdout(f),
+            StdioDescriptor::Stderr => Self::redirect_stderr(f),
+        }
     }
 }
 
--- a/filedescriptor/src/lib.rs
+++ b/filedescriptor/src/lib.rs
@@ -269,16 +269,12 @@ impl FileDescriptor {
     /// a `FileDescriptor` wrapped around the original stdio source.
     /// Since the redirection requires kernel resources that may not be
     /// available, this is a potentially fallible operation.
-    /// Supports StdIn, StdOut, and StdErr redirections.
+    /// Supports stdin, stdout, and stderr redirections.
     pub fn redirect_stdio<F: AsRawFileDescriptor>(
         f: &F,
         stdio: StdioDescriptor,
     ) -> anyhow::Result<Self> {
-        match stdio {
-            StdioDescriptor::Stdin => Self::redirect_stdin(f),
-            StdioDescriptor::Stdout => Self::redirect_stdout(f),
-            StdioDescriptor::Stderr => Self::redirect_stderr(f),
-        }
+        Self::redirect_stdio_impl(f, stdio)
     }
 }
 
--- a/filedescriptor/src/unix.rs
+++ b/filedescriptor/src/unix.rs
@@ -177,27 +177,29 @@ impl OwnedHandle {
         return Self::non_atomic_dup2(fd, dest_fd);
 
         #[cfg(target_os = "linux")]
-        let duped = unsafe { libc::dup3(fd, dest_fd, libc::O_CLOEXEC) };
+        {
+            let duped = unsafe { libc::dup3(fd, dest_fd, libc::O_CLOEXEC) };
 
-        if duped == -1 {
-            let err = std::io::Error::last_os_error();
-            if let Some(libc::EINVAL) = err.raw_os_error() {
-                // We may be running on eg: WSL or an old kernel that
-                // doesn't support O_CLOEXEC; fall back.
-                return Self::non_atomic_dup2(fd, dest_fd);
+            if duped == -1 {
+                let err = std::io::Error::last_os_error();
+                if let Some(libc::EINVAL) = err.raw_os_error() {
+                    // We may be running on eg: WSL or an old kernel that
+                    // doesn't support O_CLOEXEC; fall back.
+                    return Self::non_atomic_dup2(fd, dest_fd);
+                } else {
+                    bail!(
+                        "dup2 of fd {} and dest_fd {} failed: {:?}",
+                        fd,
+                        dest_fd,
+                        err
+                    )
+                }
             } else {
-                bail!(
-                    "dup2 of fd {} and dest_fd {} failed: {:?}",
-                    fd,
-                    dest_fd,
-                    err
-                )
+                Ok(OwnedHandle {
+                    handle: duped,
+                    handle_type,
+                })
             }
-        } else {
-            Ok(OwnedHandle {
-                handle: duped,
-                handle_type,
-            })
         }
     }
 
@@ -295,7 +297,7 @@ impl FileDescriptor {
         };
 
         let std_original = FileDescriptor::dup(std_descriptor)?;
-        unsafe { FileDescriptor::dup2(f.as_raw_file_descriptor(), std_descriptor) }?;
+        unsafe { FileDescriptor::dup2(f, std_descriptor) }?;
 
         Ok(std_original)
     }
--- a/filedescriptor/src/unix.rs
+++ b/filedescriptor/src/unix.rs
@@ -122,6 +122,25 @@ impl OwnedHandle {
         }
     }
 
+    fn non_atomic_dup2(fd: RawFd, dest_fd: RawFd) -> anyhow::Result<Self> {
+        let duped = unsafe { libc::dup2(fd, dest_fd) };
+        if duped == -1 {
+            bail!(
+                "dup2 of fd {} and dest_fd {} failed: {:?}",
+                fd,
+                dest_fd,
+                std::io::Error::last_os_error()
+            )
+        } else {
+            let mut owned = OwnedHandle {
+                handle: duped,
+                handle_type: (),
+            };
+            owned.cloexec()?;
+            Ok(owned)
+        }
+    }
+
     #[inline]
     pub(crate) fn dup_impl<F: AsRawFileDescriptor>(
         fd: &F,
@@ -137,6 +156,35 @@ impl OwnedHandle {
                 return Self::non_atomic_dup(fd);
             } else {
                 bail!("dup of fd {} failed: {:?}", fd, err)
+            }
+        } else {
+            Ok(OwnedHandle {
+                handle: duped,
+                handle_type,
+            })
+        }
+    }
+
+    #[inline]
+    pub(crate) unsafe fn dup2_impl<F: AsRawFileDescriptor>(
+        fd: &F,
+        dest_fd: RawFd,
+    ) -> anyhow::Result<Self> {
+        let fd = fd.as_raw_file_descriptor();
+        let duped = unsafe { libc::dup3(fd, dest_fd, libc::O_CLOEXEC) };
+        if duped == -1 {
+            let err = std::io::Error::last_os_error();
+            if let Some(libc::EINVAL) = err.raw_os_error() {
+                // We may be running on eg: WSL or an old kernel that
+                // doesn't support O_CLOEXEC; fall back.
+                return Self::non_atomic_dup2(fd, dest_fd);
+            } else {
+                bail!(
+                    "dup2 of fd {} and dest_fd {} failed: {:?}",
+                    fd,
+                    dest_fd,
+                    err
+                )
             }
         } else {
             Ok(OwnedHandle {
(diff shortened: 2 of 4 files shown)


filedescriptor/src/windows.rs
```diff
@@ -208,6 +213,43 @@ impl OwnedHandle {
             })
         }
     }
+
+    pub(crate) fn redirect_stdout_to_file(&self) -> anyhow::Result<Redirection> {
+        let std_original = unsafe{ GetStdHandle(STD_OUTPUT_HANDLE) } ;
```
wez, line 218: I think you want to `FileDescriptor::dup(std_original)` here so that you have a managed duplicate handle.  It may be slightly better to `FileDescriptor::new(std_original)` so that you take complete ownership of that handle; that is semantically closer to what this would do on unix.

filedescriptor/src/windows.rs
```diff
@@ -208,6 +213,43 @@ impl OwnedHandle {
...
+        if unsafe {SetStdHandle(STD_OUTPUT_HANDLE, self.handle)} == 0 {
```
wez, line 219: I think this should assign a duplicate of `self.handle`, otherwise when `self` is dropped, we break the output for the whole program.

Really, I think this should be a "class" function and not accept `self` but instead be passed something that is `AsRawFileDescriptor`.  Then this logic would be to `FileDescriptor::dup()` the passed handle, and if `SetStdHandle` succeeds, that duplicated handle would have `into_raw_descriptor` called on it to transfer ownership to the OS.

filedescriptor/src/windows.rs
robinbernon: Have made this change to the code now

filedescriptor/src/windows.rs
robinbernon: Have made this change to the code now

filedescriptor/src/unix.rs
```diff
@@ -217,6 +265,41 @@ impl FileDescriptor {
...
+    /// The returned handle has a separate lifetime from the source, but
+    /// references the same object at the kernel level.
+    pub unsafe fn dup2<F: AsRawFileDescriptor>(f: &F, dest_fd: RawFd) -> anyhow::Result<Self> {
+        OwnedHandle::dup2_impl(f, dest_fd).map(|handle| Self { handle })
+    }
+
+    pub(crate) fn redirect_stdin<F: AsRawFileDescriptor>(f: &F) -> anyhow::Result<FileDescriptor> {
+        let std_original = FileDescriptor::dup(libc::STDIN_FILENO)?;
+        let cloned_handle = OwnedHandle::dup(f)?;
+        unsafe { FileDescriptor::dup2(cloned_handle.handle, libc::STDIN_FILENO) }?;
```
wez, line 283: I was envisioning passing `f` directly to `dup2` and avoiding the intermediate `cloned_handle`:

```suggestion
        unsafe { FileDescriptor::dup2(f, libc::STDIN_FILENO) }?;
```

filedescriptor/src/windows.rs
```diff
@@ -298,12 +261,76 @@ impl FileDescriptor {
...
+                handle: raw_std_handle,
+                handle_type: OwnedHandle::probe_handle_type(raw_std_handle),
+            },
+        };
+
+        let cloned_handle = OwnedHandle::dup(f)?;
+        if unsafe {
+            SetStdHandle(
+                STD_INPUT_HANDLE,
+                cloned_handle.handle as *mut winapi::ctypes::c_void,
```
wez, line 277: I think this should transfer ownership to the OS, otherwise the handle will get closed when `cloned_handle` is dropped when this function returns:

```suggestion
                cloned_handle.into_raw_handle() as *mut winapi::ctypes::c_void,
```

filedescriptor/src/unix.rs
```diff
@@ -217,6 +265,41 @@ impl FileDescriptor {
...
+
+        Ok(std_original)
+    }
+
+    pub(crate) fn redirect_stdout<F: AsRawFileDescriptor>(f: &F) -> anyhow::Result<FileDescriptor> {
```
wez, line 288: FWIW, as a matter of style, I was envisioning this being a single function like this:

```rust
pub(crate) fn redirect_stdio_impl<F: AsRawFileDescriptor>(
        f: &F,
        stdio: StdioDescriptor,
    ) -> anyhow::Result<Self> {
```

that internally did:

```rust
let std_original = FileDescriptor::dup(match stdio {
   StdioDescriptor::Stdin => libc::STDIN_FILENO,
   StdioDescriptor::Stdout => libc::STDOUT_FILENO,
   StdioDescriptor::Stderr => libc::STDERR_FILENO,
))?;
```

and similarly for the Windows implementation

filedescriptor/src/unix.rs
```diff
@@ -146,6 +165,35 @@ impl OwnedHandle {
         }
     }
 
+    #[inline]
+    pub(crate) unsafe fn dup2_impl<F: AsRawFileDescriptor>(
```
wez, line 169: For a bit of complexity, I think `dup3` is a linux only interface, so I suspect that you'll need to do something like:

```rust

#[cfg(target_os = "linux")]
{
   // try to use dup3 here
}
// fall back to non-atomic dup2 here
```

filedescriptor/src/lib.rs
```diff
@@ -257,6 +264,22 @@ impl FileDescriptor {
     pub fn set_non_blocking(&mut self, non_blocking: bool) -> anyhow::Result<()> {
         self.set_non_blocking_impl(non_blocking)
     }
+
+    /// Attempt to redirect stdio to the underlying handle and return
+    /// a `FileDescriptor` wrapped around the original stdio source.
+    /// Since the redirection requires kernel resources that may not be
+    /// available, this is a potentially fallible operation.
+    /// Supports StdIn, StdOut, and StdErr redirections.
```
wez, line 272: nit: the case of these doesn't match those in the enum.  I'd be fine omitting this line TBH, but if we keep it, I'd suggest making this consistent! :)

filedescriptor/src/unix.rs
robinbernon: Ah ok sure. Just thought the execution would be slightly faster with what I've done currently as it avoids having to move the enum into the interior methods. Will change this though to do it this way instead then 👍

filedescriptor/src/unix.rs
robinbernon: It's already in the unix mod. This is cfg toggled in lib so should never get called when running in windows.

filedescriptor/src/lib.rs
robinbernon: Sure, makes sense :) will make these all lowercase then.

filedescriptor/src/unix.rs
wez: FWIW, that enum is super cheap to move around because it fits in a register.  Also: the runtime cost of this code path will be dominated by the cost of the file descriptor related syscalls, so even if the enum was big it still wouldn't matter.

filedescriptor/src/unix.rs
robinbernon: If you want me to build a windows version I'm not sure how to deal with cloexec management. What would I need to do for this?

wez: what I meant was, the only unix with `dup3` is linux.  macos and freebsd don't have it, so you'll need to explicitly check for linux to call `dup3`.

filedescriptor/src/unix.rs
robinbernon: Done

filedescriptor/src/windows.rs
robinbernon: Done

filedescriptor/src/unix.rs
robinbernon: Cool, have now made the change

filedescriptor/src/unix.rs
robinbernon: Ah apologies, completely misread this. Will change now.

filedescriptor/src/lib.rs
robinbernon: Done

filedescriptor/src/unix.rs
robinbernon: Done

filedescriptor/src/unix.rs
```diff
@@ -217,6 +266,33 @@ impl FileDescriptor {
...
+        stdio: StdioDescriptor,
+    ) -> anyhow::Result<Self> {
+        let std_descriptor = match stdio {
+            StdioDescriptor::Stdin => libc::STDIN_FILENO,
+            StdioDescriptor::Stdout => libc::STDOUT_FILENO,
+            StdioDescriptor::Stderr => libc::STDERR_FILENO,
+        };
+
+        let std_original = FileDescriptor::dup(std_descriptor)?;
+        unsafe { FileDescriptor::dup2(f, std_descriptor) }?;
```
wez, line 296: I think this may also need `.into_raw_fd()` here to avoid closing the fd when this function returns.

It would be great to add a regression test for this :)

filedescriptor/src/unix.rs
```diff
@@ -146,6 +166,39 @@ impl OwnedHandle {
...
+    pub(crate) unsafe fn dup2_impl<F: AsRawFileDescriptor>(
+        fd: &F,
+        dest_fd: RawFd,
+    ) -> anyhow::Result<Self> {
+        let fd = fd.as_raw_file_descriptor();
+
+        #[cfg(not(target_os = "linux"))]
+        return Self::non_atomic_dup2(fd, dest_fd);
+
+        let duped = unsafe { libc::dup3(fd, dest_fd, libc::O_CLOEXEC) };
```
wez, line 179: this line has to be enclosed in `#[cfg(target_os = "linux")]` block in order to compile+link on !linux systems

robinbernon: Sorry not sure what you mean here. As is, this forces dup2 for non linux, otherwise, if linux, then it proceeds to attempt dup3 first. Is this not what you want here?

filedescriptor/src/unix.rs
robinbernon: I've made the change to add ```.into_raw_fd()``` into the code. Not sure what do make for the regression test. I had a plan instead to implement this now in a branch in the gag crate. Was gonna run my tests for these through the gag create instead.

filedescriptor/src/unix.rs
wez: Like I sketched out in the original comment; you need to enclose the block of code that calls dup3 in the cfg check, then under it, fall back to the non atomic dup2.

The issue to resolve is that you can only reference `dup3` on linux.  The program won't compile on any other platform unless dup3 is in a linux only block.

robinbernon: The code above already does what you want here => if not linux then skip dup3 and do only dup2

robinbernon: image

wez: The intent of your change is correct, but the way it is expressed won't compile on mac or freebsd because `dup3` doesn't exist there.  You need to ensure that `dup3` is only referenced on linux, but right now it is referenced on all unix systems.

filedescriptor/src/unix.rs
```diff
@@ -217,6 +270,33 @@ impl FileDescriptor {
...
+        stdio: StdioDescriptor,
+    ) -> anyhow::Result<Self> {
+        let std_descriptor = match stdio {
+            StdioDescriptor::Stdin => libc::STDIN_FILENO,
+            StdioDescriptor::Stdout => libc::STDOUT_FILENO,
+            StdioDescriptor::Stderr => libc::STDERR_FILENO,
+        };
+
+        let std_original = FileDescriptor::dup(std_descriptor)?;
+        unsafe { FileDescriptor::dup2(f.into_raw_fd(), std_descriptor) }?;
```
wez, line 296: This doesn't compile because all we know about `F` is that it is `AsRawFileDescriptor`.  That means that we can call `as_raw_file_descriptor` to borrow (but not take ownership of) the `f` parameter.

```suggestion
        unsafe { FileDescriptor::dup2(f.as_raw_file_descriptor(), std_descriptor) }?;
```_

Comment (wez) [reply filedescriptor/src/unix.rs:296]: Actually, I think we just want to pass this through:

```suggestion
        unsafe { FileDescriptor::dup2(f, std_descriptor) }?;
```

robinbernon: Done

filedescriptor/src/unix.rs
robinbernon: Done

wez: GH actions need approval on each change, which is a bit of a drag, however, the freebsd CI is showing that this doesn't compile; it you structure it by enclosing the appropriate code in a block as I said in the original comment, you'll have the most success.

You can keep iterating against that until it goes green and then I can approve and run the rest of the platforms!

wez (changes requested):
Please fixup the freebsd build per earlier comments; thanks!

filedescriptor/src/unix.rs
robinbernon: @wez could you run the pipeline again please?

robinbernon:
> Please fixup the freebsd build per earlier comments; thanks!

Think it's fixed now. Can you please re-run the pipeline?

robinbernon:
@wez is there more info on why freebsd-12 failed here?

robinbernon:
@wez looks like it's not even getting to the  build stage on freebsd-12, is there something wrong with the environment setup here?

wez:
> @wez looks like it's not even getting to the build stage on freebsd-12, is there something wrong with the environment setup here?

Appears to be: https://github.com/rust-lang/rustup/issues/2774

Let's see if https://github.com/wez/wezterm/commit/caed905626123cf5cec1368377330413a3e1ab1a resolves it; you can try rebasing the PR on that commit.

filedescriptor/src/unix.rs
```diff
@@ -146,6 +166,43 @@ impl OwnedHandle {
...
+                        "dup2 of fd {} and dest_fd {} failed: {:?}",
+                        fd,
+                        dest_fd,
+                        err
+                    )
+                }
+            } else {
+                return Ok(OwnedHandle {
+                    handle: duped,
+                    handle_type: Self::probe_handle_type(duped),
```
wez, line 200: I think this can match the code in `non_atomic_dup2` above:

```suggestion
                    handle_type: (),
```

filedescriptor/src/unix.rs
```diff
@@ -217,6 +274,33 @@ impl FileDescriptor {
...
+        stdio: StdioDescriptor,
+    ) -> anyhow::Result<Self> {
+        let std_descriptor = match stdio {
+            StdioDescriptor::Stdin => libc::STDIN_FILENO,
+            StdioDescriptor::Stdout => libc::STDOUT_FILENO,
+            StdioDescriptor::Stderr => libc::STDERR_FILENO,
+        };
+
+        let std_original = FileDescriptor::dup(&std_descriptor)?;
+        unsafe { FileDescriptor::dup2(f, std_descriptor) }?;
```
wez, line 300:
```suggestion
        unsafe { FileDescriptor::dup2(f, std_descriptor) }?.into_raw_fd();
```

filedescriptor/src/windows.rs
```diff
@@ -255,6 +261,36 @@ impl FileDescriptor {
...
+        };
+
+        let raw_std_handle = unsafe { GetStdHandle(std_handle) } as *mut _;
+        let std_original = unsafe { FileDescriptor::from_raw_handle(raw_std_handle) };
+
+        let cloned_handle = OwnedHandle::dup(f)?;
+        if unsafe {
+            SetStdHandle(
+                std_handle,
+                cloned_handle.into_raw_handle() as *mut winapi::ctypes::c_void,
```
wez, line 282: I think you can simplify this:

```suggestion
                cloned_handle.into_raw_handle() as *mut _,
```

robinbernon: Done

filedescriptor/src/unix.rs
robinbernon: Done

filedescriptor/src/unix.rs
robinbernon: Done

robinbernon:
@wez one more run please?

robinbernon:
Guess I must have forgotten to fmt. Have done this now soz. Could you please try again?

wez:
I think you forgot to push :)

PR #799, opened 2021-05-15
Fix spawning new tabs/windows in directory with urlunsafe paths

Arvedui:
The CWD is internally stored as URL. The path component is therefore stored percent_encoded. It must be decoded before use.


wezterm-gui/src/termwindow/spawn.rs
```diff
@@ -127,7 +128,10 @@ impl super::TermWindow {
         } else {
             match cwd {
                 Some(url) if url.scheme() == "file" => {
-                    let path = url.path().to_string();
+                    let path = percent_decode_str(url.path())
+                        .decode_utf8()
+                        .unwrap()
```
wez, line 133: We can't tolerate an unwrap here; it will kill everything that the user is doing if a program has set a bogus cwd.

A failure to decode to utf8 should result in the `path` being treated as `None`.

There's a similar section of code over there that should also have the same treatment:
https://github.com/wez/wezterm/blob/main/wezterm-mux-server-impl/src/sessionhandler.rs#L615

wez (changes requested):
Thanks for digging in!
We'll need to remove the `unwrap`, but this otherwise sounds OK.
Please also update the other (unfortunate) duplicate block of code!

Arvedui:
Thanks for your feedback. Changed the patch accordingly. I did not test the changes in sessionhandler.rs though, because it is not clear to me how to do that.

wez:
Thanks!

PR #838, opened 2021-05-31
Better antialiasing for Undercurl

ModProg:
The reason I am making this request again, is that after using it for a while I noticed that the simpler calculation leaves artefacts:
Current (with dark pixels around undercurl):
image
With the more complex calculation:
image

old on the left, mine on the right:
image


wez:
Thanks; yeah, it does look better!

PR #851, opened 2021-06-03
Use configured default program when running subcommand "start"

exactly-one-kas:
Running `wezterm-gui start --cwd .` starts the system's default shell, regardless of the configured `default_prog`. This PR fixes that.

### DIFF
--- a/wezterm-gui/src/main.rs
+++ b/wezterm-gui/src/main.rs
@@ -326,7 +326,13 @@ fn run_terminal_gui(opts: StartCommand) -> anyhow::Result<()> {
 
         let cmd = if need_builder {
             let mut builder = if opts.prog.is_empty() {
-                CommandBuilder::new_default_prog()
+                let config = config::configuration();
+
+                if let Some(prog) = &config.default_prog {
+                    CommandBuilder::from_argv(prog.iter().map(Into::into).collect())
+                } else {
+                    CommandBuilder::new_default_prog()
+                }
             } else {
                 CommandBuilder::from_argv(opts.prog)
             };
--- a/wezterm-gui/src/main.rs
+++ b/wezterm-gui/src/main.rs
@@ -10,6 +10,7 @@ use mux::Mux;
 use portable_pty::cmdbuilder::CommandBuilder;
 use promise::spawn::block_on;
 use std::ffi::OsString;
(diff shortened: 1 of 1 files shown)


wezterm-gui/src/main.rs
```diff
@@ -326,7 +327,13 @@ fn run_terminal_gui(opts: StartCommand) -> anyhow::Result<()> {
 
         let cmd = if need_builder {
```
wez, line 328: I think doing this is a little cleaner:

```suggestion
            let config = config::configuration();
            let prog = opts.prog.iter().map(|s| s.as_os_str()).collect::<Vec<_>>();
            let mut builder = config.build_prog(if prog.is_empty() { None } else { Some(prog) })?;
            if let Some(cwd) = opts.cwd {
                builder.cwd(cwd);
            }
            Some(builder)
```

wezterm-gui/src/main.rs
```diff
@@ -508,7 +517,22 @@ fn run() -> anyhow::Result<()> {
             SetStdHandle(STD_ERROR_HANDLE, stderr.into_raw_file_descriptor());
             */
 
-            std::env::set_current_dir(config::HOME_DIR.as_path())?;
+            // Only change the current directory to the user's home path if
```
wez, line 520: This doesn't appear to be related to the issue that you set out to resolve in the PR.
Please remove it from here and track this separately, along with an explanation of why it is needed.

FWIW, I don't think this is necessary: this is controlling the current directory of the GUI process itself, and not its child processes.

wez (changes requested):
Thanks for digging in!
I've made a suggested alternative way to fix this bug below.
I'd prefer it if you separated out the other change to the GUI current dir as I don't think it is related to this issue and I'm not sure of your intent with that change.

Thanks!

wezterm-gui/src/main.rs
exactly-one-kas: This is for starting wezterm in the current working directory from a GUI context. My use case is starting a terminal from my file manager, which starts the process in the currently opened directory. Since there's no console available, the current directory gets reset to the user's home directory. I'll put this into a separate PR.

wez:
Thanks; I pulled in and amended the first part of this!

Issue #133, opened 2020-01-26
vttest 


erf:
## Describe the bug

Are you using [vttest](https://invisible-island.net/vttest/) ? It is a great way to test your terminal features. Some of them fails now.

Also maybe this [comment](https://github.com/alacritty/alacritty/issues/240#issuecomment-284181449) might help.

## Environment (please complete the following information):

 - macOS 10.15.2 (19C57)
 - Frontend:  not sure
 - 20200113-222147-724ad3a 

## To Reproduce

Download vttest from: https://invisible-island.net/vttest/

run `vttest` from terminal

press `1` to run the first test. press `Enter` several times to skip to next test and show errors.

wez:
I ran it in the past; the problem I had with it was that it requires a human to look at it and know whether the output is correct; there is no scalable way to automate its conformance tests.  I agree that we should have correct behavior, but I'm more motivated to correct behavior that manifests for real applications before behaviors that no one is actually hitting. 

With that in mind, if you're experiencing something broken in an application, please file a separate issue about that with repro steps and that will get priorized ahead of vttest conformance!

wez:
I fixed up the cursor positioning conformance.  There are a couple of screens that switch to 132 column mode that look messed up simply because we don't (and lkely won't) support 132 mode at all.
I also ran through the screen tests, and they look reasonable; some don't make a lot of sense to me (scroll positioning test doesn't have any information on what correct should look like), and we're not handling some of the character set shifts, but in a UTF-8 world I don't think that is a super high priority.

erf:
Thanks, i opened up a new issue, regarding the vis editor.

Maybe you could make a set of unit test files which tries to run spesific vt100 commands and read back the result?

Feel free to close this.

wez:
> Thanks, i opened up a new issue, regarding the vis editor.

Thanks!
 
> Maybe you could make a set of unit test files which tries to run spesific vt100 commands and read back the result?

Yep, that's the approach used by our unit tests already!

erf:
I noticed the background colour is not set in the vttest 2, screen 3.

wez:
@erf regarding the path: that may be because your shell configuration doesn't set up your PATH in all cases; by default your $SHELL is spawned and the PATH is inherited from your desktop environment.  

You can force wezterm to spawn a login shell on macOS by default placing this in your `~/.wezterm.toml`; this will tell your shell to process login related startup items and will likely make your PATH more consistent:

```toml
[...]

PR #891, opened 2021-06-19
Finish up DEC Special Graphics character set

klamonte:
Hi!  This is my first PR ever in Rust, so it likely needs work.  But the point of it:

1. DEC Special Graphics is saved/restored by DECSC/DECRC.  (And it's actually not just one bool, it should be the selected character sets for G0, G1, G2, G3, and GR; and by default selectable via ASCII SO and SI.)
2. Added the remaining mappings.  I do not know why some of the glyphs aren't making it to the screen correctly.  These are all standard Unicode glyphs:  ◆▒␉␌␍␊°±␤␋┘┐┌└┼⎺⎻─⎼⎽├┤┴┬│≤≥π≠£·  Many of these are also accessible to ncurses applications via ACS_DIAMOND, ACS_CKBOARD, etc.

With these changes (and assuming you know how to fix the glyphs for ␉ ␌ ␍ ␊ ␋), this would finish out the screen features SAVE/RESTORE CURSOR screen, and get better on the VT100 character sets screen.

### DIFF
--- a/term/src/terminalstate.rs
+++ b/term/src/terminalstate.rs
@@ -2965,9 +2965,7 @@ impl<'a> Performer<'a> {
         for g in unicode_segmentation::UnicodeSegmentation::graphemes(p.as_str(), true) {
             let g = if self.dec_line_drawing_mode {
                 match g {
-                    // AZL: I do not know why the diamond is red in vttest.
-                    "`" => "♦",
-
+                    "`" => "◆",
                     "a" => "▒",
 
                     // AZL: I do not know why the next four Unicode glyphs are
--- a/term/src/terminalstate.rs
+++ b/term/src/terminalstate.rs
@@ -35,6 +35,13 @@ struct TabStop {
     tab_width: usize,
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum CharSet {
+    Ascii,
+    Uk,
+    DecLineDrawing,
+}
+
 impl TabStop {
     fn new(screen_width: usize, tab_width: usize) -> Self {
         let mut tabs = Vec::with_capacity(screen_width);
@@ -103,7 +110,8 @@ struct SavedCursor {
     wrap_next: bool,
     pen: CellAttributes,
     dec_origin_mode: bool,
-    dec_line_drawing_mode: bool,
+    g0_charset: CharSet,
+    g1_charset: CharSet,
     // TODO: selective_erase when supported
 }
 
@@ -277,7 +285,11 @@ pub struct TerminalState {
     current_mouse_button: MouseButton,
     last_mouse_move: Option<MouseEvent>,
     cursor_visible: bool,
-    dec_line_drawing_mode: bool,
+
+    /// Support for US, UK, and DEC Special Graphics
+    g0_charset: CharSet,
+    g1_charset: CharSet,
+    shift_out: bool,
 
     tabs: TabStop,
 
@@ -478,7 +490,9 @@ impl TerminalState {
             mouse_tracking: false,
             last_mouse_move: None,
             cursor_visible: true,
-            dec_line_drawing_mode: false,
+            g0_charset: CharSet::Ascii,
+            g1_charset: CharSet::DecLineDrawing,
+            shift_out: false,
             current_mouse_button: MouseButton::None,
             tabs: TabStop::new(size.physical_cols, 8),
             title: "wezterm".to_string(),
@@ -2754,7 +2768,8 @@ impl TerminalState {
             wrap_next: self.wrap_next,
             pen: self.pen.clone(),
             dec_origin_mode: self.dec_origin_mode,
-            dec_line_drawing_mode: self.dec_line_drawing_mode,
+            g0_charset: self.g0_charset,
+            g1_charset: self.g1_charset,
         };
         debug!(
             "saving cursor {:?} is_alt={}",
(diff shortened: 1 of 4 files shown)


klamonte:
Reference also https://github.com/wez/wezterm/pull/49 and #133.

term/src/terminalstate.rs
```diff
@@ -103,6 +103,7 @@ struct SavedCursor {
     wrap_next: bool,
     pen: CellAttributes,
     dec_origin_mode: bool,
+    dec_line_drawing_mode : bool,
```
wez, line 106:
```suggestion
    dec_line_drawing_mode: bool,
```

The GH actions CI is complaining about this PR because it wants it to be `cargo fmt`'d; I think it's just this line that is upsetting it

term/src/terminalstate.rs
```diff
@@ -2961,17 +2965,45 @@ impl<'a> Performer<'a> {
         for g in unicode_segmentation::UnicodeSegmentation::graphemes(p.as_str(), true) {
             let g = if self.dec_line_drawing_mode {
                 match g {
+                    "`" => "◆",
+                    "a" => "▒",
+
+                    // AZL: I do not know why the next four Unicode glyphs are
```
wez, line 2971: Is that a local font problem?  If I copy and paste the characters from your PR description into the terminal I see those glyphs in my terminal on this mac.  If you do the same, do you see those glyphs on your system, or is the issue that the don't show up when used in the appropriate mode when rendered via the terminal?

wezterm will log information about glyphs that it failed to resolve.  It can take several seconds to do that if you're in a debug build, because it has to parse and compute glyph coverage maps for a potentially large list of fallback fonts, and that's really slow in a debug build.

wez (commented):
Thanks!

Would you mind adding a basic test for this?

I think you could get away with doing something like this:
https://github.com/wez/wezterm/blob/main/term/src/test/mod.rs#L632-L642
but printing the relevant sequences to get into the alt charset and print out the chars, and then assert that they show up as the intended unicode chars.

You can run the `term` tests directly via `cargo test -p wezterm-term`

term/src/terminalstate.rs
klamonte: I saw nothing logged to stderr on the glyphs.  wezterm ls-fonts yields:

```
 Primary font:
wezterm.font_with_fallback({
  -- <built-in>, BuiltIn
  "JetBrains Mono",

  -- /usr/share/fonts/truetype/noto/NotoColorEmoji.ttf, FontConfig
  "Noto Color Emoji",

  -- <built-in>, BuiltIn
  "Last Resort High-Efficiency",

})
```

The other fonts are all the same: JetBrains Mono, Noto, Last Resort.

When I paste " ␉ ␌ ␍ ␊ ␋" into cat I see these instead:

glyphs

I've got the format fixed, and will add the test.  (This is fun! :) )

wez: I think the glyphs you're seeing are coming from a fallback font; they seem like alternative pictorial versions of `␉ ␌ ␍ ␊ ␋`.   I think I should add a thing to `ls-fonts` to have it explain which fonts are used for a given text input to help understand cases like this!


> I've got the format fixed, and will add the test. (This is fun! :) )

Great!

wez: `wezterm ls-fonts --text "␉ ␌ ␍ ␊ ␋"` will explain which fonts were used now in `main`

klamonte:
OK, this update handles VT100 ASCII, UK, and DEC character sets.  It passes the vttest VT100 character set test, and can support ncurses ACS chars.

What do you think?

termwiz/src/escape/esc.rs
```diff
@@ -68,10 +68,19 @@ pub enum EscCode {
     /// DECPNM - Normal Keypad
     DecNormalKeyPad = esc!('>'),
 
-    /// Designate Character Set – DEC Line Drawing
-    DecLineDrawing = esc!('(', '0'),
-    /// Designate Character Set – US ASCII
-    AsciiCharacterSet = esc!('(', 'B'),
```
wez, line 74: There are a couple of tests in the termwiz crate that need to be adjusted for this change:
https://github.com/wez/wezterm/blob/main/termwiz/src/render/terminfo.rs#L891

You can run those tests using `cargo test -p termwiz` (or just run all of the tests: `cargo test`)

wez (commented):
This is looking great thanks!  I think the only thing outstanding is that there are couple of tests in the `termwiz` crate that need adjusting to the `EscCode` changes, but this otherwise looks ready to land.

Thanks!

klamonte:
When I use Terminus font, the scan line glyphs (⎺⎻─⎼⎽) look good for one moment, and are then overwritten with wrongly-placed lines that look like (⎺⎻──⎼).  I thought I saw some logic for wezterm to draw its own glyphs; is there a way to turn that feature off, so that I can ensure I am testing my font and not that drawing code?

wez:
> When I use Terminus font, the scan line glyphs (⎺⎻─⎼⎽) look good for one moment, and are then overwritten with wrongly-placed lines that look like (⎺⎻──⎼). I thought I saw some logic for wezterm to draw its own glyphs; is there a way to turn that feature off, so that I can ensure I am testing my font and not that drawing code?

Hmm, that sounds like a fallback font is being loaded asynchronously.
If you run: `WEZTERM_LOGG=wezterm_font=trace,info wezterm` it should show you what's happening in the font crate and perhaps give some insight

PR #900, opened 2021-06-23
Only switch CWD on Windows if no explicit relative working dir is given

exactly-one-kas:


wez:
Can you tell me more about why you want to make this change?
This code affects the working directory of the wezterm process rather than directly affecting the child processes that it spawns.
Is there a particular scenario where things aren't working for you?

exactly-one-kas:
This prevented Wezterm from being spawned in the current directory of my file manager (Double Commander). Since it doesn't support placeholders, I had to put `wezterm-gui.exe start --cwd .` as the command.

Since the CWD of the wezterm process is used to resolve a relative path given by `--cwd`, resetting it to the home directory prevents this from working properly.

The change shouldn't affect Wezterm being started by double-clicking the EXE or a start menu link (without any command line options given)

wez:
Thanks for the context!

What do you think about simply deleting the original `set_current_dir` call entirely?

exactly-one-kas:
Personally, I wouldn't mind, since I'm starting the application via AutoHotKey (whose CWD is already my home dir) and as a tool with a specific working directory explicitly set.

I'm not sure how users using system-wide launchers (eg the start menu) would feel, since the starting directory would be the program's install dir or `C:\Windows\System32`

wez:
I removed that logic and made the behavior consistent with unix in the pty layer in https://github.com/wez/wezterm/commit/7cd3dcaee45e852ec752eadedf1ecc5c95825141

The intended behavior is:

* If the cwd is specified, use that.
* Otherwise, use the home directory

I don't know if this is going to satisfy you, as this appears to be the same behavior as we had anyway!

What do you think?

exactly-one-kas:
LGTM

However, `CreateProcessW()` expects a full path and behaves weirdly if the shell / parent process switched drives before. I'll make a PR for that.

PR #904, opened 2021-06-27
Improve VT102 compliance

klamonte:
The following features are implemented by these changes, closing most of the items in #133:

- Render blinking text, both normal (SGR 5) and rapid (SGR 6)
- Render reverse video (DECSCNM)
- Respond to DECREQTPARM
- Parse sequences for double-height / double-width awareness.  These are not rendered currently, but are available for future use and/or downstream projects.
- ENQ answerback config option
- AutomaticNewline mode (ANM)


wezterm-gui/src/termwindow/mod.rs
```diff
@@ -1086,6 +1092,90 @@ impl TermWindow {
...
+                        }
+                    }
+                }
+            }
+        }
+
+        (false, None)
+    }
+
+    fn maintain_text_blink_rapid(&mut self, now: Instant) -> (bool, Option<Instant>) {
```
wez, line 1140: it would be nice to combine this with the slow blink assessment so that we can avoid iterating the individual cells twice.

It's a bit unfortunate that the (presumably common case) of no blinking cells is `O(2 * viewport_area)`.   Two possible strategies to improve this:

* Add a rapid and slow blink bits to LineBits and maintain them in logically the same place as `scan_and_create_hyperlinks` and `invalidate_implicit_hyperlinks`
* Have the render logic maintain the timing similar to how animated gif frames are dispatched. https://github.com/wez/wezterm/blob/main/wezterm-gui/src/termwindow/render.rs#L80 and search for `has_animation` to get a feel for that.  [update_next_frame_time](https://github.com/wez/wezterm/blob/main/wezterm-gui/src/termwindow/render.rs#L137) is called whenever we render a cell with an anim gif.  We could do something similar when we recognize that we have a blinking cell and maintain the timing as part of that scan, avoiding additional per-cell scans

I think I like the idea of the second one more.

wezterm-gui/src/termwindow/render.rs
```diff
@@ -724,11 +729,32 @@ impl super::TermWindow {
                 let mut bg = bg_color;
                 let mut bg_default = bg_is_default;
 
-                if attrs.reverse() {
+                // Check the line reverse_video flag and flip.
+                if (attrs.reverse() && !params.line.is_reverse())
```
wez, line 733: It's still early here, but can this be phrased like this?

```rust
if attrs.reverse() == !params.line.is_reverse() {
```

wezterm-gui/src/termwindow/render.rs
```diff
@@ -724,11 +729,32 @@ impl super::TermWindow {
...
+                // part of blinking then set fg = bg.  This is a cheap
+                // means of getting it done without impacting other
+                // features.
+                let blink_rate = match attrs.blink() {
+                    Blink::None => 0,
+                    Blink::Slow => params.config.text_blink_rate,
+                    Blink::Rapid => params.config.text_blink_rate_rapid,
+                };
+                if blink_rate != 0
+                    && (self.config.text_blink_rate != 0 || self.config.text_blink_rate_rapid != 0)
```
wez, line 750: `params.config` is the same as `self.config`, so these `!= 0` should be redundant; `blink_rate` will be set to the configured value and if it is non-zero it is because the config was non-zero. This conditional should be ok to be;

```rust
if blink_rate != 0 {
```

termwiz/src/surface/line.rs
```diff
@@ -97,6 +129,13 @@ impl Line {
                 .map(|chunk| {
                     let mut line = Line {
                         cells: chunk.to_vec(),
+                        // AZL TODO:
+                        //
+                        // How to "| self.bits.LineBits::REVERSE"
```
wez, line 134: I think you want:

```rust
   bits: LineBits::DIRTY | (self.bits & LineBits::REVERSE)
```

termwiz/src/surface/line.rs
```diff
@@ -46,6 +68,16 @@ impl Line {
         Self { bits, cells }
     }
 
+    pub fn with_width_reverse(width: usize, reverse: bool) -> Self {
+        let mut cells = Vec::with_capacity(width);
+        cells.resize(width, Cell::default());
+        let mut bits = LineBits::DIRTY;
+        if reverse {
```
wez, line 75: You can save a couple of lines:

```rust
bits.set(LineBits::REVERSE, reverse);
```

term/src/terminalstate.rs
```diff
@@ -1949,10 +1963,26 @@ impl TerminalState {
...
 
-            Mode::SetDecPrivateMode(DecPrivateMode::Code(DecPrivateModeCode::ReverseVideo))
-            | Mode::ResetDecPrivateMode(DecPrivateMode::Code(DecPrivateModeCode::ReverseVideo)) => {
-                // I'm mostly intentionally ignoring this in favor
-                // of respecting the configured colors
+            Mode::SetDecPrivateMode(DecPrivateMode::Code(DecPrivateModeCode::ReverseVideo)) => {
+                // Turn on reverse video for all of the lines on the
+                // display.
+                self.screen.reverse_video_mode = true;
+                for y in 0..self.screen.physical_rows as i64 {
```
wez, line 1970: In rust, array dimension accesses are bounds-checked by default which means that there is some (small) overhead for C style iteration.  It's preferable to iterate containers directly so that the bounds check happens just once and the individual steps are essentially equivalent to bumping a pointer with no other overhead:

```rust
for line in &mut self.screen.lines {
   line.set_reverse(true);
}
```

term/src/terminalstate.rs
```diff
@@ -1949,10 +1963,26 @@ impl TerminalState {
...
+                    let line = self.screen.line_mut(line_idx);
+                    line.set_reverse(true);
+                }
+            }
+
+            Mode::ResetDecPrivateMode(DecPrivateMode::Code(DecPrivateModeCode::ReverseVideo)) => {
+                // Turn off reverse video for all of the lines on the
+                // display.
+                self.screen.reverse_video_mode = false;
+                for y in 0..self.screen.physical_rows as i64 {
```
wez, line 1981:
```rust
for line in &mut self.screen.lines {
   line.set_reverse(false);
}
```

term/src/terminalstate.rs
wez: This is `O(scrollback_size)` whenever the reverse video mode is changed, and some people love having really huge scrollback.  Is reverse video really a property of the line?  Does it make sense to expose `TerminalState::get_reverse_video()` as an accessor to avoid this update pass?

For rendering purposes, we could still keep `LineBits::REVERSE` but teach `terminal_get_lines` to set it when retrieving the lines that need to be rendered:

https://github.com/wez/wezterm/blob/main/mux/src/renderable.rs#L72-L93

```rust
/// Implements Pane::get_lines for Terminal
pub fn terminal_get_lines(
    term: &mut Terminal,
    lines: Range<StableRowIndex>,
) -> (StableRowIndex, Vec<Line>) {
    let screen = term.screen_mut();
    let reverse = term.get_reverse_video(); // added
    let phys_range = screen.stable_range(&lines);
    (
        screen.phys_to_stable_row_index(phys_range.start),
        screen
            .lines
            .iter_mut()
            .skip(phys_range.start)
            .take(phys_range.end - phys_range.start)
            .map(|line| {
                let cloned = line.clone();
                line.clear_dirty();
                cloned.set_reverse(reverse);  // added
                cloned
            })
            .collect(),
    )
}
```

If something like that makes sense, then `terminalstate.rs` would only need to track whether the mode is on or not, and not need to set it on the lines themselves, or deal with that when rewrapping.

term/src/terminalstate.rs
```diff
@@ -1977,6 +2007,13 @@ impl TerminalState {
                 self.insert = false;
             }
 
+            Mode::SetMode(TerminalMode::Code(TerminalModeCode::AutomaticNewline)) => {
```
wez, line 2010: would you mind adding a unit test for this?
It can be similar to the one from your previous PR!

term/src/terminalstate.rs
```diff
@@ -3318,6 +3368,23 @@ impl<'a> Performer<'a> {
             Esc::Code(EscCode::DecSaveCursorPosition) => self.dec_save_cursor(),
             Esc::Code(EscCode::DecRestoreCursorPosition) => self.dec_restore_cursor(),
 
+            Esc::Code(EscCode::DecDoubleHeightTopHalfLine) => {
+                let idx = self.screen.phys_row(self.cursor.y);
+                self.screen.line_mut(idx).set_double_height_top();
```
wez, line 3373: it would be great to add a some test coverage for these attributes, especially because we're not rendering them :-)

I don't think we have any existing tests or conveniences for inspecting `LineBits`, but you should still be able to use the same `term.print` stuff and then use `assert_eq!` to verify that the right things are set on the lines in the `TestTerm`.

wez (commented):
It's great to see this, thanks!

I read the bitflags manual and found some nice helper methods that make the bit manipulation a bit easier to grok, and peppered some suggestions throughout!

Some of those suggestions may be redundant if we can reasonably defer setting `LineBits::REVERSE` based on the mode at render time; I think some of these additions may not be needed if that is the case.

The main thing I'm "worried" about is the blink stuff: at the moment it seems like it will be expensive in the common case of no blinking text, but I have a couple of suggestions for that.

term/src/terminalstate.rs
klamonte: A few thoughts:

* Setting/clearing reverse ought to only impact lines on the visible screen, not every line in the scrollback.  If the entire scrollback is begin affected (or scanned), then I need to fix that.
* I have the line-level flag there so that reversed lines that enter the scrollback retain the reverse flag.  But that's really just an opinion: the real VT100 had no scrollback, and reverse was only ever a whole-screen setting.  I should see what xterm does: whether it keeps the reverse on for lines that go into scrollback or not.
* The flag on the screen is only so that new lines are also set to reverse.  But if we decide that reverse goes away for lines that enter scrollback, then we can eliminate the line-level flag entirely and just set a reverse flag at physical screen render time.

klamonte: A small bit of exploring shows that for xterm, konsole, and VTE, reverse video is a property of the screen only, and furthermore when set all lines rendered, including those in scrollback, are reversed.

I will remove the line-level attribute and leave it on the screen only.

klamonte: OK, reverse is nicer, thank you!  However, there is one gap remaining: on lines that do not stretch to the full physical pixels, we do not see the reversed background.  Example:

reverse_incomplete

I need to add a bit after render.rs:933, so that the selection and cursor are rendered, and then if there is more physical width remaining fill it with bg_color.  Any suggestions?

wezterm-gui/src/termwindow/mod.rs
klamonte: It is a bit non-optimal, but it looks to me like the timer mechanism will only be checking every {interval} time, right?  So with both slow and rapid blink enabled, using the defaults of 500/250, it would be 6 scans per second (0, 250, 500, 750 for rapid) + (0, 500 for slow).

Could the has_animation method result in different timings for specific cells, such that two cells that are both rapid blink might not get updated on the same rendered frame?

Either optimization is a bit more advanced with Rust and this codebase than I am able to do right now.  Would it be worthwhile to wrap up this PR with blinking disabled by default for performance, and then optimize on a future issue/PR?

klamonte:
Unit tests added, cleanup around bit sets and simplified logic.

We are now at the following unresolved items:
* Fix conflict with wezterm-gui/src/termwindow/render.rs
* Fix reverse video on lines that do not extend to the full width of the display.
* Pick a path forward on blinking: 1) keep as is now, with the scans, possibly disabled by default; 2) change to line-level flags similar to hyperlinks to make scans go faster; 3) change to something like has_animation to only schedule an update if a blinking cell is visible; 4) other.

wez:
I'm conscious that I've been making conflicting changes, and that it's a PITA to wrangle merge conflicts.
With that in mind, I cherry-picked the commits from this pr into the `vttest` branch in my repo: https://github.com/wez/wezterm/tree/vttest

A casualty of this is the code that manages timing for text blinking; I'd already made changes that broke your implementation when fixing up a general render performance problem.  I'm happy to take a stab at making that work tomorrow.
I can also take a look at extending the reverse video to the full line width.

wez:
I've push this to main; many thanks!
I went with option 3 for blinking and 7192fe7696e8be88f6e0f9343265aab173d0a040 was the missing piece for reverse video!

wez:
@klamonte: re: double height as a line attribute, I'm considering reorganizing how a work with `Line` and more formally tracking the logical pre-wrapped line as it makes some things easier wrt. future bidi support but also cheaper because selection and search operations need to synthesize the logical line and that happens fairly frequently.  The question on my mind is: what should the semantics be around lines that are double height wrt. re-wrapping them during a resize?

I'm inclined towards tracking the double-height-ness on the original logical line, but I'm not sure if that is totally cool.
What do you think?

klamonte:
That sounds fine to me.

Double-height will be a per-line attribute on both lines (top and bottom), and the top-line version will (according to someone with a hardware terminal when I asked around circa 2012) not render the underline attribute.  One could even take top-halves and bottom-halves of different glyphs and make wholly new glyphs out of them -- though they would all be double-width.  So even in hardware terminals it was a "trick" for the screen/printer rather than a semantic equivalence.

PR #906, opened 2021-06-27
Remove duplicate colorscheme file

adrian5:
Looks like this accidentally slipped in. It's not in the [upstream repo](https://github.com/mbadolato/iTerm2-Color-Schemes/tree/master/wezterm) either.

* Paraiso¹ Dark :heavy_check_mark: 
* Parasio Dark :x: 

¹ Portuguese for _paradise_.

Btw, do you accept new color schemes to be added?


wez:
Ah good catch; seems like it was renamed and our sync process doesn't handle it!

> Btw, do you accept new color schemes to be added?

I'd prefer it if they were added in the upstream repo as it makes sync'ing easier, and that repo is really all about color schemes.

In theory, having local schemes that are not present there means that we might lose our local copies when we next sync, although your PR proves that that mechanism isn't working :-p

PR #916, opened 2021-07-01
docs: Add note on issues with binding mouse Up event only

bew:
As explained in the added note, there is an issue with binding only the Up event of a mouse binding, where the Down event is sent to the running program, but not the Up event, making weird unexpected behaviors when using such Up bindings.

This is something that had bothered me for a few months, and that I finally fixed _properly_ a few weeks ago in https://github.com/bew/dotfiles/commit/eccb7b507282, by disabling the Down event.

I think that blindly suggesting to bind something to the mouse Up event without disabling the Down event is 'dangerous', and the user reading the doc should be aware of this, so he can add a Down-disabling binding and avoid the unexpected behaviors.

### DIFF
--- a/docs/config/mouse.md
+++ b/docs/config/mouse.md
@@ -70,6 +70,8 @@ return {
       mods="CTRL",
       action="OpenLinkAtMouseCursor",
     },
+    -- NOTE that binding only the 'Up' event can give unexpected behaviors.
+    -- Read more below on the gotcha of binding an 'Up' event only.
   },
 }
 ```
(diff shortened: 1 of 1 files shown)


docs/config/mouse.md
```diff
@@ -70,6 +70,13 @@ return {
       mods="CTRL",
       action="OpenLinkAtMouseCursor",
     },
+    -- NOTE: if you only bind the 'Up' event in wezterm, the 'Down' event will
```
wez, line 73: I feel like this would be better off outside of the code comment and be part of the main prose below!

bew: Maybe yes, but I wanted to put it next to the Ctrl-Click binding to ensure readers see it.
Or I put the explanation below, and only a small note here, saying to read more below?

wez: I think having the main explanation outside the code seems best, but yeah, have a short comment in the example pointing the reader to the rest of the docs.

bew:
Should be good now, what do you think?

PR #918, opened 2021-07-03
glyph: Comment the actual char for easy visualization

bew:
It helps visualize the char of the glyph when reading the code :smiley:

### DIFF
--- a/wezterm-gui/src/glyphcache.rs
+++ b/wezterm-gui/src/glyphcache.rs
@@ -360,8 +360,8 @@ impl BlockKey {
     pub fn from_char(c: char) -> Option<Self> {
         let c = c as u32;
         Some(match c {
-            // BOX DRAWINGS LIGHT HORIZONTAL
-            0x2500 => Self::Poly(&[Poly { // ─
+            // [─] BOX DRAWINGS LIGHT HORIZONTAL
+            0x2500 => Self::Poly(&[Poly {
                 path: &[
                     PolyCommand::MoveTo(BlockCoord::Zero, BlockCoord::Frac(1, 2)),
                     PolyCommand::LineTo(BlockCoord::One, BlockCoord::Frac(1, 2)),
@@ -369,8 +369,8 @@ impl BlockKey {
                 intensity: BlockAlpha::Full,
                 style: PolyStyle::Outline,
             }]),
-            // BOX DRAWINGS HEAVY HORIZONTAL
-            0x2501 => Self::Poly(&[Poly { // ━
+            // [━] BOX DRAWINGS HEAVY HORIZONTAL
+            0x2501 => Self::Poly(&[Poly {
                 path: &[
                     PolyCommand::MoveTo(BlockCoord::Zero, BlockCoord::Frac(1, 2)),
                     PolyCommand::LineTo(BlockCoord::One, BlockCoord::Frac(1, 2)),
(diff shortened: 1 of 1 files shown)


wezterm-gui/src/glyphcache.rs
```diff
@@ -361,7 +361,7 @@ impl BlockKey {
         let c = c as u32;
         Some(match c {
             // BOX DRAWINGS LIGHT HORIZONTAL
-            0x2500 => Self::Poly(&[Poly {
+            0x2500 => Self::Poly(&[Poly { // ─
```
wez, line 364: I'd prefer it if the comment was on the line above the codepoint value; for the items that are labelled with the name, let's put it in with the comment; either at the LHS or the RHS of the line (RHS show in this suggestion, but I wonder if LHS might be better)

```suggestion
            // BOX DRAWINGS LIGHT HORIZONTAL  ─
            0x2500 => Self::Poly(&[Poly {
```

wezterm-gui/src/glyphcache.rs
```diff
@@ -3277,117 +3277,117 @@ impl BlockKey {
...
-            0x2592 => Self::Full(BlockAlpha::Medium),
-            0x2593 => Self::Full(BlockAlpha::Dark),
-            0x2594 => Self::Upper(1),
-            0x2595 => Self::Right(1),
-            0x2596 => Self::Quadrants(Quadrant::LOWER_LEFT),
-            0x2597 => Self::Quadrants(Quadrant::LOWER_RIGHT),
-            0x2598 => Self::Quadrants(Quadrant::UPPER_LEFT),
-            0x2599 => {
+            0x2590 => Self::Right(4), // ▐
+            0x2591 => Self::Full(BlockAlpha::Light), // ░
```
wez, line 3288:
```suggestion
            // ░
            0x2591 => Self::Full(BlockAlpha::Light),
```

wez (changes requested):
Thanks for this; now that we have these glyphs in wezterm I'm all for being able to see them in the code!

The `cargo fmt` check is failing because the commenting positioning isn't how it likes it; I think it would be good to put the comments before the codepoint -- see suggestions!

wezterm-gui/src/glyphcache.rs
bew: Hmm was thinking about this, but it wasn't easy to automate in vim, so I made this first impl for now.
I think LHS might be better, also I'd like to surround the char with `(X)` or `[X]` to better show where the char begins/ends (useful for the partial blocks)
What do you think?

wezterm-gui/src/glyphcache.rs
bew: So I add a comment line above ALL the oneline glyphs overrides?

wez: I think the comment works better when it precedes the codepoint value, as it is in the same horizontal position, which makes it easier to visually scan.

In hindsight, I should have put the codepoint description there for all of those glyphs when I started adding them; it didn't feel necessary when I started, but it definitely made sense when I did the box drawing glyphs.

If you're up for automating adding those in, that would be fantastic, but it's not required in this PR!

wezterm-gui/src/glyphcache.rs
wez: `[X]` sounds good, and I agree: LHS feels easier to visually scan

wezterm-gui/src/glyphcache.rs
bew: I can automate it quite easily if I have a list of all the codepoint descriptions and their corresponding 0x1234 notation, find me that and I'll do it in this PR 🙃

bew:
Okay, everything should be good!

Here is a little relaxing screencast of navigating in the code....
(me? crazy? nooo)

https://user-images.githubusercontent.com/9730330/124365675-cde51d80-dc49-11eb-805c-a9903aa742b2.mp4

wez:
This is great, many thanks!

PR #921, opened 2021-07-03
glyph: Make heavy lines in custom glyphs heavier

bew:
I think with 3.0 it is perfect for normal or bigger font sizes.
For smaller font size, it seems the stroke width is not adapted at all to the font size, is that intentional? Can we do something about it? (detect when the ratio is too big?)

See the difference:

<details>
<summary>BEFORE</summary>

### BEFORE
heavy_not_visible

</details>

<details>
<summary>AFTER</summary>

### AFTER
heavy_visible

</details>

<details>
<summary>AFTER (smaller font size)</summary>

### AFTER (smaller font size)
heavy_visible_small_font_size

</details>

<details>
<summary>AFTER (big font size)</summary>

### AFTER (big font size)
heavy_visible_big_font_size

</details>

---
The test file regroups every custom glyphs we currently have that mention `LIGHT` or `HEAVY` in their description, grouped by shapes


wez:
> For smaller font size, it seems the stroke width is not adapted at all to the font size

The width that is passed in comes from the underline height metric in the font which is proportional with the selected font size, but for small sizes it's possible for it to be < 1 pixel.  At that size, it's possible for 2x and even 3x something tiny to all get rounded up to the same ~1 pixel size.

Is that what you're seeing?  If so they perhaps it makes sense to `.round()` or `.ceil()` that value before passing it down from here:

https://github.com/wez/wezterm/blob/main/wezterm-gui/src/glyphcache.rs#L4264-L4268

bew:
I don't really know, honestly I think even like this it's ok.
Maybe if someone uses a small font size and have problems with this that I don't see, then we can do something about it?

bew:
Thanks @wez \o/

Issue #894, opened 2021-06-20
Follow OS theme for titlebar color


rashil2000:
**Is your feature request related to a problem? Please describe.**
The dark system theme on Windows is quite half-assed, so I tend to avoid it as much as possible. So I use light theme everywhere, but wezterm's always-dark titlebar does not look very good with any light config theme (for example BlulocoLight in wezterm.lua).

**Describe the solution you'd like**
Ideally it should be configurable in the config file itself, but I understand that the implementation of this on all platforms might be a hassle. So a really simple way would be to just follow the OS (Windows here) and provide a white and black version for the titlebar.

**Describe alternatives you've considered**
I have considered hiding the titlebar altogether, but that's not really ideal (it hides the window title, and moving/minimizing/closing all require some keyboard shortcuts)

**Additional context**
None

rashil2000:
Hi! 
Are there any plans for this?

wez:
It's something that I want to see happen, but haven't had time to do myself.  If you wanted to take a crack at making this work on windows and submit a PR, the relevant things are:

* https://github.com/wez/wezterm/blob/main/window/src/os/windows/window.rs#L762 this bit of code is what forces the title bar into dark mode
* Some logic would need to be added to determine what the system theme is set to and then call the above to apply it.  Maybe https://stackoverflow.com/a/51336913/149111
* Some logic would need to be added to determine *when* the system theme changes and then do the above.  https://docs.microsoft.com/en-us/windows/win32/winmsg/wm-wininichange is probably the thing to hook up in https://github.com/wez/wezterm/blob/main/window/src/os/windows/window.rs#L1873 
* Ideally there would a method on [WindowOps](https://github.com/wez/wezterm/blob/main/window/src/lib.rs#L128) to retrieve the current theme, and then expose that via the lua API to make it easy to do https://github.com/wez/wezterm/issues/806

rashil2000:
Hey, thanks for the detailed overview.

I see that a registry query is already being used here https://github.com/wez/wezterm/blob/main/window/src/os/windows/window.rs#L1065, we can follow similar logic to get the theme.

However I would note that I have no experience with Rust. Though it's definitely on my list, I'll try to take a shot at it if it's still open by that time.

rashil2000:
I have made tiny changes so that wezterm now follows OS theme. Note that it doesn't listen for any events, as such wezterm needs to be restarted if theme is changed. The diff is small enough to be pasted here:

```diff
[...]

PR #951, opened 2021-07-17
Remove hardcoded logic for titlebar color

rashil2000:
Initial work for #894


wez:
FWIW, I think adding:

```rust
   WM_SETTINGCHANGE => enable_theme(hwnd),
```

in around here:

https://github.com/wez/wezterm/blob/main/window/src/os/windows/window.rs#L1887-L1889

might be all you need to do to get it changing in response to the user changing the theme

rashil2000:
Done!

wez:
This works well, many thanks!

PR #961, opened 2021-07-19
Absolute working directory paths for CreateProcessW()

exactly-one-kas:
As mentioned in #900, CreateProcessW() expects a full path and behaves weirdly if the shell / parent process switched drives before with a relative path.


wez:
Thanks!

PR #972, opened 2021-07-23
tmux-cc: add support for layout-change

skyline75489:


tmux-cc/src/tmux.pest
```diff
@@ -23,6 +24,7 @@ window_renamed = { "%window-renamed " ~ window_id ~ " " ~ any_text }
 session_changed = { "%session-changed " ~ session_id ~ " " ~ any_text }
 session_renamed = { "%session-renamed " ~ any_text }
 session_window_changed = { "%session-window-changed " ~ session_id ~ " " ~ window_id }
+layout_change = { "%layout-change " ~ window_id ~ " " ~ (window_layout ~ " " ~ window_layout ~ " " ~any_text | window_layout) }
```
skyline75489, line 27: There's two types of `layout_change`. In tmux 2.x it is:

`%layout-change <window_id> <window_layout>`

in tmux 3.x it's

`%layout-change <window_id> <window_layout> <window_visible_layout> <window_raw_flags>`

I'm not quite sure the exact version where this new version is used. But I think the code should be compatible with both versions.

See https://github.com/tmux/tmux/blob/6c2bf0e22119804022c8038563b9865999f5a026/control-notify.c#L51

tmux-cc/src/lib.rs
```diff
@@ -82,6 +82,12 @@ pub enum Event {
         window: TmuxWindowId,
         name: String,
     },
+    LayoutChange {
+        window: TmuxWindowId,
+        layout_id: String,
+        width: u64,
+        height: u64
+    }
```
wez, line 90: I'm not sure of their significance/utility, but ideally this struct would have `Option`al fields to hold the `<window_visible_layout>` `<window_raw_flags>` from tmux 3; can you add those too?

wez (commented):
Thanks for picking this up!

tmux-cc/src/lib.rs
skyline75489: I don’t think the last two parameters  are really useful at the moment, because the implementation will need to first be capable of handling the old format. I think we can just use the old format now and leave the last two unused for the initial implementation. 

On that note, how do your think wezterm should behave on layout-change, for example the desired layout is smaller than the current window? I want to make tmux-cc happen and I could use some suggestions here. I am very new to the codebase.

wez: > I don’t think the last two parameters are really useful at the moment, because the implementation will need to first be capable of handling the old format. I think we can just use the old format now and leave the last two unused for the initial implementation.

Implementation wise, I agree.  The `tmux-cc` crate is just a parser though, so it would be great if we can have it faithfully parse the data out even if we don't use it right away.

> On that note, how do your think wezterm should behave on layout-change, for example the desired layout is smaller than the current window? I want to make tmux-cc happen and I could use some suggestions here. I am very new to the codebase.

If tmux's size is smaller than wezterm's, then I don't think we need to do much that is special: we can pad out the "missing" areas with blanks when we render.

wezterm has a binary tree layout system in the `Tab` struct (in `mux/src/tab.rs`) for keeping track of panes, and one of the challenges of hooking up tmux-cc mode is mapping tmux's panes to it.

The binary tree has some nice properties with respect to constraining the splits for manually adjusting the layout.

In theory, wezterm doesn't really need to know about any of the specifics of tmux's layouts so long as tmux reports the effective size and positions of the panes when the layout changes.  I *think* that all tmux layouts should be able to be mapped to some binary tree (the tmux implementation is itself based on a tree, but not strictly a binary tree), I just haven't had the time to figure out the algorithm for that yet. That mapping operation is logically equivalent to the `Tab::sync_with_pane_tree` method that is used in wezterm's mux protocol, but the difference is that wezterm's mux data is already tree-shaped, whereas the tmux data is a tree rendered as a list.

wez:
The build wants you to `cargo fmt` these changes! :)

tmux-cc/src/lib.rs
skyline75489: Thanks for all the suggestions! 

I've managed to add support for visiable_layout and flags. I'm new to rust so I might be writing something silly :sweat_smile:

tmux-cc/src/lib.rs
```diff
@@ -150,19 +157,25 @@ fn parse_guard(mut pairs: Pairs<Rule>) -> anyhow::Result<(i64, u64, i64)> {
...
+            if let (Some(layout_id), Some(width), Some(height)) =
+                (layout_id_option, width_option, height_option)
+            {
+                return Option::Some(WindowLayout {
+                    layout_id,
+                    width,
+                    height,
+                });
+            }
+            return Option::None;
```
wez, line 176:
```suggestion
            return None;
```

tmux-cc/src/lib.rs
```diff
@@ -150,19 +157,25 @@ fn parse_guard(mut pairs: Pairs<Rule>) -> anyhow::Result<(i64, u64, i64)> {
...
-            let width = pairs.next().unwrap().as_str().parse::<u64>()?;
-            let height = pairs.next().unwrap().as_str().parse::<u64>()?;
-            Ok((layout_id, width, height))
+            let layout_id_option = pairs.next()?.as_str().parse::<String>().ok();
+            let width_option = pairs.next()?.as_str().parse::<u64>().ok();
+            let height_option = pairs.next()?.as_str().parse::<u64>().ok();
+            if let (Some(layout_id), Some(width), Some(height)) =
+                (layout_id_option, width_option, height_option)
+            {
+                return Option::Some(WindowLayout {
```
wez, line 170:
```suggestion
                return Some(WindowLayout {
```

tmux-cc/src/lib.rs
```diff
@@ -150,19 +157,25 @@ fn parse_guard(mut pairs: Pairs<Rule>) -> anyhow::Result<(i64, u64, i64)> {
...
+                    height,
+                });
+            }
+            return Option::None;
         }
-        _ => anyhow::bail!(
-            "parse_window_layout can only parse Rule::window_layout, got {:?}",
-            pair
-        )
+        _ => Option::None,
```
wez, line 178:
```suggestion
        _ => None,
```

tmux-cc/src/lib.rs
```diff
@@ -264,8 +277,17 @@ fn parse_line(line: &str) -> anyhow::Result<Event> {
         Rule::layout_change => {
             let mut pairs = pair.into_inner();
             let window = parse_window_id(pairs.next().unwrap())?;
-            let (layout_id, width, height) = parse_window_layout(pairs.next().unwrap())?;
-            Ok(Event::LayoutChange { window, layout_id, width, height })
+            let layout = pairs.next().and_then(parse_window_layout).unwrap();
+            let visible_layout = pairs.next().and_then(parse_window_layout);
+            let raw_flags = pairs
+                .next()
+                .and_then(|r| Option::Some(r.as_str().to_owned()));
```
wez, line 284: I think you can simplify this:

```suggestion
                .map(|r| r.as_str().to_owned());
```

tmux-cc/src/lib.rs
```diff
@@ -663,15 +685,27 @@ here
...
                     window: 1,
-                    layout_id: "b25d".to_owned(),
-                    width: 80,
-                    height: 24
+                    layout: WindowLayout {
+                        layout_id: "b25d".to_owned(),
+                        width: 80,
+                        height: 24
+                    },
+                    visible_layout: Option::None,
```
wez, line 693:
```suggestion
                    visible_layout: None,
```

tmux-cc/src/lib.rs
```diff
@@ -663,15 +685,27 @@ here
...
+                    raw_flags: Option::None
```
wez, line 694:
```suggestion
                    raw_flags: None
```

tmux-cc/src/lib.rs
```diff
@@ -663,15 +685,27 @@ here
...
                     window: 1,
-                    layout_id: "cafd".to_owned(),
-                    width: 120,
-                    height: 29
+                    layout: WindowLayout {
+                        layout_id: "cafd".to_owned(),
+                        width: 120,
+                        height: 29
+                    },
+                    visible_layout: Option::Some(WindowLayout {
```
wez, line 703:
```suggestion
                    visible_layout: Some(WindowLayout {
```

tmux-cc/src/lib.rs
```diff
@@ -663,15 +685,27 @@ here
...
+                        layout_id: "cafd".to_owned(),
+                        width: 120,
+                        height: 29
+                    }),
+                    raw_flags: Option::Some("*".to_owned())
```
wez, line 708:
```suggestion
                    raw_flags: Some("*".to_owned())
```

wez:
Thanks!

PR #977, opened 2021-07-24
tmux-cc: force DcsPassThrough in tmux-cc mode

skyline75489:
The thing about tmux-cc is that the pane-output actually contains escape sequences.

> The output has any characters less than ASCII 32 and the \ character replaced with their octal equivalent, so \ becomes \134. Otherwise, it is exactly what the application running in the pane sent to tmux. It may not be valid UTF-8 and may contain escape sequences which will be as expected by tmux

See: https://github.com/tmux/tmux/wiki/Control-Mode#pane-output

When the app is in tmux-cc mode, all characters received must be passed through the the DCS handler. The tmux-cc mode can only be terminated by "a %exit line and a corresponding ST (\033\) sequence"

### DIFF
--- a/termwiz/src/escape/parser/mod.rs
+++ b/termwiz/src/escape/parser/mod.rs
@@ -23,6 +23,7 @@ struct SixelBuilder {
 struct ParseState {
     sixel: Option<SixelBuilder>,
     dcs: Option<ShortDeviceControl>,
+    tmux_cc: bool
 }
 
 /// The `Parser` struct holds the state machine that is used to decode
@@ -55,7 +56,13 @@ impl Parser {
             callback: &mut callback,
             state: &mut self.state.borrow_mut(),
         };
-        self.state_machine.parse(bytes, &mut perform);
+
+        if perform.state.tmux_cc {
+            self.state_machine.force_dcs_passthrough(bytes, &mut perform);
+        }
+        else {
+            self.state_machine.parse(bytes, &mut perform);
+        }
     }
 
     /// A specialized version of the parser that halts after recognizing the
@@ -173,6 +180,9 @@ impl<'a, F: FnMut(Action)> VTActor for Performer<'a, F> {
                 data: vec![],
             });
         } else {
+            if byte == b'p' && params == vec![1000] {
+                self.state.tmux_cc = true;
+            }
             (self.callback)(Action::DeviceControl(DeviceControlMode::Enter(Box::new(
                 EnterDeviceControlMode {
                     byte,
--- a/vtparse/src/lib.rs
+++ b/vtparse/src/lib.rs
@@ -679,6 +679,14 @@ impl VTParser {
             self.parse_byte(*b, actor);
         }
     }
+
+    /// Force DCS pass through for tmux cc
+    pub fn force_dcs_passthrough(&mut self, bytes: &[u8], actor: &mut dyn VTActor) {
+        for b in bytes {
+            let action = Action::Put;
+            self.action(action, *b, actor);
+        }
+    }
 }
 
 #[cfg(test)]


skyline75489:
The issue in the current implementation is that if `%output` contains a `\x9c`, the state machine (in localpane) will return to ground state, because it is the correct thing to do with normal terminal output. However with tmux-cc mode, that's not the case.

My implementation here is obviously a naive one. I hope my description of the issue is clear.

termwiz/src/escape/parser/mod.rs
```diff
@@ -173,6 +180,9 @@ impl<'a, F: FnMut(Action)> VTActor for Performer<'a, F> {
                 data: vec![],
             });
         } else {
+            if byte == b'p' && params == vec![1000] {
```
wez, line 183: I think we can avoid making a vec for this; the vec is undesirable because it represents an unconditional heap allocation just for a comparison.

```suggestion
            if byte == b'p' && params == [1000] {
```

termwiz/src/escape/parser/mod.rs
```diff
@@ -173,6 +180,9 @@ impl<'a, F: FnMut(Action)> VTActor for Performer<'a, F> {
...
+                self.state.tmux_cc = true;
```
wez, line 184: I don't see anything that will exit this state, so as-written, this feels like it will "break" the pane when the mode is exited :-/

vtparse/src/lib.rs
```diff
@@ -679,6 +679,14 @@ impl VTParser {
             self.parse_byte(*b, actor);
         }
     }
+
+    /// Force DCS pass through for tmux cc
+    pub fn force_dcs_passthrough(&mut self, bytes: &[u8], actor: &mut dyn VTActor) {
+        for b in bytes {
```
skyline75489, line 685: Is there a way to batch pass through the entire buffer? That would help the performance a lot.

termwiz/src/escape/parser/mod.rs
skyline75489: I think in iTerm2 you need to manually break the state. In Wezterm I guess "press q"? I'll see where to add this.

vtparse/src/lib.rs
wez, line 685: Ah, this bit is in vtparse.  Layering wise this makes things very difficult.  I'd suggest doing this rerouting in the termwiz crate where we have the higher level state instead of here

termwiz/src/escape/parser/mod.rs
```diff
@@ -55,7 +56,13 @@ impl Parser {
             callback: &mut callback,
             state: &mut self.state.borrow_mut(),
         };
-        self.state_machine.parse(bytes, &mut perform);
+
+        if perform.state.tmux_cc {
+            self.state_machine.force_dcs_passthrough(bytes, &mut perform);
```
wez, line 61: I haven't looked to make sure this suggestion is good, but I think that we could add `DeviceControlMode::Bytes(Vec<u8>)` and pass that bytes through at the termwiz layer rather than having this special state cut across to the vtparse layer.

For that to work, I think we need a bit more state for tmux in here, so that we can match for a sequence like `\n%exit` to break out of tmux-cc mode.

When in tmux-cc mode, we can simply bypass the vtparse parser and redirect straight through to `DeviceControlMode::Bytes` which would get handled in the same way as `DeviceControlMode::Data`, but with the advantage that it's a bit more efficient to pass the vec through there, rather than the individual `Data` items.

vtparse/src/lib.rs
skyline75489: Understood. I'll try to think of something else. By the way I have an experimental branch tmux-cc at https://github.com/skyline75489/wezterm/tree/dev/tmux-cc-prototype, which handles the output part, very roughly but works. Take a look if you want :) 

(bear with all my amateur Rust code)

skyline75489:
Closed in favor of #1090

PR #985, opened 2021-07-28
tmux-cc: handle list-pane response

skyline75489:


mux/Cargo.toml
```diff
@@ -48,3 +48,4 @@ winapi = { version = "0.3", features = [
 
 [dev-dependencies]
 k9 = "0.11"
+pretty_assertions = "0.6"
```
skyline75489, line 51: I hope it's OK to add it here.

mux/src/tmux.rs
```diff
@@ -139,12 +145,9 @@ impl TmuxDomainState {
                         let cmd = self.cmd_queue.borrow_mut().pop_front().unwrap();
                         let domain_id = self.domain_id;
                         *self.state.borrow_mut() = State::Idle;
-                        promise::spawn::spawn(async move {
-                            if let Err(err) = cmd.process_result(domain_id, &response) {
-                                log::error!("error processing result: {}", err);
-                            }
-                        })
-                        .detach();
+                        if let Err(err) = cmd.process_result(domain_id, &response, &mut self.inner.borrow_mut()) {
```
skyline75489, line 148: I don't know enough Rust to make it compile. It asks for 'static' lifetime of `self`.

I see the intention here is to not block the current thread while processing the data. I feel this right now is also OK? Because it helps debugging a bit if it does NOT switch to another thread. Plus no one can really use tmux-cc at this moment.

mux/src/tmux.rs
```diff
@@ -107,16 +108,21 @@ impl TmuxCommand for ListAllPanes {
...
         }
 
-        log::error!("panes in domain_id {}: {:?}", domain_id, items);
+        log::error!("panes in domain_id {}: {:?}", domain_id, state.panes);
         Ok(())
     }
 }
 
+struct TmuxInnerState {
+    panes: Vec<TmuxPane>
```
skyline75489, line 117: Maybe a better structure would be HashMap<PaneIndex, TmuxPane>? I'm not sure.

mux/src/tmux.rs
wez: The CI didn't get to compile this because the changes weren't `cargo fmt`'d, so I can't see the actual error message :-(

I think you might need to *reborrow* this by adding a `*` to deref the `RefMut` to the underlying state struct: `&mut* self.inner.borrow_mut()`, but it may be the case that you need to put the `RefMut` into a temporary to keep it alive.

I'm just speculating though!

wez (commented):
Is the idea to maintain a view of tmux's pane state, and then synchronize that entire state with wezterm's tab/pane state?

skyline75489:
> Is the idea to maintain a view of tmux's pane state, and then synchronize that entire state with wezterm's tab/pane state?

Yeah. That’s basically the idea. However I think it’s still too early to finalize the design. I am trying to write something actionable and testable and hope that will be useful in the future.

mux/src/tmux.rs
skyline75489: Sry I still can't make it work :sweat: . It tells me:

```text
`self` has an anonymous lifetime `'_` but it needs to satisfy a `'static` lifetime requirement
```

mux/src/tmux.rs
```diff
@@ -245,3 +251,72 @@ impl Domain for TmuxDomain {
...
+            flags: 0,
+            error: false,
+            output: "
+$0 @0 %0 0 2 37 104 38 0 0
+$1 @1 %1 0 2 37 104 38 0 0
+$2 @2 %2 0 2 2 80 24 0 0
+            ".to_owned()
+        };
+
+        let tmux_inner_state = &mut domain_state.inner.borrow_mut();
```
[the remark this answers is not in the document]
skyline75489: Thanks. I'm really new to rust. But somehow this compiles and runs? I'll try to learn more rust later.

skyline75489:
Closed in favor of #1090

PR #989, opened 2021-07-31
Added watch_path

AusCyberman:
Added a watch_path global variable and function `add_to_config_reload_watch_list` which takes a variadic list of paths to append to the global watch_list

and then inside `lib.rs` checks variable and watches those specific files.
Also added a flake.nix, which was required for me to be able to develop on this, it can't build the project but you can cargo build and then you need to run the nixosGLPatch, which kinda works, but it's good enough


flake.lock
```diff
@@ -0,0 +1,98 @@
+{
```
wez, line 1: I'm not really a fan of these unrelated files landing in the root of the repo.
I'd prefer to have these non-rust changes split out from this PR (as they don't relate to the main thesis of this PR), and either not have them in the repo at all, or to be encapsulated in a sub-directory if possible.  For example, do they make sense to put under a directory like `assets/nix`?

wez (changes requested):
Many thanks for taking a run at this!
I have some stylistic suggestions, and I think we should use the lua registry rather than directly use lua globals.

I'd prefer not to have the flake/nix artifacts in the repo if possible, as there's nothing to keep them in working order, and they appear to be largely machine generated anyway.  I'm not opposed to having a more convenient way for nix users to iterate locally, but I don't want to resent nix users every time I run `ls` :-)

flake.lock
AusCyberman: i'm pretty sure that would work, i just need them to exist under version control otherwise nix flakes cannot properly manage my developer environment

config/src/lua.rs
```diff
@@ -71,19 +71,28 @@ pub fn make_lua_context(config_file: &Path) -> anyhow::Result<Lua> {
...
                 .to_str()
                 .ok_or_else(|| anyhow!("config dir path is not UTF-8"))?,
         )?;
+        wezterm_mod.set(
+            "add_to_config_reload_watch_list",
+            lua.create_function(|lua, args: Variadic<String>| {
+                let globals = lua.globals();
+                let watch_path: String = globals.get("watch_path")?;
+                log::info!("watch_path: {}, args: {:?}", watch_path, args);
+                globals.set("watch_path", format!("{};{}", watch_path, args.join(";")))?;
```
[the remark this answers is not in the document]
AusCyberman: i kinda thought that it makes sense to set it as a global, so people can manually set it, like i could get rid of the default path or write an external function that does it for me, as opposed to having to go through the function, but still leave it as an option

config/src/lib.rs
```diff
@@ -474,16 +474,28 @@ impl ConfigInner {
                 // call to `with_lua_config` to reference this lua context
                 // even though we are (probably) resolving this from a background
                 // reloading thread.
-                if let Some(lua) = lua {
+                lua.map(|lua| {
```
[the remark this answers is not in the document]
AusCyberman: fair, but as i said, it doesn't really work with the idea of being able to wipe or arbitrarily set the watch_path

flake.lock
bew: @AusCyberman for the flakes, when it's untracked you can tell git you intend to add it **but without staging the files** right away, with `git add -N` (or the long form `git add --intend-to-add`)
It'swhat I do for a few projects where I want to have a Nix flake but not commit it for a PR

AusCyberman: i've already added them tho, so what now?

bew: To clean up this PR from Nix stuff, I'd recommend something like this:
```sh
git rm --cached flake.*  # remove files from git
git commit -m "Remove Nix flakes files" flake.*
git add -N flake.*  # re-add them to git without staging them
```

bew: Hmm and you may have a few more files to remove, like cargo.nix and maybe also crate-hashes.json ? (I'm not entirely familiar with the rust toolchains in Nix)

nixosGLPatch
```diff
@@ -0,0 +1 @@
+patchelf --add-needed /nix/store/zf44435sdgyys6rxanbbyns62n7fzv7j-libglvnd-1.3.3/lib/libEGL.so target/debug/wezterm-gui
```
bew, line 1: Remove this file as well!

config/src/lua.rs
```diff
@@ -71,19 +71,36 @@ pub fn make_lua_context(config_file: &Path) -> anyhow::Result<Lua> {
...
         wezterm_mod.set(
             "config_dir",
             config_dir
                 .to_str()
                 .ok_or_else(|| anyhow!("config dir path is not UTF-8"))?,
         )?;
+        wezterm_mod.set(
+            "add_to_config_reload_watch_list",
+            lua.create_function(|lua, args: Variadic<String>| {
+                let watch_path: String = lua.named_registry_value("wezterm-watch-paths")?;
```
wez, line 89: Rather than splitting/joining strings, let's just make this an array:

```suggestion
                let mut watch_paths: Vec<String> = lua.named_registry_value("wezterm-watch-paths")?;
```

config/src/lua.rs
```diff
@@ -71,19 +71,36 @@ pub fn make_lua_context(config_file: &Path) -> anyhow::Result<Lua> {
...
+                log::debug!("watch_path: {}, args: {:?}", watch_path, args);
+                lua.set_named_registry_value(
+                    "wezterm-watch-paths",
+                    format!("{};{}", watch_path, args.join(";")),
+                )
+            })?,
+        )?;
+        wezterm_mod.set(
+            "set_config_reload_watch_list",
```
wez, line 98: I'm kinda leaning towards not including this; the way I see this working is purely additive.  The ability to set the complete list means that there is potential for one module to trample on things set by other modules, which feels bad if there isn't a corresponding getter function, and I think also adding *that* will have an outsize maintenance cost compared to the basic need to just add a path.

config/src/lua.rs
AusCyberman: does it automatically add dividers or what?

config/src/lua.rs
AusCyberman: I think that adding a getter does make sense. as I would want to be able to make it ignore the default file, especially performance-wise.

wez:
Thanks!

PR #1043, opened 2021-08-18
IME support on X11

H-M-H:
Add support for Input Method Editors (IME) on X11.

### DIFF
--- a/.github/workflows/gen_centos7.yml
+++ b/.github/workflows/gen_centos7.yml
@@ -88,7 +88,7 @@ jobs:
       run: "source /opt/rh/devtoolset-9/enable && cargo build --all --release"
     - name: "Test (Release mode)"
       shell: bash
-      run: "cargo test --all --release"
+      run: "source /opt/rh/devtoolset-9/enable && cargo test --all --release"
     - name: "Package"
       shell: bash
       run: "bash ci/deploy.sh"
--- a/.github/workflows/gen_centos7.yml
+++ b/.github/workflows/gen_centos7.yml
@@ -33,7 +33,7 @@ jobs:
       run: |
 
             yum install -y wget curl-devel expat-devel gettext-devel openssl-devel zlib-devel centos-release-scl-rh perl-ExtUtils-MakeMaker make
-            yum install -y devtoolset-9-gcc
+            yum install -y devtoolset-9-gcc devtoolset-9-gcc-c++
 
             if test ! -x /usr/local/git/bin/git ; then
                 cd /tmp
--- a/.github/workflows/gen_centos7.yml
+++ b/.github/workflows/gen_centos7.yml
@@ -31,9 +31,10 @@ jobs:
     - name: "Install Git from source"
       shell: bash
       run: |
-            
-            yum install -y wget curl-devel expat-devel gettext-devel openssl-devel zlib-devel gcc perl-ExtUtils-MakeMaker make
-            
+
+            yum install -y wget curl-devel expat-devel gettext-devel openssl-devel zlib-devel centos-release-scl-rh perl-ExtUtils-MakeMaker make
+            yum install -y devtoolset-9-gcc
+
             if test ! -x /usr/local/git/bin/git ; then
                 cd /tmp
                 wget https://github.com/git/git/archive/v2.26.2.tar.gz
@@ -41,9 +42,9 @@ jobs:
                 cd git-2.26.2
                 make prefix=/usr/local/git install
             fi
-            
+
             ln -s /usr/local/git/bin/git /usr/local/bin/git
-                    
+
 
     - name: "Install curl"
       shell: bash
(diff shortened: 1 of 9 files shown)


window/src/os/x11/window.rs
```diff
@@ -872,6 +896,22 @@ impl XWindowInner {
...
+        self.update_ime_position();
+    }
+
+    fn update_ime_position(&mut self) {
+        if !self.has_focus {
+            return;
+        }
+        self.conn().ime.borrow_mut().update_pos(
+            self.window_id,
+            self.last_cursor_position.max_x() as i16,
```
wez, line 910: I think this might look slightly better if the x value is in the middle of the rect (eg: `(max_x() + min_x()) / 2`) as, at least on Gnome, the candidate window points to the right hand edge which looks slightly "off":

Screenshot from 2021-08-18 08-04-16

I don't know if other IMEs look different though!

window/src/os/x11/connection.rs
```diff
@@ -441,7 +454,17 @@ impl XConnection {
 
         let default_dpi = RefCell::new(compute_default_dpi(&xrm, &xsettings));
 
-        let conn = XConnection {
+        xcb_imdkit::ImeClient::set_logger(|msg| log::debug!("Ime: {}", msg));
```
wez, line 457: the debug log lines have extra newlines; this will eliminate them, but you may want to `trim_end` the string that you're passing to the logger callback in your crate so that this implementation detail is hidden :)

```suggestion
        xcb_imdkit::ImeClient::set_logger(|msg| log::debug!("Ime: {}", msg.trim_end()));
```

window/src/os/x11/window.rs
```diff
@@ -872,6 +896,22 @@ impl XWindowInner {
         }
     }
 
+    fn set_text_cursor_position(&mut self, cursor: Rect) {
+        self.last_cursor_position = cursor;
```
wez, line 900: re: discussion about laggy input and IME positioning being expensive, perhaps a simple:

```rust
   if self.last_cursor_position == cursor {
       return;
   }
```

will help with this!

H-M-H: Yes, I think this does the trick!

window/src/os/x11/window.rs
H-M-H: I checked how other gui frameworks do it and consensus seems to be to set the IME position under the cursor to the left, so I changed the x position to `min_x`.

wez:
The centos7 build appears to be failing because the fcitx ime stuff requires C11 compiler features and those aren't available in the default compiler on centos7.

I think the resolution for that platform is to arrange for the CI to install `devtoolset-9` which includes this compiler:
https://centos.pkgs.org/7/centos-sclo-rh-x86_64/devtoolset-9-gcc-9.1.1-2.6.el7.x86_64.rpm.html
and then prefix PATH with `/opt/rh/devtoolset-9/root/usr/bin`.  I'd suggest hacking on https://github.com/wez/wezterm/blob/main/.github/workflows/gen_centos7.yml to figure that out.  Then we can figure out how to phrase that in the python script that generates the workflows once it's nailed down.

The FreeBSD build appears to be failing because it can't find the headers it needs. I think another or a differently named package is needed there, but I don't know what that is :)
Two possible candidates, but pkgs.org can't show the list of files in them, so I don't know for sure
https://freebsd.pkgs.org/12/freebsd-amd64/xcb-proto-1.14.1.txz.html
https://freebsd.pkgs.org/12/freebsd-amd64/libxcb-1.14_1.txz.html

H-M-H:
Downloading and looking into `libxcb` reveals that this does include the missing headers on FreeBSD. Haven't looked at Centos7 yet though.

.github/workflows/gen_centos7.yml
```diff
@@ -84,7 +85,7 @@ jobs:
       run: "cargo fmt --all -- --check"
     - name: "Build (Release mode)"
       shell: bash
-      run: "cargo build --all --release"
+      run: "source /opt/rh/devtoolset-9/enable && cargo build --all --release"
```
wez, line 88: almost there!  I think you also need this for the `cargo test` below

wez:
Looks good!  I have a commit queued up that implements the CI changes in the python generator, and I'm happy to merge this now if you don't plan to make any last minute tweaks!

window/Cargo.toml
```diff
@@ -73,6 +73,7 @@ smithay-client-toolkit = {version = "0.14", default-features=false, optional=tru
 wayland-protocols = {version="0.28", optional=true}
 wayland-client = {version="0.28", optional=true}
 wayland-egl = {version="0.28", optional=true}
+xcb-imdkit = { version="0.1", git = "https://github.com/H-M-H/xcb-imdkit-rs" }
```
wez, line 76: BTW, when I cut the next release, I will set this a specific commit hash from your repo if you haven't made a crates.io release by that point.  That way my release tag will be repeatable!

H-M-H:
Excellent! I will probably publish my `xcb-imdkit-rs` to crates.io today or tomorrow, so you may want to wait until then.

wez:
I'll merge it now so that folks can play around with it in the nightly and because I'm too lazy to juggle a branch just for the CI changes here :-)
We can easily pin the version when you're ready!

Issue #1052, opened 2021-08-20
Provide Debian 11 Bullseye release


kar1:
**Is your feature request related to a problem? Please describe.**
When I run the latest nightlies on Debian 11 Bullseye I get the error:
```sh
/usr/bin/wezterm-gui: error while loading shared libraries: libxcb-util.so.0: cannot open shared object file: No such file or directory
```
In Debian 11 there is no `libxcb-util.so.0` but a `libxcb-util.so.1`. I think it's just a matter of building on a Debian 11 so it gets linked to the new libxcb-util version there. I have tried building from source on my Debian 11 system without any problems.

**Describe the solution you'd like**
A precompiled Debian 11 .deb version of Wezterm.

**Describe alternatives you've considered**
I'm not sure if you would like me to try and build it on github's ci system? I would copy the files:

```
gen_debian10.3.yml
gen_debian10.3_continuous.yml
gen_debian10.3_tag.yml
```

and replace debian10.3 with debian11 and see what happens :-)? Should I do that now or maybe next time? Is there a way I can build it on Github Actions without building the other releases and without uploading etc. Please advice, thanks.

wez:
If you want to make a PR for this the steps are:

* Add a debian 11 line around here: https://github.com/wez/wezterm/blob/main/ci/generate-workflows.py#L557
* Run `ci/generate-workflows.py`
* Add the changes to the commit and a PR from it
* See if any other special cases are needed for debian11, and adapt, repeat!

kar1:
Thanks for the help. Lets see if it runs #1053

wez:
Once that lands, the next step is to add a Debian11 line to https://raw.githubusercontent.com/wez/wezterm/main/docs/install/linux.markdown

Since we don't yet have a stable release, what I usually do is make that cell say something like "nightly only", and then remember to go back and fill it in once a new release is made!

kar1:
Thank you.

wez:
Thank you!

kar1:
The new debian11 link in the docs is not being processed.

I think I missed adding a debian11 line here:
https://github.com/wez/wezterm/blob/1e52016359338f88507a35dcf014aea0f7e116bf/ci/subst-release-info.py#L20

Is that correct? Should I push to the same branch?

PS - I really appreciate you taking the time to help me on this.

jameschensmith:
[...]

PR #1053, opened 2021-08-20
ci: Added Debain 11 target

kar1:
Provide Debian 11 Bullseye release #1052

### DIFF
--- a/.github/workflows/gen_debian11.yml
+++ b/.github/workflows/gen_debian11.yml
@@ -1,0 +1,92 @@
+
+name: debian11
+
+on:
+  pull_request:
+    branches:
+    - main
+    paths-ignore:
+    - '.cirrus.yml'
+    - 'docs/*'
+    - 'ci/build-docs.sh'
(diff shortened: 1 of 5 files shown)


.github/workflows/gen_debian11.yml
```diff
@@ -0,0 +1,92 @@
...
+            ~/.cargo/git
+            target
+
+         key: "debian11-None-2-${{ runner.os }}-${{ hashFiles('**/Cargo.lock') }}-cargo"
+    - name: "Install System Deps"
+      shell: bash
+      run: " env PATH=$PATH ./get-deps"
+    - name: "Check formatting"
+      shell: bash
+      run: "cargo fmt --all -- --check"
```
SuperSandro2000, line 71: Isn't it wasteful to do this for every OS? Formatting should be the same for all, right?

wez: Formatting doesn't cost anything compared to the rest of the stuff that CI does, so I'm not worried about this

PR #1055, opened 2021-08-21
Debian11 download link preprocessing

kar1:
The download link wasn't getting processed in the docs for debian11. I think it's because I forgot adding this line:

```python
 r"Debian11.deb$": "debian11_deb",
```

to the `CATEGORIZE` variable in `ci/subst-release-info.py`

### DIFF
--- a/.github/workflows/gen_debian11.yml
+++ b/.github/workflows/gen_debian11.yml
@@ -1,0 +1,92 @@
+
+name: debian11
+
+on:
+  pull_request:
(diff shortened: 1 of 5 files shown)


kar1:
I'm a little unsure about the commit history here. I just pushed the commit on my branch, but I can see that it includes my former commits. Should I have sync'ed my branch before commit'ing or just made a entirely new branch, after the pr merge?

PR #1057, opened 2021-08-21
ci: Forgot adding Debain 11 to the docs link builder

kar1:
The download link wasn't getting processed in the docs for debian11. I think it's because I forgot adding this line:

 r"Debian11.deb$": "debian11_deb",

to the CATEGORIZE variable in ci/subst-release-info.py


wez:
Ah thanks! I thought I'd done something smarter than this to recognize the different version numbers, but apparently not!

PR #1059, opened 2021-08-21
Feature: add popup support

malkomalko:
Talked about in issue: https://github.com/wez/wezterm/issues/270

Keeping this as a WIP as I explore and push up how to make this happen.

### DIFF
--- a/wezterm-mux-server-impl/src/sessionhandler.rs
+++ b/wezterm-mux-server-impl/src/sessionhandler.rs
@@ -472,6 +472,14 @@ impl SessionHandler {
                 .detach();
             }
 
+            Pdu::SpawnPopup(spawn_popup) => {
+                let sender = self.to_write_tx.clone();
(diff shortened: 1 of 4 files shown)


wezterm-mux-server-impl/src/sessionhandler.rs
```diff
@@ -795,3 +811,64 @@ async fn domain_spawn_v2(spawn: SpawnV2, sender: PduSender) -> anyhow::Result<Pd
         size: tab.get_size(),
     }))
 }
+
+async fn domain_spawn_popup(spawn_popup: SpawnPopup, sender: PduSender) -> anyhow::Result<Pdu> {
```
malkomalko, line 815: This is a copy and paste of `domain_spawn` above for the moment to get build working as I work my way through the understanding of this code.  As I start implementation next of the cli command, this more than likely will change, and if it doesn't, I'll just reuse the previous implementation.

PR #1066, opened 2021-08-22
no close-on-exec for stdio file descriptors

mcr:
redirect_stdio_impl incorrectly marks new file descriptor as close-on-exec.
This is because of the dup2/dup3 calls buried down there.  
Changing this to call cloexec() only when needed seems difficult to get right for all paths through, so instead, just unmark the file descriptor as close-on-exec.
Looked for, but did not find, any test cases on this code, would like to add a test case here.

### DIFF
--- a/filedescriptor/src/unix.rs
+++ b/filedescriptor/src/unix.rs
@@ -275,6 +275,20 @@ impl FileDescriptor {
         OwnedHandle::dup2_impl(f, dest_fd).map(|handle| Self { handle })
     }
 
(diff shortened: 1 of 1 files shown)


mcr:
https://github.com/tokio-rs/tokio/issues/4053

wez:
This seems reasonable to me, thanks!
@robinbernon any objection to this?

robinbernon:
> This seems reasonable to me, thanks!
> @robinbernon any objection to this?

Sounds good to me 👍

wez:
OK, this is now published as filedescriptor 0.8.1

PR #1078, opened 2021-08-24
Add Chocolatey as an install option

corbob:
Add Chocolatey as an install option for Windows.

### DIFF
--- a/docs/install/windows.markdown
+++ b/docs/install/windows.markdown
@@ -38,4 +38,17 @@ scoop itself can be installed like so:
 scoop bucket add extras
 scoop install wezterm
 ```
+### For `Chocolatey` users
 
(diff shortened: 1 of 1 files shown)


docs/install/windows.markdown
```diff
@@ -38,4 +38,17 @@ scoop itself can be installed like so:
...
+### For `Chocolatey` users
 
+If you prefer to use [Chocolatey](https://chocolatey.org) to manage software, wezterm is availabe from the Community Repository.
+It can be installed like so:
+
+```bash
+choco install wezterm -y
+```
+
+Alternatively, if you'd prefer the Portable installation:
```
wez, line 50: Could you have this say something about why someone might want the portable version?
The setup.exe installer deploys registry entries for "open in wezterm" and that doesn't have an equivalent/alternative in the portable zip install.  What are the benefits of the portable install when using choco?

wez (commented):
Thanks for this!

PR #1081, opened 2021-08-25
Remove TryExec to fix desktop file on Debian Bookworm

SuperSandro2000:
My Debian Bookworm with Plasma KDE is really really unhappy about this line and refuses to do open wezterm at all. When I remove it everything works.

image


wez:
I think this is probably OK, bug please also file a bug for the KDE folks.  `TryExec` is documented in the .desktop file spec, and used by a lot of applications.

https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#recognized-keys

PR #1090, opened 2021-08-30
[WIP] Tmux control mode

Dixeran:
A proof of concept mainly inspired by previous tmux-cc related PR and @skyline75489 's tmux_cc branch.

The idea is to place the tmux parser directly into escape::Parser, so we can 
1. bypass VTParse from here 
2. recover from that state if we encounter any error parsing tmux events(now we can press 'q' and nicely quit tmux mode)

Thus, DCS handler will handle the parsed TmuxEvents, things after that will remain trivial.


skyline75489:
Man, I'm excited about this PR. Hope this gets merged eventually 👍

Dixeran:
I think this can be reviewed now since it's been at a basic functional status. All comments are appreciated!

```
/*how tmux things interact with other modules*/

                        ┌──────────────────┐
                        │  Tmux Server     │
                        └────┬─▲───────────┘
                             │ │
                             │ │
                             │ │                       ┌───────────────────────────────────┐
                             │ │                       │ GUI Window For Tmux               │
                        ┌────▼─┴───────────┐           │                                   │
         ┌──────────────┤Original LocalPane│           │                                   │
         │              └──────▲───────────┘           │  ┌─────────────────────────────┐  │
         │                     │                       │  │TermTab - LocalPane          │  │
         │                     │                       │  │             │               │  │
┌────────▼───────┐             │                       │  │             │               │  │
│  Tmux Parser   │             │                       │  │       ┌─────▼────────┐      │  │
└────────┬───────┘             │                       │  │       │              │      │  │
         │                     │                       │  │       │  Tmux Pty    │      │  │
         │              ┌──────┴───────────┐           │  │       │              │      │  │
         │              │                  ├───────────┼──┼──────►└─────┬────────┘      │  │
         └──────────────►  Tmux Domain     │           │  │             │               │  │
          Tmux events   │                  ◄───────────┼──┼─────────────┘               │  │
                        └──────────────────┘           │  │                             │  │
                                                       │  │                             │  │
                                                       │  │                             │  │
                                                       │  │                             │  │
                                                       │  │                             │  │
                                                       │  └─────────────────────────────┘  │
                                                       │                                   │
                                                       │                                   │
                                                       └───────────────────────────────────┘
```

mux/src/tmux_commands.rs
```diff
@@ -0,0 +1,304 @@
...
+    pub pane: TmuxPaneId,
+}
+impl TmuxCommand for SendKeys {
+    fn get_command(&self) -> String {
+        let mut s = String::new();
+        for &byte in self.keys.iter() {
+            write!(&mut s, "0x{:X}\r", byte).expect("unable to write key");
+        }
+        format!("send-keys -t {} {}", self.pane, s)
+        // FIXME: An unexpected duplicated command will prompt next line, why?
```
Dixeran, line 298: After some search, I found this really hard to solve. Currently Tmux use `ESC k` to set the window name, which we don't support yet. It's really annoying to live with that window title echoed after every command..

Relate issue in mosh: https://github.com/mobile-shell/mosh/issues/992

skyline75489:
Build is failing on Windows for some reason.

Although tmux itself isn't available natively on Windows, one can still use tmux in WSL and over SSH.

mux/src/tmux_commands.rs
```diff
@@ -0,0 +1,304 @@
...
+#[derive(Debug)]
+pub(crate) struct SendKeys {
+    pub keys: Vec<u8>,
+    pub pane: TmuxPaneId,
+}
+impl TmuxCommand for SendKeys {
+    fn get_command(&self) -> String {
+        let mut s = String::new();
+        for &byte in self.keys.iter() {
+            write!(&mut s, "0x{:X}\r", byte).expect("unable to write key");
```
wez, line 295: is the `\r` in here intentional?

mux/src/tmux_commands.rs
wez: That `ESC k TITLE ST` sequence is pretty awkward because it doesn't match the typical OSC style escapes that use ST (string terminator).  That makes it difficult to fit into the parser at the right layer :-(

Do you know what is emitting that sequence?
http://manpages.ubuntu.com/manpages/precise/en/man1/tmux.1.html#names%20and%20titles
suggests that automatic-rename may be a factor: can we tell tmux to disable that option when we set up control mode?

mux/src/tmux_pty.rs
```diff
@@ -0,0 +1,118 @@
...
+    fn wait(&mut self) -> std::io::Result<portable_pty::ExitStatus> {
+        let (lock, var) = &*self.active_lock;
+        let mut released = lock.lock().unwrap();
+        while !*released {
+            released = var.wait(released).unwrap();
+        }
+        return Ok(ExitStatus::with_exit_code(0));
+    }
+
+    fn process_id(&self) -> Option<u32> {
```
wez, line 78: There's a windows specific method to impl here too, that can return `None`:

https://github.com/wez/wezterm/blob/3f6ff534d3ad3df8bcf93dd02e1dd6dc3da1d7c2/pty/src/lib.rs#L240-L243

mux/src/tmux_pty.rs
```diff
@@ -0,0 +1,118 @@
...
+        Ok(Box::new(TmuxPty {
+            domain_id: self.domain_id,
+            master_pane: self.master_pane.clone(),
+            rx: self.rx.clone(),
+            cmd_queue: self.cmd_queue.clone(),
+            active_lock: self.active_lock.clone(),
+        }))
+    }
+
+    fn process_group_leader(&self) -> Option<libc::pid_t> {
```
wez, line 115: This is unix specific!

https://github.com/wez/wezterm/blob/3f6ff534d3ad3df8bcf93dd02e1dd6dc3da1d7c2/pty/src/lib.rs#L111-L112

```suggestion
    #[cfg(unix)]
    fn process_group_leader(&self) -> Option<libc::pid_t> {
```

termwiz/src/escape/parser/mod.rs
```diff
@@ -81,11 +84,39 @@ impl Parser {
     }
 
     pub fn parse<F: FnMut(Action)>(&mut self, bytes: &[u8], mut callback: F) {
-        let mut perform = Performer {
-            callback: &mut callback,
-            state: &mut self.state.borrow_mut(),
-        };
-        self.state_machine.parse(bytes, &mut perform);
+        let tmux_state: bool = self.state.borrow().tmux_state.is_some();
```
wez, line 87: nit: rename for clarity:

```suggestion
        let is_tmux_mode_active = self.state.borrow().tmux_state.is_some();
```

termwiz/src/escape/parser/mod.rs
```diff
@@ -81,11 +84,39 @@ impl Parser {
...
+        if tmux_state {
+            if let Some(unparsed_str) = {
```
wez, line 89: This `if let Some(_) = { big block } { // then do something }` construct is hard to read.
Please factor out the `big block` portion into a helper method that returns `Option<String>` so that this section of code is a bit clearer and is focused on routing that unparsed string to the performer.

tmux-cc/src/lib.rs
```diff
@@ -489,30 +489,42 @@ impl Parser {
...
+                    if let Some(e) = option_event {
+                        events.push(e);
+                    }
+                }
+                Err(err) => {
+                    // TODO: concat remained bytes
+                    return Err(anyhow::anyhow!(format!(
+                        "{}{}",
+                        err,
+                        std::str::from_utf8(&bytes[i..]).unwrap()
```
wez, line 519: prefer https://doc.rust-lang.org/std/string/struct.String.html#method.from_utf8_lossy so that we can avoid the unwrap in case we have non-utf8 input, and still be able to display a reasonable error message without crashing all of the users panes!

termwiz/Cargo.toml
```diff
@@ -32,6 +32,7 @@ serde = {version="1.0", features = ["rc", "derive"], optional=true}
 sha2 = "0.9"
 terminfo = "0.7"
 thiserror = "1.0"
+tmux-cc = {version = "0.1", path = "../tmux-cc"}
```
wez, line 35: dependency management thinking: `termwiz` is its own crate on crates.io.
Adding a dep here means that we'll also either:

* need to publish `tmux-cc` to crates.io (and I'm not sure that that makes sense as it doesn't seem obviously reusable)
* move tmux-cc to be a module inside termwiz, and potentially make it an optional feature

mux/src/tmux_commands.rs
```diff
@@ -0,0 +1,304 @@
...
+        };
+        let tmux_domain = match domain.downcast_ref::<TmuxDomain>() {
+            Some(t) => t,
+            None => anyhow::bail!("Tmux domain lost"),
+        };
+
+        let pane_map = tmux_domain.inner.remote_panes.borrow();
+        if let Some(pane) = pane_map.get(&self.0) {
+            let lock = pane.lock().unwrap();
+            lock.tx.send(result.output.to_owned()).unwrap();
```
wez, line 279: there are quite a few `unwrap` calls overall.  In a method that returns `Result` we should be able to use either `?` or `.ok_or_else(|| anyhow!("reason"))?` to avoid taking down the whole program.

In cases where the unwrap represents an impossible situation, please replace the unwrap with an `.expect("reason")`.

That will help with reviewing the code, because it will make your assumptions/invariants/intent clearer by capturing them in the code!

mux/src/tmux_pty.rs
```diff
@@ -0,0 +1,118 @@
+use crate::{
+    tmux::{RefTmuxRemotePane, TmuxCmdQueue, TmuxDomainState},
+    tmux_commands::SendKeys,
+};
+use flume;
```
wez, line 5: using just the crate name doesn't really "do" anything, so you should be able to remove it!

```suggestion
```

mux/src/tmux_pty.rs
```diff
@@ -0,0 +1,118 @@
...
+
+pub(crate) struct TmuxReader {
+    rx: flume::Receiver<String>,
+}
+
+impl Read for TmuxReader {
+    fn read(&mut self, mut buf: &mut [u8]) -> std::io::Result<usize> {
+        match self.rx.recv() {
+            Ok(str) => {
+                return buf.write(str.as_bytes());
```
wez, line 20: This buffer reader has a potential problem: if we recv a String that is larger than buf it won't fit and we'll discard the excess.

I think we can make this safer by using a `socketpair` from our `filedescriptor` crate.  Take a look at `read_from_pane_pty` to see how it uses a socketpair as a channel for the bytes.

The tmuxreader could potentially just one of the `FileDescriptor`s from the socket pair and simplify some of the code in here.

mux/src/tmux_pty.rs
```diff
@@ -0,0 +1,118 @@
...
+    pub active_lock: Arc<(Mutex<bool>, Condvar)>,
+}
+
+impl Write for TmuxPty {
+    fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
+        let pane_id = {
+            let pane_lock = self.master_pane.lock().unwrap();
+            pane_lock.pane_id
+        };
+        log::info!("pane:{}, content:{:?}", &pane_id, buf);
```
wez, line 45: nit: info is too high a log level for this :-)

```suggestion
        log::trace!("pane:{}, content:{:?}", &pane_id, buf);
```

mux/src/tmux.rs
```diff
@@ -179,21 +144,50 @@ impl TmuxDomainState {
             *self.state.borrow_mut() = State::WaitingForResponse;
         }
     }
+
+    pub fn kick_off(domain_id: usize) {
```
wez, line 148: I'd suggest renaming to better reflect the purpose; adding a doc comment would be nice if things are more complicated! :-)

```suggestion
    pub fn schedule_send_next_command(domain_id: usize) {
```

wez:
Thanks for taking this on!
It's a big PR so I'll probably take a couple of review passes this weekend.

There's some cleanup needed in general before we merge; there are a few things that are adding commented-out code that should be pruned out, and it's relatively light on comments overall that make it a little bit harder to understand the flow.
Your ascii art diagram helps a bit with that :)

I've made a few suggestions already, but I may come back with more feedback as more of what's happening in the PR "clicks" in mind!

Dixeran:
> Build is failing on Windows for some reason.
> 
> Although tmux itself isn't available natively on Windows, one can still use tmux in WSL and over SSH.

Fixed. However I still can't switch into tmux control mode inside WSL, and there is no `DeviceControlMode::Enter` performed. Would it possible that it was filtered by ConHost?

mux/src/tmux_commands.rs
Dixeran: Oh..It should be at the end of the whole line. Fixed.

mux/src/tmux_commands.rs
Dixeran: I found that it was my zsh. Set `DISABLE_AUTO_TITLE="true"` disable this sequence.

termwiz/Cargo.toml
Dixeran: Woa, didn't expect that. Move into `termwiz` seems more reasonable, maybe better to move it after all other issues fixed?

skyline75489:
> Would it possible that it was filtered by ConHost

Yes, some of the sequences will be filtered due to how ConPTY works. I think this is on the roadmap of the ConPTY project.

mux/src/tmux_pty.rs
```diff
@@ -0,0 +1,159 @@
...
+                        self.head_cursor = 0;
+                        return self.read(buf);
+                    } else {
+                        return buf.write(str.as_bytes());
+                    }
+                }
+                Err(_) => {
+                    return Ok(0);
+                }
+            }
```
Dixeran, line 55: Add a buffer here to prevent write out of range. Will this solve the problem?

mux/src/tmux_pty.rs
Dixeran: Can't get point of `socketpair` here, but I tried to add a buffer to avoid the "overrange" problem.

wez:
It's taking me a while to get through and review this: I've been a bit preoccupied with preparation for a long distance house move in real life, so wanted to let you know that that's why I'm going slow

mux/src/tmux_pty.rs
wez: I haven't looked deeply at this change yet, but my knee-jerk reaction is that that looks very complex.  I think replacing the channel based reader/writer with a socketpair will make most of the complexity disappear!

Dixeran:
I fully understand that this PR will take some time, I'm thinking of writing a more detailed design description within the next few days.

joexue:
can not wait it.

Felixoid:
Kudos to this PR, I'd really like to have it working!

hurricanehrndz:
This work looks amazing!

wez:
Thanks for this!

I rebased it and fixed up the build for recent changes in `main`, addressed the socketpair/buffering issue we discussed and fixed up send-keys and capturep so that they correctly reference panes, and pushed that to `main`.

The current state is:
* Attaching now shows the pane content
* The cursor position is wrong
* Typing sends input to the pane

There's still a bunch of things to fully reconcile with the items in https://github.com/wez/wezterm/issues/336

One of the interesting recent developments in `main` is that we now have a *workspace* concept which is analogous to the tmux session concept, which should make some of the integration a bit easier to model.

wez:
I'm closing this as this batch of changes is now in `main`, and we can make additional PRs to follow up.
Thanks again for your work here!

Dixeran:
Really glad to hear that! Can't wait to move forward.

skyline75489:
Thanks @Dixeran and @wez for the amazing work here! Can't wait to see this land :)

OmegaRogue:
whats the current state of this in `main`? It doesnt seem to do anything after attaching to a tmux session yet

wez:
It is incomplete as indicated in https://github.com/wez/wezterm/pull/1090#issuecomment-1014913205

PR #1091, opened 2021-08-30
Add `showkey` recomendation to SendString docu

ModProg:
Not sure if this is the best tool for this, but it's what I found, and this could save someone some time in the future.

### DIFF
--- a/docs/config/lua/keyassignment/SendString.md
+++ b/docs/config/lua/keyassignment/SendString.md
@@ -32,3 +32,12 @@ return {
 }
 ```
 
+An easy way to get the correct escape sequence is running `showkey -a`.
+
+This would be the output for pressing `CTRL+L`:
+
+```
+^L 	 12 0014 0x0c
+```
+
+You can then use `"\x0c"` in your `wezterm.lua`.


wez:
I'm not keen on taking this as-is, because `showkey` is very linux specific, and the usage here feels a bit tautological (sorry for the long word) in that showkey shows you what you pressed, so it feels a bit backwards to reverse engineer what you pressed when you want to change what something does when you press it.

We have debug key events to show what you pressed: https://wezfurlong.org/wezterm/config/lua/config/debug_key_events.html

but what I think you're looking for here is a bit different; would you rather have a function that, given a key combination expressed using `{key="L", mods="CTRL"}` form, returned what wezterm would encode that combination as?

Or extending that a bit more, something like this would replace the example currently at the
bottom of the SendString docs:

```lua
return {
  keys = {
     -- Rebind OPT-Left, OPT-Right as ALT-b, ALT-f respectively to match Terminal.app behavior
     {key="LeftArrow", mods="OPT", action=wezterm.action{SendKey={key="b", mods="ALT"}}},
     {key="RightArrow", mods="OPT", action=wezterm.action{SendKey={key="f", mods="ALT"}}},
  }
}
```

wez:
https://wezfurlong.org/wezterm/config/lua/keyassignment/SendKey.html is now in the nightly.
I think we can close this PR, but I'd be happy to hear more about cases where showkey makes things easier to figure out, so that we can close that gap in wezterm!

ModProg:
We only reason I would now see use in this is when you want to map a key sequence, I agree that `SendKey` is a better solution for most people (myself included). than needing to find strange ansi codes manually.

wez:
I've also added https://wezfurlong.org/wezterm/config/lua/keyassignment/Multiple.html for the key sequence scenario!

PR #1096, opened 2021-09-01
(macOS) respect alt key configs for ime

### LINKED ISSUE #615: When non-English IME is active, keyboard shortcuts are also interpreted by IME
## Describe the bug

*Regardless of* `use_ime` configuration, when the system IME switches to a non-English keyboard layout (e.g., [Korean Sebeolsik](https://en.wikipedia.org/wiki/Keyboard_layout#/media/File:KB_Sebeolsik_Flnal.svg)), all keyboard shortcuts becomes unusable because IME intercepts them.

Maybe related with #318, though IME is highly platform-specific one.

## Environment (please complete the following information):

 - OS: macOS
 - Version: wezterm 20210314-114017-04b7cedd-120-g217dbfb9
 - The active keyboard layout name (eg: ENG, DEU and so on): ENG / Korean Sebeolsik

## To Reproduce

Switch the system IME to a non-English layout and try to use keyboard shortcuts that have different key mappings in that layout.
For example, Cmd+1 to switch to the first tab should work regardless of the IME state, but in the Korean Sebulsik (3-beolsik) layout, "1" becomes "ᇂ" and the shortcut does not work.

## Configuration

```lua

return {
    use_ime = true,  -- the bug is reproducible with false as well!
    color_scheme = "Andromeda",
        "JetBrains Mono",
        "Menlo",
        "NanumBarunGothic",
    }),
}
```

## Expected behavior

Normally, keyboard shortcuts should bypass the IME and interpreted as-is.

jakelinnzy:
This should partially solve #615.

Currently, the config items `send_composed_key_when_(left|right)_alt_is_pressed` are ignored when `use_ime` is true.

e.g. In wezterm.lua:

```lua
return {
    use_ime = true,
    send_composed_key_when_left_alt_is_pressed = false,
    send_composed_key_when_right_alt_is_pressed = true,
}
```

Pressing `LeftAlt + a` used to produce `å`, but now it should send `<M-a>` to the shell. The behaviour of `RightAlt` is unchanged.

### DIFF
--- a/window/src/os/macos/window.rs
+++ b/window/src/os/macos/window.rs
@@ -1736,7 +1736,12 @@ impl WindowView {
             return;
         };
 
-        let use_ime = config::configuration().use_ime;
+        let config_handle = config::configuration();
+        let use_ime = config_handle.use_ime;
+        let send_composed_key_when_left_alt_is_pressed =
+            config_handle.send_composed_key_when_left_alt_is_pressed;
+        let send_composed_key_when_right_alt_is_pressed =
+            config_handle.send_composed_key_when_right_alt_is_pressed;
 
         // `Delete` on macos is really Backspace and emits BS.
         // `Fn-Delete` emits DEL.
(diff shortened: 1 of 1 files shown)


wez:
Thanks!

Issue #1098, opened 2021-09-02
Support OS appearance events on GNOME


EpocSquadron:
I'd really like to be able to switch terminal theme with the rest of my UI in the context of GNOME desktop on linux. This currently does not work in either X11 or Wayland.

I'm not sure how this can be acheived, but I know that Firefox is able to detect the appearance change and even propagates it to webpages with the prefers-color-scheme CSS media query. I do know that Firefox makes use of the GTK toolkit to provide GNOME integration more effectively though. I know that's not the case here (which is also why client side decorations are out of place when run in wayland mode in GNOME).

Despite that knowledge, I tracked down some discussion on the implementation of dark mode detection in webkit, which contains C code that might be useful. https://bugs.webkit.org/show_bug.cgi?id=196685

Besides that, we can query gsettings if present at window creation and match the theme that way. https://wiki.archlinux.org/title/Dark_mode_switching#gsettings

Either way, dark mode in GTK appears to be implemented as paired themes. Themes with a (-dark) variant would work with either of the above methods as the heuristic is really "is -dark in the name of the current theme". Other themes or themes that are only dark or light will not work with the detection.

For the window event, I don't know if there's any method to listen for theme changes in GTK without being a GTK app, but I also didn't research that.

lunaryorn:
I came across this, and would like to point out that the upcoming Gnome 42 release provides the dark mode preference through the generic XDG settings portal, which in addition to Gnome is going to be supported by at least KDE and elementary as well.

This portal should work consistently across X11 and Wayland, and provides the current setting as well as a change signal over DBus, see <https://flatpak.github.io/xdg-desktop-portal/#gdbus-org.freedesktop.portal.Settings>

I think this would allow implementing `window:get_appearance()` on Wayland with no dependencies other than DBus, and given that wezterm does notifications on Linux I presume it already uses DBus one way or another.  The change signal would be a bit of a challenge I guess, because receiving DBus signals usually needs an open connection to the bus and some kind of mainloop.

@wez  Would you accept a pull request which implements `window:get_appearance` for Wayland based on the XDG portal API, and perhaps the `window

wez:
Yeah, I'd like to see that!  https://github.com/wez/wezterm/blob/main/wezterm-toast-notification/src/dbus.rs is our integration with dbus: it does listen for events to be able to handle things like clicking to open up the changelog URL when we have a release update.  It uses the `zbus` crate for that.

lunaryorn:
[...]

PR #1100, opened 2021-09-03
Document workaround for get_appearance on GNOME Wayland

EpocSquadron:
This adds a note about how to acheive dark mode detection on GNOME Wayland systems, for future generations. Follows up on #1098

### DIFF
--- a/docs/config/lua/window/get_appearance.md
+++ b/docs/config/lua/window/get_appearance.md
@@ -45,40 +45,40 @@ end)
 return {
 }
 ```
-### Wayland GNOME Workaround
 
-If you happen to be running a Wayland session through GNOME desktop,
-a workaround exists in which you can query the `gsettings` command
-for the theme name. While GNOME does not specifically support
-appearance, it is common for themes to provide a dark variant (and
-this is the case for the default "Adwaita" theme), which are typically
-named with "-dark" appended to the theme name. With this heuristic,
-we can determine if GNOME is in dark mode or not, and abuse the
-[update-right-status](../window-events/update-right-status.md)
-event to poll for changes. When coupled with the [Night
-Theme](https://nightthemeswitcher.romainvigier.fr/) extension or
-similar, you can enjoy a Wezterm that changes theme based on day night
-cycle with the rest of your system.
+### Wayland GNOME Appearance
+
+The GNOME desktop environment provides the `gsettings` tool that can
+inform us of the selected appearance even in a Wayland session. We can
+substitute the call to `window:get_appearance` above with a call to the
+following function, which takes advantage of this.
 
 ```lua
-function·compute_scheme_gnome_wayland()¬
-··local·success,·stdout·=·wezterm.run_child_process(¬
-····{"gsettings",·"get",·"org.gnome.desktop.interface",·"gtk-theme"}¬
-··)¬
-··if·stdout:match("-dark")·then¬
-····return·"Builtin Solarized Dark"¬
-··else¬
-····return·"Builtin Solarized Light"¬
-··end¬
-end¬
-¬
-wezterm.on("update-right-status",·function(window,·pane)¬
-··local·overrides·=·window:get_config_overrides()·or·{}¬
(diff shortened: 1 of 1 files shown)


docs/config/lua/window/get_appearance.md
```diff
@@ -45,4 +45,40 @@ end)
...
+we can determine if GNOME is in dark mode or not, and abuse the
+[update-right-status](../window-events/update-right-status.md)
+event to poll for changes. When coupled with the [Night
+Theme](https://nightthemeswitcher.romainvigier.fr/) extension or
+similar, you can enjoy a Wezterm that changes theme based on day night
+cycle with the rest of your system.
+
+```lua
+function·compute_scheme_gnome_wayland()¬
+··local·success,·stdout·=·wezterm.run_child_process(¬
```
wez, line 65: Please remove these special spacing characters from the example; they should be simple whitespace! :)

docs/config/lua/window/get_appearance.md
```diff
@@ -45,4 +45,40 @@ end)
 return {
 }
 ```
+### Wayland GNOME Workaround
```
wez, line 48: I'd prefer to phrase this section something like:

```
### Wayland GNOME Appearance

Since Wayland doesn't specifically support appearance, we can use this
alternative method to determine the appearance and adapt the color
scheme accordingly.
```

docs/config/lua/window/get_appearance.md
```diff
@@ -45,4 +45,40 @@ end)
...
+[update-right-status](../window-events/update-right-status.md)
+event to poll for changes. When coupled with the [Night
+Theme](https://nightthemeswitcher.romainvigier.fr/) extension or
+similar, you can enjoy a Wezterm that changes theme based on day night
+cycle with the rest of your system.
+
+```lua
+function·compute_scheme_gnome_wayland()¬
+··local·success,·stdout·=·wezterm.run_child_process(¬
+····{"gsettings",·"get",·"org.gnome.desktop.interface",·"gtk-theme"}¬
```
wez, line 66: I think it would be helpful to factor out the bit that determines the appearance.

For example:

```lua
function query_appearance_gnome()
   local success, stdout = wezterm.run_child_process({"gsettings", ....})
   -- lowercase and remove whitespace
   stdout = stdout.lower():gsub("%s+", "")
   local mapping = {
      highcontrast = "LightHighContrast",
      highcontrastinverse = "DarkHighContrast",
      adwaita = "Light",
      ["adwaita-dark"] = "Dark"
   }
   if stdout == "highcontrast" then
     return "LighHighContrast"
   end
   local appearance = mapping[stdout]
   if appearance then
      return appearance
   end
   if stdout:find("dark") then
      return "Dark"
   end
   return "Light"
end
```

See also: https://github.com/wez/wezterm/blob/main/window/src/os/x11/connection.rs#L170-L175

Then the example can reuse the `scheme_for_appearance` function from the earlier example on this page, and use the `update-right-status` hook using:

```lua
  local appearance = query_appearance_gnome()
  local scheme = scheme_for_appearance(appearance)
```

wez (changes requested):
Thanks for playing with this!  I have some suggested changes!

docs/config/lua/window/get_appearance.md
EpocSquadron: I was having trouble copying from my editor, so I used wezterm's copy mode to get the snippet into another editor instance, which displays whitespace in such a way so I didn't notice it. Will remove shortly.

docs/config/lua/window/get_appearance.md
EpocSquadron: All set in latest commit.

docs/config/lua/window/get_appearance.md
EpocSquadron: I tested this out locally and after a little debugging this works well even with the High Contrast switch in GNOME's Accessibility menu. :100:  Refactored in latest commit.

wez:
Thanks!

PR #1104, opened 2021-09-05
get-deps: Zypper needs -l (--auto-agree-with-licenses)

hpjansson:
Without this switch, zypper would automatically answer "no" to any license agreements, halting the update.


wez:
Thanks!

PR #1126, opened 2021-09-08
gui: cancel mouse click streak when the cursor has moved to a different cell (fixes #910)

### LINKED ISSUE #910: De-selection does not work under some circumstances
## Describe the bug

Canceling selection does not work under some circumstances.

## Environment (please complete the following information):

 - OS: Linux X11
 - Version: `20210628-091813-e05ddb17`
 - The active keyboard layout name: ENG

## To Reproduce

Steps to reproduce the behavior.

### Method 1
1. triple-click any line to select it;
2. move the cursor to another line and click.

If you act fast so that the gap between 1 and 2 is small enough, the line selected in 1 won't be deselected.

### Method 2
1. press down the cursor and move the cursor to drag an area
2. move the cursor to another line and click.

If you act fast so that the gap between 1 and 2 is small enough, wezterm will always select the area of 2 rather than deselected.

## Configuration

```
return {
    colors = {
        foreground = "#F8F8F2",
        background = "#282A36",
        selection_bg = "#44475a",
        selection_fg = "#ffffff",
        ansi = {"#282A36", "#ff5555", "#50fa7b", "#f1fa8c", "#bd93f9", "#ff79c6", "#8be9fd", "#bfbfbf"},
        brights = {"#4d4d4d", "#ff6e67", "#5af78e", "#f4f99d", "#caa9fa", "#ff92d0", "#9aedfe", "#e6e6e6"},
    },
    check_for_updates = false,
    exit_behavior = "Close",
        "Fantasque Sans Mono",
    }),
    font_size = font_size,
    force_reverse_video_cursor = true,
    freetype_interpreter_version = 35,
    freetype_render_target = "HorizontalLcd",
    scrollback_lines = 35000,
    term = "xterm-256color",
}
```

## Expected behavior
wezterm should deselect the area in 1.

unrelentingtech:
Fast-clicking users generally expect to be able to rapidly do regular selections or otherwise cancel double/triple/quad clicks, so significant mouse movement should end the streak.

Allowing tiny pixel movements to account for touchpads is necessary.
Fortunately, the position here is already in character grid cells, which provides enough margin for this.
Other code editors like gedit also seem to do this based on the character grid.

---

@zhou13 hopefully this solves #910

### DIFF
--- a/term/src/input.rs
+++ b/term/src/input.rs
@@ -45,10 +45,12 @@ pub struct MouseEvent {
 /// which is the number of successive clicks of the same mouse button
 /// within the `CLICK_INTERVAL`.  The streak is reset to 1 each time
 /// the mouse button differs from the last click, or when the elapsed
-/// time exceeds `CLICK_INTERVAL`.
+/// time exceeds `CLICK_INTERVAL`, or when the cursor position
+/// changes to a different character cell.
 #[derive(Debug, Clone)]
 pub struct LastMouseClick {
     pub button: MouseButton,
+    position: (usize, i64),
     time: Instant,
     pub streak: usize,
 }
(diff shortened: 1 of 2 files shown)


wez:
Thanks!  Hopefully this is sufficient; I think there might be a chance of this being frustrating in the case where the click started close to eg: the left edge of a cell and the pointer drifts 1-2 pixels to the left.  If that's the case, then an alternative is to consider pixel positioning and allow drifting within the cell metrics dimensions expressed in pixels.

PR #1127, opened 2021-09-08
window: implement window movement by dragging on Wayland

unrelentingtech:
Piggybacking off of `CopyAndPaste` for the required last input serial, maybe it would make sense to store it in a more "neutral" place eventually…


wez:
Thanks!

PR #1132, opened 2021-09-10
Add additional tab activation key assignments

friederbluemle:
Thank you for creating the most awesome terminal emulator! :+1:

This small PR adds a few more default keybindings.

All of them are de facto standards on Linux and/or Windows.

### Cycle tabs

- Activate next tab with <kbd>Ctrl</kbd> + <kbd>PageDown</kbd> and <kbd>Ctrl</kbd> + <kbd>Tab</kbd>
- Activate previous tab with <kbd>Ctrl</kbd> + <kbd>PageUp</kbd> and <kbd>Ctrl</kbd> + <kbd>Shift</kbd> + <kbd>Tab</kbd>

### Directly select a tab

Using <kbd>Alt</kbd> + <kbd>NUMBER</kbd> (e.g. <kbd>Alt</kbd> + <kbd>1</kbd> for first tab, <kbd>Alt</kbd> + <kbd>2</kbd> for second tab, etc).

Example programs: Firefox, Google Chrome, VS Code, Atom, various file managers (Thunar, PCManFM), etc

In a separate commit, it also adds `.idea/` to `.gitignore` (used by CLion and other IntelliJ-based IDEs).


wez:
Thanks for diving in!

wez:
Heads up: I removed the ALT-number default assignments in the nightly; they're causing problems with non-US layouts in #1542 and #1543

friederbluemle:
Thanks for the heads up!
I wonder how this works with other tab-based interfaces.. Most applications have Alt+number bindings to directly activate tabs (at least on US and German keyboard layouts).
I'm not familiar with French or Norwegian keyboards, but based on images I saw they both appear to have an "<kbd>Alt Gr</kbd>" key to the right of the space bar. That is similar to a German keyboard, which also has two different <kbd>Alt</kbd> keys. That's unlike the US keyboard, which practically has two identical <kbd>Alt</kbd> keys (at least for 99% of use cases). I wonder if the issue is only related to bindings which involve <kbd>AltGr</kbd>+number. The left Alt key on a German keyboard works like a US <kbd>Alt</kbd> key, and <kbd>AltGr</kbd> is used for additional punctuation. It would seem logical if that was also the case for French and Norwegian.
Since both #1542 and #1543 are from mac users, I'm also wondering if it has something to do with Apple keyboards, which of course are slightly different than standard PC keyboards ("option"/"command" vs "super"/"alt").

As a side note, while Ctrl-Shift+number works fine on my Windows test machine, it does **not** work for me in Linux (both have US keyboards). I'm not sure what the reason is or if this is a know issue..

PR #1133, opened 2021-09-10
Implement increment step resize on Wayland too

unrelentingtech:
Simple, just round up in the configure request handling.

…oh wait, not *that* simple — have to be careful with HiDPI since the incoming `resize_increments` are in physical pixels. If we want to avoid ugly 1px rounding gaps, we have to cut them out with glScissor. (If you have any ideas about passing the data through to the renderer than adding a field to `Dimensions`, do suggest.)

And this has uncovered an interesting little bug (see first commit)


window/src/os/wayland/window.rs
```diff
@@ -547,12 +549,23 @@ impl WaylandWindowInner {
...
-                let pixel_width = self.surface_to_pixels(w.try_into().unwrap());
-                let pixel_height = self.surface_to_pixels(h.try_into().unwrap());
+                let mut pixel_width = self.surface_to_pixels(w.try_into().unwrap());
+                let mut pixel_height = self.surface_to_pixels(h.try_into().unwrap());
+
+                if self.window_state.can_resize() {
+                    if let Some((x, y)) = self.resize_increments {
+                        let desired_pixel_width = pixel_width - (pixel_width % x as i32);
+                        let desired_pixel_height = pixel_height - (pixel_height % y as i32);
+                        w = self.pixels_to_surface(desired_pixel_width) as u32;
```
wez, line 565: I've been semi-idly glancing at this with the TV on in the background, so these may not be fully formed thoughts...

This transform to and from the scaled size feels tricky to me, especially with it leading to the use of glScissor.

I wonder if it would be a bit simpler overall if we could just set `pixel_width = desired_pixel_width` instead of this adjustment here.

To make that work, I think `resize_increments` needs to be recorded in terms of the scaled size, and we could perform that adjustment in `set_resize_increments`.

That way we could avoid the `crop_area` fanning out into `Dimensions` and `glScissor`.

What do you think?

unrelentingtech: This final code is definitely not the first thing I've tried :)

- Rounding the logical size up to "unscaled" cell size (divided by the factor) leads to *huge* gaps, because the rounding error accumulates over *all* the rows/columns
- Using the desired size as pixel (buffer) size and just "also" rounding the logical size leads to a mismatch that results in nothing getting rendered (stuck on the first frame, forever quarter size with one square cursor)
  - That was tested with the bad rounding that results in the huge gaps, but a mismatch's a mismatch, I'm almost certain that a 1px mismatch would result in the same issue
- There's a very simple option of just rounding the logical size up to the physical cell size… resulting in steps of 2 rows/cols on 2x scale, etc. which is not great

unrelentingtech:
It would be fine to merge at least just the first two commits, leaving the 1px border imperfection on HiDPI in favor of not touching the GL code.

wez:
Thanks for suggesting that: I'll take the first two commits for now.  My main objection to the scissor related code is how it fans out into `Dimensions`.  I haven't had bandwidth to think about an alternative for this yet, so let's take the main functionality for now and we can solve the other bit later!

wez:
picked and pushed as c5da21984703a08d5f296aec4053cecc5869bbb3 and 275dee4aa3bb5a572ca79cf0eb2dbe9eea3465d6
Thanks!

PR #1143, opened 2021-09-15
docs: Ensure Ctrl-Shift-Space binding is readable

bew:
Necessary, because with the default dark background, the space is invisible:
invisible-space


wez:
Thanks!

PR #1151, opened 2021-09-19
lua: Add `wezterm.action_callback` to help define key callbacks

bew:
Following up on https://matrix.to/#/!PirwUBcuIlTXwNveYz:matrix.org/$ets1eEk2ixZ3MgxHLrrpOoo2Rg3VI6Cg4z82IYxsOqA?via=matrix.org

I couldn't find a way to get the filename/linenumber of the call to use it for the event name without trying to use unsafe code.
It's actually not that bad, because as error in the callback will show a traceback with correct filename/linenumber.

:upside_down_face:

### DIFF
--- a/docs/config/lua/wezterm/action_callback.md
+++ b/docs/config/lua/wezterm/action_callback.md
@@ -1,0 +1,39 @@
+# `wezterm.action_callback(callback)`
+
+*Since: nightly*
+
+This function is a helper to register a custom event and return an action triggering it.
+
+It is helpful to write custom key bindings directly, without having to declare
+the event and use it in a different place.
+
+The implementation is essentially the same as:
+```lua
+function wezterm.action_callback(callback)
+  local event_id = "..." -- the function generates a unique event id
+  wezterm.on(event_id, callback)
+  return wezterm.action{EmitEvent=event_id}
+end
+```
+
+See [wezterm.on](./on.md) and [wezterm.action](./action.md) for more info on what you can do with these.
+
+
+## Usage
+
+```lua
+local wezterm = require 'wezterm';
+
+return {
+  keys = {
+    {
+      mods = "CTRL|SHIFT",
+      key = "i",
+      action = wezterm.action_callback(function(win, pane)
+        wezterm.log_info("Hello from callback!")
+        wezterm.log_info("WindowID:", win:window_id(), "PaneID:", pane:pane_id())
+      end)
+    },
+  }
+}
+```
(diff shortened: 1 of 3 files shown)


docs/config/lua/wezterm/on.md
```diff
@@ -22,6 +22,8 @@ There is no way to de-register an event handler.  However, since the Lua
 state is built from scratch when the configuration is reloaded, simply
 reloading the configuration will clear any existing event handlers.
 
+See [wezterm.action_callback](./action_callback.md) for a helper to define a custom action callback.
```
wez, line 25: What do you think about adjusitng the custom events example below to use the new helper?

docs/config/lua/wezterm/action_callback.md
```diff
@@ -0,0 +1,39 @@
+# `wezterm.action_callback(callback)`
+
+*Since: nightly*
```
wez, line 3: nit: I tend to use this phrasing so that I can more easily grep to find things when I make a release:

```suggestion
*Since: nightly builds only*
```

wez:
Looks good, thanks! What do you think about expanding the examples to show it off?

docs/config/lua/wezterm/on.md
bew: I'm not sure, I think it's important to have an example using `wezterm.on` and `wezterm.action{EmitEvent=...}` explicitely, in case one wants to do something specific that would not be supported by `wezterm.action_callback` (like using the event id in another custom action, to emit it manually with `wezterm.emit(...)`).

I can write another simpler example below it though (the same as in the docs for `wezterm.action_callback` ?) to show its usage in this doc file?

wez: How about both in the same example?  If the line starting `wezterm.on("trigger-vim-with-scrollback", function(window, pane)` becomes a helper function definition, rather than a closure: `function trigger_vim_with_scrollback(window, pane)`,
then we could show the new `action_callback` example first:

```lua
return {
  keys = {
    {key="E", mods="CTRL",
      action=wezterm.action_callback(function(win, pane)
           trigger_vim_with_scrollback(win, pane)
      end)
    }
}
```

then show the older/discrete way as an alternative:

```lua
wezterm.on("trigger-vim-with-scrollback", trigger_vim_with_scrollback)
return {
  keys = {
    {key="E", mods="CTRL",
      action=wezterm.action{EmitEvent="trigger-vim-with-scrollback"}},
  }
}
```

PR #1159, opened 2021-09-21
Do not run upload action on forks

friederbluemle:
A small update to the "continuous" GitHub Actions workflow files.

It adds a conditional to the `Upload to Nightly Release` step in order to prevent it from running on forks.

Right now, these workflows will fail in all forked repositories (e.g. https://github.com/friederbluemle/wezterm/runs/3658906316), as the required secrets are missing (and, of course, there is no point in releasing from forks anyway).

### DIFF
--- a/ci/generate-workflows.py
+++ b/ci/generate-workflows.py
@@ -47,15 +47,18 @@ class RunStep(Step):
 
 
 class ActionStep(Step):
-    def __init__(self, name, action, params=None, env=None):
+    def __init__(self, name, action, params=None, env=None, condition=None):
         self.name = name
         self.action = action
         self.params = params
         self.env = env
+        self.condition = condition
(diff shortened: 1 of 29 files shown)


ci/generate-workflows.py
```diff
@@ -372,6 +375,7 @@ def upload_asset_nightly(self):
             ActionStep(
                 "Upload to Nightly Release",
                 action="wez/upload-release-assets@releases/v1",
+                condition="github.event.repository.fork == false",
```
wez, line 378: Probably should add this to the upload in `upload_asset_tag` below too!

friederbluemle: Makes sense, I added it!

wez:
Thanks!

PR #1166, opened 2021-09-24
Change first color so that it does not match the BG

maxammann:
If we choose the same color for BG as for the first color in the ansi/bright colors, then I'm unable to see the zsh autosuggestions: https://github.com/zsh-users/zsh-autosuggestions


bew:
Note that these colorschemes come from https://github.com/mbadolato/iTerm2-Color-Schemes, I'd suggest to contribute your change there

PR #1167, opened 2021-09-24
Extend wezterm-ssh with Sftp functionality

chipsenkbeil:
Work-in-progress to extend wezterm-ssh with sftp functionality that I need for distant-ssh2.

At the moment, this does the following:

1. Extends `SessionRequest` with additional requests for each type of sftp action that can be taken
2. Adds a `File` struct that implements `smol::io::{AsyncRead, AsyncWrite}`, communicating back to the session thread to perform the actions on `ssh2::File` instances
3. Exposes a variety of dependency types that are directly used in public APIs to avoid consumers from needing to manually import those dependencies themselves
    - `filedescriptor::FileDescriptor`, 
    - `portable_pty::Child`
    - `ssh2::{FileStat, OpenFlags, OpenType, RenameFlags}`

This currently is not fully tested or working as there are a lot of async hangups that I need to debug. It may also be worth porting over some of the testing logic I've been implementing with sshd programmatically to use with this crate to maintain our sanity with it working as expected e2e.

1. The design is a flat exposure of `ssh2::Sftp` methods instead of returning an sftp channel directly, or some wrapper around the sftp channel. 
    - We could alternatively switch to returning an Sftp wrapper object (holding a `SessionSender`) with the methods exposed on it that just performs the sends, and that may be cleaner than having them all directly on the session. @wez, thoughts? We could have `SessionRequest` contain just an `Sftp(Sftp)` variant where the Sftp value is another enum containing the submessage, if we want to not have everything be entirely flat.
    - Right now, we allocate a singular sftp channel upon first receiving an sftp-related message. I think this makes sense, but I don't know if there are any drawbacks. @wez?
2. The design I've seen from the existing `SessionRequest` handlers is to include a `Sender<...>` to act as a reply for the async request. These don't handle sending back error information, so I don't know if that is done elsewhere. 
    - My only thought behind these not causing async to hang indefinitely is that the sender gets dropped at the end of the request regardless, but this doesn't communicate any specific error information to the async caller, does it?
3. The file read/write I'm just calling directly in a blocking fashion. To be honest, your file descriptor and polling work was a little more than I wanted to read through last night, but I'm also not sure how much it applies and/or would work with `ssh2::File` from the sftp channel. @wez?


chipsenkbeil:
image

chipsenkbeil:
@wez this now passes all tests when run on my local machine, but there still is some hangup with authentication from within a Github action. Do you have any ideas here? It's just using the publickey method and has all of the key information configured post-generation to my knowledge.

chipsenkbeil:
Authentication error may be due to home directory permissions: https://stackoverflow.com/a/60367309

wezterm-ssh/src/lib.rs
```diff
@@ -9,3 +9,8 @@ pub use config::*;
 pub use host::*;
 pub use pty::*;
 pub use session::*;
+
+// NOTE: Re-exported as is exposed in a public API of this crate
+pub use filedescriptor::FileDescriptor;
+pub use portable_pty::Child;
+pub use ssh2::{FileStat, OpenFlags, OpenType, RenameFlags};
```
wez, line 16: One of the things that I'd like to be possible in the near future is allowing for wezterm-ssh to be built against `libssh` instead of `libssh2` via a crate feature (libssh has better and broader support for different ssh key types), so exposing the raw ssh2 types directly would be undesirable.

Could you make wrapper types for these that avoid exposing the ssh2 types directly?
I know that's a bit of a boilerplatey PITA!

As part of that, I think it would be good to keep an eye on `libssh`'s API to try to avoid committing us to an API that cannot be represented if we're using that under the covers.  I think it should be mostly fine as both libraries are coupled to the ssh2 protocol.

And while thinking API style thoughts: the ssh2 crate exposes libssh2's terminology. I think it would be good to make a pass over the names of things and tweak them to more closely match Rust's terminology for filesystem operations.

wezterm-ssh/Cargo.toml
```diff
@@ -20,15 +23,21 @@ log = "0.4"
...
 
 # Not used directly, but is used to centralize the openssl vendor feature selection
 async_ossl = { path = "../async_ossl" }
 
 [dev-dependencies]
+assert_fs = "1.0.4"
+indoc = "1.0.3"
 k9 = "0.11.0"
+once_cell = "1.8.0"
+predicates = "2.0.2"
```
wez, line 37: I tend to use just `"2.0"` when specifying deps, as the `"2.0.2"` form doesn't actually pin to `2.0`, and every couple of weeks I run `cargo update` when reminded by dependabot, so the chances of these crates actually being `2.0.2` in practice is very low.

wezterm-ssh/Cargo.toml
```diff
@@ -20,15 +23,21 @@ log = "0.4"
 portable-pty = { version="0.5", path = "../pty" }
 regex = "1"
 smol = "1.2"
-ssh2 = {version="0.9", features=["openssl-on-win32"]}
-#ssh2 = {version="0.9"}
+# ssh2 = {version="0.9", features=["openssl-on-win32"]}
```
wez, line 26: I can't merge this PR without `openssl-on-win32` without breaking wezterm on windows.
Would it make things saner for you while iterating with `distant` if this dep directly referenced my fork instead of using the patched version in the top level `Cargo.toml`?

wezterm-ssh/src/pty.rs
```diff
@@ -275,4 +292,18 @@ impl crate::session::SessionInner {
         }
         Ok(())
     }
+
+    pub fn channel_eof_and_close(
```
wez, line 296: Can we rename this to `send_eof_and_close`?
It feels easier to intuit what actions are being taken to me!

wezterm-ssh/src/session/sftp.rs
```diff
@@ -0,0 +1,419 @@
...
+    /// Specialized type for file-based operations
+    File(FileRequest),
+}
+
+pub(crate) struct OpenMode {
+    pub filename: PathBuf,
+    pub flags: OpenFlags,
+    pub mode: i32,
+    pub open_type: OpenType,
+    pub reply: Sender<File>,
```
wez, line 310: re: your question about propagating errors, you could use `Sender<Result<File>>` if the `Result`'s `Err` type is `Send + Sync`.

I would suggest using the `thiserror` crate to define an exportable error variant of our own to represent the significant types of errors that folks would want to handle in code, rather than using the ssh2 crate error types (which are honestly a bit awkward: lots of possible errors in the API, but relatively few of them are used in practice).

Again: a bit of a boilerplatey PITA.

wez (commented):
Overall, I think this looks good, thanks!

I think it's "fine" to make the individual sftp operations block in the session dispatcher: I don't think there's ever any real parallelism with the I/O on the other end of the session anyway, so I don't think it is going to have too much impact on overall throughput, and the code is definitely much simpler with blocking around those operations.

re: e2e tests, I think we'll need to gate running those so that we only decide to do that for a particular distro/flavor, as my experience with them in the ssh2 crate was frustrating trying to wrangle things across linux and macos (and macos in particular!).

An alternative approach to spawning sshd might be to use the thrussh ssh server to test against: stand up isolated instances of that should be pretty portable, but risks not actually testing against the ssh server implementation that most sites use, and I'm not sure how much of sftp that supports either.

wezterm-ssh/Cargo.toml
chipsenkbeil: I wasn't planning to leave this bit once everything passes :) Was just so I could test quickly before you mentioned the patched version you've got! I'll switch over to the patched version myself.

wezterm-ssh/src/pty.rs
chipsenkbeil: I'm actually going to remove this one as this was a quick hack to see if it fixed my problem.

wezterm-ssh/Cargo.toml
chipsenkbeil: @wez I do this out of habit given that even though a library promises that nothing breaks on a patch revision, you may assume some functionality or behavior on the first version you use (e.g. 2.0.2), which if someone else pulls in the minimum version (e.g. 2.0.0) the crate may fail to compile or worse behave unexpectedly. This always guarantees 2.0.2 or above for patch revisions. Is there a reason you want 2.0 specifically?

Did you still want me to switch?

wezterm-ssh/src/lib.rs
chipsenkbeil: Yeah, I can make wrapper types. To my knowledge, sftp was only introduced in ssh 2, meaning that it wouldn't be available in libssh, right? Just a heads up that distant would require ssh2 and couldn't use libssh.

Would it make sense to make a public module called ssh2 or something that wezterm exposes instead with those types? That way it could be optionally included with a feature and would be clear what is exposed:

```rust
#[cfg(feature = "ssh2")]
pub mod ssh2 {
    pub use ::ssh2::FileStat;
    pub use ::ssh2::OpenFlags;
    pub use ::ssh2::OpenType;
    pub use ::ssh2::RenameFlags;
}
```

Or did you want wrapper types regardless? I can do either, just checking.

wezterm-ssh/src/session/sftp.rs
```diff
@@ -0,0 +1,419 @@
...
+}
+
+impl fmt::Debug for OpenMode {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        // NOTE: OpenType does not implement debug,
+        //       so we create a string representation
+        let open_type_string = match self.open_type {
+            OpenType::Dir => String::from("OpenType::Dir"),
```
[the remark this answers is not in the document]
chipsenkbeil: Good call on the string allocation. For some reason, I was getting compilation errors when I **thought** I was using a str, but must have been doing something a little different!

As for e2e, the authentication still fails for me on CI, so I'm going to have to dig in. I looked at thrussh, but does it support any of the auth methods? From what I remember from reading through the readme, it lacked a lot of the auth methods. Granted, for e2e verifying post-auth functionality, we don't care how we authenticate as long as it works.

wezterm-ssh/src/session/sftp.rs
chipsenkbeil: Yeah, I think I'll revise to do that. It's really annoying right now on distant as the only error you get is a message about the receiver being closed without clarity about the real problem.

wezterm-ssh/Cargo.toml
wez: Unless you know that the thing you need is in `2.0.2` and not in any lower version, it's just a bit less cluttered to use `2.0`.
I tend only to be more specific in the case that something is broken in a given version of a dep.

wezterm-ssh/src/lib.rs
wez: I know it's more work, but I'd prefer wrapper types so that it is easier to evolve the API (and naming!) for different ssh backends.

`libssh` supports sftp: https://api.libssh.org/master/libssh_tutor_sftp.html

(`libssh` support SSH protocol 2!)

chipsenkbeil: > (`libssh` support SSH protocol 2!)

Oh nice! I'll whip up some wrapper types, then.

wezterm-ssh/src/session/sftp.rs
chipsenkbeil: Tackled this for SFTP specifically. Out of scope for broader ssh.

wezterm-ssh/Cargo.toml
chipsenkbeil: Reverted back to including `openssl-on-win32`.

wezterm-ssh/Cargo.toml
chipsenkbeil: Updated deps to just use major + minor.

wezterm-ssh/src/lib.rs
chipsenkbeil: Created wrapper types. Things like `OpenType`, `FileStat`, `OpenFlags`, and `RenameFlags` all exist in libssh as well with same values, so wrapper type mirroring those values (mostly) should be enough.

wezterm-ssh/src/session/sftp/file.rs
```diff
@@ -0,0 +1,388 @@
+use super::{FileStat, SessionRequest, SessionSender, SftpRequest};
+use smol::{
+    channel::{bounded, Sender},
```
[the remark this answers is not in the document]
chipsenkbeil: Flattened.

chipsenkbeil:
@wez I've run out of ideas regarding the authentication error. I was able to specify an explicit user for ssh and ensure that `/run/sshd` exists, which sometimes doesn't if sshd is not started from an init script. You can now see that some systems pass and others don't, sometimes just from an OS version difference.

 I was able to use [`act`](https://github.com/nektos/act) to run locally and inspect `sshd.log` to see an error about `/run/sshd`.

Should we gate the tests and call it a day? Or do something else? I really don't want to spend any more time on this if I can help it.

wez:
re: gating tests, I'm totally fine with restricting to running on the easier to test environments

chipsenkbeil:
> re: gating tests, I'm totally fine with restricting to running on the easier to test environments

How can I go about gating? FreeBSD and Windows are off the table. For centos8 and fedora, I'd have to check why they're failing by running locally with `act` and inspecting the `sshd.log` to see if there's anything else odd going on. Otherwise, we'd mask those.

wez:
There are lots of ways to gate running tests that are all a bit horrible.

Maybe the cleanest is to use something like https://github.com/stanislav-tkach/os_info and have a helper function that encodes which OS/distro the tests should run on, and then have the tests return early.

chipsenkbeil:
> There are lots of ways to gate running tests that are all a bit horrible.
> 
> Maybe the cleanest is to use something like https://github.com/stanislav-tkach/os_info and have a helper function that encodes which OS/distro the tests should run on, and then have the tests return early.

Let me see if I can figure out why Centos 8 and the Fedora pair are failing with the authentication error. If I'm lucky - something stands out in the sshd.log - and I can fix it, then gating will be as simple as adding a `#[cfg_attr(not(any(target_os = "macos", target_os = "linux")), ignore)]`

chipsenkbeil:
@wez, I managed to partially run a docker container with Fedora & CentOS and they were both complaining that PAM was disabled in sshd.log. Trying to flip it as having it enabled didn't cause any of the tests to fail on my local machine. When you have a chance, can you trigger the most recent commit to run against the full suite? I want to see how enabling PAM does, if anything.

chipsenkbeil:
@wez looks like there's some divergent behavior from openssh's implementation of realpath. Good news is that the authentication errors went away! I'm wondering if this also resolved the issue for FreeBSD, which is currently disabled.

chipsenkbeil:
@wez, FreeBSD still fails with the authentication error, so going to keep it skipped. Fixed the last test, so crossing fingers everything should now pass on this next run.

chipsenkbeil:
Finally! Alright, @wez, I believe this one is good to go from my end.

wez (changes requested):
Structurally this is looking good, thanks!

My feedback is mostly in two areas:
* Naming of the sftp methods to better match the rust std lib
* Error code/mode constants possibly being platform specific magic numbers; would be good to have a more deliberate mapping from the underlying libssh2 constants to the items exposed in our API

wezterm-ssh/src/session/sftp/types.rs
```diff
@@ -0,0 +1,349 @@
+const OCTAL_FT_DIR: u32 = 0o040000;
+const OCTAL_FT_FILE: u32 = 0o100000;
+const OCTAL_FT_SYMLINK: u32 = 0o120000;
```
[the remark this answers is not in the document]
chipsenkbeil: These are used both in libssh and libssh2.

* [libssh](https://git.libssh.org/projects/libssh.git/tree/include/libssh/sftp.h#n999)
* [libssh2](https://github.com/libssh2/libssh2/blob/master/include/libssh2_sftp.h#L139)

wezterm-ssh/src/session/sftp.rs
```diff
@@ -0,0 +1,440 @@
...
+            })))
+            .await?;
+        let result = rx.recv().await??;
+        Ok(result)
+    }
+
+    /// Resolve the real path for `path`.
+    ///
+    /// See [`Sftp::realpath`] for more information.
+    pub async fn realpath(&self, path: impl Into<PathBuf>) -> SftpChannelResult<PathBuf> {
```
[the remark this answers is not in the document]
chipsenkbeil: Sure, I can switch over to this.

wezterm-ssh/src/session/sftp/error.rs
```diff
@@ -0,0 +1,138 @@
...
+
+    /// Attempt to convert an arbitrary code to an sftp error, returning
+    /// `Ok` if matching an sftp error or `Err` if the code represented a
+    /// success or was unknown
+    fn try_from(code: i32) -> Result<Self, Self::Error> {
+        match code {
+            // 0 means okay in libssh and libssh2, which isn't an error
+            0 => Err(Ok(())),
+
+            1 => Ok(Self::Eof),
```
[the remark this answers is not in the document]
chipsenkbeil: 0 through 13 are defined in both libssh and libssh2 consistently. For 14 through 21, these only show up in libssh2 from what I can see.

* [libssh](https://git.libssh.org/projects/libssh.git/tree/include/libssh/sftp.h#n960)
* [libssh2](https://github.com/libssh2/libssh2/blob/master/include/libssh2_sftp.h#L192)

wezterm-ssh/src/session/sftp.rs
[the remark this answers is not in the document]
chipsenkbeil: @wez as a heads up, using camino makes the API a lot uglier, requiring a lot of conversions by the user. Guaranteeing utf8 paths doesn't clarify the `windows <-> unix` difference from my understanding. Is this something we really want in this API? It's going to cause me more headache in distant and the end result is converting any utf8 path to the std path before calling sftp underneath and converting back to a utf8 path any result.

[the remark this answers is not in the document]
chipsenkbeil: @wez I've converted the API to use generics for `TryInto<Utf8PathBuf, E>` to provide the simplest input I can think of. Only consequence is that this requires a `PathBuf` up front versus a `Path` whereas the old `Into<PathBuf>` accepted both. Thoughts?

PR #1185, opened 2021-09-30
Remove unused Error import

pLaborderie:
Hello, I noticed while installing WezTerm that there was a warning about an unused import.
I have removed this Error import in TermWiz.
I ran `cargo check` and `cargo test --all` with no errors.


wez:
Thanks!

PR #1187, opened 2021-10-02
ci: Simplify and clean up generate-workflows.py

friederbluemle:
## Summary

A bunch of whitespace and indentation fixes and cleanups in the generated GitHub Actions workflow files.

## Why?

I noticed trailing spaces and other irregularities while working on #1159. I couldn't help but figure out and fix what was causing it. ;)

## How to review

Most of the files are generated automatically. The only file that was updated manually (and needs to be reviewed) is `ci/generate-workflows.py` with 97 line changes.

There should be no functional/behavioral changes in CI before and after this PR.

Quick note regarding

```
    strategy:
      fail-fast: false
```

Since none of the jobs use a [matrix strategy](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix) (instead, individual workflows are used here), these lines have no effect and can be removed.

### DIFF
--- a/ci/generate-workflows.py
+++ b/ci/generate-workflows.py
@@ -3,7 +3,7 @@ import os
 import sys
 
 
-def yv(v):
+def yv(v, depth=0):
     if v is True:
         return "true"
     if v is False:
@@ -13,15 +13,18 @@ def yv(v):
 
     if isinstance(v, str):
         if "\n" in v:
-            spacer = " " * 12
-            return "|\n" + spacer + v.replace("\n", "\n" + spacer) + "\n"
+            indent = "  " * depth
+            result = ""
+            for l in v.splitlines():
+                result = result + "\n" + (f"{indent}{l}" if l else "")
+            return "|" + result
         return '"' + v + '"'
 
     return v
 
 
 class Step(object):
-    def render(self, f, env):
+    def render(self, f, env, depth=0):
         raise NotImplementedError(repr(self))
 
 
@@ -31,10 +34,11 @@ class RunStep(Step):
         self.run = run
         self.shell = shell
 
-    def render(self, f, env):
-        f.write(f"    - name: {yv(self.name)}\n")
+    def render(self, f, env, depth=0):
+        indent = "  " * depth
+        f.write(f"{indent}- name: {yv(self.name)}\n")
         if self.shell:
-            f.write(f"      shell: {self.shell}\n")
+            f.write(f"{indent}  shell: {self.shell}\n")
 
         run = self.run
 
@@ -43,7 +47,7 @@ class RunStep(Step):
                 if self.shell == "bash":
                     run = f"export {k}={v}\n{run}\n"
 
-        f.write(f"      run: {yv(run)}\n")
+        f.write(f"{indent}  run: {yv(run, depth + 2)}\n")
 
 
 class ActionStep(Step):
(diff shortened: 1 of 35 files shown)


ci/generate-workflows.py
```diff
@@ -54,19 +58,22 @@ def __init__(self, name, action, params=None, env=None, condition=None):
...
+        f.write(f"{indent}  uses: {self.action}\n")
         if self.condition:
-            f.write(f"      if: {self.condition}\n")
+            f.write(f"{indent}  if: {self.condition}\n")
         if self.params:
-            f.write("      with:\n")
+            f.write(f"{indent}  with:\n")
             for k, v in self.params.items():
-                f.write(f"         {k}: {yv(v)}\n")
+                d = depth + 4 if v is str and "\n" in v else depth + 3
```
wez, line 70: This special casing based on the value feels like we're leaking implementation details of the `yv` function out to the caller.  Indentation of the `v` should be handled by `yv` itself based on the provided depth.

The 3 and 4 constants here are a bit unexpected to me; this sort of pattern of passing down the depth usually indicates the actual call depth and not some externally adjusted depth.

ci/generate-workflows.py
```diff
@@ -54,19 +58,22 @@ def __init__(self, name, action, params=None, env=None, condition=None):
...
+                f.write(f"{indent}    {k}: {yv(v, d)}\n")
```
wez, line 71: I feel like this should just be:

```suggestion
                f.write(f"{indent}    {k}: {yv(v, depth + 1)}\n")
```

and then `yv` would do the appropriate indent based on the depth

ci/generate-workflows.py
```diff
@@ -90,9 +97,9 @@ def __init__(self, runs_on, container=None, steps=None, env=None):
         self.steps = steps
         self.env = env
 
-    def render(self, f):
+    def render(self, f, depth=3):
```
wez, line 100: The default value here should be zero to match the definition of `Step.render`; overriding with 3 feels a bit magical to me.

```suggestion
    def render(self, f, depth=0):
```

wez (changes requested):
Thanks for looking at this!

I like the general approach, but I think there are a couple of places that bleed implementation details when they don't need to (see inline comments/suggestions).

I use [black](https://pypi.org/project/black/) to format `generate-workflows.py`, but the CI doesn't verify/enforce that today.  Please run `black` over that file :-)

ci/generate-workflows.py
```diff
@@ -43,7 +47,7 @@ def render(self, f, env):
                 if self.shell == "bash":
                     run = f"export {k}={v}\n{run}\n"
 
-        f.write(f"      run: {yv(run)}\n")
+        f.write(f"{indent}  run: {yv(run, 5)}\n")
```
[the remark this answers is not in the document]
friederbluemle: Good point. 5 looks magical. It should actually be `depth + 2` here, because we're writing the run statements of a (potential) multiline block:

```
- name:     <-- depth "0"
  shell:    <-- depth +1  
  run: |    <-- depth +1
    line1   <-- depth +2
    line2   <-- depth +2

File: ci/generate-workflows.py [hunk 5]
```diff
@@ -31,10 +34,11 @@ def __init__(self, name, run, shell="bash"):
         self.run = run
         self.shell = shell
 
-    def render(self, f, env):
-        f.write(f"    - name: {yv(self.name)}\n")
+    def render(self, f, env, depth=0):
+        indent = "  " * depth
+        f.write(f"{indent}- name: {yv(self.name)}\n")
```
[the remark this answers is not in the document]
friederbluemle: The fundamental "problem" is that the single `yv` function takes on two related, but still quite different tasks:

1. Format a single value (e.g. a boolean, or a string without newlines), in which case the depth would have no meaning, and the result would just be inserted right where `yv` is called
2. Format a multiline string, in which case the result would be written to the _next_ line, at the indentation controlled by depth

This was the case before this PR as well, I didn't change it. If you have a good suggestion for a larger refactoring, I'd be happy to take it on.

ci/generate-workflows.py
friederbluemle: Nicely spotted :) This was one line that both took me the longest to figure out, and I wasn't really happy with as well. I had another look, and it turned out the special casing is not necessary here (I think it was at one point during the refactoring).

ci/generate-workflows.py
friederbluemle: Yes, it is possible to use a simpler expression here, thanks! The value is `depth + 3`:

```
- name: "Cache cargo"             <--- depth "0"
  uses: actions/cache@v2.1.6      <--- depth +1
  with:                           <--- depth +1
    path: |                       <--- depth +2
      ~/.cargo/registry           <--- depth +3
      ~/.cargo/git                <--- depth +3
      target                      <--- depth +3
```

ci/generate-workflows.py
friederbluemle: Agree! Probably missed this before my final review. I changed it to 0.

friederbluemle:
Thanks a lot for the thorough review :+1: I changed some of the code and addressed the comments.

I installed black and ran `black ci/generate-workflows.py`, and this is what I got:
```
All done! ✨ 🍰 ✨
1 file left unchanged.
```
So seems like the file is formatted correctly already? Anything else I need to specifiy?

wez:
Thank you!

PR #1190, opened 2021-10-02
Add full assignment example. Clarify Shift v SHIFT.

follower:
Thanks for your work on `wezterm`, I've been trying it out off & on and this time I got as far as setting up key assignments! :)

### Context

Unfortunately setting up key assignments turned out to be more frustrating than it needed to be due to the order of the required information in the docs (and my lack of Lua knowledge).

So, this PR adds a complete example earlier in the documentation flow to hopefully prevent others to running into the same situation.

### Details

The existing example uses "Nop" which can be supplied as string but other key assignments require use of `wezterm.action`.

For people not familiar with Lua (like me :) ) a full example with correct syntax is helpful at this point in the documentation.

Additionally, it was not immediately obvious to me that the Shift key requires different capitalization in different contexts. The resulting error messages also don't mention the significance of the capitalization. So this commit adds a note about this detail.

Hopefully these additions will make the initial key assignment experience more approachable & less frustrating. :)

(The key combination used in this example matches that used by the elementary OS terminal.)

_[I'm not particular about the specific wording used so feel free to modify as desired--especially as (for expectation management purposes :) ) I tend not to be great at following up drive-by PRs.]_

Thanks!

### DIFF
--- a/docs/config/keys.md
+++ b/docs/config/keys.md
@@ -264,6 +264,24 @@ Possible Modifier labels are:
 
 You can combine modifiers using the `|` symbol (eg: `"CMD|CTRL"`).
 
+> Note: To specify the shift key in `key` use the capitalization `Shift` but
+>       for use in `mods` use the capitalization `SHIFT`.
+
+A complete example to demonstrate use of `wezterm.action`, use of the shift key
+as a modifier key, and a key assignment which requires an argument:
+
+```lua
+local wezterm = require 'wezterm';
+
+return {
+  keys = {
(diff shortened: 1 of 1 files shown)


wez:
Thanks for the feedback.  I'm not sure that the proposed changes really clarify too much: it seems to me that the issue is in seeing which actions you want to take and not seeing how to use them.  I expanded the docs a little to link to the actions from that initial Nop example, and encourage clicking through to read the individual action examples that are present in the docs.

PR #1227, opened 2021-10-13
Add a recipes page. Fixes #1223

### LINKED ISSUE #1223: Docs need a recipe section
It is hard to get started in a tmux/GNU screen style terminal multiplexer configuration.

A recipes section for users looking to start with specific workflows or coming from other terminals/multiplexers. e.g, I came to wezterm looking for tmux style usage but could not find how to do it (unix domain and automatic connect) anywhere on blogs etc. and literally just happened to stumble on the hopefully correct way of doing so. Otherwise I'd have given up.

Another recipe could be tmux and/or screen equivalent key mappings.

Another example recipe could be how to rotate background image e.g. I cooked some crude solution up with picking a random image from a directory on config re/load.

etc.

Continue using kitty with tmux if I can't figure wezterm's mux setup out.

**Additional context**

junkblocker:


wez:
While I think this is a good start, I think we need more recipes before we can really call it a recipes section!
I'd love to see a bit more brainstorming on what set of recipes we should have, and then we can just go and flesh it out.
For now, I'm going to close this PR.
Thanks!

PR #1275, opened 2021-10-27
pty: add `CommandBuilder::env_clear`, `CommandBuilder::env_remove` functions

### LINKED ISSUE #1274: [portable-pty] no `CommandBuilder::env_clear` function

macOS

### WezTerm version

portable-pty=0.5.0

### Describe the bug

There is no way to avoid inheriting the parent environment in `portable-pty`'s `CommandBuilder`. The relevant lines are here:

https://github.com/wez/wezterm/blob/cf1c4240c29962f63aa24475d569229c565a0d18/pty/src/cmdbuilder.rs#L337-L340

For comparison, Rust has a `Command::env_clear` function: https://doc.rust-lang.org/std/process/struct.Command.html#method.env_clear

We use this for testing https://github.com/arxanas/git-branchless (cc @bcongdon). In principle, we could clear out the global environment, but in practice, it causes tests to encounter race conditions, because they run in parallel.

### To Reproduce

_No response_

### Configuration

N/A

### Expected Behavior

_No response_

### Logs

_No response_

### Anything else?

_No response_

arxanas:
Closes #1274.

### DIFF
--- a/mux/src/ssh.rs
+++ b/mux/src/ssh.rs
@@ -453,7 +453,7 @@ impl Domain for RemoteSshDomain {
             Some(cmd.as_unix_command_line()?)
         };
         let mut env: HashMap<String, String> = cmd
-            .iter_env_as_str()
+            .iter_extra_env_as_str()
             .map(|(k, v)| (k.to_string(), v.to_string()))
             .collect();
         env.insert("WEZTERM_PANE".to_string(), pane_id.to_string());
@@ -613,7 +613,7 @@ impl Domain for RemoteSshDomain {
             Some(cmd.as_unix_command_line()?)
         };
         let mut env: HashMap<String, String> = cmd
-            .iter_env_as_str()
+            .iter_extra_env_as_str()
             .map(|(k, v)| (k.to_string(), v.to_string()))
             .collect();
         env.insert("WEZTERM_PANE".to_string(), pane_id.to_string());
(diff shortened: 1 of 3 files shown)


wez:
Things are just getting back to sanity after moving house; sorry for the delay in responding!
I think this looks good, but there is one more important difference to reconcile; in the ssh pty implementations:

https://github.com/wez/wezterm/blob/main/pty/src/ssh.rs#L200
and
https://github.com/wez/wezterm/blob/main/mux/src/ssh.rs#L455-L459

the logic was effectively to iterate over only items that had been added to the originally empty environment and send those to the remote host.

With this PR we're now going to send all of the local environment to the remote host, which isn't desirable.

What do you think about adding a boolean to the `EnvEntry` to indicate that a given entry was part of the base environment, and then we can change `iter_env_as_str` to skip the base environment?  Should probably either rename that method or introduce a parameter to control whether the base env is included.

arxanas:
@wez This should be good to review now.

wez:
Looks good, many thanks for this!

wez:
Are you desperate for a release to be published to crates.io, or are you happy with a git repo reference for the time being?
Just trying to gauge when I need to sit down and make a pass over releasing that crate :)

wez:
Ah, turned out to be easy to bump; I've published 0.6.0

arxanas:
@wez Thanks for doing the release, it makes things a lot easier on my end.

PR #1301, opened 2021-11-07
Unzoom when switching panes

yyogo:
Add the `unzoom_on_switch_pane` config flag, which determines whether switching panes with `ActivatePaneDirection` while the active pane is zoomed in should first unzoom and then switch (similar to tmux' behavior), or keep the pane zoomed and do nothing.

I set it to true by default, it may be desirable to keep it false to keep current behavior.

### DIFF
--- a/docs/config/lua/config/unzoom_on_switch_pane.md
+++ b/docs/config/lua/config/unzoom_on_switch_pane.md
@@ -1,0 +1,9 @@
+# `unzoom_on_switch_pane = true`
+
+If set to `false`, the 
+[`ActivatePaneDirection`](../keyassignment/ActivatePaneDirection.md) command
+will have no effect if the active pane is zoomed. 
+
+If `true`, the active pane will be unzoomed first and then switched.
+
+See also: [`TogglePaneZoomState`](../keyassignment/TogglePaneZoomState.md)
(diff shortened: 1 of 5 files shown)


mux/src/tab.rs
```diff
@@ -1038,7 +1039,11 @@ impl Tab {
     /// edge intersection.
     pub fn activate_pane_direction(&self, direction: PaneDirection) {
         if self.zoomed.borrow().is_some() {
-            return;
+            if configuration().unzoom_on_switch_pane {
+                self.toggle_zoom();
+            } else {
+                return;
+            }
```
wez, line 1046: I prefer early-return style:

```suggestion
            if !configuration().unzoom_on_switch_pane {
                return;
            }
            self.toggle_zoom();
```

docs/config/lua/config/unzoom_on_switch_pane.md
```diff
@@ -0,0 +1,9 @@
+# `unzoom_on_switch_pane = true`
+
```
wez, line 2:
```suggestion

*Since: nightly builds only*

```

wez (commented):
Thanks for this!

PR #1307, opened 2021-11-08
Manjaro requires additional base-devel package

mculp:
When running `cargo build --release` on a fresh install of Manjaro 21.1.6, I got the following message:

```sh
   Compiling libc v0.2.107
   Compiling proc-macro2 v1.0.32
   Compiling unicode-xid v0.2.2
   Compiling autocfg v1.0.1
   Compiling cfg-if v1.0.0
   Compiling syn v1.0.81
   Compiling version_check v0.9.3
error: linker `cc` not found
  |
  = note: No such file or directory (os error 2)

error: could not compile `libc` due to previous error
warning: build failed, waiting for other jobs to finish...
error: build failed
```

I installed `base-devel` afterwards and was able to continue.

### DIFF
--- a/get-deps
+++ b/get-deps
@@ -96,6 +96,11 @@ fi
 
 if test -e /etc/arch-release ; then
   PACMAN="$SUDO pacman"
+  
+  if test -e /etc/manjaro-release ; then
+    $PACMAN -S --noconfirm --needed base-devel
+  fi
+  
   $PACMAN -S --noconfirm --needed \
     'cargo' \
     'cmake' \
--- a/get-deps
+++ b/get-deps
@@ -96,12 +96,8 @@ fi
 
 if test -e /etc/arch-release ; then
   PACMAN="$SUDO pacman"
-  
-  if test -e /etc/manjaro-release ; then
-    $PACMAN -S --noconfirm --needed base-devel
-  fi
-  
   $PACMAN -S --noconfirm --needed \
+    'base-devel' \
     'cargo' \
     'cmake' \
     'fontconfig' \


get-deps
```diff
@@ -96,6 +96,11 @@ fi
 
 if test -e /etc/arch-release ; then
   PACMAN="$SUDO pacman"
+  
+  if test -e /etc/manjaro-release ; then
+    $PACMAN -S --noconfirm --needed base-devel
+  fi
+  
   $PACMAN -S --noconfirm --needed \
     'cargo' \
```
wez, line 105:
```suggestion
    'base-devel' \
    'cargo' \
```

get-deps
```diff
@@ -96,6 +96,11 @@ fi
 
 if test -e /etc/arch-release ; then
   PACMAN="$SUDO pacman"
+
```
wez, line 99: I'm not an Arch expert, but https://archlinux.org/groups/x86_64/base-devel/ suggests that `base-devel` is part of arch itself, so I think you could move this to the list below

get-deps
```diff
@@ -96,6 +96,11 @@ fi
 
 if test -e /etc/arch-release ; then
   PACMAN="$SUDO pacman"
+  
+  if test -e /etc/manjaro-release ; then
```
wez, line 100:
```suggestion
```

get-deps
```diff
@@ -96,6 +96,11 @@ fi
...
+    $PACMAN -S --noconfirm --needed base-devel
```
wez, line 101:
```suggestion
```

get-deps
wez:
```suggestion
```

get-deps
```diff
@@ -96,6 +96,11 @@ fi
...
+  fi
```
wez, line 102:
```suggestion
```

get-deps
```diff
@@ -96,6 +96,11 @@ fi
...
+
```
wez, line 103:
```suggestion
```

wez (commented):
Thanks for this!

get-deps
mculp: @wez Yeah, I just assumed the existing list worked on arch so I didn't want to install anything extra on vanilla arch if it was only necessary on manjaro

PR #1312, opened 2021-11-11
Improve powerline semicircle custom glyphs

### LINKED ISSUE #1311: powerline semicircle custom glyphs are too pointy
sorry if the wording is strange, I'm using google translator.
I have a problem with the zsh prompt with the theme of powerlevel10k, the semicircle shaped sides are "weird" more pointed than they appear in other terminals

left Tilix / right Wezterm
both terminals with the same source

Screenshot_20211110_194038

Do you know of any solution? , beyond that I have not had another problem, I am liking the terminal a lot, it has everything I am looking for

bew:
Closes #1311 

The main idea is to split the semi circle in 2 arcs, the top one and the bottom one.

Final result:
wezterm-better-semicircle-chars--as-designed

---
<details>
<summary>Alternative render for stroke semicircle</summary>

NOTE: At one point I had this result:
wezterm-better-semicircle-chars--smooth-to-the-edge

But compared to the reference result https://github.com/ryanoasis/powerline-extra-symbols#glyphs , it looks wrong..

</details>


sdrik:
Outline glyphs look good with large font sizes (eg. 100), but not so great with small font sizes (eg. 11).
See #1313 for an alternate solution built upon this PR.

image

bew:
Replaced by #1313

PR #1313, opened 2021-11-11
Improve powerline semicircle custom glyphs (Builds upon #1312 with better outline glyphs)

### LINKED ISSUE #1311: powerline semicircle custom glyphs are too pointy

sdrik:
Closes #1311

Final result at font size 100 :

image

Final result at font size 11 :

image


bew:
Hello, good call on the small size font, completely missed that ><

I want to point out that with your solution, the semicircle appears off-centered:
wezterm-better-semicircle-chars--other-solution

I'm thinking it would be a bit better if it could be centered horizontally?

---
NOTE: For the enums I see you didn't take the `ZeroWithOffset` / `OneWithOffset` from my PR, (and I know it's quiet subjective), I think those are more explicit than playing with `FracWithOffset` with non-fractions. Tough If you manage to center it in the cell it probably won't matter anymore since you'd always have fractional coords..

sdrik:
It will be difficult with my method to get it perfectly centered, but here is what can be easily achieved:
image

bew (approved):
Looks good to me :+1:

bew (changes requested):
Actually no, I think you can simplify the outline ones, you don't need any FracWithOffset, since all points are basically outside of the cell.
_Also in your last image, I noticed the line doesn't exactly pass at the same X at the top and bottom, because you used FracWithOffset for the top X, and Frac for the bottom X.._

I suggest this diff: (use `Frac` only, and put coords on one line since they're shorter now)
```diff
diff --git a/wezterm-gui/src/customglyph.rs b/wezterm-gui/src/customglyph.rs
index 4dcab31c..ad0b8e27 100644
--- a/wezterm-gui/src/customglyph.rs
+++ b/wezterm-gui/src/customglyph.rs
@@ -3506,19 +3506,13 @@ impl BlockKey {
                 style: PolyStyle::Fill,
             }]),
             // [] Powerline outline left semicircle
             0xe0b5 => Self::Poly(&[Poly {
                 path: &[
-                    PolyCommand::MoveTo(
-                        BlockCoord::Frac(-1, 4),
-                        BlockCoord::FracWithOffset(-1, 3, LineScale::Mul(1)),
-                    ),
+                    PolyCommand::MoveTo(BlockCoord::Frac(-1, 4), BlockCoord::Frac(-1, 3)),
                     PolyCommand::QuadTo {
-                        control: (
-                            BlockCoord::FracWithOffset(7, 4, LineScale::Mul(-1)),
-                            BlockCoord::Frac(1, 2),
-                        ),
+                        control: (BlockCoord::Frac(7, 4), BlockCoord::Frac(1, 2)),
                         to: (BlockCoord::Frac(-1, 4), BlockCoord::Frac(4, 3)),
                     },
                 ],
                 intensity: BlockAlpha::Full,
                 style: PolyStyle::Outline,
@@ -3541,23 +3535,14 @@ impl BlockKey {
                 style: PolyStyle::Fill,
             }]),
             // [] Powerline outline right semicircle
             0xe0b7 => Self::Poly(&[Poly {
                 path: &[
-                    PolyCommand::MoveTo(
-                        BlockCoord::FracWithOffset(5, 4, LineScale::Mul(1)),
-                        BlockCoord::FracWithOffset(-1, 3, LineScale::Mul(1)),
-                    ),
+                    PolyCommand::MoveTo(BlockCoord::Frac(5, 4), BlockCoord::Frac(-1, 3)),
                     PolyCommand::QuadTo {
-                        control: (
-                            BlockCoord::FracWithOffset(-3, 4, LineScale::Mul(1)),
-                            BlockCoord::Frac(1, 2),
-                        ),
-                        to: (
-                            BlockCoord::FracWithOffset(5, 4, LineScale::Mul(1)),
-                            BlockCoord::Frac(4, 3),
-                        ),
+                        control: (BlockCoord::Frac(-3, 4), BlockCoord::Frac(1, 2)),
+                        to: (BlockCoord::Frac(5, 4), BlockCoord::Frac(4, 3)),
                     },
                 ],
                 intensity: BlockAlpha::Full,
                 style: PolyStyle::Outline,
             }]),
```

sdrik:
You're right, offsets are not needed anymore since the stroke has been shifted.
With my initial solution, they were necessary because I was getting things like that (not visible with large font sizes):
image

bew (approved):

wez:
Looks good; thanks to you both!

PR #1317, opened 2021-11-14
Quickselect: allow limiting number of lines before/after viewport to consider

yyogo:
Add `lines_{before,after}_viewport` option to `QuickSelectArgs`, which controls the search scope around the viewport. When unspecified, the entire scrollback is considered.

Quickselect can take a while to run when the scrollback is full. Scoping makes it run faster and consider less matches so that shorter prefixes are used more often. When the viewport is changed (on scroll/resize), quickselect will update the available matches and shortcuts.

### DIFF
--- a/wezterm-gui/src/overlay/quickselect.rs
+++ b/wezterm-gui/src/overlay/quickselect.rs
@@ -523,7 +523,20 @@ impl QuickSelectRenderable {
 
     fn compute_range(&self) -> Range<StableRowIndex> {
         let dims = self.delegate.get_dimensions();
+        let top = if let Some(lines_before) = self.args.lines_before_viewport {
+            self.viewport
+                .unwrap_or(dims.physical_top)
                 .saturating_sub(lines_before as isize)
+        } else {
+            dims.scrollback_top
+        };
+        let bottom = if let Some(lines_after) = self.args.lines_after_viewport {
+            (top + dims.viewport_rows as StableRowIndex)
+                .saturating_sub(1)
+                .saturating_add(lines_after as isize)
+        } else {
+            dims.scrollback_top + dims.scrollback_rows as StableRowIndex
+        };
         top..bottom
     }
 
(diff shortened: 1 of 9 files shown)


wez:
thinking about this a bit more, I think the search should probably be scoped to the current viewport position + some number of lines above and below it

yyogo:
So should there be a config var `quickselect_scope_lines` for how many lines around the viewport are used?

wez:
Yeah, that sounds like a better name for the config option!

wez:
I think that setting should specify the number of lines above and below.  For example, if it is set to 100 then the search range is the 100 lines before the viewport, the viewport itself, and then 100 lines after the viewport.

wez:
Now that we have [QuickSelectArgs](https://wezfurlong.org/wezterm/config/lua/keyassignment/QuickSelectArgs.html) I think this PR should add optional fields for the number of lines before and the number of lines after the viewport.  If omitted, the whole viewport is searched.  I think this will give the most flexibility for users!

wezterm-gui/src/overlay/quickselect.rs
```diff
@@ -258,6 +258,7 @@ impl QuickSelectOverlay {
                 render.dirty_results.add(*pos);
             }
             render.viewport = viewport;
+            render.update_search();
```
wez, line 261: For the case where no range was set in the args, then I think this will re-search the entire scrollback each time the window is scrolled? That feels undesirable!
Is your goal to update the highlighted text as the scrollback is moved?
If so, I think it might be best to make this conditional on the range arguments and update only if the new viewport is now outside of the last-searched range

wezterm-gui/src/overlay/quickselect.rs
```diff
@@ -660,14 +678,6 @@ impl QuickSelectRenderable {
                             let mut r = search_overlay.renderer.borrow_mut();
                             r.results = results.take().unwrap();
                             r.recompute_results();
-                            let num_results = r.results.len();
-
-                            if !r.results.is_empty() {
```
wez, line 665: Removing this code changes the behavior for the case where no range is specified.
What's your rationale for this? Does it make sense to only skip this when range(s) were passed in?
Should this be a separate argument for quickselect?

wez (commented):
Thanks! This is looking good!  I have a couple of comments!

wez:
I'm going to close this out for now; please reopen it when you have time to rebase and respond to my earlier comments!

PR #1338, opened 2021-11-23
fix: do not propagate $APPIMAGE and $APPDIR env vars to child processes

srevinsaju:
Some apps use `$APPIMAGE` and `$APPDIR` environment variables to figure out if they are running as an AppImage. When apps (which are not appimages) were launched from Wezterm AppImage, it tried to get resources from Wezterm's mount (since they tried to get the path relative to the $APPDIR`)

### DIFF
--- a/config/src/lib.rs
+++ b/config/src/lib.rs
@@ -1839,12 +1839,12 @@ impl Config {
         // across the win32/wsl boundary
         let mut wsl_env = std::env::var("WSLENV").ok();
 
-        // Since we are running inside the appimage, we have "$APPIMAGE"
-        // and "$APPDIR" set inside the wezterm appimage. This gets
+        // If we are running as an appimage, we will have "$APPIMAGE"
+        // and "$APPDIR" set in the wezterm process. These will be
         // propagated to the child processes. Since some apps (including
         // wezterm) use these variables to detect if they are running in
-        // an appimage, the child processes get misconfigured
-        // We should just unset them too.
+        // an appimage, those child processes will be misconfigured.
+        // Ensure that they are unset.
         // https://docs.appimage.org/packaging-guide/environment-variables.html#id2
         cmd.env_remove("APPIMAGE");
         cmd.env_remove("APPDIR");
--- a/config/src/lib.rs
+++ b/config/src/lib.rs
@@ -1839,6 +1839,17 @@ impl Config {
         // across the win32/wsl boundary
         let mut wsl_env = std::env::var("WSLENV").ok();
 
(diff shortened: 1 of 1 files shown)


wez:
Do you need `$APPIMAGE` cleared? You only mentioned `$APPDIR`.  There is other code in wezterm that inspects `$APPIMAGE` because it needs to adjust its behavior when running from an appimage.

It's also worth pointing out that the `wezterm` executable spawns the `wezterm-gui` executable that is also located in the appimage; at the moment it looks as though this PR may cause the appimage to get unmounted out from under a gui process launched in this way:

https://github.com/wez/wezterm/blob/3b30230d5810634940bb4d4e349314ad366e4762/wezterm/src/main.rs#L349-L360

I think the path forwards here is:

1. Wait for https://github.com/wez/wezterm/pull/1275 to land (I plan to review and merge this today)
2. Around [here](https://github.com/wez/wezterm/blob/main/config/src/lib.rs#L1829) (prior to applying the `set_environment_variables` list to the environment) use the new `env_remove` method to remove the set of variables listed [in the appimage spec](https://docs.appimage.org/packaging-guide/environment-variables.html#id2).  This should preserve the APPIMAGE-ness of wezterm itself, but prevent bleeding this environment to newly spawned panes.

However, I don't have full information on the issue you're aiming to resolve with this PR, so I don't know if this is sufficient!

srevinsaju:
Agreed, your suggestion looks appropriate. I shall wait for the merge! Thank you for your awesome terminal! :smile:

wez:
#1275 has been merged!

srevinsaju:
Thank you! I have updated the patch according to your suggestions :innocent:

config/src/lib.rs
```diff
@@ -1839,6 +1839,17 @@ impl Config {
         // across the win32/wsl boundary
         let mut wsl_env = std::env::var("WSLENV").ok();
 
+        // Since we are running inside the appimage, we have "$APPIMAGE"
+        // and "$APPDIR" set inside the wezterm appimage. This gets
+        // propagated to the child processes. Since some apps (including
+        // wezterm) use these variables to detect if they are running in
+        // an appimage, the child processes get misconfigured
+        // We should just unset them too.
```
wez, line 1847:
```suggestion
        // If we are running as an appimage, we will have "$APPIMAGE"
        // and "$APPDIR" set in the wezterm process. These will be
        // propagated to the child processes. Since some apps (including
        // wezterm) use these variables to detect if they are running in
        // an appimage, those child processes will be misconfigured.
        // Ensure that they are unset.
```

wez:
Thanks!

PR #1346, opened 2021-11-27
Update Clink hyperlink to newer version

rashil2000:
This is a newer and very actively maintained official fork (also used by Cmder now) of Clink.


wez:
Looks good, thanks!

PR #1376, opened 2021-12-08
DECSTR should reset DECLRMM

ninjalj:
While the description at DEC STD-070 04-31 does not mention Left Right
Margin Mode (it does mention the left and right margins), the code at
04-34 sets it to FIXED (the reset state, the set state is SETABLE).

From esctest:
DECSETTests.test_DECSET_DECLRMM_ModeResetByDECSTR


wez:
Thanks!

PR #1388, opened 2021-12-10
Passthru exit code for other commands in $PROMPT_COMMAND

SuperSandro2000:


wez:
Would you mind saying a sentence or two about what was broken? That will help to write up the changelog!

SuperSandro2000:
My prompt is loaded later in $COMMAND_PROMPT and uses $? to get the exit status of the last run command to display in the terminal. Before this change this was always 0. I tested it with a command which does not exist and after this change it once again display 127 instead of 0.

PR #1390, opened 2021-12-12
docs(installation): fix links

vedantmgoyal2009:
Fixes links that were resulting 404 Not Found error.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -1,9 +1,9 @@
 WezTerm is available pre-built for the major platforms and, because it is open
 source, you may also build it for yourself.
 
- - [Windows](install/windows.md)
- - [macOS](install/macos.md)
- - [Linux](install/linux.md)
- - [FreeBSD](install/freebsd.md)
- - [Build from source](install/source.md)
+ - [Windows](install/windows.markdown)
+ - [macOS](install/macos.markdown)
+ - [Linux](install/linux.markdown)
+ - [FreeBSD](install/freebsd.markdown)
+ - [Build from source](install/source.markdown)
 


wez:
The links do already work (you can see that at https://wezfurlong.org/wezterm/installation.html); you need to run `ci/build-docs.sh` to generate the individual installation pages and substitute release information!

vedantmgoyal2009:
It doesn't work on GitHub:

https://user-images.githubusercontent.com/83997633/145702348-5ace1200-11c6-49de-85e4-428ff4cfa294.mp4

wez:
The docs directory is source material for the doc build which gets published to the website; they're not intended to be fully browsable via github.
This commit will break the doc build!

vedantmgoyal2009:
Okay, then I would recommend adding a **readme.md** file in the **docs** directory which would be rendered by GitHub and inform users to browse the docs at https://wezfurlong.org/wezterm instead of GitHub. Also, if possible you can exclude it in the "pages" workflow so that it won't affect the website. I have no knowledge about how your workflow works otherwise I would have made the changes in this PR only.

wez:
What led you to the docs dir?  The about section on the right links to the website and its docs, and the readme also points to the user facing docs: https://github.com/wez/wezterm#readme

I'm wondering what else we can do to help guide folks to the right place rather than have them working through the github repo, as there's a lot of content in the repo that I don't think most folks should worry about!

I like your readme suggestion, but care needs to be taken to avoid it tricking the doc build itself!  I'll look into that tomorrow!

I've also pushed a workflow commit that should trigger the page build to run on PRs to verify doc changes; I think we lost that by accident a while back.

vedantmgoyal2009:
> What led you to the docs dir?

Actually, most of the time the links featured in the **About** section get ignored by me and I directly scroll and look into the readme or docs (or any other similar) folder and browse through it. Not sure, if other users do the same but this only made me find the 404 links 😀.

image

PR #1403, opened 2021-12-15
docs: make 'borderless' searchable

eugenesvk:
Tiny edit in the docs to help future users avoid my mistake
(would've prevented https://github.com/wez/wezterm/issues/1397)


aorith (changes requested):
There's a little _typo_: `bordlerless` should be `borderless`

wez:
Thanks!

PR #1405, opened 2021-12-15
docs: replace AltDrag with AltSnap

eugenesvk:
AltDrag is old (latest release in 2015) and unusable with WezTerm (too much lag), AltSnap is the maintained fork that works with only a little hiccup


wez:
Thanks!

PR #1415, opened 2021-12-18
Support gating libssh-rs and ssh2 behind features of same name

chipsenkbeil:
Moves libssh-rs and ssh2 backends to be gated by features of the same name. This enables building wezterm-ssh with only one of two dependencies, getting around problems with compiling both for a cdylib target.

You can verify that each backend works by running with `cargo test --no-default-features --features libssh-rs` and `cargo test --no-default-features --features ssh2` from within the wezterm-ssh directory. If `libssh-rs` is not present, then the backend will be switched to `ssh2`.

### DIFF
--- a/wezterm-ssh/Cargo.toml
+++ b/wezterm-ssh/Cargo.toml
@@ -11,6 +11,7 @@ documentation = "https://docs.rs/wezterm-ssh"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [features]
+default = ["libssh-rs", "ssh2"]
 vendored-openssl = ["ssh2/vendored-openssl", "libssh-rs/vendored-openssl"]
 
 [dependencies]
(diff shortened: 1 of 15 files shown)


chipsenkbeil:
@wez the only thing not done right now is how to handle the `vendored-openssl` feature. With stable cargo, specifying the feature of a dependency as enabled will always include that feature, even if the dependency itself hasn't been included.

There was an issue about it https://github.com/rust-lang/cargo/issues/3494, with a solution called [weak dependency features](https://github.com/rust-lang/cargo/pull/8818) merged in last year, but this feature is still only available on nightly cargo.

chipsenkbeil:
@wez if there's any early comments you might have, let me know :) I'm assuming everyone is taking a break throughout the holidays, so not expecting a full review until next year. This is one of the blockers for the next release of distant, so would love to get this rolling, though!

wez:
Looks good, thanks!

chipsenkbeil:
@wez just noticed that the weak feature functionality of cargo is now released: https://blog.rust-lang.org/2022/04/07/Rust-1.60.0.html#new-syntax-for-cargo-features.

If/when you bump to the new cargo version, you could collapse the old features above, removing the need for features like `vendored-openssl-ssh2`, `vendored-openssl-libssh-rs`. Instead, you'd have `vendored-openssl = ["ssh2?/vendored-openssl", "libssh-rs?/vendored-openssl"]`.

Issue #1424, opened 2021-12-19
Feature request: option to not scroll cursor with iTerm2 images


ghost:
I'm playing with iTerm2 image output, and noticing that when I put image data on the bottom line it scrolls.  This can be seen via:

```
java -Djexer.ECMA48.iTerm2Images=true -Djexer.hideStatusBar=true -jar path/to/jexer.jar
```

Then open an image, drag it into the bottom row, and things scroll.  This is similar to sixel, but does not have to be.

The [iTerm2 document](https://iterm2.com/documentation-images.html) provides no means to prevent the cursor from being moved, and mintty did not [extend or document a way either](https://github.com/mintty/mintty/wiki/CtrlSeqs#image-support) in that implementation.  (I haven't been following the ecosystem for a while, has anyone else implemented iTerm2?)

It would be nice to have a way to put image data on the bottom row for fullscreen applications to use.  Would you have an objection to me implementing something like an optional "noScroll=1" to mean "leave the cursor where it was" for wezterm?

klamonte:
Oh how convenient!   Seeing as how do_not_move_cursor is already there, I am stubbing in a path for "doNotMoveCursor=1" to do exactly what I want.  (It's just like the inline option.)

klamonte:
Compatibility: I just validated that both mintty (current) and iTerm2 (circa 2020) do not have adverse behavior to including "doNotMoveCursor=1" in the image sequence, i.e. they behave the same with and without the parameter.

wez:
I don't have an objection to the implementation, I'm just a little cagey about diverging behavior over time.  I wonder if @gnachman would consider making a similar change to iTerm2?

wez:
I went ahead and merged your PR in the meantime :)

PR #1433, opened 2021-12-20
New optional argument 'doNotMoveCursor' for iTerm2 images.

klamonte:
For #1424

New optional argument 'doNotMoveCursor' for iTerm2 images.  This permits iTerm2 images to be drawn anywhere on screen without scrolling the cursor, including the bottom row.

Also included is a check in fcwrap.rs to_range_set(), without which was causing a panic at runtime due to subtraction from unsigned leading to overflow.

(Sidebar: I had to switch to Rust nightly to compile git head, due to the sysinfo dependency not compiling with Rust stable.  Is that intended behavior?)

### DIFF
--- a/wezterm-font/src/fcwrap.rs
+++ b/wezterm-font/src/fcwrap.rs
@@ -125,7 +125,7 @@ impl<'a> CharSetRef<'a> {
                 for j in 0..32 {
                     if mask & (1 << j) != 0 {
                         let new_code_point = base_code_point + (j + i * 32) as u32;
-                        if new_code_point - 1 > code_point {
+                        if new_code_point > 0 && new_code_point - 1 > code_point {
                             coverage.add_range_unchecked(range_start..code_point + 1);
                             range_start = new_code_point;
                         }
--- a/term/src/terminalstate/iterm.rs
+++ b/term/src/terminalstate/iterm.rs
@@ -151,7 +151,7 @@ impl TerminalState {
             style: ImageAttachStyle::Iterm,
             image_id: None,
             placement_id: None,
-            do_not_move_cursor: false,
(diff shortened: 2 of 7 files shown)


wez:
re: rust version, stable Rust 1.57 builds things just fine.  I suspect that you're using a distribution-provided version of Rust that hasn't caught up to the latest stable Rust.  I recommend using `rustup` to manage the Rust version!

Would you mind sharing which version of stable rust you were using? It sounds like we should update the `ci/check-rust-version.sh` to reflect the minimum version requirement from sysinfo.

klamonte:
My stable version currently is:

stable-x86_64-unknown-linux-gnu unchanged - rustc 1.57.0 (f1edd0429 2021-11-29)

But it had been quite a while since I did an upgrade before yesterday, so can't know for sure if the failed compile was at this version.  I immediately switched to nightly and continued on.

wezterm-font/src/fcwrap.rs
```diff
@@ -125,7 +125,7 @@ impl<'a> CharSetRef<'a> {
                 for j in 0..32 {
                     if mask & (1 << j) != 0 {
                         let new_code_point = base_code_point + (j + i * 32) as u32;
-                        if new_code_point - 1 > code_point {
+                        if new_code_point > 0 && new_code_point - 1 > code_point {
```
wez, line 128: ah, this looks like it doesn't belong in this PR!

wez: I'll push this in as a separate commit (bd09b568ed31401a5b525aad40c5d3817e55e486); hopefully this one will rebase out without issue

wez:
Thanks!

PR #1446, opened 2021-12-21
docs: add a keyword to help find an option

eugenesvk:
'tab index' doesn't find this due to the '_' characters, so I've added a phrase without them

will hopefully help avoid isues like this https://github.com/wez/wezterm/issues/1442


wez:
Thanks!

PR #1467, opened 2021-12-28
SGR-Pixels mouse encoding (1016)

klamonte:
This is for #1457 .

* DECRQM/DECRPM work for applications to detect support.
* Pixels are passed as offsets through the mouse event so that x/y retain their text coordinate meaning.  For SGR-Pixels encoding the x+offset / y+offset is converted back into screen pixel coordinates.
* I do not have test coverage, I think?  But all tests pass.

### DIFF
--- a/term/src/input.rs
+++ b/term/src/input.rs
@@ -37,10 +37,18 @@ pub struct MouseEvent {
     pub kind: MouseEventKind,
     pub x: usize,
     pub y: VisibleRowIndex,
-    pub x_offset: usize,
-    pub y_offset: usize,
+    pub x_pixel_offset: usize,
+    pub y_pixel_offset: usize,
     pub button: MouseButton,
     pub modifiers: KeyModifiers,
+}
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub struct ClickPosition {
+    pub column: usize,
+    pub row: i64,
+    pub x_pixel_offset: usize,
+    pub y_pixel_offset: usize,
 }
 
 /// This is a little helper that keeps track of the "click streak",
@@ -52,7 +60,7 @@ pub struct MouseEvent {
 #[derive(Debug, Clone)]
 pub struct LastMouseClick {
     pub button: MouseButton,
-    position: (usize, i64, usize, usize),
+    pub position: ClickPosition,
     time: Instant,
     pub streak: usize,
 }
@@ -61,7 +69,7 @@ pub struct LastMouseClick {
 const CLICK_INTERVAL: u64 = 500;
 
 impl LastMouseClick {
-    pub fn new(button: MouseButton, position: (usize, i64, usize, usize)) -> Self {
+    pub fn new(button: MouseButton, position: ClickPosition) -> Self {
         Self {
             button,
             position,
@@ -70,7 +78,7 @@ impl LastMouseClick {
         }
     }
 
-    pub fn add(&self, button: MouseButton, position: (usize, i64, usize, usize)) -> Self {
+    pub fn add(&self, button: MouseButton, position: ClickPosition) -> Self {
         let now = Instant::now();
         let streak = if button == self.button
             && position == self.position
--- a/term/src/input.rs
+++ b/term/src/input.rs
@@ -37,6 +37,8 @@ pub struct MouseEvent {
     pub kind: MouseEventKind,
     pub x: usize,
     pub y: VisibleRowIndex,
+    pub x_offset: usize,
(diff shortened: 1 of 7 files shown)


klamonte:
@wez

I have been trying to figure out why I picked offsets within a cell rather than just pixels. It fit very naturally in jexer that way, but for other terminals I just did pixel_x/y as you initially suggested.

I finally understood why: in a multihead setup with different cell sizes, offset is better because moving the mouse across the entire screen on one head will move it across all heads. During multihead there might be some small distortion right on the right/bottom edges (jitter) as the mouse travels across the screen and exceeds or skips locations on the other heads, but it's not bad in practice.

I have seen requests for wezterm to support different font sizes in different windows, and thought cell+offset might make it easier if that scenario is supported later on.

But I can also switch really to pixel_x/y if you would prefer.

term/src/input.rs
```diff
@@ -50,7 +52,7 @@ pub struct MouseEvent {
 #[derive(Debug, Clone)]
 pub struct LastMouseClick {
     pub button: MouseButton,
-    position: (usize, i64),
+    position: (usize, i64, usize, usize),
```
wez, line 55: Stylistically speaking, big tuples are hard to grok.

Would you mind turning this tuple into a struct, and having `LastMouseClick::new` and `LastMouseClick::add` use that type for the position info?

Something like this should make it difficult to misinterpret the meaning of the fields in a future pass over the code:

```rust
struct ClickPosition {
   column: usize,
   row: i64,
   x_pixel_offset: usize,
   y_pixel_offset: usize,
}
```

termwiz/src/input.rs
```diff
@@ -1267,6 +1267,21 @@ impl InputParser {
                                         modifiers,
                                     }));
                                 }
+                                // AZL - I don't quite understand this
```
wez, line 1270: Ah, this is for terminal applications who are listening for mouse reports from the terminal.
I think it's important for the application to know that the x/y are pixels for this case.
Perhaps the simplest way to encode that is to introduce a `InputEvent::PixelMouse(MouseEvent)` variant further up in this file:

https://github.com/wez/wezterm/blob/main/termwiz/src/input.rs#L63-L66

and then switch to `PixelMouse` in the callback we issue below?

wez (commented):
Thanks for this!

I did wonder about the offsets in my first pass through, but honestly they feel fine to me!
Structurally this looks good! have a stylistic suggestion, and then we also need to reconcile the termwiz input mouse event -- comments inline!

termwiz/src/input.rs
klamonte: How does this look now?

wezterm-gui/src/termwindow/mouseevent.rs
```diff
@@ -105,8 +116,24 @@ impl super::TermWindow {
                 let button = mouse_press_to_tmb(press);
 
                 let click = match self.last_mouse_click.take() {
-                    None => LastMouseClick::new(button, (x, y)),
-                    Some(click) => click.add(button, (x, y)),
+                    None => LastMouseClick::new(
```
wez, line 119: nit: you could:

```
let click_position = ClickPosition {
   ...
};
```

ahead of this block, and then use pass that down to the `new` and `add` calls to save a little bit of repetition.

wezterm-gui/src/termwindow/mouseevent.rs
```diff
@@ -165,7 +192,7 @@ impl super::TermWindow {
         if let Some(item) = ui_item {
             self.mouse_event_ui_item(item, pane, y, event, context);
         } else {
-            self.mouse_event_terminal(pane, x, y, event, context);
+            self.mouse_event_terminal(pane, x, y, x_pixel_offset, y_pixel_offset, event, context);
```
wez, line 195: how bad is the fan out from passing `ClickPosition` down here in place of the separate coordinates?  Does it add much value?

wez (commented):
I think this is looking good!  I have a minor nit and a question about further code golf, but otherwise this looks fine!

wezterm-gui/src/termwindow/mouseevent.rs
klamonte: No extra value that I can see.  I can switch these to ClickPosition also.

klamonte: mouse_event_terminal isn't referenced anywhere else.  I can also make it not pub. ?

wez:
Thanks!

PR #1473, opened 2021-12-29
Add icon alt: less paddings and roundness, more square, bigger text

bew:
Since I'm not a huge fan of the current icon, I made another one..
- less paddings
- smaller rounded corners => more square
- bigger text (to avoid useless space around the text)

I'm only adding it as a possibility, and to push it somewhere..

<details>
<summary>Compare before/after</summary>

Before:
image
After:
image

</details>


proski:
The contrast is still too low, which is the worst issue with the current icon in my opinion. Please search for "accessible icon contrast", there is a lot of information online. Please consider white on indigo or indigo on white.

This is how the existing icon looks in comparison to other icons. I made the taskbar brighter to alleviate the issue, it was even worse with the default dark gray background.

wezterm-icon

proski:
Here's my version. But I'm not a graphic designer.
A better icon can be made if the author works with a graphic designer.
But the contrast is an accessibility issue that could be hurting wezterm popularity now.
wezterm-icon

proski:
More radical remake. Two letters are pretty much the limit for a good icon, and the letters should be simple (like `Ps` in the Photoshop icon). `W` and `$` are both relatively complex. Also, having letters of different sizes helps distinguish them visually. `$` is a UNIX tradition, but otherwise it's a sign of money. `>` is wider recognized as a prompt, it's used in many terminal icons. It's also a simple arrow-like shape, a good thing for an icon.

I left the original `W` but slanted it slightly. The `>` is to the right to represent "term" in "wezterm". It's not slanted, it's standing still as if waiting for user input. Green is a reference to old green-on-black terminals. The indigo color of the original text is now the background. I made it darker and more purple to match the perceived color of the original.

wezterm-icon2

bersace:
> More radical remake. Two letters are pretty much the limit for a good icon, and the letters should be simple (like `Ps` in the Photoshop icon). `W` and `$` are both relatively complex. Also, having letters of different sizes helps distinguish them visually. `$` is a UNIX tradition, but otherwise it's a sign of money. `>` is wider recognized as a prompt, it's used in many terminal icons. It's also a simple arrow-like shape, a good thing for an icon.
> 
> I left the original `W` but slanted it slightly. The `>` is to the right to represent "term" in "wezterm". It's not slanted, it's standing still as if waiting for user input. Green is a reference to old green-on-black terminals. The indigo color of the original text is now the background. I made it darker and more purple to match the perceived color of the original.
> 
> wezterm-icon2

Actually, W$ is more unix while W> is more windows. I agree that $W looks like a PHP var while W$ reads more like WezTerm.

bew:
I'm going to close this one, it was an tentative but there are more ideas now by better skilled artists
(see https://github.com/wez/wezterm/pull/1473#issuecomment-1382676513,  #2396)

PR #1500, opened 2022-01-03
Fix update instruction for nightly macosx via brew

jankatins:
The update variant errored after rm the cask and I had to install the cask again.

### DIFF
--- a/docs/install/macos.markdown
+++ b/docs/install/macos.markdown
@@ -36,7 +36,7 @@ install:
 
 ```bash
 $ brew rm --cask wez/wezterm/wezterm-nightly
-$ brew upgrade --cask wez/wezterm/wezterm-nightly
+$ brew install --cask wez/wezterm/wezterm-nightly
 ```
 
 ## MacPorts


wez:
Thanks!

jankatins:
Unfortunately, this has the drawback that the icon vanishes from the dock and has to be reattached after each such update :-(

There is also a bit of inconsistency here: the wezterm nightly cask definition seems to suggest that running upgrade with the cask name works: `brew upgrade --cask wezterm-nightly --no-quarantine`

https://github.com/wez/homebrew-wezterm/blob/master/Casks/wezterm-nightly.rb#L46-L49

-> Have yet to try that after the nightly changes again, but I suspect it won't work because the nightly declares its '[latest](https://github.com/wez/homebrew-wezterm/blob/master/Casks/wezterm-nightly.rb#L2)' and such cask seems to get skipped and need `--greedy` or `--greedy-latest`: https://stackoverflow.com/a/31994862

This did worked for me:

```

brew upgrade --cask wezterm-nightly --no-quarantine --greedy-latest
```

just running it did redownload the nightly build and installed it (after not redownloading it yesterday evening after I did the old rm+install way -> probably the file wasn't changed but changed after the commits yesterday evening. Also checked in brew and it seems to compare timestamp ([downloaded file vs url time](https://github.com/Homebrew/brew/blob/371122da557d770aa4555f258a904aa008f9ce8b/Library/Homebrew/download_strategy.rb#L391-L405)) to only download when it's newer on the server.

PR #1503, opened 2022-01-04
Better upgrade instruction for nightly on mac

jankatins:
The old way would remove the icon form the dock. it needs greedy-latest because the cask is declared as `latest`.

See https://github.com/wez/wezterm/pull/1500 for the discussion

### DIFF
--- a/docs/install/macos.markdown
+++ b/docs/install/macos.markdown
@@ -31,12 +31,10 @@ If you'd like to use a nightly build:
 $ brew install --cask wez/wezterm/wezterm-nightly
 ```
 
-to upgrade to a newer nightly, it is simplest to remove then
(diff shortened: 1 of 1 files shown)


wez (commented):
We shouldn't need the no- quarantine flag because we now code sign the binary: can you confirm/update accordingly?

jankatins:
without the flag I get: `“WezTerm.app” is an app downloaded from the Internet. Are you sure you want to open it?`

(tested by `rm $HOME/Library/Caches/Homebrew/downloads/*WezTerm-macos-nightly.zip` and with/without the flag and restarting wezterm)

wez:
Thanks!

PR #1506, opened 2022-01-05
No longer assume target directory is default

CGMossa:
The build-script of `wezterm-gui` on windows assumed that the target directory was `wezterm/target`. However, if one specifies a global target-directory, then this is no longer the case.

### DIFF
--- a/wezterm-gui/build.rs
+++ b/wezterm-gui/build.rs
@@ -6,12 +6,17 @@ fn main() {
         use anyhow::Context as _;
         use std::io::Write;
         use std::path::Path;
-        let profile = std::env::var("PROFILE").unwrap();
         let repo_dir = std::env::current_dir()
             .ok()
             .and_then(|cwd| cwd.parent().map(|p| p.to_path_buf()))
             .unwrap();
-        let exe_output_dir = repo_dir.join("target").join(profile);
+        let exe_output_dir = std::env::var("OUT_DIR").unwrap();
+        let exe_output_dir = std::path::PathBuf::from(exe_output_dir);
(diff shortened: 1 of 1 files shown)


wez:
Thanks for this!  However, I don't feel good about this change; OUT_DIR looks something like this in the default build configuration:

```
REPODIR/target/debug/build/wezterm-gui-ca044d006020e4d8/out
```

https://doc.rust-lang.org/cargo/reference/environment-variables.html says:

> `OUT_DIR` — the folder in which all output should be placed. This folder is inside the build directory for the package being built, and it is unique for the package in question.

but it doesn't specify anything about the relationship or the structure of the build, which makes me worry that future changes in cargo will break the build.  For example, in this PR we have to walk up three levels to get to where we think we need to be.  If cargo changes something about that layout in the future, it may no longer be the right place.

By contrast, by building the path down from `REPODIR/target/PROFILE` we know that that is where the final outputs will be placed, which feels like a safer bet.

It doesn't look like there is a better env var to pick from :-/

CGMossa:
There is probably a better way to do this. But I did research a bit and
this is what I could come up with.

Feel free to close this. I'll revive it if I find an alternative way to do
it.

wez:
Thanks for taking the time to dive into this, I appreciate it!

nico-abram:
I haven't tested it but when using an explicit target dir I believe cargo uses /target/$triplet/$profile instead, so I think the current build script probably breaks when using an explicit target. Just try setting --target your-default-target

This is typically used when cross compiling so maybe it's not a big deal. But it does sound like more evidence that the current way of doing things is brittle.

PR #1507, opened 2022-01-05
Suggestions from Clippy

CGMossa:
While going through building wezterm, I found out there are a few complaints from clippy. These are the suggestions.

### DIFF
--- a/termwiz/src/terminal/windows.rs
+++ b/termwiz/src/terminal/windows.rs
@@ -10,7 +10,6 @@ use std::os::windows::io::{AsRawHandle, FromRawHandle};
 use std::sync::Arc;
 use std::time::Duration;
 use std::{mem, ptr};
-use winapi::shared::winerror::WAIT_TIMEOUT;
 use winapi::um::consoleapi;
 use winapi::um::synchapi::{CreateEventW, SetEvent, WaitForMultipleObjects};
 use winapi::um::winbase::{INFINITE, WAIT_FAILED, WAIT_OBJECT_0};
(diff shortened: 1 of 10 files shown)


termwiz/src/terminal/windows.rs
```diff
@@ -802,8 +801,8 @@ impl Terminal for WindowsTerminal {
                         "failed to WaitForMultipleObjects: {}",
                         IoError::last_os_error()
                     );
-                } else if result == WAIT_TIMEOUT {
-                    return Ok(None);
+                // } else if result == WAIT_TIMEOUT {
```
wez, line 804: I don't agree with most of clippy's suggestions.  They're biased to ultimate minimalism, but they cause some information to be lost.  For example, in this case, it is very important for `WAIT_TIMEOUT` to be handled.
In the future, if some work is done here around other cases, the proposed change to implicitly handle it in the common case will be lost and there is risk of regression.

CGMossa:
No problem.

You can close this or decline it as well.

The deep dive was worth it for me atleast.

wezterm-font/src/locator/mod.rs
```diff
@@ -44,6 +44,12 @@ pub enum FontDataSource {
     },
 }
 
+impl std::hash::Hash for FontDataSource {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        core::mem::discriminant(self).hash(state);
```
wez, line 49: This feels like a net negative change: why remove the derived Hash in favor of whatever this is doing?

CGMossa: The the partial eq and hash has to match. Making both explicit should atleast make sure that we ignore fields equally, etc.
This should be equivalent to the derived result.

wez (changes requested):
Thanks for taking the time to put this together.

However, I don't want most of these suggestions; they're generally not good for reasons that clippy doesn't understand.

The `write` -> `write_all` suggestions seem righteous, and I'll just fix those now to save you from coming back to this!

PR #1514, opened 2022-01-06
Docs: Note on toast notification duration for Windows

plscks:
Notes that toast notification timeout duration is not honored on Windows. I spent a little time trying to figure out why my toast notifications were hanging out for 25 seconds and it turns out that on windows the only options for timeout are `duration="long"` and the implicit `duration="short"` in the toast tag at [wezterm-toast-notification/src/windows.rs](https://github.com/wez/wezterm/blob/ca0e35826280e21793cba84fc782071ed1728a90/wezterm-toast-notification/src/windows.rs#L40). So I figured a note would be good for anyone else messing with toast notifications on windows.

### DIFF
--- a/docs/config/lua/window/toast_notification.md
+++ b/docs/config/lua/window/toast_notification.md
@@ -11,7 +11,9 @@ An optional *timeout* parameter can be provided; if so, it specifies how long
 the notification will remain prominently displayed in milliseconds.  To specify
 a timeout without specifying a url, set the url parameter to `nil`.  The timeout
 you specify may not be respected by the system, particularly in X11/Wayland
-environments.
+environments. *Note: On Windows the specified timeout will not be honored. The
+default is a duration of 25s, unless wezterm is built without the duration 
+attribute in the toast tag in which case it will default to 7s.*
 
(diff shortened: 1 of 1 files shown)


docs/config/lua/window/toast_notification.md
```diff
@@ -11,7 +11,9 @@ An optional *timeout* parameter can be provided; if so, it specifies how long
 the notification will remain prominently displayed in milliseconds.  To specify
 a timeout without specifying a url, set the url parameter to `nil`.  The timeout
 you specify may not be respected by the system, particularly in X11/Wayland
```
wez, line 13: The docs already say that the timeout may not be respected up here.
Rather than making these document current behaviors of specific versions
of Windows, and referencing something that users cannot change in the code,
I'd rather that this just be a generic statement that you cannot rely on the timeout
being supported by any system.

plscks: Right this is reasonable, I just thought it was worth mentioning somewhere for those that might come along and face a similar issue of desiring a shorter timeout. Killer termina emulator non-the-less!

PR #1518, opened 2022-01-07
Make the threshold for predictive echoing configurable

qperret:
Some users (including myself) have a higher tolerance for latency than mispredictions for the predictive echo feature -- I always used `--predict=never` in mosh for instance. Since the threshold for activating that feature is currently hardcoded, there is no easy way (that I know of) to disable it, or to set it to a alternative value. To allow this, let's expose it in a new config knob that defaults to 100ms hence retaining the current behavior.

### DIFF
--- a/config/src/unix.rs
+++ b/config/src/unix.rs
@@ -48,6 +48,10 @@ pub struct UnixDomain {
 
     #[serde(default = "default_write_timeout")]
     pub write_timeout: Duration,
+
+    /// Don't use default_local_echo_threshold_ms() here to
+    /// disable the predictive echo for Unix domains by default.
+    pub local_echo_threshold_ms: Option<u64>,
 }
 impl_lua_conversion!(UnixDomain);
 
@@ -62,6 +66,7 @@ impl Default for UnixDomain {
             skip_permissions_check: false,
             read_timeout: default_read_timeout(),
             write_timeout: default_write_timeout(),
+            local_echo_threshold_ms: None,
             proxy_command: None,
         }
     }
--- a/docs/config/lua/SshDomain.md
+++ b/docs/config/lua/SshDomain.md
@@ -108,3 +108,21 @@ return {
 }
 ```
 
+You may now specify the round-trip latency threshold for enabling predictive
+local echo using `local_echo_threshold_ms`. If the measured round-trip latency
+between the wezterm client and the server exceeds the specified threshold, the
+client will attempt to predict the server's response to key events and echo the
+result of that prediction locally without waiting, hence hiding latency to the
+user. This option only applies when `multiplexing = "WezTerm"`.
+
+```lua
+return {
+  ssh_domains = {
+    {
+      name = "my.server",
+      remote_address = "192.168.1.1",
+      local_echo_threshold_ms = 10,
+    }
+  },
+}
+```
--- a/wezterm-client/src/domain.rs
+++ b/wezterm-client/src/domain.rs
@@ -22,6 +22,7 @@ pub struct ClientInner {
     pub client: Client,
     pub local_domain_id: DomainId,
     pub remote_domain_id: DomainId,
+    pub local_echo_threshold_ms: Option<u64>,
     remote_to_local_window: Mutex<HashMap<WindowId, WindowId>>,
     remote_to_local_tab: Mutex<HashMap<TabId, TabId>>,
     remote_to_local_pane: Mutex<HashMap<PaneId, PaneId>>,
@@ -128,6 +129,14 @@ impl ClientDomainConfig {
         }
     }
 
+    pub fn local_echo_threshold_ms(&self) -> Option<u64> {
+        match self {
+            ClientDomainConfig::Unix(unix) => unix.local_echo_threshold_ms,
+            ClientDomainConfig::Tls(tls) => tls.local_echo_threshold_ms,
+            ClientDomainConfig::Ssh(ssh) => ssh.local_echo_threshold_ms,
+        }
+    }
+
     pub fn label(&self) -> String {
         match self {
             ClientDomainConfig::Unix(unix) => format!("unix mux {}", unix.socket_path().display()),
@@ -152,7 +161,11 @@ impl ClientDomainConfig {
 }
 
 impl ClientInner {
-    pub fn new(local_domain_id: DomainId, client: Client) -> Self {
+    pub fn new(
+        local_domain_id: DomainId,
+        client: Client,
+        local_echo_threshold_ms: Option<u64>,
+    ) -> Self {
         // Assumption: that the domain id on the other end is
         // always the first created default domain.  In the future
         // we'll add a way to discover/enumerate domains to populate
@@ -162,6 +175,7 @@ impl ClientInner {
             client,
             local_domain_id,
             remote_domain_id,
+            local_echo_threshold_ms,
             remote_to_local_window: Mutex::new(HashMap::new()),
             remote_to_local_tab: Mutex::new(HashMap::new()),
             remote_to_local_pane: Mutex::new(HashMap::new()),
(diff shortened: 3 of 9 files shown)


qperret:
Hi, I've switched to wezterm recently (it's awesome, thanks so much for all the work :)), and this is one the only small things that I'd love to be able to configure. This is clearly my first contribution to the project though, so I'll be happy to do this differently if needed. Thanks!

config/src/lib.rs
```diff
@@ -1257,9 +1257,16 @@ pub struct Config {
 
     #[serde(default = "default_true")]
     pub allow_download_protocols: bool,
+
+    #[serde(default = "default_local_echo_threshold_ms")]
+    pub local_echo_threshold_ms: u64,
```
wez, line 1262: What do you think about scoping this per domain?
The code change is a little more involved:

Move this config value so that it lives over here alongside this one:
https://github.com/wez/wezterm/blob/4efcc81bee162579888d418f152b55550b7b8503/config/src/tls.rs#L81

Then add something like this:

```rust
   pub local_echo_threshold_ms: Option<u64>,
```

to `ClientInner`:
https://github.com/wez/wezterm/blob/4efcc81bee162579888d418f152b55550b7b8503/wezterm-client/src/domain.rs#L21

That will need to get initialized, so add that same optional value as a parameter to:
https://github.com/wez/wezterm/blob/4efcc81bee162579888d418f152b55550b7b8503/wezterm-client/src/domain.rs#L155

to do that, I think you'll want to add an accessor method for this value over here, similar to this name method:
https://github.com/wez/wezterm/blob/4efcc81bee162579888d418f152b55550b7b8503/wezterm-client/src/domain.rs#L122-L123

so that you can pass it down through here:
https://github.com/wez/wezterm/blob/4efcc81bee162579888d418f152b55550b7b8503/wezterm-client/src/domain.rs#L510

then with that in place, the `should_predict` method could be phrased as something like this:

```rust
fn should_predict(&self) -> bool {
  self.client.local_echo_threshold_ms.map(|thresh| self.last_input_rtt >= thresh).unwrap_or(false)
}
```

wez (commented):
Thanks for diving in!
I have a few comments!

docs/config/lua/config/local_echo_threshold_ms.md
```diff
@@ -0,0 +1,10 @@
+# `local_echo_threshold_ms`
+
+Specifies the maximum tolerated round-trip latency before using predictive
+echoing on remote terminals. The value is measured in milliseconds.
```
bew, line 4: Would it be the place to explain what `predictive echoing` means? Or put a link that explains it?
(I have no idea what it is)

Question for myself: if it's about predictions of remote terminal, why is the option named `local_echo_...`  ?

wez: Predictive local echo is a feature of remote multiplexer connections that is intended to hide latency.
The idea is that if we *predict* that a given keypress (say of the `a` key) might result in an `a` being printed at the cursor location, then we could locally show that before the remote host responds and updates the screen.  This can make a slow link feel fast if you are quickly typing prose, as the text appears to quickly update, but can also be jarring if the threshold is very close to the actual latency of the link--the cursor can appear to jiggle back and forth.

docs/config/lua/config/local_echo_threshold_ms.md
```diff
@@ -0,0 +1,10 @@
+# `local_echo_threshold_ms`
+
```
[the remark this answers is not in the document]
qperret: Sure thing.

config/src/lib.rs
qperret: Aha, yes, that sounds like a good idea. And that would perhaps even allow us to drop the `is_local()` machinery in favour of a different default for unix domains?
I'll have a go at it, probably next week though.

docs/config/lua/config/local_echo_threshold_ms.md
qperret: I'm not sure what the best place would be, but for sure some more doc sounds like a good idea. Unless @wez already has something I can refer to I can try and write something down here.

To answer your question in the meantime, the predictive echoing feature is basically a nice trick to try and hide latency when using e.g. an ssh domain where the round-trip latency can be high. Other tools (mosh for instance) have a similar feature. The idea is for the client to make a prediction about the server's response to key events and immediately echo that locally. And when the server's response arrives later on, if the client's speculation was correct it's all good, otherwise the client might need to back off what it echoed locally, which can lead to visible 'noise'.

qperret:
Apologies for the delay, I have now updated the branch. Still finding my way around, but hopefully this looks a little better now.

wezterm-client/src/domain.rs
```diff
@@ -128,6 +129,14 @@ impl ClientDomainConfig {
         }
     }
 
+    pub fn local_echo_threshold_ms(&self) -> Option<u64> {
+        match self {
+            ClientDomainConfig::Unix(unix) => unix.local_echo_threshold_ms,
```
wez, line 134: we don't use predictions for local connections (the `!self.client.is_local()` logic in `should_predict` prior to this PR), so we should set it to `None` and remove the option from the unix domain as well, to preserve those semantics:

```suggestion
            ClientDomainConfig::Unix(unix) => None,
```

docs/config/lua/SshDomain.md
```diff
@@ -108,3 +108,21 @@ return {
 }
 ```
 
+You main now specify the round-trip latency threshold for enabling predictive
```
wez, line 111:
```suggestion
You maynow specify the round-trip latency threshold for enabling predictive
```

docs/config/lua/SshDomain.md
```diff
@@ -108,3 +108,21 @@ return {
...
+local echo using `local_echo_threshold_ms`. If the measured round-trip latency
+between the wezterm client and the server exceeds the specified threshold, the
+client will attempt to predict the server's response to key events and echo the
+result of that prediction locally without waiting, hence hiding latency to the
+user.
```
wez, line 116:
```suggestion
user.  This option only applies when `multiplexing = "WezTerm"`.
```

docs/config/lua/SshDomain.md
wez, line 111: I think a similar thing should be added to the TlsDomain docs as well!

config/src/unix.rs
```diff
@@ -48,6 +48,10 @@ pub struct UnixDomain {
 
     #[serde(default = "default_write_timeout")]
     pub write_timeout: Duration,
+
+    /// Don't use default_local_echo_threshold_ms() here to
+    /// disable the predictive echo for Unix domains by default.
+    pub local_echo_threshold_ms: Option<u64>,
```
wez, line 54: as mentioned below, this option doesn't apply to unix domains, so we shouldn't add it here

wez (changes requested):
Thanks!  This is just about there!

docs/config/lua/SshDomain.md
bew: Or even more correct 😅
```suggestion
You may now specify the round-trip latency threshold for enabling predictive
```

wezterm-client/src/domain.rs
qperret: Yup, but I've set the default to `None` for unix connections -- see `config/src/unix.rs` so the semantics should be the same by default. For the record, I found that allowing users to specify it for unix domains as well if they really want to might actually be a good idea. Following the test methodology explained in https://lwn.net/Articles/751763/ (that is, use the `typometer` benchmark from https://pavelfatin.com/typing-with-pleasure/), I ran a few small tests on my wezterm setup to measure typing latency. I'm using a Lenovo Carbon X1 with Debian / X / Gnome, and got the following results:

- Unix domain (very similar to a normal/native instance) w/o predictive echo: min=43.0ms, max=65.9ms, avg=57.1ms, sd=6.9ms

- Unix domain with predictive echo (threshold set to 0): min=25.5ms, max=49.1, avg=40.0ms, sd=7.5ms

 And that seems to be very reproducible. So, I figured offering the option might be nice given that it's not that complicated, but no strong opinion if you think it's not worth it.

wez: You make a good case for keeping that in, so let's do that :)

qperret:
Update pushed:
 - documentation updated for SSH, TLS and Unix domains
 - minor code refactoring to make the CI/linter happy.
 
Thanks!

qperret:
Updated again, hopefully that'll make the linter happy this time. (For some reason I'm having troubles running it locally, I'll try and get my setup fixed, but please bear with me in the meantime :smiley: )

wez:
Thank you!

PR #1523, opened 2022-01-09
get-deps: refactor script and add alpine

panekj:
Script does not require bash to run (which is not available everywhere)
Add handling OS by inspecting `/etc/os-release` + fallback to old method
Add AlpineLinux

Signed-off-by: Jakub Panek <me@panekj.dev>

### DIFF
--- a/ci/check-rust-version.sh
+++ b/ci/check-rust-version.sh
@@ -1,49 +1,22 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
-min_rust=(1 57 0)
-rust_ver=()
-
-parse_rustc_version() {
-  local IFS
-  IFS=' '
-  local FIELDS
-  read -ra FIELDS <<< $(rustc --version)
-  IFS='.'
-  read -ra rust_ver <<< "${FIELDS[1]}"
-}
+min_rust="1.57.0"
+rust_ver="$(rustc --version | cut -d' ' -f2)"
 
 check_rust_version() {
-  parse_rustc_version
-  # rust_ver=(1 46 0) for testing purposes
-
-  if [[ "${rust_ver[0]}" -gt "${min_rust[0]}" ]] ; then
+  ver=$(printf "%s\n%s\n" "$min_rust" "$rust_ver" | sort --version-sort | head -n1)
+  if test "$ver" = "$min_rust"; then
     return 0
-  fi
-  if [[ "${rust_ver[0]}" -lt "${min_rust[0]}" ]] ; then
+  else
     return 1
   fi
-  if [[ "${rust_ver[1]}" -gt "${min_rust[1]}" ]] ; then
-    return 0
-  fi
-  if [[ "${rust_ver[1]}" -lt "${min_rust[1]}" ]] ; then
-    return 1
-  fi
-  if [[ "${rust_ver[2]}" -gt "${min_rust[2]}" ]] ; then
-    return 0
-  fi
-  if [[ "${rust_ver[2]}" -lt "${min_rust[2]}" ]] ; then
-    return 1
-  fi
-
-  return 0
 }
 
 if ! check_rust_version ; then
-  rust_ver=$(IFS=. ; echo "${rust_ver[*]}")
-  min_rust=$(IFS=. ; echo "${min_rust[*]}")
-  echo "Installed rustc version $rust_ver is less than required $min_rust"
+  echo "Installed rustc version '$rust_ver' is less than required '$min_rust'"
   echo
-  echo "Using rustup to manage your installed version of Rust is recommended"
+  echo "Check if your OS provides newer version of Rust, if not"
+  echo "use rustup to manage installed versions of Rust"
   echo "https://www.rust-lang.org/en-US/install.html"
   echo
   echo "See https://wezfurlong.org/wezterm/install/source.html for complete"
--- a/ci/generate-workflows.py
+++ b/ci/generate-workflows.py
@@ -27,7 +27,7 @@ def yv(v, depth=0):
 
 
 class Step(object):
-    def render(self, f, env, depth=0):
+    def render(self, f, depth=0):
         raise NotImplementedError(repr(self))
 
 
@@ -38,7 +38,7 @@ class RunStep(Step):
         self.shell = shell
         self.env = env
 
-    def render(self, f, env, depth=0):
+    def render(self, f, depth=0):
         indent = "  " * depth
         f.write(f"{indent}- name: {yv(self.name)}\n")
         if self.env:
@@ -49,11 +49,6 @@ class RunStep(Step):
             f.write(f"{indent}  shell: {self.shell}\n")
 
         run = self.run
-
-        if env:
-            for k, v in env.items():
-                if self.shell == "bash":
-                    run = f"export {k}={v}\n{run}\n"
 
         f.write(f"{indent}  run: {yv(run, depth + 2)}\n")
 
@@ -66,7 +61,7 @@ class ActionStep(Step):
         self.env = env
         self.condition = condition
 
-    def render(self, f, env, depth=0):
+    def render(self, f, depth=0):
         indent = "  " * depth
         f.write(f"{indent}- name: {yv(self.name)}\n")
         f.write(f"{indent}  uses: {self.action}\n")
@@ -105,8 +100,9 @@ class Job(object):
         self.env = env
 
     def render(self, f, depth=0):
+        f.write("\n    steps:\n")
         for s in self.steps:
-            s.render(f, self.env, depth)
+            s.render(f, depth)
 
 
 class Target(object):
@@ -119,6 +115,7 @@ class Target(object):
         rust_target=None,
         continuous_only=False,
         app_image=False,
+        is_tag=False,
     ):
         if not name:
             if container:
@@ -132,6 +129,16 @@ class Target(object):
         self.rust_target = rust_target
         self.continuous_only = continuous_only
         self.app_image = app_image
+        self.env = {}
+        self.is_tag = is_tag
+
+    def render_env(self, f, depth=0):
+        self.global_env()
+        if self.env:
+            indent = "    "
+            f.write(f"{indent}env:\n")
+            for k, v in self.env.items():
+                f.write(f"{indent}  {k}: {yv(v, depth + 3)}\n")
 
     def uses_yum(self):
         if "fedora" in self.name:
@@ -147,6 +154,11 @@ class Target(object):
             return True
         return False
 
+    def uses_apk(self):
+        if "alpine" in self.name:
+            return True
+        return False
+
     def needs_sudo(self):
         if not self.container and self.uses_apt():
             return True
@@ -158,14 +170,19 @@ class Target(object):
             installer = "yum"
         elif self.uses_apt():
             installer = "apt-get"
+        elif self.uses_apk():
+            installer = "apk"
         else:
             return []
         if self.needs_sudo():
             installer = f"sudo -n {installer}"
-        return [RunStep(f"Install {name}", f"{installer} install -y {name}")]
+        if self.uses_apk():
+            return [RunStep(f"Install {name}", f"{installer} add {name}")]
+        else:
+            return [RunStep(f"Install {name}", f"{installer} install -y {name}")]
 
     def install_curl(self):
-        if self.uses_yum() or (self.uses_apt() and self.container):
+        if self.uses_yum() or self.uses_apk() or (self.uses_apt() and self.container):
             if "centos:stream9" in self.container:
                 return self.install_system_package("curl-minimal")
             else:
@@ -173,11 +190,14 @@ class Target(object):
         return []
 
     def install_openssh_server(self):
+        steps = []
         if self.uses_yum() or (self.uses_apt() and self.container):
-            return [
+            steps += [
                 RunStep("Ensure /run/sshd exists", "mkdir -p /run/sshd")
             ] + self.install_system_package("openssh-server")
-        return []
+        if self.uses_apk():
+            steps += self.install_system_package("openssh")
+        return steps
 
     def install_newer_compiler(self):
         steps = []
@@ -380,6 +400,10 @@ cargo build --all --release""",
             patterns += ["WezTerm-*.zip"]
         elif ("ubuntu" in self.name) or ("debian" in self.name):
             patterns += ["wezterm-*.deb", "wezterm-*.xz"]
+        elif "alpine" in self.name:
+            patterns += ["~/packages/x86_64/wezterm-*.apk"]
+            if self.is_tag:
+                patterns.append("~/.abuild/*.pub")
 
         if self.app_image:
             patterns.append("*src.tar.gz")
@@ -511,10 +535,11 @@ cargo build --all --release""",
         return steps
 
     def global_env(self):
-        env = {}
         if "macos" in self.name:
-            env["MACOSX_DEPLOYMENT_TARGET"] = "10.9"
-        return env
+            self.env["MACOSX_DEPLOYMENT_TARGET"] = "10.9"
+        if "alpine" in self.name:
+            self.env["RUSTFLAGS"] = "-C target-feature=-crt-static"
+        return
 
     def prep_environment(self, cache=True):
         steps = []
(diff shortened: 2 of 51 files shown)


get-deps
```diff
@@ -1,100 +1,132 @@
-#!/bin/bash
+#!/bin/sh
 
-if ! bash ./ci/check-rust-version.sh ; then
+test_cmd() {
+  command -v $1 >/dev/null
+}
+
+if test_cmd 'bash' && test_cmd 'rustc' && ! bash ./ci/check-rust-version.sh ; then
```
wez, line 7: It's important that they have rustc and that they have an appropriate version of rustc.
This change effectively disables these preconditions on systems that don't have bash.
Can you expand this PR to remove the `bash` dep from `ci/check-rust-version.sh` as well?

get-deps
```diff
@@ -114,46 +146,105 @@ if test -e /etc/arch-release ; then
...
+    'zip'
+  exit $?
+}
+
+if test -e /etc/os-release; then
+  . /etc/os-release
+
+  case $ID in
+    centos|fedora|rhel)
+      fedora_deps
```
wez, line 188: I'd prefer to have the decision to `exit` be moved out of the various `_deps` functions and placed here, so that the control flow is more clear.

```suggestion
      fedora_deps
      exit $?
```

if for some reason that isn't workable, then I would advocate for renaming those functions to be something like `install_fedora_deps_and_exit` to make it clearer that they don't return.

wez (commented):
Thanks for diving in!
I have a few comments! :)

get-deps
panekj: Sure 👍

get-deps
panekj: I've skipped that since it uses Bash arrays. I have an idea how to rewrite it in posix shell, just need to test it.

panekj:
Rust version check works now, it can be tested quickly this way (Alpine v3.11 has Rust v1.39.0):
```sh
docker run -ti --rm -v $PWD:/build --workdir /build alpine:3.11 ./get-deps
```

I moved stuff around even more, so it checks Rust version after installing deps (since `cargo`/`rust` can be installed through system package manager)

ci/check-rust-version.sh
```diff
@@ -1,49 +1,22 @@
...
+min_rust="1.53.0"
+rust_ver="$(rustc --version | cut -d' ' -f2)"
 
 check_rust_version() {
-  parse_rustc_version
-  # rust_ver=(1 46 0) for testing purposes
-
-  if [[ "${rust_ver[0]}" -gt "${min_rust[0]}" ]] ; then
+  ver=$(printf "%s\n%s\n" "$min_rust" "$rust_ver" | sort -V | head -n1)
+  if test "$ver" = "1.53.0"; then
```
wez, line 8: Should this be?

```suggestion
  if test "$ver" = "$min_rust"; then
```

ci/check-rust-version.sh
```diff
@@ -1,49 +1,22 @@
...
-}
+min_rust="1.53.0"
+rust_ver="$(rustc --version | cut -d' ' -f2)"
 
 check_rust_version() {
-  parse_rustc_version
-  # rust_ver=(1 46 0) for testing purposes
-
-  if [[ "${rust_ver[0]}" -gt "${min_rust[0]}" ]] ; then
+  ver=$(printf "%s\n%s\n" "$min_rust" "$rust_ver" | sort -V | head -n1)
```
wez, line 7: ah, pretty neat!
I generally prefer the longer names for relatively unusual parameters; assuming that the long name is portable enough, can we use this?

```suggestion
  ver=$(printf "%s\n%s\n" "$min_rust" "$rust_ver" | sort --version-sort | head -n1)
```

panekj: Nope, longer names are a GNU thing, and most posix environments will have only `-V` (e.g. busybox)

ci/check-rust-version.sh
panekj: yup, completely missed it 😄

get-deps
```diff
@@ -113,47 +139,116 @@ if test -e /etc/arch-release ; then
...
+    echo "Couldn't find OS by ID, found ID: $ID"
+    echo "Fallback to detecting '/etc/<name>-release'"
+    fallback_method
+    if ! test $NOTFOUND -eq 0; then
+      echo "Couldn't identify OS through '/etc/<name>-release'"
+    fi
+  ;;
+esac
+
 case $OSTYPE in
```
SuperSandro2000, line 224: SC3028: not defined in POSIX sh

get-deps
```diff
@@ -113,47 +139,116 @@ if test -e /etc/arch-release ; then
...
-    echo "For `lsb_release -ds`"
-    exit 1
+    NOTFOUND=1
   ;;
 esac
+
+if ! test $NOTFOUND -eq 0; then
+  echo "Please contribute the commands to install the deps for:"
+  if test_cmd 'lsb_release'; then
+    echo "`lsb_release -ds`"
```
SuperSandro2000, line 243:
```suggestion
    lsb_release -ds
```

ci/check-rust-version.sh
```diff
@@ -1,49 +1,22 @@
-#!/bin/bash
+#!/bin/sh
```
SuperSandro2000, line 1: I would be in favor of using ``#!/usr/bin/env bash`` because posix sh/dash is a pain to work with and has many stupid quirks.

panekj: `bash` is not available everywhere

get-deps
panekj: Can be ignored, I added check for empty case

ci/check-rust-version.sh
SuperSandro2000: ``apk add bash`` is a very simple fix for that and avoids all the quirks the many different implementations of /bin/sh have.

panekj: `apk add bash` is not solution to every system without `bash`, and installing `bash` is also not a proper solution

SuperSandro2000: I don't know any system where bash is not available but rust works and I don't think there exists one but I know that ``/bin/sh`` is a different shell with different implantations on almost every platform I used so far and is a pain to support even with shellcheck:
- Debian has dash
- alpine has either busybox sh which has some more features than posix sh or also ash
- Arch has bash
- Darwin has something different, too
- NixOS has bash in the build sandbox and dash in the real system

For example if I would change the script on Arch I wouldn't even notice that I am using bash features while I break the script on Alpine or Debian. It is just so much more easier to just use bash which has the same feature set on all platforms except on darwin where the default bash is just to old.

---

Also making scripts POSIX compatible from bash is usually not worth the time because POSIX is just missing so many convenience features and features in general that it is often more a pain because you can't use so most things. best example is new test ``[[`` which is so much more usable compared to ``[`` and eliminates many of the quoting pitfalls.

ci/check-rust-version.sh
SuperSandro2000: Probably most if not all development machines will have gnu coreutils. On Alpine you can just install it via ``apk add coreutils`` which replaces the limited busybox sort applet. It will be so much easier for everyone if we would use that and wouldn't bring other people in the pity to break Alpine compatibility by accident.
I think it would be a really good choice to do the following for alpine in beginning of ``get-deps``:
```
sudo apk add coreutils bash
```

ci/check-rust-version.sh
panekj: > I know that `/bin/sh` is a different shell with different implantations on almost every platform

It only matters if you write non-POSIX code. `bash` differences are even more painful than `/bin/sh`, because of breaking changes or bugs, and there is much more wrong with `bash` than that.

> Also making scripts POSIX compatible from bash is usually not worth the time because POSIX is just missing so many convenience features and features in general that it is often more a pain because you can't use so most things.

`bashisms` are terrible (as well as other GNU shenanigans), I just made script POSIX compliant and it isn't that hard, it took me about an hour or two, none of `bash` features are needed or useful to install deps or check rust version. The script is readable and not hard to understand

> For example if I would change the script on Arch I wouldn't even notice that I am using bash features while I break the script on Alpine or Debian. 

If I change something while still being POSIX compliant, nothing breaks nowhere :>

In the end, this is not an elaborate script doing "shell magic", skipping on `bash` is perfectly fine (and not only here imo but this is not a place to discuss that).
POSIX `sh` will still be the lowest common denominator.

ci/check-rust-version.sh
panekj: `sudo` is deprecated and doesn't have to be installed everywhere (see line 9-15), `apk` is alpine specific, you can have universal script across systems at very minimal cost

> Probably most if not all development machines will have gnu coreutils.

The script doesn't have to be used on development machine. If someone just wants to use/install `wezterm`, they don't need `bash` nor `coreutils`.

panekj:
@wez would you like me to add workflow for building/testing `wezterm` on alpine/musl?

wez:
> @wez would you like me to add workflow for building/testing `wezterm` on alpine/musl?

Yes please!  The workflows are generated from `ci/generate-workflows.py` as I find it much easier to program in python than in YAML :-)

I would suggest starting by adding a line like this, but using `alpine` or whatever is appropriate as the base docker image:

https://github.com/wez/wezterm/blob/f48e2d9a858c54374b19c7a1e61bef7ae90aad65/ci/generate-workflows.py#L647

then adjusting the other methods in that file to get the environment ready.

Run `ci/generate-workflows.py` to update the workflows, then `git add` the new alpine workflows and include them in this PR.
Please use `black` for autoformatting of `ci/generate-workflows.py`.

ci/check-rust-version.sh
panekj: While I'm still in the process of fixing `wezterm` builds on Alpine (ref: https://gitlab.alpinelinux.org/panekj/aports/-/pipelines/105982), reasons you can't use provided binaries are:
- AppImage doesn't work with musl
- No pre-built binaries are provided for musl
- `wezterm` can't be built statically on musl due to proc-macros, thus it's not portable across musl systems

panekj:
Eh, I'm fighting with `cargo test --all --release`. For some reason it doesn't link `fontconfig` and `freetype` but only during tests.

panekj:
`rust` installed through `rustup` in workflows fails to link them properly, `rust` provided through Alpine repos works fine

panekj:
Forgot to recreate workflows after updating script earlier

panekj:
Few things:
- `nodejs`, `zstd` and `wget` are added so GitHub doesn't break (since Alpine is `musl`, pre-installed `nodejs` on GitHub Actions won't work)
- Currently `openssh-server` is installed as CI dependency, but it's required for `wezterm-ssh` tests, shouldn't it be in `get-deps`?
- `-C target-feature=-crt-static` is required for `musl` systems (on Alpine it's patched into Rust), I added it like it's done for `"BUILD_REASON"` but alternatively it could be done as `echo "RUSTFLAGS='-C target-feature=-crt-static'" | tee -a $GITHUB_ENV`
- I added `bash` and `coreutils` since it's required for rest of CI system, should I revert `ci/check-rust-version.sh`?

wez:
> * Currently `openssh-server` is installed as CI dependency, but it's required for `wezterm-ssh` tests, shouldn't it be in `get-deps`?

Hmm, that dep is only needed for tests.  I think it's OK for it be captured in the CI workflows, but agree that it can feel a bit surprising.  Maybe add a `--testing` (or better named) option to `get-deps` that installs packages required to run tests?

> * `-C target-feature=-crt-static` is required for `musl` systems (on Alpine it's patched into Rust), I added it like it's done for `"BUILD_REASON"` but alternatively it could be done as `echo "RUSTFLAGS='-C target-feature=-crt-static'" | tee -a $GITHUB_ENV` in `prep_environment()`

I think when I did that original env exporting stuff, there wasn't a GITHUB_ENV thing.  I'd be happy to see things simplify and set up the environment at the start!

> * I added `bash` and `coreutils` since it's required for rest of CI system, should I revert `ci/check-rust-version.sh`?

I like how your updated version is more succinct: I really don't like clever bash scripts.  Unless you're worried about portability of the new version, I think your updated version is probably fine.  Maybe just change the shebang back to bash? :-)

panekj:
> I think when I did that original env exporting stuff, there wasn't a GITHUB_ENV thing. I'd be happy to see things simplify and set up the environment at the start!

That sounds right, `$GITHUB_ENV` and `$GITHUB_PATH` are fairly recent and were a security response to previously available `echo "::set-path"`
Although I think it would be better to set those values in workflow or job level `env:`, and the `RunStep` could have own `env:` key but that would be work for another PR :)

> Unless you're worried about portability of the new version [...] Maybe just change the shebang back to bash? :-)

Done, it should do fine in bash

ci/generate-workflows.py
```diff
@@ -393,6 +393,8 @@ def asset_patterns(self):
             patterns += ["WezTerm-*.zip"]
         elif ("ubuntu" in self.name) or ("debian" in self.name):
             patterns += ["wezterm-*.deb", "wezterm-*.xz"]
+        elif "alpine" in self.name:
+            patterns += ["/root/packages/wezterm-*.apk", "/root/.abuild/*.rsa.pub"]
```
wez, line 397: What is the purpose of uploading the `.rsa.pub` here?

panekj: It's how artefacts are packaged in Alpine CI (`.apk` + public key)
`*.rsa.pub` can be added to `/etc/apk/keys` for the signer to be "trusted" and package can be installed without issues
This can be bypassed with `--allow-untrusted` though

wez: For nightly builds, if we regenerate the key on every build, I can see that being annoying for users to keep updating to the new random key.

I think the key should probably be a repo-level secret so that it is consistent across the builds, with the public portion checked in to the repo. We can publish the public key on each release upload so that it is available alongside the `.apk`.

panekj: I agree but I never dealt with signing/keys in CIs so not sure how to get about that. I assume I could just do below if you set a secret with priv key(?):
```sh
echo "${{ secrets.WEZTERM_SIGN_KEY }}" > /root/.abuild/sign.rsa
```

ci/check-rust-version.sh
SuperSandro2000: > bash differences are even more painful than /bin/sh, because of breaking changes or bugs, and there is much more wrong with bash than that.

I am not aware of any breaking changes or bugs in bash in the last years. I think you are just generally against bash and GNU in general. A binary path which is different on many systems and has many different implementations with widely varying feature sets is supposed to be more stable than one shell?

> I just made script POSIX compliant and it isn't that hard, it took me about an hour or two

Yeah, it took you one or two hours with the alternative being that you install one of the most widely used unix programs. Seams to be a not very fair tradeoff.

> If I change something while still being POSIX compliant

True but more people than you change things and many of them do not even have a fully posix shell on their system to check if things still work.

ci/generate-workflows.py
SuperSandro2000: Just be aware that secrets are not available for PRs.

wez: I think it will be hard for you to iterate on secrets as you're not the repo owner, so my suggestion to get this through is:

* Do the keygen and upload for tagged releases
* Skip the key upload for nightly builds, and suggest that folks need to use `--allow-untrusted` in any docs that we put together

We can work on adopting secrets as a follow up

panekj: @wez will it be okay to include key pattern in `upload_asset_tag` function?

panekj:
```py
    def upload_asset_tag(self):
        steps = []

        patterns = self.asset_patterns()
        if "alpine" in self.name:
            patterns.append("$HOME/.abuild/*.rsa.pub")
        glob = " ".join(patterns)
```

wez: Yeah, I think that makes sense for `upload_asset_tag`

panekj: after all I figured out how to add it there
https://github.com/panekj/wezterm/blob/8fb11b6baa2de48550e985720085b09975aca90a/ci/generate-workflows.py#L403-L406

ci/generate-workflows.py
```diff
@@ -511,10 +535,11 @@ def update_homebrew_tap(self):
         return steps
 
     def global_env(self):
-        env = {}
         if "macos" in self.name:
-            env["MACOSX_DEPLOYMENT_TARGET"] = "10.9"
-        return env
+            self.env["MACOSX_DEPLOYMENT_TARGET"] = "10.9"
+        if "alpine" in self.name:
+            self.env["RUSTFLAGS"] = "'-C target-feature=-crt-static'"
```
panekj, line 541: too many quotes :fearful:

wez:
Thanks!

PR #1532, opened 2022-01-11
Update workflows to validate that tests can compile and run with mutally exclusive libssh-rs and ssh2 features

chipsenkbeil:
Picked Ubuntu 18.04 instead of ubuntu-latest since a bunch of other targets have the `os` variable set to ubuntu-latest by default.

Let me know your thoughts on this one.

### DIFF
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -474,7 +474,7 @@ pub struct Config {
     #[serde(default = "default_true")]
     pub scroll_to_bottom_on_input: bool,
 
-    #[serde(default = "default_use_ime")]
+    #[serde(default = "default_true")]
     pub use_ime: bool,
     #[serde(default)]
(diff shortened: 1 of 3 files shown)


wez:
I was thinking of this being a separate manually curated workflow file (say, using https://github.com/wez/wezterm/blob/main/.github/workflows/verify-pages.yml as a "model") that just runs these two test invocations

wez:
https://github.com/wez/wezterm/blob/main/.github/workflows/termwiz.yml is probably a better basis (just made that for a different issue).  

I think you might need to swap out the checkout action to this if the build needs to pull in submodules:

https://github.com/wez/wezterm/blob/507267aeeecd1a1a70b6d92313b2abb86368d983/.github/workflows/gen_centos8.yml#L41-L44

PR #1534, opened 2022-01-12
FEATURE: Allow ScrollByPage f64 values

hahuang65:
`ScrollByPage` can accept non-integer values in the configuration.
This allows fractional page scrolling, such as by half a page.

The default remains the same, at 1 page.

### DIFF
--- a/docs/config/lua/keyassignment/ScrollByPage.md
+++ b/docs/config/lua/keyassignment/ScrollByPage.md
@@ -14,4 +14,18 @@ return {
 }
 ```
 
+*Since: nightly builds only*
 
+You may now use floating point values to scroll by partial pages.  This example shows
+how to make the `PageUp`/`PageDown` scroll by half a page at a time:
+
+```lua
+local wezterm = require 'wezterm';
+
+return {
+  keys = {
+    {key="PageUp", mods="SHIFT", action=wezterm.action{ScrollByPage=-0.5}},
+    {key="PageDown", mods="SHIFT", action=wezterm.action{ScrollByPage=0.5}},
+  }
+}
+```
--- a/wezterm-gui/src/termwindow/mod.rs
+++ b/wezterm-gui/src/termwindow/mod.rs
@@ -1756,7 +1756,7 @@ impl TermWindow {
         Ok(())
     }
 
-    fn scroll_by_page(&mut self, amount: isize) -> anyhow::Result<()> {
+    fn scroll_by_page(&mut self, amount: f64) -> anyhow::Result<()> {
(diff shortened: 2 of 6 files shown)


docs/config/lua/keyassignment/ScrollByPage.md
```diff
@@ -2,6 +2,7 @@
 
 Adjusts the scroll position by the number of pages specified by the argument.
 Negative values scroll upwards, while positive values scroll downwards.
+Float values will scroll by partial pages (e.g. 0.5 will scroll half a page down).
```
wez, line 5: Thanks for adding some docs!  What I like to do for these is make a section like this:

```
*Since: nightly builds only*

You may now use floating point values to scroll by partial pages.  This example shows
how to make the `PageUp`/`PageDown` scroll by half a page at a time:

<insert example here>
```

I would put that whole block below the existing example

wezterm-gui/src/termwindow/mod.rs
```diff
@@ -1756,17 +1757,17 @@ impl TermWindow {
...
             None => return Ok(()),
         };
         let dims = pane.get_dimensions();
         let position = self
             .get_viewport(pane.pane_id())
-            .unwrap_or(dims.physical_top)
-            .saturating_add(amount * dims.viewport_rows as isize);
-        self.set_viewport(pane.pane_id(), Some(position), dims);
+            .unwrap_or(dims.physical_top) as f64
+            + (f64::from(amount) * dims.viewport_rows as f64);
```
wez, line 1769: I think you can deref the `NotNan` to get at the inner value; can you try this?
I think this is slightly easier to read:

```suggestion
            + (*amount * dims.viewport_rows as f64);
```

wezterm-gui/src/termwindow/mod.rs
```diff
@@ -1756,17 +1757,17 @@ impl TermWindow {
         Ok(())
     }
 
-    fn scroll_by_page(&mut self, amount: isize) -> anyhow::Result<()> {
+    fn scroll_by_page(&mut self, amount: NotNan<f64>) -> anyhow::Result<()> {
```
wez, line 1760: I think it looks a little less magical to pass `f64` directly here, and extract the value from the `NotNan` at the call site.

```suggestion
    fn scroll_by_page(&mut self, amount: f64) -> anyhow::Result<()> {
```

wez (commented):
This is looking good!

hahuang65:
@wez this could use another look. I haven't figured out why the formatter is complaining about unrelated code. All my stuff is up to date. Is it possible that Rust 1.57.0 is why?

PR #1544, opened 2022-01-13
Add wezterm-ssh independent CI task to verify libssh-rs and ssh2 features can work independently

chipsenkbeil:
Update workflows to validate that tests can compile and run with mutally exclusive libssh-rs and ssh2 features


wez:
Thanks!

PR #1549, opened 2022-01-15
Filter out per-pixel mouse movement when not in SGR-Pixels mode

klamonte:
Fix for #1457


term/src/terminalstate/mouse.rs
```diff
@@ -186,7 +186,23 @@ impl TerminalState {
...
+                y_pixel_offset: 0,
+                button: event.button,
+                modifiers: event.modifiers,
+            },
+        };
+        let reportable = (self.any_event_mouse || !self.current_mouse_buttons.is_empty())
+            && (self.mouse_encoding == MouseEncoding::SgrPixels
+                || (self.mouse_encoding != MouseEncoding::SgrPixels
+                    && (event.x != check_last_mouse_move.x)
+                    || (event.y != check_last_mouse_move.y)));
```
wez, line 205: One of the nice things you can do with `match` is to phrase things almost like a state table.
I don't know if this is 100% the logic you want here, but this feels a bit easier to follow: we're matching both the mouse movement and the encoding together:

```suggestion
        let moved = match (&self.last_mouse_move, self.mouse_encoding) {
            (None, _) => true,
            (Some(last), MouseEncoding::SgrPixels) => last.x != event.x || last.y != event.y || last.x_pixel_offset != event.x_pixel_offset || last.y_pixel_offset != event.y_pixel_offset,
            (Some(last), _) => last.x != event.x || last.y != event.y, 
        };

        let reportable = (self.any_event_mouse || !self.current_mouse_buttons.is_empty())
            && moved;
```

klamonte: Love it! That is so much clearer, and exactly the design intent.

wez:
Thanks!

PR #1571, opened 2022-01-20
fix socketpair() implementation to build on Illumos

Nieuwejaar:
The PF_LOCAL #define isn't available on Illumos.  The correct value to use on that OS is AF_UNIX.  I suspect that AF_UNIX is the correct value to use everywhere, but I didn't want to make a universal change without having more systems to test on.

As written, this change should only affect Illumos systems.  As a sanity check, I've also built on Linux and macos.

### DIFF
--- a/filedescriptor/src/unix.rs
+++ b/filedescriptor/src/unix.rs
@@ -392,7 +392,12 @@ pub fn socketpair_impl() -> Result<(FileDescriptor, FileDescriptor)> {
 #[doc(hidden)]
 pub fn socketpair_impl() -> Result<(FileDescriptor, FileDescriptor)> {
     let mut fds = [-1i32; 2];
-    let res = unsafe { libc::socketpair(libc::PF_LOCAL, libc::SOCK_STREAM, 0, fds.as_mut_ptr()) };
+    #[cfg(target_os = "illumos")]
+    let domain = libc::AF_UNIX;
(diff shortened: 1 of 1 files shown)


wez:
Thanks!
FWIW, I'd accept this unconditionally doing AF_UNIX.  The CI covers freebsd, macos, windows and linux which I think is sufficient for the !illumos cases.

Nieuwejaar:
> Thanks! FWIW, I'd accept this unconditionally doing AF_UNIX. The CI covers freebsd, macos, windows and linux which I think is sufficient for the !illumos cases.

I'm not sure what that means.  Are you asking me to resubmit the PR with an unconditional change?  Or are you good with this change as it is, and I should submit future PRs as unconditional changes?  I'm happy either way - I just didn't want to leave this hanging here if you're waiting for me to follow-up.

wez:
Please update this PR to unconditionally use AF_UNIX :-)

Nieuwejaar:
Got it.  Done.

wez:
Thanks!  Are you doing this in order to use wezterm, or are you using the filedescriptor crate directly?

Nieuwejaar:
I'm using the lineedit crate, which pulled in the filedescriptor crate.

wez:
OK! I've published filedescriptor 0.8.2

PR #1572, opened 2022-01-21
Add support for blesh hooks

SuperSandro2000:
blesh provides its own hooks which are better integrated into it. No need to duplicate the functionality and slow down the shell unnecessarily. See https://github.com/akinomyoga/ble.sh/wiki/Manual-%C2%A71-Introduction#user-content-fn-blehook

### DIFF
--- a/assets/shell-integration/wezterm.sh
+++ b/assets/shell-integration/wezterm.sh
@@ -1,3 +1,5 @@
+# shellcheck shell=bash
+
 # This file hooks up shell integration for wezterm.
 # It is suitable for zsh and bash.
 #
@@ -10,6 +12,7 @@
 # WEZTERM_SHELL_SKIP_SEMANTIC_ZONES - disables zones
 # WEZTERM_SHELL_SKIP_CWD - disables OSC 7 cwd setting
 
+# shellcheck disable=SC2166
 if [ -z "${BASH_VERSION}" -a -z "${ZSH_NAME}" ] ; then
   # Only for bash or zsh
   return 0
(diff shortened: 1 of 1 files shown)


SuperSandro2000:
I played a bit around with it and jumping from prompt to prompt works like a charm.

wez:
Thanks!

assets/shell-integration/wezterm.sh
```diff
@@ -380,7 +383,11 @@ fi;
 
 } # end of __wezterm_install_bash_prexec
 
-__wezterm_install_bash_prexec
+# blesh provides it's own preexec mechanism which is recommended over bash-preexec
+# See https://github.com/akinomyoga/ble.sh/wiki/Manual-%C2%A71-Introduction#user-content-fn-blehook for more details
+if [[ ! -v BLE_VERSION ]]; then
```
akinomyoga, line 388: `[[ -v var ]]` is Bash 4.2 feature. Unfortunately, this will produce error messages in lower versions of Bash such as Bash 3.2 in macOS. `[[ ! ${BLE_VERSION-} ]]` is the way that works in all the versions of Bash 3.0+.

wez: I pushed https://github.com/wez/wezterm/commit/78ea214a96da7e9e8c8e5f3539bb788902eaff53 which uses the same `-n "$foo"` test that we use for testing for zsh

SuperSandro2000: Thanks. I usually don't give a damn about the ancient versions of Darwin.

PR #1574, opened 2022-01-21
Added missing flag in documentation about listing system fonts

ngmlabs:
"Troubleshooting Fonts" section in documentation is missing the `--list-system` flag in output with system fonts.


wez:
Hah! Good catch! Thanks!

PR #1577, opened 2022-01-23
Support DECSDM

klamonte:
For #217 :

* DECSDM sets/resets Sixel Display Mode correctly.
* For sixel, images are placed at 0,0 when DECSDM is set.
* Sixel and iTerm2 images currently replace rather than blit-over, need to get that fixed.


term/src/terminalstate/sixel.rs
```diff
@@ -123,9 +131,10 @@ impl TerminalState {
             z_index: 0,
             image_id: None,
             placement_id: None,
-            do_not_move_cursor: false,
+            do_not_move_cursor: self.sixel_display_mode,
         }) {
             log::error!("set sixel image: {:#}", err);
         }
+        self.cursor = old_cursor;
```
wez, line 138: should this be conditional on sixel_display_mode?  Feels like we're going to revert the cursor position in the case where we did want it to move?

klamonte: Up to you.  old_cursor is only changed for sixel_display_mode.  There is also sixel_scrolls_right mode which I would assume DECSDM overrides based on ctlseqs: "Sixel scrolling leaves cursor to right of graphic." ; and that is dealt with in assign_image_to_cells.

The abstraction is leaking. Do we want a single place to choose both start and end position?  That would put it in assign_image_to_cells; but then we're putting a really sixel-only-ish logic in there.  Would it be better to make bottom_right a parameter from outside assign_image_to_cells so that scrolling and DECSDM are handled together in sixel() ?

```
       if !params.do_not_move_cursor {
            // Sixel places the cursor under the left corner of the image,
            // unless sixel_scrolls_right is enabled.
            // iTerm places it after the bottom right corner.
            let bottom_right = match params.style {
                ImageAttachStyle::Kitty | ImageAttachStyle::Iterm => true,
                ImageAttachStyle::Sixel => self.sixel_scrolls_right,
            };

            if bottom_right {
                self.set_cursor_pos(
                    &Position::Relative(width_in_cells as i64),
                    &Position::Relative(-1),
                );
            }
        }
```

klamonte: Re-reading that made me think that we should just put start/end position as parameters to assign_image_to_cells, and make all of the image formats pass it in.  Because that's really what this is: they have their own rules, and shouldn't be mixed up.  assign_image_to_cells should just be slice-and-dice: then it could be re-used anywhere else you might want to splat a bitmap on screen later on.

wez: what I meant was: this appears to unconditionally reset `self.cursor` to old cursor, which would undo any cursor movement that might have been done by `assign_image_to_cells` in the `do_not_move_cursor: false` case.

term/src/terminalstate/mod.rs
```diff
@@ -379,8 +379,24 @@ struct UnicodeVersionStackEntry {
 
 fn default_color_map() -> HashMap<u16, RgbColor> {
     let mut color_map = HashMap::new();
+    // Match colors to the VT340 color table:
+    // https://github.com/hackerb9/vt340test/blob/main/colormap/showcolortable.png
     color_map.insert(0, RgbColor::new_8bpc(0, 0, 0));
-    color_map.insert(3, RgbColor::new_8bpc(0, 255, 0));
+    color_map.insert(1, RgbColor::new_8bpc(0x33, 0x33, 0xcc));
```
wez, line 385: if you fancy a game of Rust golf, this is at a point where I think we could reduce some boilerplate, something like:

```rust
   for (idx, r, g, b) in &[
               (0, 0x00, 0x00, 0x00),
               (1, 0x33, 0x33, 0xcc), 
               (2, 0xcc, 0x23, 0x23), 
     ] {
       color_map.insert(idx, RgbColor::new_8bpc(r, g, b));
   }
```

klamonte: Fore! :)

term/src/terminalstate/sixel.rs
klamonte: Oops!  You're right.  I didn't see it because I never allow sixel to scroll.

term/src/terminalstate/mod.rs
klamonte: The compiler did not like the "&".  It wanted more like:

```Rust
for (idx, r, g, b) in [
               (0, 0x00, 0x00, 0x00),
               (1, 0x33, 0x33, 0xcc), 
               (2, 0xcc, 0x23, 0x23), 
     ] {
       color_map.insert(idx, RgbColor::new_8bpc(r, g, b));
   }
```

Does that work for you?

wez: Yeah, that's totally fine

term/src/terminalstate/image.rs
```diff
@@ -184,6 +184,75 @@ impl TerminalState {
...
+                    (_, Transparent) => {},
+                    (ImageAttachStyle::Kitty, _) => cell.attrs_mut().attach_image(img),
+                    (ImageAttachStyle::Sixel | ImageAttachStyle::Iterm, Opaque) => {
+                        cell.attrs_mut().set_image(img)
+                    },
+                    (ImageAttachStyle::Sixel | ImageAttachStyle::Iterm, Mixed) => {
+                        cell.attrs_mut().blit_over_image(img)
+                    },
+                };
+                */
```
klamonte, line 255: @wez Here is where I am stuck.  (And gonna press the "Add single comment" button rather than "Start a review"... Sorry still don't know GH that well.)

klamonte:
@wez I'm thinking about renaming my GitHub username inside the next couple weeks, which might have odd ramifications for open PRs.

If I eliminated the really long comment and you merged here, then you would gain:

* Detection of DECSDM support via DECRQM/DECRPM.
* Correct sixel image placement when DECSDM is set.
* VT340 default sixel colors.

The "transparent-over-nontransparent" stuff probably impacts iTerm2, maybe Kitty, and will end up more invasive than this stuff.  I feel like it may be better to put that part off to later.

What do you think?

wez:
Yeah, let's split out the transparency stuff to a separate PR; the rest of this one looks good!

klamonte:
Did you want one more merge from upstream?

wez:
No need for another merge: I squashed and pushed!
Thanks again!

jgcodes2020:
Just for the record, I was experimenting with sixel, and couldn't get wezterm to show [this sixel image](https://gist.github.com/jgcodes2020/440427b878e7c1be7910ae60f7112eff) that I generated using a program. As far as I know, it should be valid sixel.

PR #1586, opened 2022-01-24
assets/shell-integration: improve the integration with ble.sh

akinomyoga:
Here are fixes for the shell integration. These are additional adjustments to 56e9a9af7 and 78ea214a9. Let me open this as a draft PR. There can be even more fixes because the check of the shell integration with `ble.sh` is still ongoing (Cc: @SuperSandro2000).

- 608aad49f In case that the shell option `set -u` is set, the optional variables are referenced as `${varname-}`
- 9df1b53db The precmd/preexec hooks are registered only when it is not yet registered. This avoids duplicate precmd/preexec handlers after `wezterm.sh` is sourced multiple times. Maybe we also want to modify the code for `bash-preexec`?
- e2c0816c2 By default, ble.sh assumes that the cursor position and the terminal contents are not changed by calling PRECMD/PROMPT_COMMAND. However, it seems the escape sequence sent from the wezterm precmd handler causes a newline when the cursor position is not at the beginning of the line. This breaks the internal layout calculation of ble.sh. To tell ble.sh that PRECMD can change the layout, we set a blesh option.

**Update 2022-03-04**: I'm currently waiting for https://github.com/rcaloras/bash-preexec/pull/129 being processed.

### DIFF
--- a/assets/shell-integration/wezterm.sh
+++ b/assets/shell-integration/wezterm.sh
@@ -13,12 +13,12 @@
 # WEZTERM_SHELL_SKIP_CWD - disables OSC 7 cwd setting
 
 # shellcheck disable=SC2166
-if [ -z "${BASH_VERSION}" -a -z "${ZSH_NAME}" ] ; then
+if [ -z "${BASH_VERSION-}" -a -z "${ZSH_NAME-}" ] ; then
   # Only for bash or zsh
   return 0
 fi
 
-if [ "${WEZTERM_SHELL_SKIP_ALL}" = "1" ] ; then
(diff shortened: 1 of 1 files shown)


wez:
I'm going to close this for now: when you're ready to submit it, please rebase and reopen.

akinomyoga:
OK! Sorry for keeping this open for a long time. I actually wanted to adjust it after https://github.com/rcaloras/bash-preexec/pull/129 is settled, but it seems to take some time for it to receive another review from some other contributor of `bash-preexec`. I'll reopen it after https://github.com/rcaloras/bash-preexec/pull/129 is finally merged or closed.

PR #1599, opened 2022-01-29
window: correctly handle keyboard input after seat updates

LawnGnome:
When the user switches to another virtual console and then back into a
Wayland session, a number of seat updates are received by the listener
installed by `Environment::listen_for_seats()`. These updates
essentially flip the keyboard and pointer flags off and then back on in
the seat data, which indicates that new handlers need to be assigned to
the keyboard and pointer objects.

However, because one update is received for _each_ flag being toggled,
this means that the listener is fired twice with `has_keyboard` set to
true, which means that two keyboard handlers end up being installed.
Users then experience each keypress being delivered twice to the
terminal.

This commit adds a map to track the keyboard object that has been
assigned on each seat, thereby preventing duplicate assignments and (by
extension) duplicate keypresses being registered. This logic is
essentially the same as what's used in the `kbd_input` example in SCTK,
which doesn't have this issue.

Something similar is required for pointer handling, which also breaks
after switching to another virtual console and back into Wayland, but I
was scared off by the TODO in the `listen_for_seats` callback and didn't
investigate this further.

### DIFF
--- a/window/src/os/wayland/connection.rs
+++ b/window/src/os/wayland/connection.rs
@@ -60,14 +60,18 @@ impl WaylandConnection {
             toolkit::new_default_environment!(MyEnvironment, desktop)?;
 
         let mut pointer = None;
+        let mut seat_keyboards = HashMap::new();
 
         for seat in environment.get_all_seats() {
-            if let Some((has_kbd, has_ptr)) = toolkit::seat::with_seat_data(&seat, |seat_data| {
-                (
-                    seat_data.has_keyboard && !seat_data.defunct,
-                    seat_data.has_pointer && !seat_data.defunct,
-                )
-            }) {
+            if let Some((has_kbd, has_ptr, name)) =
+                toolkit::seat::with_seat_data(&seat, |seat_data| {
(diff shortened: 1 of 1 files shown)


wez:
Thanks for this!  It sounds like it will resolve https://github.com/wez/wezterm/issues/1497
Did you notice a problem in practice with the mouse/pointer in this situation?

LawnGnome:
Thanks! I went looking for a bug, but obviously missed that.

> Did you notice a problem in practice with the mouse/pointer in this situation?

I did, yes — I have vim configured to support mouse input, and it stopped working after switching virtual consoles.

wez:
Sorry for leaving this hanging out there: I'd like to also see the mouse point stuff resolved "at the same time", as even though this improves things, we're still a bit broken in this situation.  Would you mind taking a stab at that?

LawnGnome:
Sorry Wez — I've been away most of the month and not paying a tonne of attention to my personal notifications. It looks like you had a stab at the pointer side of things: is there anything I can do to help there?

PR #1636, opened 2022-02-11
Add support for urgency hints

noctux:
Note: Please regard this PR more as a WIP for starting a discussion, and not as a completely polished submission.

Often, programs want to request the user's focus. The common way to do so is by using the "bell". Currently, wezterm offers the user three major options for this event: AudibleBell (play a sound), VisualBell (blink the whole terminal pane) and sending a ToastNotification from the lua callback.

However, nearly all platforms furthermore posess a native mechanism to request focus:
- X11: Set the `URGENCY`-flag in `WM_FLAGS` (as specified in the [icccm standard](https://tronche.com/gui/x/icccm/sec-4.html#s-4.1.2.4)
- Wayland: Via the [xdg-activation protocol](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/blob/main/staging/xdg-activation/xdg-activation-v1.xml)
- Windows: Via [FlashWindow](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-flashwindow), which causes the Taskbar icon to flash
- OSX:Bounce the icon in the bar (Probably, I've never really used that platform, and haven't found any references by casually googling)

Often, these mechanisms allow for a tight integration with the overall system, e.g. by offering keybindings to switch to the alerted window, etc.pp., so I as a user would be interested in exposing "urgency" to the system.

The most generic way I've found for adding this behavior is by exposing the signaling capability to the lua bindings, that way, the user is completely flexible in how and when she wants to actually use this mechanism:

```lua
wezterm.on("bell", function(window, pane)
	window:alert();
end)
```

For now, this PR only implements this behavior on X11, because this happens to be what I am using. For wayland, it seems that the xdg-activation protocol is not yet supported in wayland-rs (at least to my understanding, I've never really worked with wayland so far, but at least grepping the repository does not yield any matches for 'activation'). I do not have access to Windows or OSX machines.

On the implementation itself: I really hate how I seem to be forced to manually reimplement a copy mechanism for `WmHints` in `XWindowInner::alert`, however, I have not seen any option (sort of transmuting the object, which seems very wrong) to keep the old urgency values. Unfortunately, the documentation on `xcb_util::icccm::(get|set)_wm_hints` seems very sparse, but the implementation seems to not update but completely replace the wm-hints, so this dance seems necessary. Other libraries offer this functionality in a more straightforward fashion (e.g. [here](https://github.com/alacritty/alacritty/pull/812/files#diff-046e4a8d68b341cbc69ff710f5c2a0530f57ad4a1a2c78a3f723efba02533530R327) is alacritty's implementation using xlib), so I'm really not sure if the xcb_utils-wrapping is strange or I'm just a bid slow on the uptake :D

Furthermore, I'm not sure if `alert()` is the best name for the functionality.

Thanks for reading this far (and WezTerm in general)! As a sidenote, I'm not really a rust programmer (yet), so feel free to point out any strange things within the code.
  ~ Simon

More related urls:
[Here](https://codeberg.org/dnkl/foot/commit/92e517ae34243d9c294e966221f5a4589684a9b5) is foot's implementation of xdg-activation support, it does not seem tooo bad, but I wonder if this can be done (in a clean fashion) without involving/buggering wayland-rs.

### DIFF
--- a/window/src/os/x11/window.rs
+++ b/window/src/os/x11/window.rs
@@ -920,7 +920,7 @@ impl XWindowInner {
         self.set_fullscreen_hint(!fullscreen).ok();
     }
 
-    fn alert(&mut self) {
+    fn set_attention_hint(&mut self, new_state: bool) {
         let hintreply =
             xcb_util::icccm::get_wm_hints(self.conn().conn(), self.window_id).get_reply();
         let hints = match hintreply {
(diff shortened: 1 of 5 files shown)


wez:
Thanks for diving in!

Naming: I'm leaning towards something like `request_attention` rather than `alert`.  What do you think?

For macOS, it's a bit weird because the application icon in the dock is "global" rather than per window, so there may be some additional consideration needed to connect the bouncing icon to activating the intended window when multiple are active.

For wayland, it looks like https://docs.rs/wayland-protocols/latest/wayland_protocols/staging/xdg_activation/index.html exposes the protocol in Rust.

re: wmhints building, I agree that the read-modify-write interface is horrible.  I don't have any great suggestions that wouldn't be better than just fixing the interface in the xcb-util crate; perhaps it's worth sending a PR there to do that? :)  For example, it seems reasonable to add a alternative constructor that makes a builder from an existing hints instance.

I think we should have methods that both set and clear the attention state on `Window`

window/src/os/x11/window.rs
```diff
@@ -920,6 +920,70 @@ impl XWindowInner {
...
+            // but subsequent setting then works.
+            Err(_) => xcb_util::icccm::WmHints::empty().build(),
+        };
+        // Note: unfortunately, hints.is_urgent seems buggy as of xcb-util 0.3.0,
+        //       always returning false, so we for now do not do the "smart" thing
+        //       of checking is_urgent != new_state as shown below, but instead always
+        //       assign hints unconditionally, setting urgency as per new_state.
+        //
+        // let is_urgent = hints.is_urgent().unwrap_or(false);
+        // if is_urgent != new_state {
```
wez, line 940: I'd like to avoid the unbalanced brace here, so I'd suggest removing this line!

```suggestion
```

wez:
What do you think about `request_attention` and `clear_attention_request` as the pair of methods for this?
The `bool` is "OK" but doesn't feel very expressive :)

wez:
I'm going to close this out for now; please reopen it when you have time to rebase and revisit.

ncfavier:
I'd like to give this a shot but Rust does not like me: after rebasing on main I added `xcb-util` to `window/Cargo.toml`, but I still cannot import `xcb_util::icccm` for some reason.

If I do

```rust
use crate::xcb_util::icccm;
```

then I get "unresolved import `crate::xcb_util`".

ncfavier:
Okay, I needed to enable `xcb-util`'s `icccm` feature and now `::xcb_util::icccm` works fine. But now it's complaining that two different versions of `xcb` are being used. Looks like we use xcb 1.3 while xcb-utils wants xcb 0.10.

ncfavier:
Tried using `xcb-wm` instead, but instantly ran into https://github.com/arminfriedl/xcb-wm/issues/5.

PR #1657, opened 2022-02-22
docs: fix truncate call in example and fix typo

Riyyi:
- truncate_to_width doesn't exist in the current codebase
- small typo "window managed" -> "window manager"


wez:
Thanks!

PR #1658, opened 2022-02-22
docs: update minimum requires Rust version for build

arekfu:
The doc used to indicate that v1.51 is required, but even v1.55 fails because of `bidi`:
```
❯ cargo build
error: failed to load manifest for workspace member `/home/catA/dm232107/src/wezterm/bidi`

Caused by:
  failed to parse manifest at `/home/catA/dm232107/src/wezterm/bidi/Cargo.toml`

Caused by:
  feature `edition2021` is required

  The package requires the Cargo feature called `edition2021`, but that feature is not stabilized in this version of Cargo (1.55.0 (32da73ab1 2021-08-23)).
  Consider trying a newer version of Cargo (this may require the nightly release).
  See https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#edition-2021 for more information about the status of this feature.
```
I tried version 1.58 and it works, but v1.56 should be okay (`edition2021` was standardized in v1.56).

### DIFF
--- a/ci/check-rust-version.sh
+++ b/ci/check-rust-version.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-min_rust=(1 53 0)
+min_rust=(1 56 0)
 rust_ver=()
 
 parse_rustc_version() {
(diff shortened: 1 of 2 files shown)


wez:
https://github.com/wez/wezterm/blob/main/ci/check-rust-version.sh#L3 checks for 1.53.  If that's no longer correct, would you mind also updating that line to match the version in this PR?

arekfu:
I can confirm that 1.55 fails and 1.56 works.

wez:
Thanks!

PR #1669, opened 2022-02-26
fix x button display without cell size

junnplus:
Signed-off-by: ye.sijun <junnplus@gmail.com>

before:
image

after:
image

### DIFF
--- a/wezterm-gui/src/termwindow/render.rs
+++ b/wezterm-gui/src/termwindow/render.rs
@@ -687,14 +687,16 @@ impl super::TermWindow {
                         ElementContent::Text(_) => unreachable!(),
                         ElementContent::Poly { .. } => unreachable!(),
                         ElementContent::Children(mut kids) => {
+                            let x_dimension =
+                                Dimension::Pixels(metrics.cell_size.width as f32 * 0.5);
                             let x_button = Element::new(
                                 &font,
                                 ElementContent::Poly {
                                     line_width: metrics.underline_height.max(2),
                                     poly: SizedPoly {
                                         poly: X_BUTTON,
-                                        width: Dimension::Cells(0.5),
-                                        height: Dimension::Cells(0.5),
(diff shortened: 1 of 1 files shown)


wez:
Thanks for diving in here!

Fixing the aspect to be square makes sense.  Stylistically, I'd prefer to see this diff phrased this way:

```diff
diff --git a/wezterm-gui/src/termwindow/render.rs b/wezterm-gui/src/termwindow/render.rs
index 3d331fdc3..fe7d24a2b 100644
--- a/wezterm-gui/src/termwindow/render.rs
+++ b/wezterm-gui/src/termwindow/render.rs
@@ -687,14 +687,17 @@ impl super::TermWindow {
                         ElementContent::Text(_) => unreachable!(),
                         ElementContent::Poly { .. } => unreachable!(),
                         ElementContent::Children(mut kids) => {
+                            // Make the button square
+                            let x_dimension =
+                                Dimension::Pixels(metrics.cell_size.width as f32 * 0.5);
                             let x_button = Element::new(
                                 &font,
                                 ElementContent::Poly {
                                     line_width: metrics.underline_height.max(2),
                                     poly: SizedPoly {
                                         poly: X_BUTTON,
-                                        width: Dimension::Cells(0.5),
-                                        height: Dimension::Cells(0.5),
+                                        width: x_dimension,
+                                        height: x_dimension,
                                     },
                                 },
                             )
```

another thing to note here is that using `0.6` results in clipping the top of the `x` graphic when run with the default configuration; you can easily see/verify for yourself by running `./target/debug/wezterm -n`.

The reason is that the rest of the layout assumes `0.5` cell height and uses padding and margins that are also based on the height/width respectively.

`0.5` looks better, but it still has the potential to look weird if the font's aspect ratio is unusual.

A more robust version of this PR would need to compute the correct pixel dimensions for all of the margins and padding used in the x button; would you mind taking a pass at that?

PR #1670, opened 2022-02-26
Fix the Path style with tilde key prefix not working in background image config 

Zheaoli:
While I try to config the background image with `tilde key prefix` such as below

```lua

return {
    "Monaco",
    { family = "JetBrains Mono", weight = "Regular" },
    "Hack Nerd Font Mono",
    "MesloLGS NF",
    "PowerlineSymbols",
  }),
  font_size = 12.0,
  window_background_opacity = 0.95,
  window_background_image = "~/Documents/background.png",
  window_background_image_hsb = {
    -- Darken the background image by reducing it to 1/3rd
    brightness = 0.01,

    -- You can adjust the hue by scaling its value.
    -- a multiplier of 1.0 leaves the value unchanged.
    hue = 1.0,

    -- You can adjust the saturation also.
    saturation = 1.0,
  },
  use_ime = true,
}
```

The config module will parse the path to `$HOME/.config/wezterm/~/Documents/background.png`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -662,6 +662,7 @@ dependencies = [
  "bstr 0.2.17",
  "chrono",
  "colorgrad",
+ "dirs 4.0.0",
  "dirs-next",
  "enum-display-derive",
  "filenamegen",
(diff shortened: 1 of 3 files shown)


wez:
Thanks for diving in!

The supported way to reference your home directory is like this:

```lua

return {
  window_background_image = wezterm.home_dir .. "/Documents/background.png",
}
```

(See: https://wezfurlong.org/wezterm/config/lua/wezterm/home_dir.html)

I've been resisting adding full tilde expansion because it is more complicated than it first appears--it supports resolving home directories for arbitrary users, which may fail for various reasons on systems that use NIS+, and has some irritating complications with Windows style paths.  Furthermore, there are quite a few different places where the config references filenames, so adding it to just one place would then make all the others inconsistent.

I'm not inclined to add automatic tilde expansion to the config layer as a result of this: I'm not sure it is worth the engineering effort when the example above will do the job quite simply.

I'd be open to seeing lua helper function(s) to make working with path strings more convenient if you don't like the idea of raw string concatenation!

Zheaoli:
Thanks for your patience, I think your concern is right! we should think about the non-Unix-like platform's path style if we need to add a path hook in the codebase.

I will close my PR first!

Issue #1674, opened 2022-02-28
Improve RESIZE window decorations on Windows


davidrios:
**Is your feature request related to a problem? Please describe.**
Currently, the terminal behaves badly when using `window_decorations = "RESIZE"` on Windows:

- It loses the ability to minimize, maximize.
- Shortcuts like Win+Arrows to maximize, snap to sides, etc, don't work.
- There's an ugly white border on top of the window.

**Describe the solution you'd like**
Make the window behave in a more natural way.

**Additional context**
This is related to the necessary work for window buttons integrated in the tab bar on Windows.

davidrios:
The fixes for this are already merged.

PR #1675, opened 2022-02-28
Improve RESIZE window decorations on Windows

davidrios:
The problems are described in #1674.

This change fixes all of them. The code is what's generally done to make an app with a custom title bar, with tabs like Firefox for instance:

- It handles `nccalcsize`, so that you draw your own borders/window decorations.
- It handles `nchittest` to deal with mouse input in your custom title bar.

This is based on this article: https://kubyshkin.name/posts/win32-window-custom-title-bar-caption/

### DIFF
--- a/assets/windows/manifest.manifest
+++ b/assets/windows/manifest.manifest
@@ -1,9 +1,43 @@
-<?xml version='1.0' encoding='UTF-8' standalone='yes'?>
-<assembly xmlns="urn:schemas-microsoft-com:asm.v1"
-    manifestVersion="1.0" xmlns:asmv3="urn:schemas-microsoft-com:asm.v3" >
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<assembly
+  manifestVersion="1.0"
+  xmlns="urn:schemas-microsoft-com:asm.v1"
+  xmlns:asmv3="urn:schemas-microsoft-com:asm.v3"
+>
   <asmv3:application>
     <asmv3:windowsSettings xmlns="http://schemas.microsoft.com/SMI/2016/WindowsSettings">
       <dpiAwareness>PerMonitorV2</dpiAwareness>
     </asmv3:windowsSettings>
   </asmv3:application>
+  <dependency>
+    <dependentAssembly>
+      <assemblyIdentity
+        type="win32"
+        name="Microsoft.Windows.Common-Controls"
+        version="6.0.0.0"
+        processorArchitecture="*"
+        publicKeyToken="6595b64144ccf1df"
+        language="*"
+      />
+    </dependentAssembly>
+  </dependency>
+  <compatibility xmlns="urn:schemas-microsoft-com:compatibility.v1">
+    <application>
+      <!-- Windows 10 and Windows 11 -->
+      <supportedOS Id="{8e0f7a12-bfb3-4fe8-b9a5-48fd50a15a9a}"/>
+    </application>
+  </compatibility>
+  <trustInfo xmlns="urn:schemas-microsoft-com:asm.v3">
+    <security>
+      <requestedPrivileges>
+        <!--
+          UAC settings:
+          - app should run at same integrity level as calling process
+          - app does not need to manipulate windows belonging to
+            higher-integrity-level processes
+          -->
+        <requestedExecutionLevel level="asInvoker" uiAccess="false" />
+      </requestedPrivileges>
+    </security>
+  </trustInfo>
 </assembly>
--- a/wezterm-gui/src/termwindow/mod.rs
+++ b/wezterm-gui/src/termwindow/mod.rs
@@ -355,6 +355,18 @@ pub struct TermWindow {
 }
 
 impl TermWindow {
+    fn load_os_parameters(&mut self) {
+        if let Some(ref window) = self.window {
+            self.os_parameters = match window.get_os_parameters(&self.config, self.window_state) {
+                Ok(os_parameters) => os_parameters,
+                Err(err) => {
+                    log::warn!("Error while getting OS parameters: {:#}", err);
+                    None
+                }
+            };
+        }
+    }
+
     fn close_requested(&mut self, window: &Window) {
         let mux = Mux::get().unwrap();
         match self.config.window_close_confirmation {
@@ -402,6 +414,7 @@ impl TermWindow {
     fn focus_changed(&mut self, focused: bool, window: &Window) {
         log::trace!("Setting focus to {:?}", focused);
         self.focused = if focused { Some(Instant::now()) } else { None };
+        self.load_os_parameters();
 
         if self.focused.is_none() {
             self.last_mouse_click = None;
@@ -447,7 +460,7 @@ impl TermWindow {
             }
         }
 
-        self.os_parameters = window.get_os_parameters(&self.config).unwrap_or(None);
+        self.load_os_parameters();
 
         window.show();
 
@@ -1436,14 +1449,7 @@ impl TermWindow {
         };
 
         if let Some(window) = self.window.as_ref().map(|w| w.clone()) {
-            self.os_parameters = match window.get_os_parameters(&config) {
-                Ok(os_parameters) => os_parameters,
-                Err(_) => {
-                    log::warn!("Error while getting OS parameters");
-                    None
-                }
-            };
-
+            self.load_os_parameters();
             self.apply_scale_change(&dimensions, self.fonts.get_font_scale(), &window);
             self.apply_dimensions(&dimensions, None, &window);
             window.config_did_change(&config);
--- a/wezterm-gui/src/termwindow/mod.rs
+++ b/wezterm-gui/src/termwindow/mod.rs
@@ -285,6 +285,7 @@ pub struct TermWindow {
     pub window: Option<Window>,
     pub config: ConfigHandle,
     pub config_overrides: serde_json::Value,
+    os_parameters: Option<parameters::Parameters>,
     /// When we most recently received keyboard focus
     focused: Option<Instant>,
     fonts: Rc<FontConfiguration>,
@@ -445,6 +446,8 @@ impl TermWindow {
                 log::error!("failed to create OpenGLRenderState: {}", err);
             }
         }
+
+        self.os_parameters = window.get_os_parameters(&self.config).unwrap_or(None);
 
         window.show();
 
@@ -690,6 +693,7 @@ impl TermWindow {
 
         let myself = Self {
             config_subscription: None,
+            os_parameters: None,
             gl: None,
             window: None,
             window_background,
@@ -1432,6 +1436,14 @@ impl TermWindow {
         };
 
         if let Some(window) = self.window.as_ref().map(|w| w.clone()) {
+            self.os_parameters = match window.get_os_parameters(&config) {
+                Ok(os_parameters) => os_parameters,
+                Err(_) => {
+                    log::warn!("Error while getting OS parameters");
+                    None
+                }
+            };
+
             self.apply_scale_change(&dimensions, self.fonts.get_font_scale(), &window);
             self.apply_dimensions(&dimensions, None, &window);
             window.config_did_change(&config);
--- a/window/src/os/windows/window.rs
+++ b/window/src/os/windows/window.rs
@@ -748,9 +748,7 @@ impl WindowOps for Window {
         window_state: WindowState,
     ) -> anyhow::Result<Option<Parameters>> {
         let hwnd = self.0 .0;
-        if hwnd.is_null() {
-            return Err(anyhow::anyhow!("HWND is null"));
-        }
+        anyhow::ensure!(!hwnd.is_null(), "HWND is null");
 
         let has_focus = unsafe { GetFocus() } == hwnd;
         let is_full_screen = window_state.contains(WindowState::FULL_SCREEN);
--- a/window/src/os/windows/window.rs
+++ b/window/src/os/windows/window.rs
@@ -752,7 +752,7 @@ impl WindowOps for Window {
             return Err(anyhow::anyhow!("HWND is null"));
         }
 
-        let has_focus = !unsafe { GetFocus() }.is_null();
+        let has_focus = unsafe { GetFocus() } == hwnd;
         let is_full_screen = window_state.contains(WindowState::FULL_SCREEN);
 
         // let title_font = unsafe {
--- a/window/src/os/windows/window.rs
+++ b/window/src/os/windows/window.rs
@@ -753,19 +753,16 @@ impl WindowOps for Window {
         let has_focus = unsafe { GetFocus() } == hwnd;
         let is_full_screen = window_state.contains(WindowState::FULL_SCREEN);
 
-        // let title_font = unsafe {
-        //     let hdc = GetDC(hwnd);
-        //     if hdc.is_null() {
-        //         return Err(anyhow::anyhow!("Could not get device context"));
-        //     }
-        //     let result = match get_title_log_font(hwnd, hdc) {
-        //         Some(lf) => Some(wezterm_font::locator::gdi::parse_log_font(&lf, hdc)?),
-        //         None => None,
-        //     };
-        //     ReleaseDC(hwnd, hdc);
-        //     result
-        // };
-        let title_font = None;
+        let title_font = unsafe {
+            let hdc = GetDC(hwnd);
+            anyhow::ensure!(!hdc.is_null(), "Could not get device context");
+            let result = match get_title_log_font(hwnd, hdc) {
+                Some(lf) => wezterm_font::locator::gdi::parse_log_font(&lf, hdc).ok(),
+                None => None,
+            };
+            ReleaseDC(hwnd, hdc);
+            result
+        };
 
         let hkcu = RegKey::predef(HKEY_CURRENT_USER);
         let use_accent = hkcu
--- a/window/src/os/windows/window.rs
+++ b/window/src/os/windows/window.rs
@@ -1,6 +1,6 @@
 use super::*;
 use crate::connection::ConnectionOps;
-use crate::parameters::Parameters;
+use crate::parameters::{self, Parameters};
 use crate::{
     Appearance, Clipboard, DeadKeyStatus, Dimensions, Handled, KeyCode, KeyEvent, Length,
     Modifiers, MouseButtons, MouseCursor, MouseEvent, MouseEventKind, MousePress, Point,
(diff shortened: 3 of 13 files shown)


window/src/os/windows/window.rs
```diff
@@ -815,6 +815,104 @@ unsafe fn wm_ncdestroy(
...
+            return None;
+        }
+
+        if inner.saved_placement.is_none() {
+            let dpi = GetDpiForWindow(hwnd);
+            let frame_x = GetSystemMetricsForDpi(SM_CXFRAME, dpi);
+            let frame_y = GetSystemMetricsForDpi(SM_CYFRAME, dpi);
+            let padding = GetSystemMetricsForDpi(SM_CXPADDEDBORDER, dpi);
+
+            let params = (_lparam as *mut NCCALCSIZE_PARAMS).as_mut().unwrap();
```
wez, line 837:
```suggestion
            let params = (lparam as *mut NCCALCSIZE_PARAMS).as_mut().unwrap();
```

window/src/os/windows/window.rs
```diff
@@ -815,6 +815,104 @@ unsafe fn wm_ncdestroy(
     None
 }
 
+unsafe fn wm_nccalcsize(
+    hwnd: HWND,
+    _msg: UINT,
+    _wparam: WPARAM,
+    _lparam: LPARAM,
```
wez, line 822:
```suggestion
    lparam: LPARAM,
```

window/src/os/windows/window.rs
```diff
@@ -815,6 +815,104 @@ unsafe fn wm_ncdestroy(
...
+) -> Option<LRESULT> {
+    if let Some(inner) = rc_from_hwnd(hwnd) {
+        let inner = inner.borrow_mut();
+
+        if !(_wparam == 1 && inner.config.window_decorations == WindowDecorations::RESIZE) {
```
wez, line 827:
```suggestion
        if !(wparam == 1 && inner.config.window_decorations == WindowDecorations::RESIZE) {
```

window/src/os/windows/window.rs
```diff
@@ -815,6 +815,104 @@ unsafe fn wm_ncdestroy(
     None
 }
 
+unsafe fn wm_nccalcsize(
+    hwnd: HWND,
+    _msg: UINT,
+    _wparam: WPARAM,
```
wez, line 821: The underscore prefix is used to indicate unused parameters, but these parameters are actually used!

```suggestion
    wparam: WPARAM,
```

window/src/os/windows/window.rs
```diff
@@ -815,6 +815,104 @@ unsafe fn wm_ncdestroy(
...
+    }
+
+    None
+}
+
+unsafe fn wm_nchittest(
+    hwnd: HWND,
+    _msg: UINT,
+    _wparam: WPARAM,
+    _lparam: LPARAM,
```
wez, line 860:
```suggestion
    msg: UINT,
    wparam: WPARAM,
    lparam: LPARAM,
```

window/src/os/windows/window.rs
```diff
@@ -815,6 +815,104 @@ unsafe fn wm_ncdestroy(
...
+) -> Option<LRESULT> {
+    if let Some(inner) = rc_from_hwnd(hwnd) {
+        let inner = inner.borrow_mut();
+
+        if inner.config.window_decorations != WindowDecorations::RESIZE {
+            return None;
+        }
+
+        // Let the default procedure handle resizing areas
+        let result = DefWindowProcW(hwnd, _msg, _wparam, _lparam);
```
wez, line 870:
```suggestion
        let result = DefWindowProcW(hwnd, msg, wparam, lparam);
```

window/src/os/windows/window.rs
```diff
@@ -815,6 +815,104 @@ unsafe fn wm_ncdestroy(
...
+            }
+            _ => {}
+        }
+
+        // The adjustment in NCCALCSIZE messes with the detection
+        // of the top hit area so manually fixing that.
+        let dpi = GetDpiForWindow(hwnd);
+        let frame_y = GetSystemMetricsForDpi(SM_CYFRAME, dpi);
+
+        let cursor_point = MAKEPOINTS(_lparam as u32);
```
wez, line 885:
```suggestion
        let cursor_point = MAKEPOINTS(lparam as u32);
```

window/src/os/windows/window.rs
```diff
@@ -815,6 +815,104 @@ unsafe fn wm_ncdestroy(
...
+        // Let the default procedure handle resizing areas
+        let result = DefWindowProcW(hwnd, _msg, _wparam, _lparam);
+
+        match result {
+            HTNOWHERE | HTRIGHT | HTLEFT | HTTOPLEFT | HTTOP | HTTOPRIGHT | HTBOTTOMRIGHT
+            | HTBOTTOM | HTBOTTOMLEFT => {
+                return Some(result);
+            }
+            _ => {}
+        }
```
wez, line 878:
```suggestion
        if matches!(result,
            HTNOWHERE | HTRIGHT | HTLEFT | HTTOPLEFT | HTTOP | HTTOPRIGHT | HTBOTTOMRIGHT
            | HTBOTTOM | HTBOTTOMLEFT) {
                return Some(result);
        }
```

wez (changes requested):
Thanks for diving in to this!
I have a few stylistic suggestions, but I think the code looks good.  I'll try it out a bit later!

window/src/os/windows/window.rs
davidrios: Oh, that's right, I forgot `_` is to signal the variable is not used. I'm still new to Rust and my editor is not working well with this repo.

davidrios:
I made a new one that supersedes this one with more features, but the code for them is more dubious: #1677

wez:
This looks good, but one thing feels a bit jarring; the top border is now invisible, but the other sides are not.

before:
image

after:
image

chrome:
image

Is there a tweak to get the top border to be consistent with the bottom border size?

Regarding the other PR: I'm interested in following up, but would like to land this one first so that those additional changes are smaller and can be tackled separately!

davidrios:
That's an issue discussed in the referenced article:

> One thing you probably notice is that the top "border" is missing.
> ... Since I knew I saw it working for some custom-themed applications I use like Google Chrome and IntelliJ I tried to look closer at how they do it. Turns out they fake it!
> Even Microsoft-developed Edge browser fakes this. It does show that there is probably no real solution if Microsoft's own engineers resort to hacks.

Basically you have to draw that border manually, simulating the three others 😂. I tested drawing with GDI like the article, but I got lots of artifacts and flickering when resizing the window, so I'm not sure how to do it. I can give it a go if you have any suggestion.

wez:
Hmm, that's unfortunate!

In https://github.com/wez/wezterm/pull/1677#discussion_r817705800 I just mentioned that the window crate could expose information about placement and rendering of system UI things; I was thinking primarily of the min/max/close buttons, but I suppose it could also extend to an optional top border edge.

With that info provided, the wezterm-gui layer could render the border as part of its normal paint routine.

I think there might be some annoying math/transformation issues: the top border height should impact the usable client area calculation (which in turn affects the number of cell rows available) and adjust the y pixel values used throughout rendering.

davidrios:
Like this?
https://github.com/davidrios/wezterm/commit/3c0422802f2968e450fcecd3f92f83f0e4d0bc1d

davidrios:
They fixed the top bar issue on Windows 11, lol, and they trim the top pixel line from the client area, presumably to compensate for programs that drew their fake border and to not clash with the border drawn by the system.

wez:
> Something like this POC? [davidrios@3c04228](https://github.com/davidrios/wezterm/commit/3c0422802f2968e450fcecd3f92f83f0e4d0bc1d)

I left some comments over on that commit: yeah, I like the look of the code!

One thing that I think we might need to accommodate for this extra pixel is adjusting the render positioning.
Consider an extreme example where the discrepancy in the border is something like 32 pixels in height.
With your changes we can now render the border in the right place, but the window contents will be in the wrong y position.

I think we may want one set of dimensions for the border that we need to render, and another set that affects our calculations of where to render the content.

Look for `top_pixel_y`, `left_pixel_x` and `pixel_width`  in `RenderScreenLineOpenGLParams` to see where we compute coordinates: I think we want to adjust those by some OS dependent amount that may be different from the sizes that we draw for the border.  The tab bar positioning logic should also adjust for that.

window/src/os/windows/window.rs
```diff
@@ -989,10 +1241,12 @@ unsafe fn wm_kill_focus(
...
 ) -> Option<LRESULT> {
     if let Some(inner) = rc_from_hwnd(hwnd) {
+        let mut inner = inner.borrow_mut();
+        inner.events.dispatch(WindowEvent::FocusChanged(false));
+        let appearance = inner.appearance;
         inner
-            .borrow_mut()
             .events
-            .dispatch(WindowEvent::FocusChanged(false));
+            .dispatch(WindowEvent::AppearanceChanged(appearance));
```
wez, line 1249: This causes the config to be reloaded each time the window focus is changed, which results in a noticeable delay in responsiveness when focusing the window and for the bottom window border color to lag behind the others.

To resolve this:

* Remove the `AppearanceChanged` dispatch from here and from `wm_set_focus` above
* Add a call to fetch the OS parameters directly in `TermWindow::focus_changed`

window/src/os/windows/window.rs
```diff
@@ -716,72 +735,129 @@ impl WindowOps for Window {
...
+        let hwnd = self.0 .0;
+        if hwnd.is_null() {
+            return Err(anyhow::anyhow!("HWND is null"));
+        }
 
-            let res = GetThemeSysFont(theme, TMT_CAPTIONFONT, &mut log_font);
-            if !theme.is_null() {
-                CloseThemeData(theme);
-            }
+        let has_focus = !unsafe { GetFocus() }.is_null();
```
wez, line 744: I think this may produce incorrect results when there are multiple windows.  How about this instead?

```suggestion
        let has_focus = unsafe { GetFocus() } == hwnd;
```

window/src/os/windows/window.rs
```diff
@@ -417,6 +431,7 @@ impl Window {
             dead_pending: None,
             saved_placement: None,
             track_mouse_leave: false,
+            last_resize_type: SIZE_RESTORED,
```
wez, line 434: Not strictly related to this PR, but https://github.com/wez/wezterm/issues/1502 is probably caused by the `nCmdShow` parameter in `WinMain` not being respected when launched via a shortcut with the maximized state.

I don't know (haven't looked) to see if there is a smarter way to know when we want to do that or apply it.

Anyway: doesn't need changes here in this PR, but wanted to float that by you in case you have insights on what to do about it :)

window/src/os/windows/window.rs
```diff
@@ -569,6 +584,10 @@ impl WindowInner {
                 })
                 .detach();
             }
+            promise::spawn::spawn(async move {
+                PostMessageW(hwnd, extra_constants::UM_APPEARANCE_CHANGED, 0, 0);
+            })
+            .detach();
```
wez, line 590: Similar to my comments on the focus state changes, I think this is a bit heavyweight to do when toggling fullscreen: it should remain nice and snappy.

What I suggest is removing this and adjusting `TermWindow::resize` to re-fetch the OS parameters if `self.window_state != window_state`.  I think that captures the intent of propagating the border change when full screen mode is toggled.

```suggestion
```

window/src/os/windows/window.rs
```diff
@@ -716,72 +735,129 @@ impl WindowOps for Window {
...
-                if res == S_OK {
-                    CloseThemeData(theme);
-                    return Some(log_font);
-                }
-            }
+    fn get_os_parameters(&self, config: &ConfigHandle) -> anyhow::Result<Option<Parameters>> {
+        let hwnd = self.0 .0;
+        if hwnd.is_null() {
+            return Err(anyhow::anyhow!("HWND is null"));
+        }
```
wez, line 742:
```suggestion
        anyhow::ensure!(!hwnd.is_null(), "HWND is null");
```

window/src/os/windows/window.rs
```diff
@@ -716,72 +735,129 @@ impl WindowOps for Window {
...
-            let hwnd = self.0 .0;
+        };
+
+        let title_font = unsafe {
             let hdc = GetDC(hwnd);
-            let result = match populate_log_font(hwnd, hdc) {
-                Some(lf) => wezterm_font::locator::gdi::parse_log_font(&lf, hdc).ok(),
+            if hdc.is_null() {
+                return Err(anyhow::anyhow!("Could not get device context"));
+            }
```
wez, line 771:
```suggestion
            anyhow::ensure!(!hdc.is_null(), "Could not get device context");
```

window/src/os/windows/window.rs
```diff
@@ -716,72 +735,129 @@ impl WindowOps for Window {
...
+            let result = match get_title_log_font(hwnd, hdc) {
+                Some(lf) => Some(wezterm_font::locator::gdi::parse_log_font(&lf, hdc)?),
```
wez, line 773: The `?` can cause us to return from here without releasing `hdc` and leak it.

I'd suggest going back to the prior implementation of this bit; I don't think there's a ton of value in propagating an error here, as the code already handles the `None` case.

window/src/os/windows/window.rs
```diff
@@ -716,72 +735,129 @@ impl WindowOps for Window {
...
+        let use_accent = hkcu
+            .open_subkey("SOFTWARE\\Microsoft\\Windows\\DWM")?
+            .get_value::<u32, _>("ColorPrevalence")?;
+        let settings = UISettings::new()?;
+        let top_border_color = if has_focus {
+            if use_accent == 1 {
+                wuicolor_to_linearrgba(settings.GetColorValue(UIColorType::Accent)?)
+            } else {
+                if is_win10() {
+                    LinearRgba(0.01, 0.01, 0.01, 0.67)
```
wez, line 790: These hard coded numbers are a bit unfortunate.
Can we get the colors we need from https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getsystemmetrics instead?

window/src/os/windows/window.rs
```diff
@@ -817,6 +893,152 @@ unsafe fn wm_ncdestroy(
...
+    };
+
+    if unsafe { GetWindowPlacement(hwnd, &mut placement) } != winapi::shared::minwindef::TRUE {
+        false
+    } else {
+        placement.showCmd == SW_SHOWMAXIMIZED as u32
+    }
+}
+
+fn is_win10() -> bool {
```
wez, line 1029: To avoid making these calls repeatedly, I think we can cache this in a global and then the rest of the code can consult `IS_WIN10`.

Or alternatively, rename `is_win10` to `is_win10_impl` and then make `is_win10` return `IS_WIN10` if you'd prefer to avoid having lots of shouty case.

```suggestion
lazy_static! {
  static ref IS_WIN10: bool = is_win10();
}  

fn is_win10() -> bool {
```

window/src/os/windows/window.rs
```diff
@@ -947,6 +1180,23 @@ unsafe fn wm_windowposchanged(
     Some(0)
 }
 
+unsafe fn wm_sizeraw(hwnd: HWND, _msg: UINT, wparam: WPARAM, _lparam: LPARAM) -> Option<LRESULT> {
```
wez, line 1183: I wonder if we can eliminate this?

`inner.last_resize_type` is effectively the same thing as `inner.appearance`, so I think it would be nice to remove that redundancy, but if we can't it would be good to have some comments in the code that explain the nuance.

If we can remove tracking `last_resize_type`, then the change I suggested for `TermWindow::resize` would allow removing the dispatch of `AppearanceChanged` here as well.

wezterm-gui/src/termwindow/mod.rs
```diff
@@ -1432,6 +1436,14 @@ impl TermWindow {
         };
 
         if let Some(window) = self.window.as_ref().map(|w| w.clone()) {
+            self.os_parameters = match window.get_os_parameters(&config) {
+                Ok(os_parameters) => os_parameters,
+                Err(_) => {
+                    log::warn!("Error while getting OS parameters");
```
wez, line 1442: It's almost always a good idea to include context in error messages!

```suggestion
                Err(err) => {
                    log::warn!("Error while getting OS parameters: {:#}", err);
```

assets/windows/manifest.manifest
```diff
@@ -1,9 +1,43 @@
...
   <asmv3:application>
     <asmv3:windowsSettings xmlns="http://schemas.microsoft.com/SMI/2016/WindowsSettings">
       <dpiAwareness>PerMonitorV2</dpiAwareness>
     </asmv3:windowsSettings>
   </asmv3:application>
+  <dependency>
+    <dependentAssembly>
+      <assemblyIdentity
+        type="win32"
+        name="Microsoft.Windows.Common-Controls"
```
wez, line 16: this is a little surprising to see here: do we need this dependency listed?

wez (changes requested):
Thanks for this: this is looking really good and we're really close!

I have a few suggesestions: the main thing is avoiding triggering a config reload in some situations that should remain feeling fast.

I know that I said that we could for the most part concentrate the os parameter fetching to config reload events in an earlier round of comments.  The context there was to avoid the relatively expensive font parsing on each paint call.  For resize related events I think it is OK to retrieve the os parameters when we realize that we need to.

I'm a little cagey about that logfont load/parsing in the context of a live resize though, and I think it might be worth splitting that portion out and separately caching it, so that when the overall set of os parameters are retrieved, we can simply copy/reference the already parsed font information.

That feels like more work than is perhaps needed, so maybe we could split the font info out from the rest of the parameters?  What do you think?

assets/windows/manifest.manifest
davidrios: It's not *needed*, but I guess it's good practice anyway.

davidrios:
> I'm a little cagey about that logfont load/parsing in the context of a live resize though, and I think it might be worth splitting that portion out and separately caching it, so that when the overall set of os parameters are retrieved, we can simply copy/reference the already parsed font information.
> 
> That feels like more work than is perhaps needed, so maybe we could split the font info out from the rest of the parameters? What do you think?

Is it acceptable to cache it in a global static,like your suggestion for the is_win10?

window/src/os/windows/window.rs
davidrios: No, they aren't available there. Win32 was basically deprecated by Microsoft in favor of UWP, so lots of things there don't work on Windows 10+

window/src/os/windows/window.rs
davidrios: I tested and only got problems if I had a configuration setting initial_cols/initial_rows. Anyway, I've made a change in `check_and_call_resize_if_needed` that solves it. I think it's supposed to be that way.

wez:
I rebased this locally and pushed it; many thanks! I really appreciate your work here!

PR #1676, opened 2022-02-28
fix: tmux -CC capturep with extra \n

Dixeran:
If the original content is

```
wezterm▊ // <- cursor right after the word
```

then the returned content of `capturep` command will be 

```
%begin
wezterm // <- implicit \n
%end
```

thus the rendered content will be

```
wezterm
▊
```

### DIFF
--- a/mux/src/tmux_commands.rs
+++ b/mux/src/tmux_commands.rs
@@ -308,12 +308,14 @@ impl TmuxCommand for CapturePane {
 
         let unescaped = termwiz::tmux_cc::unvis(&result.output).context("unescape pane content")?;
         let unescaped = unescaped.replace("\n", "\r\n");
+        // capturep contents returned from guarded lines which always contain a tailing '\n'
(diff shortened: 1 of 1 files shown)


mux/src/tmux_commands.rs
```diff
@@ -308,12 +308,14 @@ impl TmuxCommand for CapturePane {
 
         let unescaped = termwiz::tmux_cc::unvis(&result.output).context("unescape pane content")?;
         let unescaped = unescaped.replace("\n", "\r\n");
+        // capturep contents returned from guarded lines which always contain a tailing '\n'
+        let trim_unescaped = &unescaped[0..unescaped.len() - 1];
```
wez, line 312: I think this should probably happen before the `\n` is rewritten into `\r\n` so that we don't leave a stray `\r` on the end of this string?

Also, it would be good to make this safer if we end up with unescaped as an empty string; `unescaped.len().saturating_sub(1)` won't wrap to a really large number in that case.

wez:
Thanks!

PR #1677, opened 2022-02-28
Improve RESIZE window decorations on Windows (more features)

davidrios:
The problems are described in #1674.

This change fixes all of them. The code is what's generally done to make an app with a custom title bar, with tabs like Firefox for instance:

- It handles `nccalcsize`, so that you draw your own borders/window decorations.
- It handles `nchittest` to deal with mouse input in your custom title bar.

In addition to #1675, this one also makes the empty area of the tab bar act as a native title bar, with features:

- Double click to max/minimize
- Drag to corners of the screen to snap/maximize
- Drag when maximized to restore original size
- Shake while dragging to minimize/restore other windows
- Right click to open the context menu

This is based on this article: https://kubyshkin.name/posts/win32-window-custom-title-bar-caption/

I hacked a way to check if the mouse is over the empty area of the tab bar, it's unclear if that's a good enough solution or if some other plumbing should be implemented to enable that.

### DIFF
--- a/window/src/lib.rs
+++ b/window/src/lib.rs
@@ -64,6 +64,12 @@ pub enum MouseCursor {
     SizeLeftRight,
 }
 
+pub enum TabBarItem {
+    None,
+    Tab { tab_idx: usize, active: bool },
+    NewTabButton,
+}
+
 /// Represents the preferred appearance of the windowing
 /// environment.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
@@ -283,4 +289,6 @@ pub trait WindowOps {
     fn get_title_font_and_point_size(&self) -> Option<(wezterm_font::parser::ParsedFont, f64)> {
         None
     }
+
+    fn mouse_event_tab_bar(&self, _item: TabBarItem, _event: MouseEvent) {}
 }
--- a/window/src/os/windows/window.rs
+++ b/window/src/os/windows/window.rs
@@ -863,15 +863,15 @@ unsafe fn wm_nchittest(
         let mut inner = inner.borrow_mut();
 
         if inner.config.window_decorations != WindowDecorations::RESIZE {
-            return None
+            return None;
         }
 
         // Let the default procedure handle resizing areas
         let result = DefWindowProcW(hwnd, _msg, _wparam, _lparam);
 
         match result {
-            HTNOWHERE | HTRIGHT | HTLEFT | HTTOPLEFT | HTTOP | HTTOPRIGHT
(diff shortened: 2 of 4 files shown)


window/src/lib.rs
```diff
@@ -64,6 +64,12 @@ pub enum MouseCursor {
     SizeLeftRight,
 }
 
+pub enum TabBarItem {
```
wez, line 67: I don't feel great about introducing a tab bar item type here, because the `window` crate doesn't really do anything with tabs.

See my other comment: I think we can eliminate this from this PR!

window/src/os/windows/window.rs
```diff
@@ -1093,6 +1279,53 @@ fn apply_mouse_cursor(cursor: Option<MouseCursor>) {
     }
 }
 
+unsafe fn check_title_rmouse(hwnd: HWND, msg: UINT, screen_coords: ScreenPoint) -> Option<LRESULT> {
```
wez, line 1282: For https://github.com/wez/wezterm/issues/1485 I've been thinking of teaching the wezterm-gui layer how to produce its own context menu windows (I don't *really* want to do my own, but the native context menus generally block the rest of the GUI until a selection is made which poses some implementation challenges on various OSs).

In that world, the choice to show and what to show are all handled in the wezterm-gui layer rather than in the window layer.

My semi-formed thoughts around this:

* The title bar area has some ideal locations for the minimize, maximize and close buttons (on the RHS on Windows, on the LHS on macOS, and not prevailing choice on linux), so it would make sense for the window crate to expose positioning information to the application layer so that the fancy tab bar would make space for them and/or have a way to trigger rendering them.  That doesn't need to happen here in this PR, I'm just sharing what I've been thinking.
* The decision about when to show a context menu should really be done in the application layer, which knows where its UIItems reside
* We could add a `WindowOps::show_windows_system_context_menu` function that does the `GetSystemMenu`/`TrackPopupMenuEx` bit of this function, and then teach `mouse_event_tab_bar` to call it for right clicks on `TabBarItem::None` as a way to get this functionality working in the meantime, and then ultimately replace it with the full blown context menu feature once that is done

davidrios: Good points. I'll implement like your suggestion from the last 2 items.

As for the first one, do you mean to render native window control buttons? I'm not sure how that would work on other OSes, but on Windows there's no API to make the system draw the window control buttons, you have to draw them yourself. It's funny, because after knowing that, I started noticing that indeed every app that has a custom title bar does exactly that. They try to make it similar to the native Windows 10 one, but each one is slightly different and have slightly different behaviors. 😂

There's actually a new way, but it's only available for UWP/Windows 11 apps.

PR #1678, opened 2022-03-01
fix: tmux -CC sync cursor position after attach

Dixeran:
This combine with #1676 should make the cursor position correct after attaching to an existing tmux session. Not sure if it's fine to expose the `set_cursor_pos` method from `TerminalState` and add a `set_cursor_position` method to the `Pane` trait (which is default empty implemented).


wez:
I'm not sure about this change!

I think the tmux integration should probably move to look a bit more like the wezterm multiplexer client implementation and have separate Pane/Renderable implementations that know how to retrieve the appropriate lines from the viewport/scrollback and report the current cursor position.  I'm mostly about that to expose the historical scrollback without forcing the client to parse all of it on startup.

I think it will be a little bit tricky to get that started, but I think in the long term it will make it easier to support/implement!

PR #1679, opened 2022-03-02
Detect mouse leaving the window

### LINKED ISSUE #1434: Disable inactive_tab_hover style when not hovering

Currently the `inactive_tab_hover` style sticks even when mouse is not hovering over, which isn't great. It only seems to reset when you hover on a different tab or type something

`inactive_tab_hover` should only be active when the mouse cursor is actually hovering over the tab and reset to the default style when hovering is over

None

**Additional context**
None

davidrios:
Detect when mouse leaves the window, for instance to clear the hover state of the items in the tab bar.

This fixes #1434 on Windows, Wayland, X11 and macOS

### DIFF
--- a/window/src/os/windows/window.rs
+++ b/window/src/os/windows/window.rs
@@ -1172,17 +1172,7 @@ unsafe fn mouse_leave(hwnd: HWND, _msg: UINT, _wparam: WPARAM, _lparam: LPARAM) 
         let mut inner = inner.borrow_mut();
 
         inner.track_mouse_leave = false;
-
-        let coords = Point::new(-1, -1);
-        let event = MouseEvent {
-            kind: MouseEventKind::Move,
-            coords,
-            screen_coords: client_to_screen(hwnd, coords),
-            mouse_buttons: MouseButtons::default(),
-            modifiers: Modifiers::default(),
-        };
-
-        inner.events.dispatch(WindowEvent::MouseEvent(event));
+        inner.events.dispatch(WindowEvent::MouseLeave);
 
         Some(0)
     } else {
--- a/window/src/os/windows/window.rs
+++ b/window/src/os/windows/window.rs
@@ -61,6 +61,7 @@ pub(crate) struct WindowInner {
     in_size_move: bool,
     dead_pending: Option<(Modifiers, u32)>,
     saved_placement: Option<WINDOWPLACEMENT>,
+    track_mouse_leave: bool,
 
     keyboard_info: KeyboardLayoutInfo,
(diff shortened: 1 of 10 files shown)


window/src/os/windows/window.rs
```diff
@@ -1143,10 +1160,30 @@ unsafe fn mouse_move(hwnd: HWND, _msg: UINT, wparam: WPARAM, lparam: LPARAM) ->
...
+    }
+}
+
+unsafe fn mouse_leave(hwnd: HWND, _msg: UINT, _wparam: WPARAM, _lparam: LPARAM) -> Option<LRESULT> {
+    if let Some(inner) = rc_from_hwnd(hwnd) {
+        let mut inner = inner.borrow_mut();
+
+        inner.track_mouse_leave = false;
+
+        let coords = Point::new(-1, -1);
```
wez, line 1176: Could we get the actual mouse screen coordinates and compute the effective client coordinates here instead?

I'm thinking about future features such as dragging and dropping tabs and I think it might be surprising if this event interleaved and made it appear as though the mouse suddenly moved off screen.

davidrios: I thought about that too, so I did some testing. Turns out the mouse leave event is not fired when the mouse leaves the window while captured, and it only fires after all mouse events are done, so it's pretty safe. I did that `-1` as a hacky way to signal that the mouse is out of the window, but I think it would be better if there's a specific API for that.

wez: I think using [GetCursorPos](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getcursorpos) to get the screen coordinates and then translating those to window coordinates feels more robust overall

davidrios: But then again, that would imply Windows-specific code. An API to signal mouse out of the window, which is independent of mouse position, would likely be useful to all supported OSes. Maybe a `WindowEvent::LeaveWindow` or something similar, as opposed to a `MouseEvent`, because in this case all other mouse related stuff are irrelevant.

davidrios: I've implemented my suggested API and used it to fix the bug on Windows, Wayland and X11.

wez:
This looks great!  Did you test this on all systems?

davidrios:
Yeah, tested on them all. I couldn't reproduce the issue on macOS, but implemented the logic anyway, for the sake of completeness. Maybe it only affects newer versions, I could only test on Catalina, since it was the latest version I could hack into(sh) my laptop.

PR #1689, opened 2022-03-07
Use newer windows crate

davidrios:
Upgrade the windows crate version to 0.33.0.

This is in anticipation of its use in another PR, and also it's future-proof.


wez:
Thanks!

PR #1708, opened 2022-03-09
Support primary selection on Wayland

### LINKED ISSUE #1423: Copy/paste to PrimarySelection under wayland
The ability to interact with the PrimarySelection under wayland.

Using this the [primary-selection-unstable-v1](https://wayland.app/protocols/primary-selection-unstable-v1) to interact with it, if the wayland server supports this protocoll

lunaryorn:
Adds support for primary selection under wayland, using the protocol wrappers in [smithay_client_toolkit](https://docs.rs/smithay-client-toolkit/latest/smithay_client_toolkit/primary_selection/index.html), which support both existing protocols for primary selection (the older Gtk protocol and the "official" one).

I'd love to see this merged because I'm using primary selection a lot.  I'd appreciate a thorough review though as I'm not very familiar with Wayland.

Closes #1423

### DIFF
--- a/window/src/os/wayland/copy_and_paste.rs
+++ b/window/src/os/wayland/copy_and_paste.rs
@@ -1,11 +1,17 @@
-use anyhow::{anyhow, Error};
+use anyhow::{anyhow, bail, Context, Error};
 use filedescriptor::{FileDescriptor, Pipe};
 use smithay_client_toolkit as toolkit;
+use std::io::Write;
 use std::os::unix::io::AsRawFd;
+use std::os::unix::prelude::{FromRawFd, IntoRawFd};
 use std::sync::{Arc, Mutex};
+use toolkit::primary_selection::*;
 use toolkit::reexports::client::protocol::wl_data_offer::{Event as DataOfferEvent, WlDataOffer};
-use toolkit::reexports::client::protocol::wl_data_source::WlDataSource;
-use wayland_client::Attached;
+use wayland_client::protocol::wl_data_device_manager::WlDataDeviceManager;
+use wayland_client::protocol::wl_data_source::Event as DataSourceEvent;
+
+use crate::connection::ConnectionOps;
+use crate::Clipboard;
 
 #[derive(Default)]
 pub struct CopyAndPaste {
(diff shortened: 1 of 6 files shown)


wez:
Thanks for this!
The code looks plausible :)
@unrelentingtech: would you mind taking a look at this?

window/src/os/wayland/window.rs
```diff
@@ -867,12 +868,31 @@ impl WindowOps for WaylandWindow {
...
-            let read = inner.copy_and_paste.lock().unwrap().get_clipboard_data()?;
+            let conn = Connection::get().unwrap().wayland();
+            let selection = if let Clipboard::PrimarySelection = clipboard {
+                conn.primary_selection.as_ref()
+            } else {
+                None
+            };
+            let read = match selection {
+                None => inner.copy_and_paste.lock().unwrap().get_clipboard_data(),
+                Some(selection) => {
```
wez, line 884: is it possible to move some of the more verbose parts of the getting/setting code into the `CopyAndPaste` struct?
I feel like `PrimarySelection` could potentially merge with that struct and that `get_clipboard_data` could be passed the `Clipboard` type and internally do the right thing

lunaryorn: Sure, will do.  I'll see if I can find a bit of time for this over the next days 🙂

unrelentingtech:
Yeah, other than the suggestion to reduce duplication, looks good

window/src/os/wayland/window.rs
lunaryorn: @wez I've moved things around a bit.  There's still a bit of duplication because the clipboard and primary selection types are different, and it also looks as if the data device code for clipboard access directly uses wayland protocol types whereas the new primary selection code uses some higher-level interface from the wayland toolkit library, which uses a slight different flow to access the selection offer and to create the selection source.

I've pushed it as a separate commit to be able to go back to the earlier version if you have objections; if you agree with the latest change I can squash it all together and clean up.

window/src/os/wayland/copy_and_paste.rs
```diff
@@ -35,14 +41,83 @@ impl CopyAndPaste {
...
+        };
+
+        match primary_selection {
+            Some((manager, device)) => {
+                let source = PrimarySelectionSource::new(
+                    &manager,
+                    &[TEXT_MIME_TYPE.to_string()],
+                    move |event, _dispatch_data| match event {
+                        PrimarySelectionSourceEvent::Cancelled => {
+                            unimplemented!();
```
wez, line 93: this will crash the program if this ever triggers: can we log a warning and NOP here instead?

wez (approved):
I think this looks good, thanks!  I'd like to see the `unimplemented!` bit go away, but that should be a minor change.

window/src/os/wayland/copy_and_paste.rs
lunaryorn: Ah yeah, actually I wanted to implement this 🙈 Sorry totally forgot this when reshuffling all the code.

Thanks for catching this, I'll push an update to this PR.

wez:
One more thing: would you mind updating https://github.com/wez/wezterm/blob/main/docs/config/lua/keyassignment/CopyTo.md and https://github.com/wez/wezterm/blob/main/docs/config/lua/keyassignment/PasteFrom.md to reflect these changes?

I'd suggest something like adding this at the bottom of the page:

```
*Since: nightly builds only*

`PrimarySelection` is now also supported on Wayland systems that support  [primary-selection-unstable-v1](https://wayland.app/protocols/primary-selection-unstable-v1)
```

lunaryorn:
Sure will do.  Should I also mention that it also supports the older proprietary Gtk protocol, too?  I'm not sure how relevant that is, though. it looks as if it was removed from most compositors about a year ago…

Given that the protocol is part of wlroots since about 2018, in KDE since about 2020, in Gnome as well for probably some time as well, does it even make sense to mention the protocol?  It's likely supported on all major compositors by now :thinking:

wez:
definitely worth mentioning IMO!

window/src/os/wayland/copy_and_paste.rs
lunaryorn: Added the missing piece :slightly_smiling_face:

lunaryorn:
I've also added documentation.

wez:
Thanks!

lunaryorn:
Thank you, it was a pleasure :slightly_smiling_face:

lunaryorn:
@wez Just one thing: Would you mind to cut a release so that Arch can update its package?  I know there's the nightly PKGBUILD but if it's all the same to you I really prefer to have wezterm straight out of the standard repos :innocent:

wez:
The next release is gated on tidying up some things that are being tracked in https://github.com/wez/wezterm/milestone/3

lunaryorn:
Thanks for the new release 🙏. Much obliged.

PR #1712, opened 2022-03-11
Add context menu extension for Nautilus

### LINKED ISSUE #1092: Add integration to Nautilus file manager
`cd` to deep directory structure is hard.

Add right click context menu to open wezterm using current directory from Nautilus.

[..]

**Additional context**
Example from [Tilix](https://github.com/gnunn1/tilix).
nautilus-context-menu

lunaryorn:
Closes #1092

### DIFF
--- a/assets/wezterm-nautilus.py
+++ b/assets/wezterm-nautilus.py
@@ -14,6 +14,7 @@
 # with this program; if not, write to the Free Software Foundation, Inc.,
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 
+import os.path
 
 from gi import require_version
 require_version('Nautilus', '3.0')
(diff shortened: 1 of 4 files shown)


unrelentingtech:
I don't think this should have a hard dependency on systemd.

ci/deploy.sh
```diff
@@ -156,6 +156,7 @@ install -Dm644 assets/shell-integration/* -t %{buildroot}/etc/profile.d
 install -Dm644 assets/icon/terminal.png %{buildroot}/usr/share/icons/hicolor/128x128/apps/org.wezfurlong.wezterm.png
 install -Dm644 assets/wezterm.desktop %{buildroot}/usr/share/applications/org.wezfurlong.wezterm.desktop
 install -Dm644 assets/wezterm.appdata.xml %{buildroot}/usr/share/metainfo/org.wezfurlong.wezterm.appdata.xml
+install -Dm644 assets/wezterm-nautilus.py %{buildroot}/usr/share/nautilus-python/extensions/wezterm-nautilus.py
```
wez, line 159: I think you need to add `/usr/share/nautilus-python/extensions/wezterm-nautilus.py` to the `%files` list below

assets/wezterm-nautilus.py
```diff
@@ -0,0 +1,86 @@
...
+from gi import require_version
+require_version('Nautilus', '3.0')
+from gi.repository import Nautilus, GObject, Gio, GLib
+
+
+class OpenInWezTermAction(GObject.GObject, Nautilus.MenuProvider):
+    def __init__(self):
+        super().__init__()
+        session = Gio.bus_get_sync(Gio.BusType.SESSION, None)
+        self._systemd = Gio.DBusProxy.new_sync(session, Gio.DBusProxyFlags.NONE,
```
wez, line 27: I think it would be good to not require systemd here.  Assuming that this proxy creation fails if systemd is not present, then I'd suggest wrapping this in a `try`/`catch`, then in `_open_terminal` below use `subprocess` to spawn the terminal if `self_systemd is None`.

Would be good to verify that the subprocess branch of the code actually works! :-)

lunaryorn: I've added a check for systemd before using the systemd API.

ci/deploy.sh
lunaryorn: Fixed :slightly_smiling_face:

wez:
Thanks!

PR #1767, opened 2022-03-26
Fix GitHubDark cursor fg color

ModProg:
Otherwise, it is white on light gray and very much not readable.

### DIFF
--- a/assets/colors/GitHub Dark.toml
+++ b/assets/colors/GitHub Dark.toml
@@ -4,7 +4,7 @@ foreground = "#8b949e"
 background = "#101216"
 cursor_bg = "#c9d1d9"
 cursor_border = "#c9d1d9"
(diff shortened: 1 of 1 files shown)


wez:
That file is synchronized from data in https://github.com/mbadolato/iTerm2-Color-Schemes and this edit will be lost when I next re-sync.  Can you get a PR in over there first, and then we can sync it back?

ModProg:
> Can you get a PR in over there first, and then we can sync it back?

https://github.com/mbadolato/iTerm2-Color-Schemes/pull/356

ModProg:
It's merged now: https://github.com/mbadolato/iTerm2-Color-Schemes/pull/357

wez:
Thanks!  I've re-synced!

PR #1771, opened 2022-03-26
Implement native window dragging on Windows

davidrios:
Implement native window dragging by the tabbar on Windows, when window decoration = `RESIZE`, so things like drag to snap and double click to max/minimize works.

Also prevent dragging by the tabbar when the window is maximized (and decoration is not `RESIZE`) or in full screen, since those cause undesirable behavior.

### DIFF
--- a/wezterm-gui/src/termwindow/mouseevent.rs
+++ b/wezterm-gui/src/termwindow/mouseevent.rs
@@ -3,6 +3,7 @@ use crate::termwindow::keyevent::window_mods_to_termwiz_mods;
 use crate::termwindow::{PositionedSplit, ScrollHit, UIItem, UIItemType, TMB};
 use ::window::{
     MouseButtons as WMB, MouseCursor, MouseEvent, MouseEventKind as WMEK, MousePress, WindowOps,
+    WindowState,
 };
 use config::keyassignment::{MouseEventTrigger, SpawnTabDomain};
 use mux::pane::Pane;
@@ -352,7 +353,12 @@ impl super::TermWindow {
                 }
                 TabBarItem::None => {
                     // Potentially starting a drag by the tab bar
-                    self.window_drag_position.replace(event.clone());
+                    if !self
+                        .window_state
+                        .intersects(WindowState::MAXIMIZED | WindowState::FULL_SCREEN)
+                    {
+                        self.window_drag_position.replace(event.clone());
+                    }
                     context.request_drag_move();
                 }
             },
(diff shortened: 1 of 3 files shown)


window/src/lib.rs
```diff
@@ -249,6 +249,14 @@ pub trait WindowOps {
     /// window movement on the server side (Wayland).
     fn request_drag_move(&self) {}
 
+    /// Signal to the windowing system that the mouse is over
+    /// a window dragging area.
+    ///
+    /// This is only implemented on backends that need to
+    /// know if the mouse is in a drag area to handle the
+    /// click before forwarding the event (Windows).
+    fn set_window_drag_position(&self, _coords: ScreenPoint) {}
```
wez, line 258: What do you think about making this accept a `Rect` that describes the title bar region as opposed to a single point?
I think having a single point might be prone to occasionally "missing" in the hit test if the cursor is moving rapidly.

wezterm-gui/src/termwindow/mouseevent.rs
```diff
@@ -371,6 +377,12 @@ impl super::TermWindow {
                 }
                 TabBarItem::None => {}
             },
+            WMEK::Move => match item {
+                TabBarItem::None => {
+                    context.set_window_drag_position(event.screen_coords);
+                }
+                _ => {}
+            },
```
wez, line 385: If we pass down the `item` parameter from `mouse_event_ui_item`, we could assign the drag region based on the `x`, `y`, `width`, and `height` 

Stylistic suggestion for this bit is to use `matches!`:

```suggestion
            WMEK::Move => if matches!(item, TabBarItem::None) {
                    context.set_window_drag_position(event.screen_coords);
            },
```

window/src/lib.rs
davidrios: I think the point is better because it's precise, since the drag-able area of the tab bar is not a rectangle, it also includes the left padding for instance. I tried and couldn't cause the miss, seems like the event processing is fast enough, so I think the possibly very rare miss is acceptable.

wezterm-gui/src/termwindow/mouseevent.rs
davidrios: The formatter breaks it into:

```rust
            WMEK::Move => {
                if matches!(item, TabBarItem::None) {
                    context.set_window_drag_position(event.screen_coords);
                }
            }
``` 

So I guess it's not much of an improvement

PR #1775, opened 2022-03-27
nautilus extension: Open directories in existing terminal windows

lunaryorn:
Instead of opening every directory in a new wezterm instance check if
there's a running wezterm instance and if so, spawn every directory into
a new tab in that instance.  Otherwise spawn a new instance for the
first directory, and then spawn the remaining directories into that
instance.

Also extract the systemd interface from the menu provider class, to make
things easier to understand.

### DIFF
--- a/assets/wezterm-nautilus.py
+++ b/assets/wezterm-nautilus.py
@@ -14,6 +14,7 @@
 # with this program; if not, write to the Free Software Foundation, Inc.,
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 
+import time
 import os.path
 
 from gi import require_version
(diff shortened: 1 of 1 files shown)


assets/wezterm-nautilus.py
```diff
@@ -14,28 +14,66 @@
...
+
 
-    def _open_terminal(self, path):
+def _open_in_existing_terminal(path):
+    cmd = ['wezterm', 'cli', 'spawn', '--cwd', path]
+    child = Gio.Subprocess.new(cmd, Gio.SubprocessFlags.NONE)
+    child.wait_check()
+
+def _wezterm_running():
+    cmd = ['wezterm', 'cli', 'list']
```
lunaryorn, line 64: @wez Is that the right way to check whether wezterm's running?

assets/wezterm-nautilus.py
```diff
@@ -44,17 +82,20 @@ def _open_terminal(self, path):
...
-                    Gio.DBusCallFlags.NO_AUTO_START, 500, None)
+            self._systemd.move_to_dedicated_scope(int(child.get_identifier()))
+
+    def _open_paths_in_new_or_existing_terminal(self, paths):
+        if _wezterm_running():
+            for path in paths:
+                _open_in_existing_terminal(path)
+        else:
+            self._open_new_terminal(paths[0])
+            time.sleep(0.5)
```
lunaryorn, line 93: I'm not very happy about this sleep, but I haven't found a better way to wait for the fresh instance to start its internal mux server and be ready for wezterm cli connections.

lunaryorn:
@wez Would making wezterm a DBus-activatable application (per Gtk/Gnome interface) be an option?  That'd considerably simplify remote control under Linux/Gnome; for instance in this case specifically I could just call `org.gtk.Actions.activate("open_in_tab", array_of_directories)` via DBus, and have DBus automatically start WezTerm for me :slightly_smiling_face:

wez:
Hmm, you shouldn't need to try to figure this out from here, as `wezterm start` internally uses some heuristics to find a running gui.

If it's not working out for you in this flow, it would be good to troubleshoot!

Setting `WEZTERM_LOG=wezterm_gui=trace,wezterm_client=trace,info` in the environment before spawning wezterm should turn up some diagnostics on stderr.

https://github.com/wez/wezterm/blob/4cd893335b3cb7b7986fc17547d0a9e68dc0320e/wezterm-gui/src/main.rs#L593-L598


The logic for whether we can reuse a running gui is here:

https://github.com/wez/wezterm/blob/4cd893335b3cb7b7986fc17547d0a9e68dc0320e/wezterm-gui/src/main.rs#L383-L406

The way that we locate a running gui is that we maintain per-class-name symlinks in the state dir that point to the gui instance's socket path:

https://github.com/wez/wezterm/blob/4cd893335b3cb7b7986fc17547d0a9e68dc0320e/wezterm-client/src/discovery.rs#L271-L316

wez:
> Would making wezterm a DBus-activatable application (per Gtk/Gnome interface) be an option?

I'm open to it, provided that it doesn't require linking with Gtk directly, but may not have the bandwidth to work on it directly.  Would you mind opening an enhancement issue for that and including a high level summary of what the required changes might be?

lunaryorn:
>> Hmm, you shouldn't need to try to figure this out from here, as wezterm start internally uses some heuristics to find a running gui.

Indeed, it does :astonished:

I simply wasn't aware :see_no_evil:   Still, having tried it, I have two issues with `wezterm start`:

1. It always opens a new window, whereas I'd prefer to have a new tab :innocent: : 
2. I can't tell whether the process starts a new instance or uses an existing one, other than by parsing wezterm's stdout/stderr and looking for some specific text like "Using existing GUI" or similar.  But I'd need to know in order to be able to move a new instance into its own systemd scope.  And I'd like to keep doing that if only to limit the scope of OOM actions.

> I'm open to it, provided that it doesn't require linking with Gtk directly, but may not have the bandwidth to work on it directly. Would you mind opening an enhancement issue for that and including a high level summary of what the required changes might be?

I opened https://github.com/wez/wezterm/issues/1778 for further discussion :slightly_smiling_face:

lunaryorn:
Fine with me :pray:  I can use this version locally meanwhile (I really prefer tabs over new windows :slightly_smiling_face: ), so I don't really depend on this.

PR #1800, opened 2022-03-31
Improve touchpad precision on Wayland

davidrios:
Touchpad input on Wayland has responsiveness/precision issues, similar to what happened on Windows. This simple change improves it.


wez:
Thanks!

PR #1805, opened 2022-04-01
gui: improve mouse text selection

davidrios:
* implement mouse press capture between the terminal and UI, so when you
  start selecting text from the terminal the tabs won't activate and
  vice-versa
* selecting from the top and bottom lines won't scroll the viewport
  anymore, it will only scroll if the mouse is moved out of line bounds
* change cell selection so that it behaves like text selection usually
  does in other popular software

refs: https://github.com/wez/wezterm/issues/1199
refs: https://github.com/wez/wezterm/issues/1386
refs: https://github.com/wez/wezterm/issues/354
possibly: https://github.com/wez/wezterm/issues/1589

### DIFF
--- a/term/src/input.rs
+++ b/term/src/input.rs
@@ -36,8 +36,8 @@ pub struct MouseEvent {
     pub kind: MouseEventKind,
     pub x: usize,
     pub y: VisibleRowIndex,
-    pub x_pixel_offset: usize,
-    pub y_pixel_offset: usize,
+    pub x_pixel_offset: isize,
+    pub y_pixel_offset: isize,
     pub button: MouseButton,
     pub modifiers: KeyModifiers,
 }
@@ -46,8 +46,8 @@ pub struct MouseEvent {
 pub struct ClickPosition {
     pub column: usize,
     pub row: i64,
-    pub x_pixel_offset: usize,
-    pub y_pixel_offset: usize,
+    pub x_pixel_offset: isize,
+    pub y_pixel_offset: isize,
 }
 
 /// This is a little helper that keeps track of the "click streak",
--- a/wezterm-gui/src/termwindow/selection.rs
+++ b/wezterm-gui/src/termwindow/selection.rs
@@ -63,25 +63,41 @@ impl super::TermWindow {
     ) {
         self.selection(pane.pane_id()).seqno = pane.get_current_seqno();
         let mode = mode.unwrap_or(SelectionMode::Cell);
-        let (x, y, position) = match self.pane_state(pane.pane_id()).mouse_terminal_coords {
-            Some(coords) => (coords.0.column, coords.1, coords.0),
+        let (position, y) = match self.pane_state(pane.pane_id()).mouse_terminal_coords {
+            Some(coords) => coords,
             None => return,
         };
+        let x = position.column;
         match mode {
             SelectionMode::Cell => {
+                // Origin is the cell in which the selection action started. E.g. the cell
+                // that had the mouse over it when the left mouse button was pressed
                 let origin = self
                     .selection(pane.pane_id())
                     .origin
                     .unwrap_or(SelectionCoordinate { x, y });
                 self.selection(pane.pane_id()).origin = Some(origin);
 
-                let (start_x, end_x) = if (origin.x <= x && origin.y == y) || origin.y < y {
+                // Compute the start and end horizontall cell of the selection.
+                // The selection extent depends on the mouse cursor position in relation
+                // to the origin.
+                let (start_x, end_x) = if (x >= origin.x && y == origin.y) || y > origin.y {
+                    // If the selection is extending forwards from the origin, it includes the
+                    // origin and doesn't include the cell under the cursor. Note that the
+                    // reported cell here is offset by -50% from the real cell you see on the
+                    // screen, so this causes a visual cell on the screen to be selected when
+                    // the mouse moves over 50% of its width, which effectively means the next
+                    // cell is being reported here, hence it's excluded
                     (origin.x, x.saturating_sub(1))
                 } else {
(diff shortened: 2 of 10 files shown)


davidrios:
@wez What do you think of this approach for testing?
https://github.com/davidrios/wezterm/commit/c8a2764d460d5200f73584809d4b07df68d4e01e
I moved some things to traits to be able to implement them in testing.

term/src/input.rs
```diff
@@ -36,8 +36,8 @@ pub struct MouseEvent {
     pub kind: MouseEventKind,
     pub x: usize,
     pub y: VisibleRowIndex,
-    pub x_pixel_offset: usize,
-    pub y_pixel_offset: usize,
+    pub x_pixel_offset: isize,
```
wez, line 39: I don't have time for a full review right now, but this caught my eye: why is the type changing? I'm pretty sure that this can never legitimately be negative. @AutumnMeowMeow: can you confirm?

davidrios: The idea is for them to be able to have negative values, to encode information about the mouse leaving the area. For instance, y = 0 meaning that the cell is the top one, but y pixel offset = -10 means the mouse is 10 pixels above that cell.

wez:
> What do you think of this approach for testing?

I think it looks good!  Just the right amount of trait use IMO.

There are some selection bits in the `term` subdir that are probably dead code now, but date back to when selection was managed by `Terminal` rather than the GUI.  There might be some stuff there that is worth moving into the gui crate if you'd like to augment selection tests: https://github.com/wez/wezterm/blob/main/term/src/test/selection.rs

term/src/input.rs
AutumnMeowMeow: (Sorry if incoherent, watching Matrix 4 while drunk... 🍷🔫🔫🔫👭🔫🔫🔫🔫🔫🔫🔫🔫🔫👭🔫🔫🔫🔫🔫🔫🔫🔫🔫💞💞)

The terminal should not be reporting negative values (they are required to be positive in CSI sequences), but at one point xterm itself was reporting negative when the mouse was captured (button down) and dragged above the top-left corner of the text area -- when using SGR-Pixel mode. I didn't report that as a bug, someone else might have.  Let's see if I can find that thread....  here: https://github.com/contour-terminal/contour/issues/574#issuecomment-1008298182  Of course j4james is the real expert. :-)

Within wezterm, the offset values could work the same whether positive or negative values.

davidrios: Note that the negative offset is only used internally/where it makes sense, in other places I've clipped it to positive only values, like here: https://github.com/wez/wezterm/blob/be567fb8116419537b5957c761c663662ea62c3b/term/src/terminalstate/mouse.rs#L72

davidrios:
Anyway, I'll create another PR for the tests and some other related changes, so this one is ready for merging, as soon as all discussions are resolved.

term/src/input.rs
davidrios: On the topic of those offsets, they have an unexpected (to me) behavior, so I'm not sure if it's a bug or could have weird side effects. These are the values reported:

wezterm_cell_reporting

The blue cells are what is reported in the `x` property, because of the rounding here: https://github.com/wez/wezterm/blob/be567fb8116419537b5957c761c663662ea62c3b/wezterm-gui/src/termwindow/mouseevent.rs#L85-L87, which is done so mouse selection feel more natural, yet the `x_pixel_offset` is calculated based on the real cell, so the reported `cell, offset` are not in sync.

I'm thinking maybe it would be better to report the real cell and offset, and let the code that's using them do the relevant processing.

wezterm-gui/src/termwindow/selection.rs
```diff
@@ -58,29 +63,42 @@ impl super::TermWindow {
...
-                            .extend(end)
-                    }
-                    Some(sel) => sel.extend(end),
+                let origin = self
+                    .selection(pane.pane_id())
+                    .origin
+                    .unwrap_or(SelectionCoordinate { x, y });
+                self.selection(pane.pane_id()).origin = Some(origin);
+
+                let (start_x, end_x) = if (origin.x <= x && origin.y == y) || origin.y < y {
```
wez, line 78: Could you add some comments around this code to explain this logic?

wezterm-gui/src/termwindow/selection.rs
```diff
@@ -58,29 +63,42 @@ impl super::TermWindow {
     ) {
         self.selection(pane.pane_id()).seqno = pane.get_current_seqno();
         let mode = mode.unwrap_or(SelectionMode::Cell);
-        let (x, y) = match self.pane_state(pane.pane_id()).mouse_terminal_coords {
-            Some(coords) => coords,
+        let (x, y, position) = match self.pane_state(pane.pane_id()).mouse_terminal_coords {
+            Some(coords) => (coords.0.column, coords.1, coords.0),
```
wez, line 67: More of a nit, but: `(coords.0.column, coords.1, coords.0)` feels awkward and looks a bit opaque.

I feel like it would read better here if this was:

```rust
let (position, y) = match self.pane_state(pane.pane_id()).mouse_terminal_coords {
   Some(coords) => coords,
   None => return,
};
```

and then the code below would use `position.column` instead of `x`

What do you think?

wez (commented):
I think this looks good! Thanks!

term/src/input.rs
wez: @AutumnMeowMeow: what do you think about these positioning offsets; are they actually what you want to see in the report sent to your app?

wezterm-gui/src/termwindow/selection.rs
davidrios: Yeah, indeed it looks a bit esoteric without context 😂

wez:
Thanks!

PR #1808, opened 2022-04-02
docs: add Windows build/dev instructions

davidrios:


docs/install/windows-dev.markdown
```diff
@@ -0,0 +1,74 @@
...
+* Install [Git for windows](https://git-scm.com/download/win).
+* Install and configure [vcpkg](https://github.com/microsoft/vcpkg#quick-start-windows),
+  including running the `integrate install` command.
+
+After all pre-requisites are installed and configured, you'll need to install `OpenSSL` from
+`vcpkg`. Open a new cmd/powershell and navigate to the directory you installed `vcpkg`,
+which contains the `vcpkg.exe` executable, and execute the command:
+
+```
+vcpkg.exe install openssl:x64-windows-static
```
wez, line 19: I don't build wezterm this way and neither does the CI.  These instructions are prone to having weird problems that don't replicate in the "one true way" that I build/maintain wezterm, so I'm a bit leery of having them so prominent.

docs/install/windows-dev.markdown
```diff
@@ -0,0 +1,74 @@
...
+```
+
+Finally use the commands to build:
+
+* Build in release mode: `cargo build --release`
+* Build and run: `cargo run --release --bin wezterm-gui`
+* Build and run showing the log output:
+
+    ```
+    cargo run --release --bin wezterm-gui -- --attach-parent-console
```
wez, line 43:
```suggestion
    cargo run --release --bin wezterm
```

docs/install/windows-dev.markdown
```diff
@@ -0,0 +1,74 @@
+## Building on Windows
+
+Pre-requisites:
+
+* You'll need to install [Visual Studio](https://visualstudio.microsoft.com/), the community edition works fine. Install the `Desktop development with C++` workload.
+* Install `rustup` to get the `rust` compiler installed on your system.
+  [Install rustup](https://www.rust-lang.org/en-US/install.html). You don't need to install 
+  the `Visual Studio C++ Build tools` if you already installed Visual Studio.
+* Rust version 1.56 or later is required
```
wez, line 9: This is going to get stale and overlooked when the min version is next bumped.

For unix systems the version is encoded in `ci/check-rust-version.sh`, but since that was written cargo now supports specifying the minimum version in the manifest, so it's probably best to put it somewhere mostly central; perhaps in the `config` crate, which is built fairly early and stands a good chance of alerting the user early when they are using the wrong version.

wez (commented):
Thanks for this!

PR #1823, opened 2022-04-05
fix incorrect mouse position reporting for split panes

### LINKED ISSUE #1822: Unable to select text from pane

macOS

### WezTerm version

wezterm 20220405-085405-b2ba7383

### Describe the bug

Once pane is created within window user is unable to copy text by mouse (double clicking or dragging cursor with pressed left button) from it. Please note that triple click highlights and copies entire line as expected (from new pane).

Copying text from original window (now also a pane) is unaffected. Only new panes are.

### To Reproduce

1. Open new wezterm window
2. Split window into two panes
3. Run any command or input text into new pane to produce text to copy
4. Try to copy any word/text by either double clicking on word or dragging cursor over it from new pane

### Configuration

No relevant config, I have changed only colours and keybinding (no mouse bindings).

### Expected Behavior

Able to copy text by double clicking on words or dragging cursor over it in any open panes.

### Logs

_No response_

### Anything else?

_No response_

davidrios:
fixes #1822

### DIFF
--- a/wezterm-gui/src/termwindow/mouseevent.rs
+++ b/wezterm-gui/src/termwindow/mouseevent.rs
@@ -616,7 +616,15 @@ impl super::TermWindow {
 
         self.pane_state(pane.pane_id())
             .mouse_terminal_coords
-            .replace((position.clone(), stable_row));
+            .replace((
+                ClickPosition {
+                    column: x,
+                    row: y,
+                    x_pixel_offset: position.x_pixel_offset,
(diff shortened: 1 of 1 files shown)


davidrios:
I've just found another bug when mouse selecting vertically split planes, but this one predates my changes and it'll be trickier to fix, so I'll tackle it in another pull request.

wez:
Thanks for this!

PR #1829, opened 2022-04-06
Extend mouse capture concept to individual panes

davidrios:
Extend mouse capture concept to individual terminal panes, so that when you're dragging the mouse with a pressed button in a pane it doesn't affect the others. This also fixes selection scrolling in individual panes.


wezterm-gui/src/termwindow/mod.rs
```diff
@@ -89,7 +89,7 @@ pub fn get_window_class() -> String {
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub enum MouseCapture {
     UI,
-    Terminal,
+    TerminalPane(usize),
```
wez, line 92:
```suggestion
    TerminalPane(PaneId),
```

wez:
Thanks!

PR #1832, opened 2022-04-07
Termwiz: fix partial bracketed paste panic

tlinford:
First of all thanks for the great lib!
We started using `termwiz` in [zellij](https://github.com/zellij-org/zellij), and this came up for some users on wsl: https://github.com/zellij-org/zellij/issues/1292, because sometimes we get a bracketed paste instruction split over two reads from stdin.

The problem is that `InputParser` panics when fed a partial bracketed paste instruction.

Minimal reproduction is:
```
let mut p = InputParser::new();
let input = b"\x1b[200~1234";
let input2 = b"4567\x1b[201~";
p.parse(input, |e| println!("{:?}", e), false);
p.parse(input2, |e| println!("{:?}", e), false);
```

In debug mode it panics on the first parse:
```thread 'input::test::partial_bracketed_paste' panicked at 'attempt to subtract with overflow', termwiz/src/input.rs:1279:58```

While in release mode the panic happens on the second parse:
```thread 'input::test::partial_bracketed_paste' panicked at 'range start index 18446744073709551613 out of range for slice of length 12', termwiz/src/readbuf.rs:47:25```

Not 100% sure on the fix, I'm assuming that the buffer won't advance until the `end_paste` sequence is found, and it does solve the problem with zellij in wsl.

### DIFF
--- a/termwiz/src/input.rs
+++ b/termwiz/src/input.rs
@@ -1276,7 +1276,7 @@ impl InputParser {
                         callback(InputEvent::Paste(pasted));
                         self.state = InputState::Normal;
                     } else {
-                        self.state = InputState::Pasting(self.buf.len() - end_paste.len());
+                        self.state = InputState::Pasting(0);
                         return;
                     }
                 }
--- a/termwiz/src/input.rs
+++ b/termwiz/src/input.rs
@@ -1593,4 +1593,19 @@ mod test {
             "\x1bOP".to_string()
         );
     }
+
+    #[test]
+    fn partial_bracketed_paste() {
+        let mut p = InputParser::new();
+
+        let input = b"\x1b[200~1234";
+        let input2 = b"5678\x1b[201~";
+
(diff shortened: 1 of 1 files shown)


wez (approved):
This looks good, thanks!

I'm glad to hear that termwiz is finding use in the real world beyond my own projects!
I'm happy to receive feedback on the API in general and making it more usable, as well as bug reports... and I always love PRs!

I made a suggested change that explains why it was doing that arithmetic in the first place, and then fixes it up to avoid the panic.

I'm happy to just commit that, or for you to fold it into your PR!

wez:
I've published termwiz 0.16.0 to crates.io with this change included!

imsnif:
> I'm glad to hear that termwiz is finding use in the real world beyond my own projects!

Just to mention that this library really came through for us. We were bouncing around between STDIN parsing libraries for a while and none were a good fit before this one. Your work is very much appreciated!

PR #1840, opened 2022-04-08
wayland: improve touchpad scroll precision

davidrios:
- Save the remainder of the scroll delta to be used on the next event
when it doesn't meet the activation threshold

refs: https://github.com/wez/wezterm/issues/1827


wez:
Many thanks!

PR #1843, opened 2022-04-09
get-deps for Gentoo

tani:
I have added build denendencies for Gentoo. Cheers!

### DIFF
--- a/get-deps
+++ b/get-deps
@@ -116,7 +116,6 @@ if test -e /etc/arch-release ; then
   exit $?
 fi
 
-# DONOT forget to set USE="X"
 if test -e /etc/gentoo-release ; then
   portageq envvar USE | xargs -n 1 | grep '^X$' \
   || (echo 'X is not found in USE flags' && exit 1)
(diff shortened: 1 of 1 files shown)


get-deps
```diff
@@ -116,6 +116,28 @@ if test -e /etc/arch-release ; then
   exit $?
 fi
 
+# DONOT forget to set USE="X"
```
wez, line 119: Who is this comment for? Should this script set `USE=X`?

wez (commented):
Thanks for this!

get-deps
tani: For users. It might (or should?) be documented on the website. 

`USE=X` is the system configuration of the user environment.
The script should not set it itself for security reason.
Most graphical Gentoo system users have already enabled this flag in the system. So, this is meddlesomeness.

wez: I think "no one" will see that comment, so it should probably be removed from the script and either:

* Add something to the script that checks for it being set and generates a warning if it isn't
* Cover that this should be set in the documentation

tani: I have added error message if X is not found in USE flags

wez:
Thanks!

PR #1868, opened 2022-04-12
support drag and drop files for macos

junnplus:
Signed-off-by: Ye Sijun <junnplus@gmail.com>

Support drag and drop files for macos: https://github.com/wez/wezterm/issues/640

### DIFF
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -1470,7 +1470,7 @@ impl Default for ExitBehavior {
 
 #[derive(Debug, Deserialize, Serialize, Clone, Copy, PartialEq, Eq)]
 pub enum DroppedFileQuoting {
-    /// Do no quoting at all, just pass through the name as-is
+    /// No quoting is performed, the file name is passed through as-is
     None,
     /// Backslash escape only spaces, leaving all other characters as-is
     SpacesOnly,
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -621,6 +621,9 @@ pub struct Config {
 
     #[serde(default)]
     pub key_map_preference: KeyMapPreference,
+
+    #[serde(default)]
+    pub quote_dropped_files: DroppedFileQuoting,
 }
 impl_lua_conversion!(Config);
 
@@ -1462,5 +1465,32 @@ pub enum ExitBehavior {
 impl Default for ExitBehavior {
     fn default() -> Self {
         ExitBehavior::CloseOnCleanExit
+    }
+}
+
+#[derive(Debug, Deserialize, Serialize, Clone, Copy, PartialEq, Eq)]
+pub enum DroppedFileQuoting {
+    /// Do no quoting at all, just pass through the name as-is
+    None,
+    /// Backslash escape only spaces, leaving all other characters as-is
+    SpacesOnly,
+    /// Use POSIX style shell word escaping
+    Posix,
+}
+
+impl Default for DroppedFileQuoting {
+    fn default() -> Self {
+        Self::SpacesOnly
+    }
+}
+
+impl DroppedFileQuoting {
+    pub fn escape(self, s: &str) -> String {
+        match self {
+            Self::None => s.to_string(),
+            Self::SpacesOnly => s.replace(" ", "\\ "),
+            // https://docs.rs/shlex/latest/shlex/fn.quote.html
+            Self::Posix => shlex::quote(s).into_owned().to_string(),
+        }
     }
 }
--- a/docs/config/lua/config/quote_dropped_files.md
+++ b/docs/config/lua/config/quote_dropped_files.md
@@ -1,0 +1,27 @@
+## quote_dropped_files = "SpacesOnly"
+
+*Since: nightly builds only*
+
+Controls how file names are quoted (or not) when dragging and dropping.
+There are three possible values:
+
+* `"None"` - no quoting is performed, the file name is passed through as-is.
+* `"SpacesOnly"` - backslash-escape only spaces, leaving all other characters as-is.  This is the default.
+* `"Posix"` - use POSIX style shell word escaping.
+
+For example:
+
+|`quote_dropped_files` |file name    |quoted result  |
+|----------------------|-------------|---------------|
+|`"None"`              |`hello world`|`hello world`  |
+|`"SpacesOnly"`        |`hello world`|`hello\ world` |
+|`"Posix"`             |`hello world`|`"hello world"`|
+
+Files drag and drop support is a platform dependent feature
+
+|Platform  |Supported since    |
+|----------|-------------------|
+|macOS     |nightly builds only|
+|Windows   |Not yet            |
+|X11       |Not yet            |
+|Wayland   |Not yet            |
--- a/docs/config/lua/config/quote_dropped_files.md
+++ b/docs/config/lua/config/quote_dropped_files.md
@@ -17,7 +17,7 @@ For example:
 |`"SpacesOnly"`        |`hello world`|`hello\ world` |
 |`"Posix"`             |`hello world`|`"hello world"`|
 
-Files drag and drop support is a platform dependent feature
+Drag and drop support for files is a platform dependent feature
 
 |Platform  |Supported since    |
 |----------|-------------------|
--- a/wezterm-gui/src/termwindow/mod.rs
+++ b/wezterm-gui/src/termwindow/mod.rs
@@ -950,7 +950,11 @@ impl TermWindow {
                 };
                 let paths = paths
                     .iter()
-                    .map(|path| format!("{:?}", path.to_string_lossy().to_string()))
+                    .map(|path| {
+                        self.config
+                            .quote_dropped_files
+                            .escape(&path.to_string_lossy())
+                    })
                     .collect::<Vec<_>>()
                     .join(" ");
                 pane.trickle_paste(paths)?;
--- a/wezterm-gui/src/termwindow/mod.rs
+++ b/wezterm-gui/src/termwindow/mod.rs
@@ -943,6 +943,20 @@ impl TermWindow {
                 }
                 Ok(true)
             }
+            WindowEvent::DroppedFile(paths) => {
+                let pane = match self.get_active_pane_or_overlay() {
+                    Some(pane) => pane,
+                    None => return Ok(true),
+                };
+                let paths = paths
+                    .iter()
+                    .map(|path| format!("{:?}", path.to_string_lossy().to_string()))
+                    .collect::<Vec<_>>()
+                    .join(" ");
+                pane.trickle_paste(paths)?;
+                Ok(true)
+            }
+            WindowEvent::DraggedFile(_) => Ok(true),
         }
     }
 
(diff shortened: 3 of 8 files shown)


window/src/os/macos/window.rs
```diff
@@ -2439,6 +2448,46 @@ impl WindowView {
...
+    extern "C" fn dragging_entered(this: &mut Object, _: Sel, sender: id) -> BOOL {
+        if let Some(this) = Self::get_this(this) {
+            let mut inner = this.inner.borrow_mut();
+
+            let pb: id = unsafe { msg_send![sender, draggingPasteboard] };
+            let filenames =
+                unsafe { NSPasteboard::propertyListForType(pb, appkit::NSFilenamesPboardType) };
+            for file in unsafe { filenames.iter() } {
+                unsafe {
+                    let f = NSString::UTF8String(file);
```
wez, line 2460: Consider using `nsstring_to_str` here

window/src/lib.rs
```diff
@@ -175,6 +176,12 @@ pub enum WindowEvent {
     AppearanceChanged(Appearance),
 
     Notification(Box<dyn Any + Send + Sync>),
+
+    // A file is being dragged into the window
+    DraggedFile(PathBuf),
+
+    // A file is been dropped into the window
+    DroppedFile(PathBuf),
```
wez, line 184: I think these should be `Vec` to model the multiple file vs. single file case so that we can make some smarter decisions about how to propagate drops to the application in the terminal; see my other comments about escaping, spaces and multiple files!

```suggestion
    // A file is being dragged into the window
    DraggedFile(Vec<PathBuf>),

    // A file is been dropped into the window
    DroppedFile(Vec<PathBuf>),
```

wezterm-gui/src/termwindow/mod.rs
```diff
@@ -928,6 +928,16 @@ impl TermWindow {
                 }
                 Ok(true)
             }
+            WindowEvent::DroppedFile(path) => {
+                let pane = match self.get_active_pane_or_overlay() {
+                    Some(pane) => pane,
+                    None => return Ok(true),
+                };
+                pane.writer().write_all(path.to_string_lossy().as_bytes())?;
+                pane.writer().write_all(b" ")?;
```
wez, line 937: I think there are some UX considerations around this that need some thought:

* The extra space is probably undesirable if I type `"` then drag a file and then want to type the closing `"`.  The space is more desirable if dropping multiple files.
* If the dragged file has spaces or other characters that might be meaningful to the shell, it might be nice to send the escaped version of the filename. How do we know what the correct escaping syntax should be?
* If bracketed paste is enabled, maybe the application will handle escaping for us if we send the filename as a bracketed paste?
* If the filename is not valid UTF-8, is it even valid to send the lossy transform of the filename?  Might we want to skip those files?

wez (commented):
Thanks for taking a run at this!

window/src/lib.rs
Xuanwo: @Junnplus What will happen if we drag/drop a dir here? Do we need to take that case into consideration?

Junnplus: Dir are the same as files, just a path.

wezterm-gui/src/termwindow/mod.rs
Junnplus: Thanks for reply! Let me think about what to do with them.

Junnplus: Escaping is a little tricky, can we consider outputting a string enclosed in quotes?

wezterm-gui/src/termwindow/mod.rs
```diff
@@ -928,6 +928,20 @@ impl TermWindow {
...
                 Ok(true)
             }
+            WindowEvent::DroppedFile(paths) => {
+                let pane = match self.get_active_pane_or_overlay() {
+                    Some(pane) => pane,
+                    None => return Ok(true),
+                };
+                let paths = paths
+                    .iter()
+                    .map(|path| format!("{:?}", path.to_string_lossy().to_string()))
```
wez, line 938: I wonder if this is OK?

```suggestion
                    .map(|path| format!("{}", path.display()))
```

wezterm-gui/src/termwindow/mod.rs
```diff
@@ -928,6 +928,20 @@ impl TermWindow {
...
+                    .collect::<Vec<_>>()
+                    .join(" ");
+                pane.writer().write_all(paths.as_bytes())?;
```
wez, line 941: You can send it as a paste this way:

```suggestion
                pane.trickle_paste(paths)?;
```

wezterm-gui/src/termwindow/mod.rs
Junnplus: Cannot achieve such an effect like https://github.com/wez/wezterm/pull/1868#discussion_r849133789 , is still an unescaped string without quotes. :(

wezterm-gui/src/termwindow/mod.rs
Junnplus: Nice way.

wezterm-gui/src/termwindow/mod.rs
wez: OK, looking at iterm2, when dropping filenames with spaces, what it does it backslash escape any spaces.  So I think what we should do here is something like:

```rust
    .map(|path| path.to_string_lossy().replace(" ", "\\ "))
```

Junnplus: Not just spaces, there are other characters that need to be escaped.
image
Why not consider  outputting a string enclosed in quotes?
image

Junnplus: > Why not consider outputting a string enclosed in quotes?

@wez Any ideas? like kitty:
image

wez: A problem with quotes is that quoting rules are different between unix shells and windows systems, and we may not know which system we are interacting with.

This sounds like a good place to introduce a config option:

```rust
#[derive(Debug, Deserialize, Serialize, Clone, Copy, PartialEq, Eq)]
enum DroppedFileQuoting {
  /// Do no quoting at all, just pass through the name as-is
  None,
  /// Backslash escape only spaces, leaving all other characters as-is
  SpacesOnly,
  /// Use POSIX style shell word escaping
  Posix,
}

impl Default for DroppedFileQuoting {
  fn default() -> Self {
    Self::SpacesOnly
  }
}

impl DroppedFileQuoting {
   fn escape(s: &str) -> String {
      match self {
         Self::None => s.to_string(),
         Self::SpacesOnly => s.replace(" ", "\\ "),
         // https://docs.rs/shlex/latest/shlex/fn.quote.html
         Self::Posix => shlex.quote(s).into_owned().to_string(),
      }
   }
}
```

and then add this to the `Config` struct in `config/src/config.rs`:

```rust
   #[serde(default)]
   quote_dropped_files: DroppedFileQuoting,
```

Then the code here in this PR would map the value something like this;

```rust
   .map(|path| config.quote_dropped_files.escape(path.to_string_lossy()))
```

Junnplus: Good idea!

wez:
This is looking really good, thanks!

To finish it off, could you add some docs for the new config option?

Add a file named `docs/config/lua/config/quote_dropped_files.md` that explains the new config option, and that demonstrates the possible options; it would be great to show an example of an input string and the different quoted outputs as part of those docs!

Since this is not yet supported on all platforms, I think there should be a table similar to the one on the https://wezfurlong.org/wezterm/config/lua/config/use_ime.html page that shows the platform and supported version.

You can build the docs locally; first install mdbook:

```
cargo install mdbook mdbook-linkcheck mddbook-mermaid
```

Then you can run:

```
ci/build-docs.sh
```

to generate the docs, then run:

```
mdbook serve docs
```

to run them; open the URL in your browser to view the page, and from that point, your changes to the markdown files should cause the page to live-reload in your browser.

junnplus:
> To finish it off, could you add some docs for the new config option?

@wez I am glad to, but it seems we need an ngintly version number in the docs.

wez:
I use: `Since: nightly builds only` as a placeholder for the version for things in the nightly build; I search for exactly that string when I later make a release, then replace it with the new version number.

config/src/config.rs
```diff
@@ -1464,3 +1467,30 @@ impl Default for ExitBehavior {
         ExitBehavior::CloseOnCleanExit
     }
 }
+
+#[derive(Debug, Deserialize, Serialize, Clone, Copy, PartialEq, Eq)]
+pub enum DroppedFileQuoting {
+    /// Do no quoting at all, just pass through the name as-is
```
MuhammedZakir, line 1473:
```suggestion
    /// Do not do any quoting at all, just pass through the name as-is
```

wez: The former is actually better grammar, but perhaps the issue is just that English is weird :-p
Perhaps it is better to rephrase as "No quoting is performed, the file name is passed through as-is".

docs/config/lua/config/quote_dropped_files.md
```diff
@@ -0,0 +1,28 @@
...
+*Since: nightly builds only*
+
+Quote the file names when drag and drop.
+There are three possible values:
+
+* `"None"` - no quoting is performed, the file name is passed through as-is.
+* `"SpacesOnly"` - backslash escape only spaces, leaving all other characters as-is.
+* `"Posix"` - use POSIX style shell work escaping.
+
+For example, if you drag-drop a file `/path/hello world.txt` from Finder onto the Wezterm,
```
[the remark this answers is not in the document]
junnplus: SGTM

docs/config/lua/config/quote_dropped_files.md
```diff
@@ -0,0 +1,27 @@
...
+
+For example
+
+|`quote_dropped_files` |file name    |quoted result  |
+|----------------------|-------------|---------------|
+|`"None"`              |`hello world`|`hello world`  |
+|`"SpacesOnly"`        |`hello world`|`hello\ world` |
+|`"Posix"`             |`hello world`|`"hello world"`|
+
+Files drag and drop support is a platform dependent feature
```
MuhammedZakir, line 20:
```suggestion
Drag and drop support for files is a platform dependent feature
```

Adding a apostrophe after "Files" is enough, but this one rings better in this case? I am not a native speaker, btw.

docs/config/lua/config/quote_dropped_files.md
```diff
@@ -0,0 +1,27 @@
+## quote_dropped_files = "SpacesOnly"
+
+*Since: nightly builds only*
+
+Controls how file names are quoted (or not) when dragging and dropping.
+There are three possible values:
+
+* `"None"` - no quoting is performed, the file name is passed through as-is.
+* `"SpacesOnly"` - backslash-escape only spaces, leaving all other characters as-is.  This is the default.
+* `"Posix"` - use POSIX style shell work escaping.
```
wez, line 10:
```suggestion
* `"Posix"` - use POSIX style shell word escaping.
```

wez:
Thanks!

PR #1871, opened 2022-04-13
fix(default-keys): aligned `Scroll Down One Page` command's key binding with the docs

NikGovorov:
PageUp -> PageDown, apparently it was a copy/paste issue.


wez:
Thanks!

PR #1875, opened 2022-04-14
ssh: support BindAddress

gpanders:
This is already supported by `libssh_rs`.

### DIFF
--- a/docs/ssh.md
+++ b/docs/ssh.md
@@ -41,6 +41,7 @@ and respects the following options:
 * `Host` (including wildcard matching)
 * `UserKnownHostsFile`
 * `IdentitiesOnly`
+* `BindAddress`
 
 All other options are parsed but have no effect.  Notably, neither `Match` or
 `Include` will do anything.
--- a/wezterm-ssh/src/sessioninner.rs
+++ b/wezterm-ssh/src/sessioninner.rs
@@ -227,8 +227,8 @@ impl SessionInner {
 
     #[cfg(feature = "ssh2")]
     fn run_impl_ssh2(&mut self) -> anyhow::Result<()> {
+        use socket2::{Domain, Socket, Type};
         use std::net::ToSocketAddrs;
-        use socket2::{Socket, Domain, Type};
         let hostname = self
             .config
             .get("hostname")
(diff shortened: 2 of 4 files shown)


docs/ssh.md
```diff
@@ -41,6 +41,7 @@ and respects the following options:
 * `Host` (including wildcard matching)
 * `UserKnownHostsFile`
 * `IdentitiesOnly`
+* `BindAddress`
```
wez, line 44:
```suggestion
* `BindAddress` (with the `libssh` backend only)
```

The libssh2 backend doesn't support this, so we should clarify.

If you wanted to add support for that, it's a little involved because
Rust's `TcpStream` doesn't provide a way to do that, so you'd need to use a crate like `socket2` to create the socket, then convert that into `TcpStream`.  https://docs.rs/socket2/latest/socket2/struct.Socket.html#examples 

Those changes would be needed around line 290 of `wezterm-ssh/src/sessioninner.rs` if you did want to take a crack at it!

wez (commented):
Thanks for this!

docs/ssh.md
gpanders: I took a stab at it. I don't write much Rust so I can't promise that it's any good 😄

gpanders: I'm not sure how to test the libssh2 changes though. I've tried running `cargo build --no-default-features --features ssh2`, but when I run it it still says it's using `libssh_rs`.

gpanders:
The `ssh2` implementation doesn't work right now, it fails with `Permission denied`. Looking into it right now.

wezterm-ssh/src/sessioninner.rs
```diff
@@ -279,20 +283,23 @@ impl SessionInner {
...
+                Socket::from_raw_socket(a.into_raw_socket())
             }
         } else {
-            let socket = TcpStream::connect((hostname.as_str(), port))
-                .with_context(|| format!("connecting to {}", remote_address))?;
-            socket
-                .set_nodelay(true)
-                .context("setting TCP NODELAY on ssh connection")?;
-            socket
+            let addr = (hostname.as_str(), port).to_socket_addrs()?.next().context("address")?;
```
wez, line 294: I think the context can be improved here to show what it was trying to do; in this case it is DNS resolving the hostname:

```suggestion
            let addr = (hostname.as_str(), port).to_socket_addrs()?.next().with_context(|| format!("resolving address for {}", hostname))?;
```

wezterm-ssh/src/sessioninner.rs
```diff
@@ -279,20 +283,23 @@ impl SessionInner {
...
+            let sock = Socket::new(Domain::for_address(addr), Type::STREAM, None)?;
+            if let Some(bind_addr) = self.config.get("bindaddress") {
+                let bind_addr = (bind_addr.as_str(), 0).to_socket_addrs()?.next().context("bind address")?;
```
wez, line 297: same here!

wezterm-ssh/src/sessioninner.rs
```diff
@@ -279,20 +283,23 @@ impl SessionInner {
...
+                sock.bind(&bind_addr.into())?;
```
wez, line 298:
```suggestion
                sock.bind(&bind_addr.into()).with_context(|| format!("binding to {:?}", bind_addr))?;
```

wezterm-ssh/src/sessioninner.rs
```diff
@@ -279,20 +283,23 @@ impl SessionInner {
...
+            }
+
+            sock.connect(&addr.into())?;
```
wez, line 301:
```suggestion
            sock.connect(&addr.into()).with_context(|| format!("Connecting to {:?}", addr))?;
```

wez (commented):
Looks good!  I think the error context stuff could be refined; some suggestions inline.
Could you try connecting to bogus hostname/address, and try specifying a bogus bind address and sanity check that the error messages are reasonably informative?

gpanders:
>Could you try connecting to bogus hostname/address, and try specifying a bogus bind address and sanity check that the error messages are reasonably informative?

I think these both look good. Connecting to a bogus address:

`Error: failed to lookup address information: nodename nor servname provided, or not known`

Invalid bind address:

`Error: binding to 192.168.0.169:0: Can't assign requested address (os error 49)`

wez:
Thanks!

PR #1882, opened 2022-04-16
fix: broken scroll bar after screen is cleared/reset

davidrios:
When the screen is cleared/reset, the physical top is not reset with it,
instead the scrollback_top variable gets set with the point at which the
screen was reset / you are allowed to scroll back to. The scroll bar
code wasn't aware of that.

refs: https://github.com/wez/wezterm/issues/1866


wez:
Thanks!

PR #1885, opened 2022-04-17
Handle OpenSUSE package name variations

xpufx:
Deploy should create suse packages with correct dependencies now. Tested on opensuse thumbleweed and ubuntu mate. Needs to be invoked from a workflow for github actions integration.


wez:
Thanks!

Would you like to try adding a SUSE build to the CI?
The steps are:
1. Add an appropriate target around here with a suse docker image:
https://github.com/wez/wezterm/blob/83e0f97ce38ddae128500922ecf29de08b2ee7c4/ci/generate-workflows.py#L667
2. Run `ci/generate-workflows.py`
3. Add the new files to this PR and push
4. I need to click approve and then we see if it builds out ok!
5. You might need to adjust generate-workflows so that it knows about zypper in cases similar to the `uses_yum` cases

xpufx:
Will start clean and come back.

PR #1886, opened 2022-04-17
Per pane scroll bar

davidrios:
Implement per-pane scroll bar

* Now instead of a boolean "enable_scroll_bar" config, you can control
  if you want to enable scroll bars on all panes (AllPanes), just the
  active one (ActivePane) or None at all with the config "scroll_bar_mode"
* Scroll bar width adjusted to ~1/4 of a cell, or it would not fit in
  split panes
* Adjusted horizontal panel split mouse dragging area size, so it's
  possible to also drag a scroll bar beside it
* Right window padding has an enforced minimum size of 1/4 of a cell when
  scrollbars are enabled

Dev notes:

* Mouse events for the scroll bar have a pane_id member, so they correctly
  trigger for their respective panes
* ScrollHit changed to instantiable ScrollThumb, encapsulating more
  information about the scroll bar, so the loginc doesn't need to be
  duplicated again in the mouse event handler
* Fixed a bug where split panes would not fill the screen all the way if
  it wasn't aligned to the cell size

### DIFF
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -333,7 +333,7 @@ pub struct Config {
     pub hide_tab_bar_if_only_one_tab: bool,
 
     #[serde(default)]
-    pub enable_scroll_bar: bool,
+    pub scroll_bar_mode: ScrollBarMode,
 
     /// If false, do not try to use a Wayland protocol connection
     /// when starting the gui frontend, and instead use X11.
@@ -1462,5 +1462,18 @@ pub enum ExitBehavior {
 impl Default for ExitBehavior {
     fn default() -> Self {
         ExitBehavior::CloseOnCleanExit
+    }
+}
+
+#[derive(Debug, Deserialize, Serialize, Clone, Copy, PartialEq, Eq)]
+pub enum ScrollBarMode {
+    None,
+    ActivePane,
+    AllPanes,
+}
+
+impl Default for ScrollBarMode {
+    fn default() -> Self {
+        ScrollBarMode::None
     }
 }
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -332,6 +332,11 @@ pub struct Config {
     #[serde(default)]
     pub hide_tab_bar_if_only_one_tab: bool,
 
+    /// This config is deprecated in favour of the new scroll_bar_mode.
+    /// If true, scroll_bar_mode will be set to AllPanes, else to None.
+    #[serde(default)]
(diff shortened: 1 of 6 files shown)


config/src/config.rs
```diff
@@ -330,7 +330,7 @@ pub struct Config {
     pub hide_tab_bar_if_only_one_tab: bool,
 
     #[serde(default)]
-    pub enable_scroll_bar: bool,
+    pub scroll_bar_mode: ScrollBarMode,
```
wez, line 333: For backwards compatibility purposes, it would be nice to allow `enable_scroll_var = true` to continue working when folks upgrade to a version with these changes.

I'd propose adding:

```rust
#[serde(default)]
pub enable_scroll_bar: Option<bool>,
```

and then adding a helper method:

```
pub fn get_scroll_bar_mode(&self) -> ScrollBarMode {
   match self.enable_scroll_bar {
       Some(true) => ScrollBarMode::AllPanes,
       Some(false) => ScrollBarMode::None,
       None => self.scroll_bar_mode,
   }
}
```
}

wez (commented):
Thanks again for your work!
I haven't had a chance to review in depth yet, but did have one quick comment

config/src/config.rs
davidrios: Good idea. Done.

wez:
Thanks for this!  I finally had some time to look over this and try it out!

Let's talk about the scrollbar width :-)

image

Having the width set to a fraction of a cell makes it almost impossible to click and drag with the default settings on macOS (pictured above).

When running in `AllPanes` mode, the width of the scrollbar is no longer controlled by [window_padding](https://wezfurlong.org/wezterm/config/lua/config/window_padding.html) (see the third paragraph on that page!). I think we should be OK to keep the original width for this case, as the splits aren't getting in the way.

For the `ActivePane` case... what do you think about rendering it wider when hovering over the thumb?
I think the scrollbar could always be the original padding related width when it is aligned to the RHS of the window.  For other cases, on hover, it could render full width over the split.  The hover hit box should probably be to the left of the split divider, but perhaps it could reasonably extend a ~full cell to the left, so that it covers the left half of the cell holding the split line, and the right ~half of the cell to its left.

* Bug: changing the font scaling (CMD +) doesn't recompute the scroll bar positioning!
* Bug: reloading the config (CMD-SHIFT-R) should cause the scroll bar to recompute (eg: CMD-+ first to produce an artifact; reloading the config should cause it to clear)

PR #1892, opened 2022-04-19
docs: add split/panes keybindings

sohnryang:
Terminal splits/panes are not documented, so I added some keybinding references to the feature list.


wez:
Thanks!

PR #1911, opened 2022-04-24
add json output format option to cli commands

ratmice:
Most of this is just indentation of existing code,
The main thing to be skeptical of code wise is whether the `TlsCreds` branch at the bottom is correct/useful, since its going through that type which implements `encode` but not serialize.

Logically,  the thing to be skeptical of seems to be whether it makes sense to have a json command on `cli` vs each subcommand.
I did it on cli and implemented for the sub commands that seem to produce output.

wdyt? :shrug:

### DIFF
--- a/wezterm/src/main.rs
+++ b/wezterm/src/main.rs
@@ -8,6 +8,7 @@ use mux::tab::SplitDirection;
 use mux::window::WindowId;
 use mux::Mux;
 use portable_pty::cmdbuilder::CommandBuilder;
+use serde::Serializer as _;
 use std::ffi::OsString;
 use std::io::{Read, Write};
 use std::rc::Rc;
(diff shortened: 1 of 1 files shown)


wezterm/src/main.rs
```diff
@@ -728,8 +768,18 @@ async fn run_cli_async(config: config::ConfigHandle, cli: CliCommand) -> anyhow:
             smol::future::pending().await
         }
         CliSubCommand::TlsCreds => {
+            let out = std::io::stdout();
             let creds = client.get_tls_creds().await?;
-            codec::Pdu::GetTlsCredsResponse(creds).encode(std::io::stdout().lock(), 0)?;
+
+            match cli.output_format {
+                CliOutputFormat::Json => {
+                    serde_json::to_writer_pretty(out.lock(), &creds)?;
```
ratmice, line 776: I think this should be calling `encode` of some sort and then serializing the result as `Vec<u8>` instead probably.

ratmice:
FWIW, the impetus for this was the following script, which from sway,
if wezterm is currently focused, spawn a new terminal in the same directory,
otherwise just launch a new window in the default directory...

There is some sway specific stuff to figuring out if a terminal is focused..

```
#!/bin/sh
alias wezterm=~/.cargo/bin/wezterm
if [ $(swaymsg -t get_tree | jq '.. | select(.type?) | select(.focused==true).window_properties.class') = \"org.wezfurlong.wezterm\" ];
  then
          PANE_ID=$(wezterm cli --output-format json list-clients | jq '.clients[] | .focused_pane_id')
          WORKDIR=$(wezterm cli --output-format json list  | jq -r ".tabs[] | .Leaf | select(.pane_id == $PANE_ID).working_dir")
          wezterm cli spawn --new-window --cwd ${WORKDIR#file://}
  else
          wezterm cli spawn --new-window 
fi
```

wez:
Thanks for this!

The main thing on my mind is that by directly exposing the struct as json we're effectively promising that it will remain a stable API in the future, which may not be the case for all of these things.

What I'd suggest is introducing eg: an explicit `CliListResult` struct for the `list` output that has those fields from the current textual output as members, then have the CLI command fill that out and then format that as json.

That explicit boundary/copy step prevents accidentally leaking current implementation details out into the API, and makes it clearer in the code what is actually part of the API as well.

The same approach should also apply to the other CLI sub commands.

wezterm/src/main.rs
wez: json is always encoded as utf8 output and that byte sequence is written to the writer, so no additional encoding is needed

wez: However, I'm not sure if this particular command should have json output given that it is an implementation detail of bootstrapping a mux connection!

ratmice: Cool, I think then it makes sense to pull --output-format out into each CliSubCommand so it'd be `wezterm cli list --output-format json`?

I'll do a pass at this later, after peeking through the testsuite results.

ratmice:
I did another pass, I ended up using `CliListResultItem`, and just using iterators, so it was hopefully clear exactly what is being stabilized (That was the intent anyways).

I also changed from `--output-format` to `--format` thinking if short names are ever adopted that `-f` is better than `-o`.
This has a bit more cloning than I prefer, in order to minimize the duplication of code between these branches.
But i figured for just the command line it didn't seem worth bending over backwards to be too efficient.
At least for this pass, may very well need another

Issue #1926, opened 2022-04-27
Add a link to latest build e.g. latest.zip as part of deploy.sh


erf:
**Is your feature request related to a problem? Please describe.**

I'm building from source using `ci/deploy.sh`.

I would like to automate the install process further and it would help to be able to refer to the latest build using the same name instead of e.g. `WezTerm-macos-20220426-191000-97005b8c.zip`

**Describe the solution you'd like**

Could we make an alias / link `latest.zip` point to the latest build as part of `deploy.sh` ? ( i saw DCSS did this )

As a bonus, perhaps add a `build` folder for these build files ? (So it's easier to clean up)

**Describe alternatives you've considered**

**Additional context**

erf:
The solution was to simply set the TAG_NAME beforehand like: 
```
TAG_NAME=latest ./ci/deploy.sh
```
Thanks to wez :)

erf:
I guess i can include my macOS build script here if it's interesting for anyone, which:
1. updates to the latest source code (with dependencies)
2. builds a macOS app
3. copies the app to the Applications folder

```bash
#!/bin/sh
git pull
git submodule update --init --recursive
cargo build --release
TAG_NAME=latest ./ci/deploy.sh
rsync -av WezTerm-macos-latest/Wezterm.app /Applications/
```

PR #1932, opened 2022-04-28
Add link to latest mac build and move to build folder (solves #1926)

erf:
Add link to latest mac build and move to build folder. Solves #1926

### DIFF
--- a/ci/deploy.sh
+++ b/ci/deploy.sh
@@ -86,6 +86,7 @@ case $OSTYPE in
     set -x
     zip -r $zipname $zipdir
     set +x
+    ln -sf $zipname WezTerm-macos-latest.zip
 
     if [ -n "$MACOS_TEAM_ID" ] ; then
       echo "Notarize"


ci/deploy.sh
```diff
@@ -96,6 +95,12 @@ case $OSTYPE in
     SHA256=$(shasum -a 256 $zipname | cut -d' ' -f1)
     sed -e "s/@TAG@/$TAG_NAME/g" -e "s/@SHA256@/$SHA256/g" < ci/wezterm-homebrew-macos.rb.template > wezterm.rb
 
+    latest=WezTerm-macos-latest.zip
+    ln -sf $zipname $latest
+    mkdir -p build
+    (cd build ; rm -rf $zipdir $zipname $latest)
+    mv $zipname $zipdir $latest build
```
wez, line 102: This part will break the nightly and release uploads:

```
Run actions/upload-artifact@v3
Warning: No files were found with the provided path: WezTerm-*.zip. No artifacts will be uploaded.
```

Putting the artifacts in subdirs is problematic with the way that artifact uploading works with GH, so I don't want to do this bit.

erf: Ok. I can remove the last commit and just leave the link in the main folder.

erf: Done.

wez: Thanks!  Now, the other problem is that this is uploading two copies of the artifacts!

```
Run actions/upload-artifact@v3
With the provided path, there will be 2 files uploaded
```

I think the sanest way to deal with this is to add either command line parameter(s) or an environment variable that allows you to tell the script whether to use the naming scheme used for distribution, or whether to use a convenient name for local testing, or perhaps, just let you completely override the name.

Actually, I think this already exists; 

https://github.com/wez/wezterm/blob/efcc4b48c11fa65d17a6fa9d050b0ac2eef4f48a/ci/deploy.sh#L22-L27

Can you try running it like this:

```
TAG_NAME=latest ./ci/deploy.sh
```

and see if that has the result you're looking for?
I think that will output `WezTerm-macos-latest.zip`

erf: Ah, you are right - that solves it 😅👍

PR #1938, opened 2022-04-30
support scrolling tabs

junnplus:
Signed-off-by: Ye Sijun <junnplus@gmail.com>

Fixes: https://github.com/wez/wezterm/issues/1726

Scrolling tabs can sometimes be convenient.

### DIFF
--- a/wezterm-gui/src/termwindow/mouseevent.rs
+++ b/wezterm-gui/src/termwindow/mouseevent.rs
@@ -425,6 +425,10 @@ impl super::TermWindow {
                 }
                 _ => {}
             },
+            WMEK::VertWheel(n) => {
+                self.activate_tab_relative(if n < 1 { 1 } else { -1 }, true)
+                    .ok();
+            }
             _ => {}
         }
         self.update_title_post_status();


wezterm-gui/src/termwindow/mouseevent.rs
```diff
@@ -425,6 +425,9 @@ impl super::TermWindow {
                 }
                 _ => {}
             },
+            WMEK::VertWheel(n) => {
+                self.activate_tab_relative(-n as isize, true).ok();
```
wez, line 429: The wheel movement may be scaled to larger than 1 based on inertia or screen scaling, which can make it difficult to "tick" through the tabs one by one.

For example, on my X11 system, wheel up events tend to tick by three each time whereas wheel down events tick down by one initially.  Weird but true!

I think this should probably clamp to a single movement per event, something like this:

```suggestion
                self.activate_tab_relative(if n < 1 { -1 } else { 1 }, true).ok();
```

junnplus: SGTM

wez:
Thanks!

PR #1939, opened 2022-05-01
Fix missing string after ";" from icon and window title

kumattau:
xterm says the following.

```
OSC Ps ; Pt BEL
OSC Ps ; Pt ST
          Ps specifies the type of operation to perform:
            Ps = 0  -> Change Icon Name and Window Title to Pt.
            Ps = 1  -> Change Icon Name to Pt.
            Ps = 2  -> Change Window Title to Pt
```

OSC string is split by ";" but Pt may contain ";", so Pt should be restored by joining ";" if necessary .
Otherwise, the string after ";" is missing from icon and window title.

### DIFF
--- a/termwiz/src/escape/osc.rs
+++ b/termwiz/src/escape/osc.rs
@@ -274,10 +274,13 @@ impl OperatingSystemCommand {
 
         macro_rules! single_string {
             ($variant:ident) => {{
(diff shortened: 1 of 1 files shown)


wez:
Thanks!

PR #1953, opened 2022-05-05
Add support for drag and drop files in Windows

datasone:
This pr enables drag and drop files on Windows. Two new patterns for `quote_dropped_files` are also added to match behaviours with other shells on Windows:

- `Windows`: Double-quote the filenames with space in it, and leave others untouched. This is the same behaviour as `cmd` or `powershell`.
- `DoubleQuoteAlways`: Same with `Windows`, while the filename is always double-quoted regardless of whether the filename contains spaces or not. This is introduced for MSYS2/Cygwin, where Windows-style paths without quotes will not be treated properly.

The examples in document for `quote_dropped_files` are also changed to better convey the differences between patterns.

---

P.S. The part below is more an issue, but I just put it here as it's relavent.

As the new examples show, the current default `SpacesOnly` won't work correctly on filenames with special chars (i.e. `$`, `` ` ``, `"` and ``\`` as shown in [shlex's code](https://docs.rs/shlex/1.1.0/src/shlex/lib.rs.html#173)). I think it may be better to set `Posix` to default or improve the `SpaceOnly`, single-quoting filenames can also be a choice. If you agree with these, I would draft a pr for them.

### DIFF
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -1500,8 +1500,8 @@ impl DroppedFileQuoting {
             // https://docs.rs/shlex/latest/shlex/fn.quote.html
             Self::Posix => shlex::quote(s).into_owned().to_string(),
             Self::Windows => {
-                let chars_need_quoting = [' ','\t','\n','\x0b','\"'];
-                if chars_need_quoting.iter().any(|&c| s.contains(c)) {
+                let chars_need_quoting = [' ', '\t', '\n', '\x0b', '\"'];
+                if s.chars().any(|c| chars_need_quoting.contains(&c)) {
                     format!("\"{}\"", s)
                 } else {
                     s.to_string()
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -1476,11 +1476,19 @@ pub enum DroppedFileQuoting {
     SpacesOnly,
     /// Use POSIX style shell word escaping
     Posix,
+    /// Use Windows style shell word escaping
+    Windows,
+    /// Always double quote the file name
+    DoubleQuoteAlways,
 }
 
 impl Default for DroppedFileQuoting {
     fn default() -> Self {
-        Self::SpacesOnly
+        if cfg!(windows) {
+            Self::Windows
+        } else {
+            Self::SpacesOnly
+        }
     }
 }
 
@@ -1491,6 +1499,14 @@ impl DroppedFileQuoting {
             Self::SpacesOnly => s.replace(" ", "\\ "),
             // https://docs.rs/shlex/latest/shlex/fn.quote.html
             Self::Posix => shlex::quote(s).into_owned().to_string(),
+            Self::Windows => {
+                if s.contains(' ') {
+                    format!("\"{}\"", s)
+                } else {
+                    s.to_string()
+                }
+            }
+            Self::DoubleQuoteAlways => format!("\"{}\"", s),
         }
     }
 }
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -1479,7 +1479,7 @@ pub enum DroppedFileQuoting {
     /// Use Windows style shell word escaping
     Windows,
     /// Always double quote the file name
-    DoubleQuoteAlways,
+    WindowsAlwaysQuoted,
 }
 
 impl Default for DroppedFileQuoting {
@@ -1500,13 +1500,14 @@ impl DroppedFileQuoting {
             // https://docs.rs/shlex/latest/shlex/fn.quote.html
             Self::Posix => shlex::quote(s).into_owned().to_string(),
             Self::Windows => {
-                if s.contains(' ') {
+                let chars_need_quoting = [' ','\t','\n','\x0b','\"'];
+                if chars_need_quoting.iter().any(|&c| s.contains(c)) {
                     format!("\"{}\"", s)
                 } else {
                     s.to_string()
                 }
             }
-            Self::DoubleQuoteAlways => format!("\"{}\"", s),
+            Self::WindowsAlwaysQuoted => format!("\"{}\"", s),
         }
     }
 }
--- a/docs/config/lua/config/quote_dropped_files.md
+++ b/docs/config/lua/config/quote_dropped_files.md
@@ -3,25 +3,29 @@
 *Since: nightly builds only*
 
 Controls how file names are quoted (or not) when dragging and dropping.
-There are three possible values:
+There are five possible values:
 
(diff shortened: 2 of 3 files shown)


config/src/config.rs
```diff
@@ -1491,6 +1499,14 @@ impl DroppedFileQuoting {
...
             // https://docs.rs/shlex/latest/shlex/fn.quote.html
             Self::Posix => shlex::quote(s).into_owned().to_string(),
+            Self::Windows => {
+                if s.contains(' ') {
+                    format!("\"{}\"", s)
+                } else {
+                    s.to_string()
+                }
+            }
+            Self::DoubleQuoteAlways => format!("\"{}\"", s),
```
wez, line 1509: I'm not sure about this vs. the other methods.  Can we remove it?

wez (changes requested):
Thanks for this!
The addition of a Windows quoting style is good, but there is more to it; see my inline comment for more on that!

config/src/config.rs
wez: I see your comment about this being for msys. Since this is for Windows, can this be renamed to `WindowsAlwaysQuoted` or something like that?

window/src/os/windows/window.rs
```diff
@@ -2406,6 +2413,36 @@ unsafe fn key(hwnd: HWND, msg: UINT, wparam: WPARAM, lparam: LPARAM) -> Option<L
     None
 }
 
+unsafe fn drop_files(hwnd: HWND, _msg: UINT, wparam: WPARAM, _lparam: LPARAM) -> Option<LRESULT> {
+    if let Some(inner) = rc_from_hwnd(hwnd) {
```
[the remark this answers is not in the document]
datasone: Thanks for the review! This along with the following two mentioned code style fixes (avoid nesting and the use of buffer length function) have been applied.

window/src/os/windows/window.rs
```diff
@@ -2406,6 +2413,36 @@ unsafe fn key(hwnd: HWND, msg: UINT, wparam: WPARAM, lparam: LPARAM) -> Option<L
...
+        // Get the number of files dropped
+        let file_count = DragQueryFileW(h_drop, 0xFFFFFFFF, null_mut(), 0);
+
+        let mut filenames: Vec<PathBuf> = Vec::with_capacity(file_count as usize);
+
+        for idx in 0..file_count {
+            // The returned size of buffer is in characters, not including the terminating null character
+            let buf_size = DragQueryFileW(h_drop, idx, null_mut(), 0);
+            if buf_size > 0 {
+                let buf_size = buf_size as usize + 1;
```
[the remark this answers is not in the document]
datasone: If the buffer doesn't have enough space (i.e. `buf_size + 1`), Windows will truncate the filename string, remove the trailing characters, and add a null terminator. So I'm afraid this cannot be removed.

config/src/config.rs
```diff
@@ -1491,6 +1499,14 @@ impl DroppedFileQuoting {
             Self::SpacesOnly => s.replace(" ", "\\ "),
             // https://docs.rs/shlex/latest/shlex/fn.quote.html
             Self::Posix => shlex::quote(s).into_owned().to_string(),
+            Self::Windows => {
```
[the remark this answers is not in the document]
datasone: Though there is more complex quoting for cli parameters, I think most of them are not needed for this special case (dropping files).

[The code piece in `cmdbuilder.rs`](https://github.com/wez/wezterm/blob/2710cefb3e79bda07e86d91c27225bb0586ac0d5/pty/src/cmdbuilder.rs#L578-L580) quotes like this:
- If the string contains any of `' ', '\t', '\n', '\x0b', '\"'`, the string is surrounded by `"`s.
- The `\` characters at the end of string are doubled.
- An `\` character is added prior to the `"` character in string, if there are any `\` characters before the `"` character, they are doubled.

Both `"` and `\` are reserved character for Windows, i.e. they cannot be used in filenames. Though files with invalid filename can be created in WSL or using third-party NTFS drivers, they cannot be used properly in win32 enviornment ([this issue](https://github.com/microsoft/WSL/issues/2689) is mainly about reserved names, but it should be the same for reserved characters). So I think we can assume those characters won't appear in filenames (heavily WSL users may choose the `Posix` quoting pattern).

Besides, the dropped directory paths have no `\` suffix, so I only added the first point (more checks for filename needed to be quoted).

config/src/config.rs
datasone: Sure! `Posix` would be more useful for unix users, so this would be mostly used on Windows.

config/src/config.rs
```diff
@@ -1491,6 +1499,15 @@ impl DroppedFileQuoting {
...
+                let chars_need_quoting = [' ','\t','\n','\x0b','\"'];
+                if chars_need_quoting.iter().any(|&c| s.contains(c)) {
```
wez, line 1504: I would flip this around to be something like this, so that there is a single pass over what is likely to be the longer string; the overlay big-O should be lower when phrased this way:

```rust
s.chars().any(|c| chars_need_quoting.contains(c))
```

docs/config/lua/config/quote_dropped_files.md
```diff
@@ -3,25 +3,29 @@
...
-|`"None"`              |`hello world`|`hello world`  |
-|`"SpacesOnly"`        |`hello world`|`hello\ world` |
-|`"Posix"`             |`hello world`|`"hello world"`|
+| `quote_dropped_files`   | file name        | quoted result       |
+|-------------------------|------------------|---------------------|
+| `"None"`                | `hello ($world)` | `hello ($world)`    |
+| `"SpacesOnly"`          | `hello ($world)` | `hello\ ($world)`   |
+| `"Posix"`               | `hello ($world)` | `"hello (\$world)"` |
+| `"Windows"`             | `hello ($world)` | `"hello ($world)"`  |
+| `"WindowsAlwaysQuoted"` | `hello($world)`  | `"hello($world)"`   |
```
wez, line 22:
```suggestion
| `"WindowsAlwaysQuoted"` | `hello ($world)`  | `"hello ($world)"`   |
```

config/src/config.rs
datasone: Yes, this is better in performance.

docs/config/lua/config/quote_dropped_files.md
datasone: Though originally I thought this would better show how this works, it's good to keep consistency for this, and the documentation above should be enough, fixed.

wez:
Thanks!

PR #1971, opened 2022-05-10
filedescriptor: Try to produce similar errors across platforms

jorendorff:
filedescriptor has a special error variant `Error::Poll(err)` that is returned when `poll` fails. I had some code that calls `filedescriptor::poll` and checks the error to see if it was `EINTR`.

        match poll(...) {
            Err(filedescriptor::Error::Poll(err)) if err.kind() == ErrorKind::Interrupted => continue,
            ...
        }

I noticed while peeking at the source code for `poll` that this variant isn't being used on macOS. The error behavior is different on different platforms. This would have caused a bug in my code where `EINTR` was not handled correctly on Mac.

Maybe this is on purpose, as the error message `"poll failed"` is a bit misleading on Mac, where the crate is using `select` instead. Feel free to just close this, if so, but if the difference is considered a bug, this PR fixes it, along with two places where Windows code uses the same pattern.

### DIFF
--- a/filedescriptor/src/unix.rs
+++ b/filedescriptor/src/unix.rs
@@ -536,7 +536,7 @@ mod macos {
         };
 
         if res < 0 {
-            Err(std::io::Error::last_os_error().into())
+            Err(Error::Poll(std::io::Error::last_os_error()))
         } else {
             for item in pfd.iter_mut() {
                 if is_set(&mut read_set, item.fd) {
--- a/filedescriptor/src/windows.rs
+++ b/filedescriptor/src/windows.rs
@@ -200,7 +200,10 @@ impl OwnedHandle {
             )
         };
         if ok == 0 {
-            Err(IoError::last_os_error().into())
+            Err(Error::Dup {
+                fd: (handle as isize).into(),
+                source: IoError::last_os_error(),
+            })
         } else {
             Ok(OwnedHandle {
                 handle: duped as *mut _,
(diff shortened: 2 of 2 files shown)


wez:
Thanks for this!
Those enum variants are, somewhat unfortunately, disclosing some implementation details about which internal syscall might have been called and failed.
With that in mind, I think it might be a bit misleading to wrap some of these platform differences up in this way.

With that said, I think there is scope for improving this `Error` type; I think that adding method(s) to expose the inner `std::io::Error` and/or `std::io:ErrorKind` would be helpful for the situation you raised, so that your code could do something like:

```rust
  match poll(...) {
        Err(err) if err.io_error_kind() == Some(ErrorKind::Interrupted) => continue,
        ...
    }
```

assuming that there were methods like:

```rust
impl Error {
   pub fn io_error(&self) -> Option<&std::io::Error> {
      match self {
         Self::Poll(e) | Self::Io(e) | ... => Some(e),
         Self::IllegalFdValue(_) => None,
     }
  }

   pub fn io_error_kind(&self) -> Option<std::io::ErrorKind> {
      self.io_error_kind().map(|e| e.kind())
  }
}
```

jorendorff:
That makes sense.

Looks like I botched the Windows version of this anyway. :) I'm going to close this—sounds like a separate PR would make sense.

Thanks for the quick response!

wez:
Would you like to do a PR for the change I suggested above?

jorendorff:
I can't take it on, I'm afraid. That would solve my problem, though. It sounds good.

PR #1973, opened 2022-05-11
CI: restore Centos 7

wez:
It was removed in be1b54ba3377d3759bab1c003b192cfba3ca630c citing difficulties with GH CLI, but I think I subsequently solved that by not running it in the container.


wez:
Pushed as 7e460c3784bd0e402cfc34123cd720054dc898cc

PR #1980, opened 2022-05-12
gui: add config option for overriding color scheme scroll bar color

Enverbalalic:
Useful for color schemes like Dracula+ where the default scrollbar
thumb color is really poorly contrasted with the background color
and it makes the scroll bar effectively invisible.


wez:
Can you try something like the examples here to meet your needs?
https://wezfurlong.org/wezterm/config/lua/wezterm/get_builtin_color_schemes.html

I'm not keen to add ad-hoc overrides like this, as it makes the overall configuration much more complex assuming the pathological end state where every option is duplicated in this way.

Enverbalalic:
Ah, yes! I couldn't find this in the doc, this works.

I fully agree about the ad-hoc override, I didn't think that this PR would be accepted anyways in the current form, it seemed way too bandaid-y, but I wanted to get the ball rolling

I'll close this, thanks.

PR #1986, opened 2022-05-14
Add fuzzer for termwiz::escape::parser

5225225:
This can be ran with https://github.com/rust-fuzz/cargo-fuzz

This currently still finds at least one bug, but might as well add it now so it can be more easily ran.


wez:
Thanks!

wez:
I added a very basic workflow to run this in the CI: https://github.com/wez/wezterm/runs/6435401395?check_suite_focus=true

PR #1989, opened 2022-05-14
Rectangular selection: some fixes

Funami580:
Fixes the bugs/missing features mentioned in https://github.com/wez/wezterm/issues/1361.

That are:
* add option to select with Alt+Shift
* fix beginning the selection with the top right or bottom left corner
* fix beginning the selection with the bottom right corner and only select a single line


wez:
Looks good, thank you!

PR #1994, opened 2022-05-15
wayland: clipboard: fix pasting content

Funami580:
Fixes pasting content that was copied before wezterm was opened. Ref: https://github.com/wez/wezterm/issues/1385

### DIFF
--- a/window/src/os/wayland/pointer.rs
+++ b/window/src/os/wayland/pointer.rs
@@ -325,6 +325,9 @@ impl PointerDispatcher {
         inner
             .surface_to_pending
             .insert(surface.as_ref().id(), Arc::clone(pending));
+        if inner.active_surface_id == 0 {
+            inner.active_surface_id = surface.as_ref().id();
+        }
     }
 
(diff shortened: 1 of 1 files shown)


wez:
Thanks! However, I'm not sure about this; wayland is very particular about relating windows, input events and clipboard ownership.

It feels like this PR might be trading a current implementation bug for a future incompatibility by outsourcing the clipboard to an external crate that manages it own independent wayland connection.

I'd much rather see a PR that resolves the root cause of the bug in the wezterm/window code!

Funami580:
Okay, I tracked down the bug. In `resolve_copy_and_paste` None was returned, because the `active_surface_id` was not set yet (therefore it was 0, although my surface id was 18). This always happens when I start wezterm and don't move my mouse into the terminal and just press Ctrl+Shift+V. Would it be okay to add
```rs
inner.active_surface_id = surface.as_ref().id();
```
to the `add_window` method in pointer.rs? Or would that introduce new bugs?

wez:
Hmm, how about only assigning it in `add_window` if it is currently `0`?

PR #1997, opened 2022-05-16
Make DecreaseFontSize the inverse of IncreaseFontSize

Funami580:
Otherwise increasing the font size and then decreasing it again will not return to the original font size.

### DIFF
--- a/wezterm-gui/src/termwindow/resize.rs
+++ b/wezterm-gui/src/termwindow/resize.rs
@@ -374,7 +374,7 @@ impl super::TermWindow {
     }
 
     pub fn decrease_font_size(&mut self, window: &Window) {
-        self.adjust_font_scale(self.fonts.get_font_scale() * 0.9, window);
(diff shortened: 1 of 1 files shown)


wezterm-gui/src/termwindow/resize.rs
```diff
@@ -374,7 +374,7 @@ impl super::TermWindow {
     }
 
     pub fn decrease_font_size(&mut self, window: &Window) {
-        self.adjust_font_scale(self.fonts.get_font_scale() * 0.9, window);
+        self.adjust_font_scale(self.fonts.get_font_scale() * (1.0 / 1.1), window);
```
wez, line 377: How about this slightly simpler form?

```suggestion
        self.adjust_font_scale(self.fonts.get_font_scale()  / 1.1, window);
```

Funami580: Alright, I pushed an update.

bew:
So _that_ was the reason, I always wondered ^^

👍

wez:
Thanks!

PR #1998, opened 2022-05-16
Fix compile error by using i686-pc-windows-msvc

kumattau:
This PR is to fix compile error by using x86_64-pc-windows-msvc related to my #1976.

https://github.com/wez/wezterm/blob/3800710eadcd058aeac44e32bd42990097143d9d/window/src/os/windows/window.rs#L52

I should have defined ISC_SHOWUICOMPOSITIONWINDOW in DWORD (u32) because LPARAM (isize) cannot accept 2147483648 (0x80000000) when pointer size is 32bit.

This fix is not important at all because wezterm does not support 32bit windows.
So, if you don't need the fix, please close the PR.


wez:
Thanks!

PR #2002, opened 2022-05-17
Fix text cursor position when using multiple panes

kumattau:
This PR is to fix text cursor position when using multiple panes, which causes incorrect IME candidate window position.

The calculating the correct position of text cursor in a pane requires the offset from the tab to the pane,
but the Pane structure argument of `update_text_cursor(&mut self, pane: &Rc<dyn Pane>)` does not have the offset.

So, the offset is got from `Vec<PositionedPane>` returned by `self.get_panes_to_render()` in every `update_text_cursor()` call , which may be high cost and cause performance degradation.

Is there any other good way ?

before:
スクリーンショット 2022-05-18 014956

after:
スクリーンショット 2022-05-18 014738

### DIFF
--- a/wezterm-gui/src/termwindow/mod.rs
+++ b/wezterm-gui/src/termwindow/mod.rs
@@ -1730,10 +1730,10 @@ impl TermWindow {
         }
     }
 
-    fn update_text_cursor(&mut self, pane: &Rc<dyn Pane>) {
-        let cursor = pane.get_cursor_position();
+    fn update_text_cursor(&mut self, pos: &PositionedPane) {
         if let Some(win) = self.window.as_ref() {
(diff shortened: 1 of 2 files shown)


wez:
Ah, thanks for this!

My suggestion is to pass `&pos` through here, rather than `pos.pane`:

https://github.com/wez/wezterm/blob/ef4a95211e13e836eb1faea99bc5a2fd0716cb8f/wezterm-gui/src/termwindow/render.rs#L1626-L1628

eg: change the type of the `update_text_cursor` parameter to `&PositionedPane`

wez:
Pushed as 9b5c887874be4d40daeaf297678440978f1fa8eb (with `cargo +nightly fmt` applied)
Thanks!

PR #2006, opened 2022-05-18
Add support of IME preedit rendered by system

kumattau:
This PR is to support OverTheSpot input style on Windows and X11.
Users can choose OnTheSpot or OverTheSpot input style by using ime_input_style in configuration file.
OnTheSpot is default. It's same as before.

Background:
wezterm supports nearly OnTheSpot input style that preedit area is rendered by wezterm itself.
"nearly" means OnTheSpot expects inline input (preedit area does not overlap original text) but it is not so currently.

ref: https://www-archive.mozilla.org/projects/intl/input-method-spec

In wezterm, OnTheSpot input style is suitable in many IMEs but it may not be so in Japanese IME
because preedit area may be very long in Japanese IME but wezterm does not support (or cannot support) to render preedit area beyond the window.

So, I propose to support OverTheSpot input style.
OverTheSpot never support inline input but can support to render preedit area beyond the window.
It may useful to input very long preedit in Japanese IME.

OnTheSpot (default)
スクリーンショット 2022-05-19 012909

OverTheSpot (useful in Japanse IME)
スクリーンショット 2022-05-19 012441

NOTE:
In above screenshot of OverTheSpot, preedit area is within the window but it can be rendered beyond the window when preedit area is much longer.

This is draft. I think the followings need to be discussed:

* wezterm should support OverTheSpot or not.
* config key (ime_input_style) and values (OnTheSpot and OverTheSpot) are appropriate names or not.

I will write docs after discussion.

@wez, What do you think about ?

### DIFF
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -501,6 +501,8 @@ pub struct Config {
     pub use_ime: bool,
     #[dynamic(default)]
     pub xim_im_name: Option<String>,
+    #[dynamic(default)]
+    pub ime_preedit_rendering: ImePreeditRendering,
 
     #[dynamic(default = "default_true")]
     pub use_dead_keys: bool,
@@ -1478,5 +1480,19 @@ impl DroppedFileQuoting {
             }
             Self::WindowsAlwaysQuoted => format!("\"{}\"", s),
         }
+    }
+}
+
+#[derive(Debug, FromDynamic, ToDynamic, Clone, Copy, PartialEq, Eq)]
+pub enum ImePreeditRendering {
+    /// IME preedit is rendered by WezTerm itself
+    Builtin,
+    /// IME preedit is rendered by system
+    System,
+}
+
+impl Default for ImePreeditRendering {
+    fn default() -> Self {
+        ImePreeditRendering::Builtin
     }
 }
--- a/docs/config/lua/config/ime_preedit_rendering.md
+++ b/docs/config/lua/config/ime_preedit_rendering.md
@@ -1,0 +1,36 @@
+# `ime_preedit_rendering`
+
+*Since: nightly builds only*
+
+Control IME preedit rendering.
+IME preedit is an area that is used to display the string being preedited in IME.
+WezTerm supports the following IME preedit rendering.
+
+* `"Builtin"` - (Default) IME preedit is rendered by WezTerm itself
+
+  "Builtin" rendering provides good look and feel for many IMEs
+
+* `"System"` - IME preedit is rendered by system
+
+  "Builtin" rendering may truncate displaying of IME preedit
+  at the end of window if IME preedit is very long
+  because the rendering does not IME preedit overflow the window
+  and does not wrap IME preedit to the next line.
+  "System" rendering can be useful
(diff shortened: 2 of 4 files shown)


config/src/config.rs
```diff
@@ -497,6 +497,8 @@ pub struct Config {
     pub use_ime: bool,
     #[serde(default)]
     pub xim_im_name: Option<String>,
+    #[serde(default)]
```
wez, line 500:
```suggestion
    #[dynamic(default)]
```

config/src/config.rs
```diff
@@ -1512,3 +1514,17 @@ impl DroppedFileQuoting {
         }
     }
 }
+
+#[derive(Debug, Deserialize, Serialize, Clone, Copy, PartialEq, Eq)]
```
wez, line 1518:
```suggestion
#[derive(Debug, FromDynamic, ToDynamic, Clone, Copy, PartialEq, Eq)]
```

kumattau:
First implementation is finished, but I think the followings need to be discussed:

* wezterm should support OverTheSpot or not.
* config key (ime_input_style) and values (OnTheSpot and OverTheSpot) are appropriate names or not

wez (commented):
Thanks for this!
I think this sounds reasonable.
Both the name `ime_input_style` and the names of the enum variants sound OK to me.

Note that in `main` the config struct no longer uses serde but uses the new `wezterm-dynamic` crate instead; that speeds up compilation times significantly.  I've made some inline suggestions on how to adapt this PR to work with that.

kumattau:
@wez 

Thank you for comment.

> I think this sounds reasonable.

Glad to hear that. This PR work will go ahead.

> Note that in main the config struct no longer uses serde but uses the new wezterm-dynamic crate instead; that speeds up > compilation times significantly. I've made some inline suggestions on how to adapt this PR to work with that.

Thank you for the suggestions. I will rebase master and fix them.

> Both the name ime_input_style and the names of the enum variants sound OK to me.

IME input style can be explained in the aspect of behavior and implementation of preedit area.

* behavior
  * OnTheSpot: preedit area is inserted in original text. preedit area does not hide original text
  * OverTheSpot: preedit area is over original text. preedit area hides original text

* implementaion
  * OnTheSpot: preedit area is rendered by the application (or GUI widget). it cannot be rendered by IME system
  * OverTheSpot: preedit area can be rendered by IME system

current wezterm is that IME behaves OverTheSpot by OnTheSpot implementaion.

So, The following name might also be suggested focused on the implementation.

* ime_preedit_rendering
  * builtin
  * system

How do you like it ?

wez:
> So, The following name might also be suggested focused on the implementation.
> 
> * ime_preedit_rendering
>   
>   * builtin
>   * system
> 
> How do you like it ?

I think that's ok!

kumattau:
I have finished update. Please review after github checks.

* update config: ime_preedit_rendering = "Builtin" or "System"
* add docs

kumattau:
Sorry, I failed commit of windows. I fixed it.

wez:
Thanks!

PR #2013, opened 2022-05-20
Fix: changing background color for Dark+ theme

saccarosium:
Fixing the background color since every instance of the theme is used.

### DIFF
--- a/assets/colors/Dark+.toml
+++ b/assets/colors/Dark+.toml
@@ -1,7 +1,7 @@
 # Dark+
 [colors]
 foreground = "#cccccc"
-background = "#0e0e0e"
+background = "#1e1e1e"
 cursor_bg = "#ffffff"
 cursor_border = "#ffffff"
 cursor_fg = "#000000"


wez:
Thanks for digging in!

That file is synchronized from data in https://github.com/mbadolato/iTerm2-Color-Schemes and this edit will be lost when I next re-sync. Can you get a PR in over there first, and then we can sync it back?

https://github.com/mbadolato/iTerm2-Color-Schemes/blob/master/schemes/Dark%2B.itermcolors is the source file, then you'd need to run https://github.com/mbadolato/iTerm2-Color-Schemes/blob/master/tools/update_all.py and submit the changes as a PR to that repo.

saccarosium:
Ok, I will do that thanks

saccarosium:
I've notice that there is only one version of gruvbox in the repos, also I don't
see some famous colorschemes. I would like to add them. Do you know how to
submit other colorscheme and automatically create colorschemes to the others
terminals?

wez:
Thanks!

PR #2022, opened 2022-05-23
Fix IME candidate window position on macOS

kumattau:
This PR is to fix IME candidate window position on macOS.
Similar to #1976

before:
スクリーンショット 2022-05-24 6 11 27

after:
スクリーンショット 2022-05-24 6 13 06


kumattau:
@wez
This PR is ready for review.
The check failure is caused by CI environment problem.

Issue #2023, opened 2022-05-23
Directional gradient with angle


erf:
**Is your feature request related to a problem? Please describe.**

I'd like to express a gradient that moves diagonally from top left to bottom right linearly across the window, like expressed in [these examples](https://digitalsynopsis.com/design/beautiful-gradient-color-palettes/).

Or more generally could be expressed with a direction angle.

**Describe the solution you'd like**

``` Lua
window_background_gradient = {
		orientation = { Direction = { angle = 135 } }
}
```

This would make a gradient in a direction from start angle and through the window where, where the normal on that direction vector will always have the same color.

**Describe alternatives you've considered**

Radial orientation or orientation "Diagonally", but i think a "Direction" angle (or radian) expresses this better.

**Additional context**

Thanks so much for this gradient ! My terminal is looking pretty sweet now :)

wez:
I'd like to see this. I think it is conceptually simple, if you have a good handle on geometry. I'd love to see a PR!

erf:
Cool! Super busy at the moment, but i can take a look at it when i find time and maybe brush off some linear algebra dust

erf:
Fixed in #2038

PR #2038, opened 2022-05-26
Linear gradient

erf:
I've added a new **Directional gradient** as suggested in #2023. I've also included a example to the [window_background_gradient.md](docs/config/lua/config/window_background_gradient.md).

It's a linear gradient which can be rotated around the window center. Using radians per now. Should i change this to degrees?

You should now be able to add diagonal (or similar) gradients like in [this example](https://digitalsynopsis.com/design/beautiful-gradient-color-palettes/).

Here's the screenshot I've used in the example:
directional-gradient

This is my first time using Rust, but the other examples made it quite easy to implement. Suggestions are welcomed.


docs/config/lua/config/window_background_gradient.md
```diff
@@ -68,6 +69,24 @@ Gradients are implemented using the `colorgrad` crate.
 Take a look at <https://github.com/mazznoer/colorgrad-rs#using-web-color-format>
 for some usage examples and additional information about gradients.
 
+## Directional gradient:
+
```
wez, line 73:
```suggestion

*Since: nightly builds only*

```

wez (commented):
This looks good, thank you!

I think the angle should be specified in degrees in the config, as that is more of a lay-person's unit, and is easier to intuit what a given degree value means than the equivalent in radians!

regarding the name `Directional`, what do you think about maybe renaming it to `Linear`?
That feels closer to the CSS equivalent: https://developer.mozilla.org/en-US/docs/web/css/gradient/linear-gradient

erf:
I agree. Will fix :)

erf:
Renamed to `Linear` and use degrees now. Also added "Only in nightly" to doc.

docs/config/lua/config/window_background_gradient.md
```diff
@@ -68,6 +69,26 @@ Gradients are implemented using the `colorgrad` crate.
 Take a look at <https://github.com/mazznoer/colorgrad-rs#using-web-color-format>
 for some usage examples and additional information about gradients.
 
+## Linear gradient:
+
+*Since: nightly builds only*
+
+A linear gradient follows a linear path across the window. It can be rotated
+around the window center. The start angle is at the center left and rotated
```
wez, line 77: The "start angle is at the center left" portion of this feels hard to understand.

Can we say something like this, adjusted to match up with the actual implementation, so that it is easier to see what a given value might result in:

> `0` degrees is equivalent to `Horizontal` with the gradient moving from left to right. `90` degrees is equivalent to `Vertical` with the gradient moving from top to bottom. `180` degrees is equivalent to `Horizontal` but with the gradient moving from right to left.  `270` degrees is equivalent to `Vertical` but with the gradient going from bottom to top.
> Negative degrees are equivalent to going counter clockwise, so `-45` is equivalent to `270` degrees and results in a gradient that moving from the top left corner down to the bottom right corner.

wez:
Thanks!

PR #2071, opened 2022-06-03
Quickselect: Add support for whitespaces in paths

hugobidois:
Add support for whitespaces in paths when using Quick Select Mode.

Useful on Windows for paths like `/mnt/c/Program Files/WezTerm`


bew:
This is dangerous I think, AFAIU it would make:
`foo/bar foo/baz`
match as a single path instead of two?

And the whole thing is pretty hard: how would it know when to consider it a single path or multiple paths ?

wez:
I'm going to close this for the reasons stated above; thanks again!

Issue #2063, opened 2022-06-01
Window content does not resize on kwin "Quick Tile"


pjones123:
### What Operating System(s) are you seeing this problem on?

Linux X11

### WezTerm version

20220528-212940-84842aaa

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

When resizing the window with kwin keyboard shortcuts for "Quick Tile Window to the [direction]", the contents of the window do not change to reflect the new window size. The window border (shown by kwin) is visible, but the window is transparent in areas that do not overlap the previous position, but the contents remain unchanged in areas that overlap the previous position. The issue also sometimes occurs when launching the application, where the entire window contents are not visible until updated by mouse interaction. I have been unable to determine the conditions under which the issue occurs in this way, but it appears to happen roughly 50% of the time.

This means that using "Quick Tile Window to the Bottom" after it is correctly showing tiled to the left, the bottom right corner is transparent, and the bottom left corner shows the bottom half of the terminal.

The window immediately updates to its new position and size after movement/resize using the mouse, or by changing focus to another window and then back to the terminal.

### To Reproduce

Launch the application from KDE plasma (on X11, wayland untested) using kwin. Use configured keyboard shortcuts to invoke kwin's "Quick Tile Window to [direction]".

### Configuration

no config

### Expected Behavior

The contents of the window should immediately update to reflect the window's new size and/or position, as it does when changed by the mouse.

### Logs

21:05:12.732  INFO   wezterm_gui::termwindow > OpenGL initialized! NVIDIA GeForce GTX 960/PCIe/SSE2 4.6.0 NVIDIA 515.43.04 is_context_loss_possible=false wezterm version: 20220528-212940-84842aaa

### Anything else?

"Quick Tile Window to the Left" after launch:
![image](https://user-images.githubusercontent.com/64071912/171492567-63635eb4-467f-4a94-8fb0-56e9d540933d.png)
"Quick Tile Window to the Bottom" after being tiled to the left:
![image](https://user-images.githubusercontent.com/64071912/171492653-cb2758fe-2027-4381-ae26-57a0855f0a45.png)

pjones123:
Here's the output from that. I've added some annotations to show what I did to cause the behaviour.
```
[...]

Issue #1992, opened 2022-05-15
Wezterm does not always notice window resize


yuttie:
### What Operating System(s) are you seeing this problem on?

Linux X11

### WezTerm version

20220514-175933-64b6e63f

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

Sometimes wezterm does not recognize its window resize.
Especially toggling fullscreen mostly does not work while resizing horizontally or vertically in a incremental fashion works a bit better.
Please watch my screen recording.

It seems that continuously (frequently) changing window size makes wezterm notice resize events, e.g., by using mouse or by holding keys for shrinking/enlarging a window or toggling fullscreen.

I tried both my window manager's key binding and wezterm's Alt+Enter to toggle fullscreen state.

I'm using i3, specifically i3-gaps v4.20.1, as the window manager on Gentoo Linux.

### To Reproduce

Launch a wezterm and resize its window.

### Configuration

no config

### Expected Behavior

Resizing a terminal window should affect the terminal's content area size.

### Logs

[wezterm-20220514-175933-64b6e63f.log](https://github.com/wez/wezterm/files/8694374/wezterm-20220514-175933-64b6e63f.log)

https://user-images.githubusercontent.com/158553/168458998-ad2f24ae-ca59-4995-b07f-c5f5cd15abdf.mp4

### Anything else?

I bisected at release level and found that `WezTerm-20210502-154244-3f7122cb-Ubuntu16.04.AppImage` worked much better than `WezTerm-20210814-124438-54e29167-Ubuntu16.04.AppImage`.
Although the former do fail to response to fullscreen toggle but much less frequently.

I've checked the release notes of the latter, it seems that the most relevant issue is only #695.

yuttie:
Today, I built another set of binaries of version `20220517-184546-9b5c8878` on an Arch Linux machine, and rsync-ed them to a different machine, which is running Gentoo Linux.

I found the reported issue only reproduced on the Gentoo machine and any resize-related problems were found on the Arch Linux machine.

I'm using i3-gaps of version 4.20.1 on both machines.  So, some other difference(s) must be causing the issue.  I will try different settings on Gentoo side.

yuttie:
As the first step, I tried different Nvidia driver versions available on Gentoo.
But none of them worked.

yuttie:
[...]

PR #2075, opened 2022-06-05
x11: allow configuration of repaint delay

pjones123:
Adds a configuration option, `focus_change_repaint_delay` to allow customisation of the delay added in 9b6329b454a51bc7e11ebaa447500ce049035833.

The default delay remains 100ms, and can be disabled by setting it to `0`, if the workaround is not required on the user's system.

refs: #2063
refs: #1992

### DIFF
--- a/window/src/os/x11/window.rs
+++ b/window/src/os/x11/window.rs
@@ -557,16 +557,19 @@ impl XWindowInner {
         // the window a short time after the focus event is processed in the
         // hope that it can observe the changed window properties and update
         // without the human needing to interact with the window.
-        let window_id = self.window_id;
-        promise::spawn::spawn(async move {
-            async_io::Timer::after(std::time::Duration::from_millis(100)).await;
-            XConnection::with_window_inner(window_id, |inner| {
-                inner.sure_about_geometry = false;
-                inner.invalidate();
-                Ok(())
-            });
-        })
-        .detach();
+        let delay = self.config.focus_change_repaint_delay;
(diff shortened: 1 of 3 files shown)


window/src/os/x11/window.rs
```diff
@@ -559,14 +561,46 @@ impl XWindowInner {
...
+                            &xcb::x::TranslateCoordinates {
+                                src_window: inner.conn().root,
+                                dst_window: window_id,
+                                src_x: 0,
+                                src_y: 0,
+                            }
+                    ))?;
+
+                    if inner.pos_x != coords.dst_x() || inner.pos_y != coords.dst_y() || width != new_geom.width() || height != new_geom.height() {
+                        inner.has_focus = Some(true);
```
wez, line 594: This may not be accurate; it's assuming that moving the window gains focus.
What I'd suggest instead is to set `inner.verify_focus = true` which will trigger some logic to query the focus state from the WM/X.
That could be done unconditionally as part of this spawn block, and avoid the need to track the window position.
In other words, just unconditionally set `sure_about_geometry = false` and `verify_focus = true`

pjones123: This method was accurate, but you're definitely right that it's better to do it this way. I'd assumed that the inaccurate focus events meant that querying focus directly wouldn't work, which obviously isn't the case.

I've reverted that commit and pushed what you suggested. I've opened another PR with these changes so as not to fill the history with this commit and its reversion.

PR #2086, opened 2022-06-06
Try implement line selection for CopyMode

### LINKED ISSUE #2083: Feature request: Line-based selection in copy mode akin to vim's Shift-V

Not related to a problem.

WezTerm currently supports vim-like selection keybindings in copy mode:

* `v` - Select a region from the initial cursor position to the ending cursor position (inclusive).
* `Ctrl-v` - Select a rectangular region bounded by the initial and ending cursor positions (inclusive).

It would be nice if WezTerm also supported something like vim's `Shift-v` selection mode, which is like `v`, except that the entire lines for both the initial and ending cursor positions are included in the selection.

I have not thought about any other approaches because it seems like WezTerm currently follows vim's visual selection conventions, so continuing to do so for line-based selection seems to make the most sense.

**Additional context**

An example of what it looks like in vim:

Screenshot 2022-06-06 135415

bew:
Hello @wez! It's been a while :)

This is a tentative PR to add line selection to copy mode.
I got Shift-v wired up, changed a few things to put the line coordinates, but I'm hitting a few walls..

See the various FIXMEs I left,
:point_right: I'd appreciate some help to understand them / guidance to resolve them.

Also, wondering, if we go with logical line selection, shouldn't `j` & `k` move by logical lines as well?
OR I could just start with physical line selection for now (smaller PR!)

Closes #2083

### DIFF
--- a/wezterm-gui/src/overlay/copy.rs
+++ b/wezterm-gui/src/overlay/copy.rs
@@ -322,14 +322,37 @@ impl CopyRenderable {
     fn select_to_cursor_pos(&mut self) {
         self.clamp_cursor_to_scrollback();
         if let Some(start) = self.start {
-            let start = SelectionCoordinate {
-                x: start.x,
-                y: start.y,
+            let cursor_is_above_start = self.cursor.y < start.y;
+            let start = if self.selection_mode == SelectionMode::Line {
+                SelectionCoordinate {
+                    x: if cursor_is_above_start {
+                        usize::max_value()
+                    } else {
+                        0
+                    },
+                    y: start.y,
+                }
+            } else {
+                SelectionCoordinate {
+                    x: start.x,
+                    y: start.y,
+                }
             };
 
-            let end = SelectionCoordinate {
-                x: self.cursor.x,
-                y: self.cursor.y,
+            let end = if self.selection_mode == SelectionMode::Line {
+                SelectionCoordinate {
+                    x: if cursor_is_above_start {
+                        0
+                    } else {
+                        usize::max_value()
+                    },
+                    y: self.cursor.y,
+                }
+            } else {
+                SelectionCoordinate {
+                    x: self.cursor.x,
+                    y: self.cursor.y,
+                }
             };
 
             self.adjust_selection(start, SelectionRange { start, end });
@@ -712,18 +735,6 @@ impl CopyRenderable {
         self.select_to_cursor_pos();
     }
 
-    fn toggle_selection_by_cell(&mut self) {
-        if self.start.take().is_none() {
-            let coord = SelectionCoordinate {
-                x: self.cursor.x,
-                y: self.cursor.y,
-            };
-            self.start.replace(coord);
-            self.select_to_cursor_pos();
-            self.selection_mode = SelectionMode::Cell;
-        }
-    }
-
     fn set_selection_mode(&mut self, mode: &Option<SelectionMode>) {
         match mode {
             None => {
@@ -736,9 +747,9 @@ impl CopyRenderable {
                         y: self.cursor.y,
                     };
                     self.start.replace(coord);
-                    self.select_to_cursor_pos();
                 }
                 self.selection_mode = *mode;
+                self.select_to_cursor_pos();
             }
         }
     }
@@ -810,7 +821,6 @@ impl Pane for CopyOverlay {
                     MoveToViewportMiddle => render.move_to_viewport_middle(),
                     MoveToScrollbackTop => render.move_to_top(),
                     MoveToScrollbackBottom => render.move_to_bottom(),
-                    ToggleSelectionByCell => render.toggle_selection_by_cell(),
                     MoveToStartOfLineContent => render.move_to_start_of_line_content(),
                     MoveToEndOfLineContent => render.move_to_end_of_line_content(),
                     MoveToStartOfLine => render.move_to_start_of_line(),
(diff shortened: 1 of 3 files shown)


wezterm-gui/src/overlay/copy.rs
```diff
@@ -697,6 +723,21 @@ impl CopyRenderable {
...
+        //   redraw
+        // else:
+        //   if mode != selection_mode
+        //     set selection_mode to mode
+        //   else:
+        //     stop selecting text
+        //   redraw
+        //
+        // TODO?: rename to 'apply_selection_mode' ?
+        // (since the behavior would be more involved that simply 'set')
```
bew, line 739: note: could be done in a later PR

wezterm-gui/src/overlay/copy.rs
```diff
@@ -708,9 +749,9 @@ impl CopyRenderable {
                         y: self.cursor.y,
                     };
                     self.start.replace(coord);
-                    self.select_to_cursor_pos();
                 }
                 self.selection_mode = *mode;
+                self.select_to_cursor_pos();
```
bew, line 754: note: this change is necessary to be able to immediately see selection mode changes (otherwise the change is only visible on next movement)

wezterm-gui/src/overlay/copy.rs
```diff
@@ -1160,12 +1201,23 @@ pub fn copy_key_table() -> KeyTable {
...
             WKeyCode::Char('v'),
             Modifiers::NONE,
-            KeyAssignment::CopyMode(CopyModeAssignment::ToggleSelectionByCell),
+            KeyAssignment::CopyMode(CopyModeAssignment::SetSelectionMode(Some(
+                SelectionMode::Cell,
+            ))),
+        ),
+        (
+            WKeyCode::Char('V'),
+            Modifiers::NONE, // FIXME: Seems to work with NONE, but why G also have SHIFT ?
```
wez, line 1217: Some platforms have SHIFT set when the key is uppercase, but some do not :-/

wezterm-gui/src/overlay/copy.rs
```diff
@@ -322,18 +322,43 @@ impl CopyRenderable {
...
+                    _ => start.x,
+                },
                 y: start.y,
             };
 
             let end = SelectionCoordinate {
-                x: self.cursor.x,
+                x: match self.selection_mode {
+                    // here we'd use LineColumn::LogicalEnd
+                    SelectionMode::Line => 50, // FIXME: remove arbitrary value!
```
wez, line 345: if you want to represent the end of any line, you could use `usize::max_value()` to get a number that is bigger than any practical line length

wezterm-gui/src/overlay/copy.rs
```diff
@@ -322,18 +322,43 @@ impl CopyRenderable {
...
+            dbg!(self.selection_mode);
+            dbg!(start);
+            dbg!(end);
+
+            // FIXME: when end is above start, the x of start/end should be swapped..
+            // => how to properly do that?
 
             self.adjust_selection(start, SelectionRange { start, end });
         } else {
+            // FIXME: need to handle this as well?
```
wez, line 360: This is the case where there is no start for the selection.
I don't think there is anything more you need to do for this case

wezterm-gui/src/overlay/copy.rs
```diff
@@ -322,18 +322,43 @@ impl CopyRenderable {
     fn select_to_cursor_pos(&mut self) {
         self.clamp_cursor_to_scrollback();
         if let Some(start) = self.start {
+            // FIXME: how to get the line length from here??
```
wez, line 325: I don't think you need the length; you can use eg: `usize::max_value()`

wez (commented):
Sorry for taking a while to get around to reviewing this!
I'd suggest keeping this PR focused on just physical line selection for now.
A couple of comments inline!

wezterm-gui/src/overlay/copy.rs
bew: Weird.. well, I added the mapping for both

bew:
Thanks for the `usize::max_value()` hint, I managed to make it work \o/

Let me know when you think it's good, and I'll squash the commits to have a cleaner merge

wezterm-gui/src/overlay/copy.rs
```diff
@@ -322,14 +322,37 @@ impl CopyRenderable {
     fn select_to_cursor_pos(&mut self) {
         self.clamp_cursor_to_scrollback();
         if let Some(start) = self.start {
-            let start = SelectionCoordinate {
-                x: start.x,
-                y: start.y,
+            let cursor_is_above_start = self.cursor.y < start.y;
+            let start = if matches!(self.selection_mode, SelectionMode::Line) {
```
wez, line 326: Since `SelectionMode` impls `Eq`, this can be simplified:

```suggestion
            let start = if self.selection_mode == SelectionMode::Line {
```

wezterm-gui/src/overlay/copy.rs
```diff
@@ -322,14 +322,37 @@ impl CopyRenderable {
...
+                SelectionCoordinate {
+                    x: start.x,
+                    y: start.y,
+                }
             };
 
-            let end = SelectionCoordinate {
-                x: self.cursor.x,
-                y: self.cursor.y,
+            let end = if matches!(self.selection_mode, SelectionMode::Line) {
```
wez, line 342:
```suggestion
            let end = if self.selection_mode == SelectionMode::Line {
```

wez:
re: `error: linker `cc` not found`, I see that from time to time in the CI: there is something flakey about the docker images where not all of the packages get installed successfully. :-/  You can ignore it in the context of this PR!

wez:
Thanks!

Issue #2089, opened 2022-06-07
Mouse selection includes the first character of the next line


TomiBelan:
### What Operating System(s) are you seeing this problem on?

Windows

### WezTerm version

wezterm 20220605-063531-e7258e0e

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

When you drag the mouse downwards along the left window edge, it also selects the first character from the line where you stop dragging.

### To Reproduce

- `printf 'ABCD\nEFGH\n'`
- drag the mouse from left of "A" to left of "E"
- the selected string is: "ABCD\nE"

![image](https://user-images.githubusercontent.com/1022278/172065711-7b49f8ca-77ab-409e-ab3d-ebaa295c4c45.png)

### Configuration

no config

### Expected Behavior

I think it makes more sense to only select the first line. The clipboard should not contain "E".

If users want to select "ABCD\nE", they'll have to drag the mouse from left of "A" to between the center of "E" and the center of "F". I think this is an unusual case so I don't think a config option is needed. (https://xkcd.com/1172/)

By analogy, when you start on other X-coordinates and move one line Y-downwards, it selects a 5 character string, not 6. E.g. dragging vertically from between (the centers of) "B" and "C" to between "F" and "G" selects "CD\nEF".

### Anything else?

By the way, I also noticed that wezterm strips all newlines from the start and end of the selected text. Personally I'd prefer to select "ABCD\n" rather than "ABCD", but I'm ok with both, and I guess other users are the opposite so maybe this could be configurable. I can file a separate feature request if you'd like. Just as long as it isn't "ABCD\nE".

wez:
This should be fixed now in `main`.

It typically takes about an hour before fixes are available as nightly builds for all platforms.  Linux builds are the fastest to build and are often available within about 20 minutes.  Windows and macOS builds take a bit longer.

Please take a few moments to try out the fix and let me know how that works out.

If you prefer to use packages provided by your distribution or package manager of choice and don't want to replace that with a nightly download, keep in mind that you can download portable packages (eg: a `.dmg` file on macOS, a `.zip` file on Windows and an `.AppImage` file on Linux) that can be run without permanently installing or replacing an existing package, and can then simply be deleted once you no longer need them.

If you are eager and can [build from source](https://wezfurlong.org/wezterm/install/source.html) then you may be able to try this out more quickly.

[...]

PR #2106, opened 2022-06-10
Fix mouse selection including the first character of the next line

Funami580:
fixes the main part of #2089, but not "wezterm strips all newlines from the start and end of the selected text"


Funami580:
This may introduce a bug in the future, though. I'm not sure.
Let's say when you copy a text it won't strip newlines.
If you have a selection like this
```
a|bcd
ef
|te
```
where `|` denote the start and end of the selection, the copied text should be `bcd\nef\n`. But with this commit the copied text might be `bcd\nef` without the trailing newline?

wezterm-gui/src/selection.rs
```diff
@@ -77,10 +78,12 @@ impl SelectionRange {
                     start: SelectionCoordinate {
                         x: 0,
                         y: logical.first_row,
+                        ..Default::default()
```
wez, line 81: Most of this PR is adding in this `..Default::default()` which makes it more difficult to see where the actual logic changes are, and makes the code a bit more verbose in general.

My suggestion is to add a constructor as a shortcut for the common case:

```rust
impl SelectionCoordinates {
   pub const fn x_y(x: usize, y: StableRowIndex) -> Self {
       Self {
            x,
            y,
            treat_zero_in_x_as_before_first_cell: false
        }
    }
}
```

then most of the changes turn into something a bit more compact like:

```rust
   start: SelectionCoordinate::x_y(0, logical.first_row),
```

wezterm-gui/src/selection.rs
```diff
@@ -44,6 +44,7 @@ impl Selection {
 pub struct SelectionCoordinate {
     pub x: usize,
     pub y: StableRowIndex,
+    pub treat_zero_in_x_as_before_first_cell: bool,
```
wez, line 47: Hmm, I'm "worried" that having this be separated from `x` might be prone to bugs when interpreting `x`.

What do you think about changing the type of `x` to reflect this; something like this?

```rust
enum SelectionX {
    /// Zero-based cell index
    Cell(usize),
    /// Before the 0th cell
    BeforeZero,
}
```

```rust
  pub x: SelectionX,
```

The big advantage of this is that the compiler will now be able to tell us where in the code we need to consider these cases.

wez (commented):
Thanks!
I have some questions/suggestions!

wezterm-gui/src/selection.rs
Funami580: Done

wezterm-gui/src/termwindow/selection.rs
```diff
@@ -83,13 +83,19 @@ impl super::TermWindow {
...
+                        ..Default::default()
+                    });
                 self.selection(pane.pane_id()).origin = Some(origin);
                 self.selection(pane.pane_id()).rectangular = mode == SelectionMode::Block;
 
                 // Compute the start and end horizontall cell of the selection.
                 // The selection extent depends on the mouse cursor position in relation
                 // to the origin.
+                let (mut start_zero_before_first_cell, mut end_zero_before_first_cell) =
+                    (false, false);
```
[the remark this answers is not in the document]
Funami580: Not needed anymore

wezterm-gui/src/selection.rs
Funami580: Good idea, done

wezterm-gui/src/termwindow/selection.rs
```diff
@@ -108,30 +114,62 @@ impl super::TermWindow {
                         // screen, so this causes a visual cell on the screen to be selected when
                         // the mouse moves over 50% of its width, which effectively means the next
                         // cell is being reported here, hence it's excluded
-                        (origin.x, x.saturating_sub(1))
+                        let end = if let Some(x) = x.checked_sub(1) {
```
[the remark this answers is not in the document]
Funami580: Done

wezterm-gui/src/selection.rs
```diff
@@ -39,13 +39,153 @@ impl Selection {
...
+
+pub trait OneBased<T> {
+    /// Calculate 0-based X coordinate given a 1-based input
+    fn one_based(x: T) -> Self;
+}
+
+impl OneBased<Self> for SelectionX {
+    fn one_based(x: Self) -> Self {
+        match x {
+            Self::Cell(x) => match x.checked_sub(1) {
```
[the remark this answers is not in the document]
Funami580: I pushed an update. Is it better now?

wezterm-gui/src/selection.rs
```diff
@@ -100,7 +224,7 @@ impl SelectionRange {
                 Ordering::Greater => return Ordering::Greater,
                 // If the zone starts on the same line then check that the
                 // x position is within bounds
-                Ordering::Equal => match zone.start_x.cmp(&start.x) {
+                Ordering::Equal => match zone.start_x.partial_cmp(&start.x).unwrap() {
```
wez, line 227: The unwrap here feels dangerous; can you adjust this match block to deal with the `None` case, or just use `cmp`?

wezterm-gui/src/selection.rs
```diff
@@ -110,7 +234,7 @@ impl SelectionRange {
                 Ordering::Less => Ordering::Less,
                 // If the zone ends on the same line then check that the
                 // x position is within bounds
-                Ordering::Equal => match zone.end_x.cmp(&start.x) {
+                Ordering::Equal => match zone.end_x.partial_cmp(&start.x).unwrap() {
```
wez, line 237: also here

wez (commented):
Thanks!

wezterm-gui/src/selection.rs
Funami580: I used something similar to `cmp`. Is it fine like this?

wez:
I think it can be made a bit easier to follow by removing some of the additional traits.  This diff is against your rev 40446ebeaf4f873b59fe1f4a024bbae1972432d1

```diff
diff --git a/wezterm-gui/src/selection.rs b/wezterm-gui/src/selection.rs
index d51fee1e3..eadc30871 100644
--- a/wezterm-gui/src/selection.rs
+++ b/wezterm-gui/src/selection.rs
@@ -70,6 +70,21 @@ impl SelectionX {
             Self::BeforeZero => Self::BeforeZero,
         }
     }
+
+    /// Creates a range from the left x coordinate (inclusive)
+    /// to the right x coordinate (exclusive)
+    pub fn range(self, rhs: Self) -> Range<usize> {
+        match self {
+            Self::Cell(left) => match rhs {
+                Self::Cell(right) => left..right,
+                Self::BeforeZero => 0..0,
+            },
+            Self::BeforeZero => match rhs {
+                Self::Cell(right) => 0..right,
+                Self::BeforeZero => 0..0,
+            },
+        }
+    }
 }
 
 impl Default for SelectionX {
@@ -94,12 +109,8 @@ impl PartialEq<SelectionX> for usize {
     }
 }
 
-pub trait SelectionXOrd<Rhs = Self> {
-    fn cmp_selection(&self, other: &Rhs) -> Ordering;
-}
-
-impl SelectionXOrd for SelectionX {
-    fn cmp_selection(&self, other: &Self) -> Ordering {
+impl Ord for SelectionX {
+    fn cmp(&self, other: &Self) -> Ordering {
         match self {
             Self::Cell(x1) => match other {
                 Self::Cell(x2) => x1.cmp(x2),
@@ -115,67 +126,19 @@ impl SelectionXOrd for SelectionX {
 
 impl PartialOrd for SelectionX {
     fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
-        Some(self.cmp_selection(other))
-    }
-}
-
-impl SelectionXOrd<usize> for SelectionX {
-    fn cmp_selection(&self, other: &usize) -> Ordering {
-        match self {
-            Self::Cell(x) => x.cmp(other),
-            Self::BeforeZero => Ordering::Less,
-        }
+        Some(self.cmp(other))
     }
 }
 
 impl PartialOrd<usize> for SelectionX {
     fn partial_cmp(&self, other: &usize) -> Option<Ordering> {
-        Some(self.cmp_selection(other))
-    }
-}
-
-impl SelectionXOrd<SelectionX> for usize {
-    fn cmp_selection(&self, other: &SelectionX) -> Ordering {
-        other.cmp_selection(self).reverse()
+        Some(self.cmp(&Self::Cell(*other)))
     }
 }
 
 impl PartialOrd<SelectionX> for usize {
     fn partial_cmp(&self, other: &SelectionX) -> Option<Ordering> {
-        Some(self.cmp_selection(other))
-    }
-}
-
-pub trait SelectionXRange<T = Self> {
-    // Creates a range from the left x coordinate (inclusive)
-    // to the right x coordinate (exclusive)
-    fn range(self, rhs: T) -> Range<usize>;
-}
-
-impl SelectionXRange for SelectionX {
-    fn range(self, rhs: Self) -> Range<usize> {
-        match self {
-            Self::Cell(left) => match rhs {
-                Self::Cell(right) => left..right,
-                Self::BeforeZero => 0..0,
-            },
-            Self::BeforeZero => match rhs {
-                Self::Cell(right) => 0..right,
-                Self::BeforeZero => 0..0,
-            },
-        }
-    }
-}
-
-impl SelectionXRange<usize> for SelectionX {
-    fn range(self, rhs: usize) -> Range<usize> {
-        self.range(Self::Cell(rhs))
-    }
-}
-
-impl SelectionXRange<SelectionX> for usize {
-    fn range(self, rhs: SelectionX) -> Range<usize> {
-        SelectionX::Cell(self).range(rhs)
+        SelectionX::Cell(*self).partial_cmp(other)
     }
 }
 
@@ -246,7 +209,7 @@ impl SelectionRange {
                 Ordering::Greater => return Ordering::Greater,
                 // If the zone starts on the same line then check that the
                 // x position is within bounds
-                Ordering::Equal => match zone.start_x.cmp_selection(&start.x) {
+                Ordering::Equal => match SelectionX::Cell(zone.start_x).cmp(&start.x) {
                     Ordering::Greater => return Ordering::Greater,
                     Ordering::Equal | Ordering::Less => {}
                 },
@@ -256,7 +219,7 @@ impl SelectionRange {
                 Ordering::Less => Ordering::Less,
                 // If the zone ends on the same line then check that the
                 // x position is within bounds
-                Ordering::Equal => match zone.end_x.cmp_selection(&start.x) {
+                Ordering::Equal => match SelectionX::Cell(zone.end_x).cmp(&start.x) {
                     Ordering::Less => Ordering::Less,
                     Ordering::Equal | Ordering::Greater => Ordering::Equal,
                 },
@@ -383,10 +346,10 @@ impl SelectionRange {
                 }
             } else if row == norm.end.y {
                 // last line of multi-line
-                0.range(norm.end.x.saturating_add(1))
+                SelectionX::Cell(0).range(norm.end.x.saturating_add(1))
             } else if row == norm.start.y {
                 // first line of multi-line
-                norm.start.x.range(usize::max_value())
+                norm.start.x.range(SelectionX::Cell(usize::max_value()))
             } else {
                 // some "middle" line of multi-line
                 0..usize::max_value()
```

wez:
Heads up: https://github.com/wez/wezterm/pull/2086 likely conflicts with this a little

wez:
Looks good, thank you!

PR #2111, opened 2022-06-12
x11: allow configuration of repaint delay, fix loss of focus

pjones123:
Adds a configuration option, `focus_change_repaint_delay` to allow customisation of the delay added in 9b6329b454a51bc7e11ebaa447500ce049035833.

The default delay remains 100ms, and can be disabled by setting it to `0`, if the workaround is not required on the user's system.

The focus events used to trigger a query of geometry are inaccurate, so the window's focus state should also be queried.

refs: #2063
refs: #1992

### DIFF
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -648,6 +648,9 @@ pub struct Config {
 
     #[dynamic(default)]
     pub quote_dropped_files: DroppedFileQuoting,
+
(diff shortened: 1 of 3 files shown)


wez:
Thanks!  I renamed the option to `x11_focus_change_repaint_delay_ms` and tweaks the docs a little!

PR #2115, opened 2022-06-12
Fix minor typo in pad_right.md doc

frabjous:
`wezterm.pad_righ` in the example should be `wezterm.pad_right`.

There's another problem, shared with the pad_left page, that `"o  "` is rendered in the HTML with just one space, collapsing the two spaces to one, which is very misleading given what the example for this function does. It needs to be in a `<pre>` tag rather than just a `<code>` tag (or have the equivalent css applied), but I don't know how to fix that from markdown.


wez:
Thanks!

PR #2132, opened 2022-06-17
Add example for numeric URLs. Show use of literal strings to keep regex simple.

m000:
Small documentation addition.

### DIFF
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -1213,7 +1213,7 @@ fn default_initial_cols() -> u16 {
 fn default_hyperlink_rules() -> Vec<hyperlink::Rule> {
     vec![
         // URL with a protocol
-        hyperlink::Rule::new(r"\b\w+://(?:[\w.-]+)\.[a-z]{2,15}\S*\b", "$0").unwrap(),
+        hyperlink::Rule::new(r"\b\w+://[\w.-]+\.[a-z]{2,15}\S*\b", "$0").unwrap(),
         // implicit mailto link
         hyperlink::Rule::new(r"\b\w+@[\w-]+(\.[\w-]+)+\b", "mailto:$0").unwrap(),
         // file://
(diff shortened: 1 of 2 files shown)


wez:
Can you add an example of that the URL it matches looks like? I don't think numeric URL is a super widely used term so it's not immediately obvious!

bew:
I've never heard of that term either, I'm curious (and didn't find much info about them), what do you use them for?

m000:
Numeric URL == a URL with a numeric IP address as host. E.g. http://127.0.0.1:8000/ (popular for running [development servers locally](https://docs.djangoproject.com/en/4.0/intro/tutorial01/#the-development-server)), or http://192.168.1.1 ([popular default router address](https://www.techspot.com/guides/287-default-router-ip-addresses/))

I have updated the text to use the verbose phrasing if "numeric URLs" are not clear enough. Feel free to update again as you see fit.

m000:
Also, unless I'm missing something, I can see two issues with the "non-numeric" URL regex:
* The non-capturing parentheses seem redundant. I.e. `(?:[\\w.-]+)` could be changed to `[\\w.-]+` without issues.
* Because a domain suffix is required, the (also popular) http://localhost:8000 will not be captured.

I can update the PR to cover for these two cases, if you find them useful.

wez:
Thanks; could you add explicitly `http://127.0.0.1:8000/` to the comment in the docs to make it super clear to the reader?
I'd also love to see some more examples for other common cases!

m000:
I have explicitely listed 127.0.0.1. Changes detailed in commit message.

I didn't add an example for http://localhost:8000/ because I think this would be a bit confusing for a documentation page. Specifically, the regex that would capture such URLs would be `[[\b\w+://[\w.-]+\S*\b]]`. But this regex would also capture URLs that have TLD (first case) and URLs with IP addresses as host. So the resulting configuration would be weird. 

Perhaps switching to `[[\b\w+://[\w.-]+\S*\b]]` would be better rather than having two different cases anyway

I can't think of any other important case that is missing. In general, crafting the one regex to conquer all URLs [gets very quickly out of hand](https://mathiasbynens.be/demo/url-regex). So, a pragmatic choice has to be made, and any niche cases can be covered with additional rules by the user.

wez:
This looks great, thank you!

PR #2148, opened 2022-06-20
support drag and drop files for wayland

Funami580:


window/src/os/wayland/drag_and_drop.rs
```diff
@@ -0,0 +1,116 @@
...
+            DataDeviceEvent::Enter {
+                serial,
+                surface,
+                id,
+                ..
+            } => {
+                if let Some(offer) = id {
+                    offer.accept(serial, Some(URI_MIME_TYPE.to_string()));
+                    offer.set_actions(
+                        DndAction::None | DndAction::Copy | DndAction::Move,
```
wez, line 30: Does it make sense to allow `Move` here? It feels like that might be a bit of lie!  Perhaps also true of `Copy`.  It probably doesn't really matter, I'm just curious if this is necessary or if there are negative consequences

window/src/os/wayland/drag_and_drop.rs
```diff
@@ -0,0 +1,116 @@
...
+                        offer.finish();
+                        let copied_surface_id = *surface_id;
+                        self.offer = None;
+                        Some((copied_surface_id, read))
+                    } else {
+                        None
+                    }
+                } else {
+                    None
+                };
```
wez, line 59: It feels a little awkward to read with the different branches returning `None`. I would suggest re-phrasing this using a helper function and taking advantage of the `?` operator, something like:

```rust
struct SurfaceAndPipe {
   surface_id: u32,
   read: FileDescriptor,
}

/// Takes the current offer, if any, and initiates a receive into a pipe,
/// returning that surface and pipe descriptor.
fn create_pipe_for_drop(&mut self) -> Option<SurfaceAndPipe> {
  let offer = self.offer.take()?;
  let pipe = Pipe::new().map_err(|err| log::error!("Unable to create pipe: {:#}", err)).ok()?;
  offer.receive(URI_MIME_TYPE.to_string(), pipe.write.as_raw_fd());
  let read = pipe.read;
  offer.finish();
  Some(SurfaceAndPipe {
    surface_id: offer.surface_id,
    read
  })
}
```

Then this block might end up looking something like this:

```rust
                if let Some(res) = self.create_pipe_for_drop() {
                    std::thread::spawn(move || {
                        read_uris_from_pipe(res, window_id)
                    });
                 }
```

window/src/os/wayland/drag_and_drop.rs
```diff
@@ -0,0 +1,116 @@
...
+                                            match path {
+                                                Ok(path) => Some(path),
+                                                Err(_) => {
+                                                    log::error!(
+                                                        "while converting url to pathbuf: {}",
+                                                        line
+                                                    );
+                                                    None
+                                                }
+                                            }
```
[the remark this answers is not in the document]
Funami580: Done

window/src/os/wayland/drag_and_drop.rs
```diff
@@ -0,0 +1,116 @@
...
+use smithay_client_toolkit as toolkit;
+use std::os::unix::io::AsRawFd;
+use toolkit::reexports::client::protocol::wl_data_device::Event as DataDeviceEvent;
+use toolkit::reexports::client::protocol::wl_data_offer::WlDataOffer;
+use url::Url;
+use wayland_client::protocol::wl_data_device_manager::DndAction;
+
+#[derive(Default)]
+pub struct DragAndDrop {
+    offer: Option<(u32, WlDataOffer)>, // surface id, offer
```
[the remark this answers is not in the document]
Funami580: Done

window/src/os/wayland/drag_and_drop.rs
Funami580: Huh, to be honest, I am not too sure either. For example my file manager only provided `Copy` and `Move`, but no `None` action. So I just thought let's keep the restrictions relaxed. I would say `Copy` is a must, but I'm not too sure about `Move`. If you think I should remove it, I can do that.

Funami580:
Thanks for the review, I feel like I learned a lot regarding stylistic choices

window/src/os/wayland/drag_and_drop.rs
Funami580: Done

window/src/os/wayland/drag_and_drop.rs
```diff
@@ -0,0 +1,131 @@
...
+    fn create_pipe_for_drop(&mut self) -> Option<SurfaceAndPipe> {
+        let SurfaceAndOffer { surface_id, offer } = self.offer.take()?;
+        let pipe = Pipe::new()
+            .map_err(|err| log::error!("Unable to create pipe: {:#}", err))
+            .ok()?;
+        offer.receive(URI_MIME_TYPE.to_string(), pipe.write.as_raw_fd());
+        let read = pipe.read;
+        offer.finish();
+        Some(SurfaceAndPipe {
+            surface_id: surface_id,
```
wez, line 41:
```suggestion
            surface_id,
```

window/src/os/wayland/drag_and_drop.rs
```diff
@@ -0,0 +1,131 @@
...
+                    url.to_file_path()
+                        .map_err(|_| {
+                            log::error!(
+                                "Error converting url {} from line {} to pathbuf",
+                                url,
+                                line
+                            );
+                        })
+                        .ok()
+                }
```
wez, line 77: Can reduce nesting by using early return!

```suggestion
                if line.starts_with('#') || line.trim().is_empty() {
                    // text/uri-list: Any lines beginning with the '#' character
                    // are comment lines and are ignored during processing
                    return None;
                }
                    let url = Url::parse(line)
                        .map_err(|err| {
                            log::error!(
                                "Error parsing dropped file line {} as url: {:#}",
                                line,
                                err
                            );
                        })
                        .ok()?;
                    url.to_file_path()
                        .map_err(|_| {
                            log::error!(
                                "Error converting url {} from line {} to pathbuf",
                                url,
                                line
                            );
                        })
                        .ok()
```

wez (commented):
two minor tweaks and I think this is done!

window/src/os/wayland/drag_and_drop.rs
Funami580: Done

window/src/os/wayland/drag_and_drop.rs
Funami580: Done

wez:
Thanks!

PR #2150, opened 2022-06-20
Implement o & O to move cursor to other end of selection in copy mode

bew:
Demo :smiley: 

https://user-images.githubusercontent.com/9730330/174689403-4520b080-20f1-4a30-b39d-20e646cf877c.mp4

Inspired from vim:
```
							*v_o*
o			Go to Other end of highlighted text: The current
			cursor position becomes the start of the highlighted
			text and the cursor is moved to the other end of the
			highlighted text.  The highlighted area remains the
			same.

							*v_O*
O			Go to Other end of highlighted text.  This is like
			"o", but in Visual block mode the cursor moves to the
			other corner in the same line.  [...]
```

### DIFF
--- a/wezterm-gui/src/overlay/copy.rs
+++ b/wezterm-gui/src/overlay/copy.rs
@@ -606,8 +606,7 @@ impl CopyRenderable {
 
     fn move_to_selection_other_end_horiz(&mut self) {
         if self.selection_mode != SelectionMode::Block {
-            // This action makes sense ONLY in block mode
-            return;
+            return self.move_to_selection_other_end();
         }
         if let Some(old_start) = self.start {
             // Swap X coordinate of cursor & start of selection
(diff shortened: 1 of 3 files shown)


wezterm-gui/src/overlay/copy.rs
```diff
@@ -591,6 +591,35 @@ impl CopyRenderable {
...
+            self.cursor.y = old_start.y;
+            self.select_to_cursor_pos();
+        }
+    }
+
+    fn move_to_selection_other_end_horiz(&mut self) {
+        if self.selection_mode != SelectionMode::Block {
+            // This action makes sense ONLY in block mode
+            return;
+        }
```
wez, line 611: vim appears to treat this case as `move_to_selection_other_end`; should we do the same?

```suggestion
        if self.selection_mode != SelectionMode::Block {
            return self.move_to_selection_other_end();
        }
```

wez (commented):
Looks good, thanks!
Just one question!

bew:
Good catch 👌

PR #2154, opened 2022-06-21
github hyperlink config example

digitallyserviced:
adding a few of the nifty things from my config to docs/example pages

hyperlink example for github user/proj paths

### DIFF
--- a/docs/hyperlinks.md
+++ b/docs/hyperlinks.md
@@ -51,8 +51,10 @@ return {
       format = "https://example.com/tasks/?t=$1",
     },
     
-    -- Make github like username/project paths
-    -- inside quotes and without clickable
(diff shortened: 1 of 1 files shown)


docs/hyperlinks.md
```diff
@@ -49,6 +49,13 @@ return {
     {
       regex = [[\b[tT](\d+)\b]],
       format = "https://example.com/tasks/?t=$1",
+    },
+    
+    -- Make github like username/project paths
+    -- inside quotes and without clickable
```
wez, line 55: Can you add a couple of examples of the text that this will match here in the comment?

digitallyserviced: Hopefully not too lengthy but figured I should mention that full URL's should not be picked up if they are for other services or do not reflect a user/project like path.

wez:
Thanks!

PR #2164, opened 2022-06-23
Added winapi feature "shellapi" to procinfo

BigPapa314:
Fixes wezterm installation on windows using `cargo install --path ./wezterm` or `cargo install --path ./wezterm-gui` by adding the shellapi feature to the winapi crate dependency.

### DIFF
--- a/procinfo/Cargo.toml
+++ b/procinfo/Cargo.toml
@@ -18,6 +18,7 @@ winapi = { version = "0.3", features = [
     "memoryapi",
     "psapi",
     "processthreadsapi",
+    "shellapi",
     "tlhelp32",
 ]}
 


BigPapa314:
Hey wez, 

thanks for the fast reaction. You are doing an awesome job here.

I don't use every feature of wezterm, I simply want a terminal emulator that is capable of a **transparent background** with support of **code ligatures** (_there goes alacritty_) and **without window decorations**.

So it's ok for me to only compile wezterm.exe and wezterm-gui.exe and put them into `PATH`. 
Because that cannot be done from crates.io my workflow of installing wezterm is:
1. checkout the repository
2. `cargo install --path ./wezterm`
3. `cargo install --path ./wezterm-gui`

This compiles both binaries and copies them to `CARGO_HOME`. 

And stopped working quite some time ago with the error message:

> error[E0432]: unresolved import `winapi::um::shellapi`
  --> procinfo\src\windows.rs:18:17
   |
 18 | use winapi::um::shellapi::CommandLineToArgvW;
  |                 ^^^^^^^^ could not find `shellapi` in `um`

> For more information about this error, try `rustc --explain E0432`.
error: could not compile `procinfo` due to previous error

The current PR fixes this error.

I know I don't install wezterm the intended way, I just thought let's note the one line to fix my compilation issue.

Just to note, I do this procedure on my linux machines too, and it works fine for me.

I look forward to switching my package managers to **cargo** on linux and windows ;)

Greetings

wez:
Makes sense; thanks!

PR #2166, opened 2022-06-23
fix: properly restore cooked mode in windows

sullivan-sean:
Found a small typo (I think) in `set_cooked_mode` that redundantly turned off `ENABLE_WINDOW_INPUT` twice instead of turning off `ENABLE_MOUSE_INPUT` and `ENABLE_WINDOW_INPUT`


wez:
Looks good, thanks!

PR #2177, opened 2022-06-25
docs: Use new `wezterm.action.Action` syntax in rest of the docs

bew:


docs/config/lua/keyassignment/ExtendSelectionToMouseCursor.md
```diff
@@ -18,9 +22,13 @@ return {
 }
 ```
 
+!!! FIXME: remove this part? it doesn't work at the moment...
```
wez, line 25: Hmm. Try `ExtendSelectionToMouseCursor(nil)` ?

bew: Nope:
```
Error converting lua value returned by script /home/lesell_b/.config/wezterm/
wezterm.lua to Config struct: error converting Lua table to Config (Error
processing mouse_bindings.action (types: Config, Mouse) Expected a valid
`KeyAssignment` variant name as single key in object, but there are 0 keys)
```
note that I also tried the original syntax as previously documented, doesn't work

wez: [ee5206d](https://github.com/wez/wezterm/commit/ee5206db507f257776230fcc3c461734802b3a5a) should help; the original example doesn't work with that either, but the `act.ExtendSelectionToMouseCursor(nil)` suggestion does

bew: Okay, I didn't test that commit, but I rebased this PR, updated that doc (and removed the FIXME)

wez:
Many thanks!

PR #2195, opened 2022-06-27
Fix install docs for Ubuntu 22

foresterre:
The asset linked to for Ubuntu 22 is the Ubuntu 20 package. This commit fixes that by refering to the correct Ubuntu 22 asset instead.

Looking at https://github.com/wez/wezterm/blob/c29a513a441ce301838567d9cde6e1a75cb57eb9/ci/subst-release-info.py, that's all that is needed to fix the link.


wez:
Good catch, thanks!

PR #2199, opened 2022-06-28
Publish to Windows Package Managar (WinGet)

vedantmgoyal2009:
This action automatically generates manifests for WinGet Community Repository ([microsoft/winget-pkgs](https://github.com/microsoft/winget-pkgs)) and submits them.

Before merging this:
1. Please add a GitHub token with `public_repo` scope as a repository secret and rename the secret name in the workflow.
2. Fork [microsoft/winget-pkgs](https://github.com/microsoft/winget-pkgs) under @wez.
3. Publish releases using a PAT and not GitHub Actions.
image

### DIFF
--- a/.github/workflows/winget.yml
+++ b/.github/workflows/winget.yml
@@ -1,0 +1,14 @@
+name: Publish to WinGet
+on:
+  release:
+    types: [released]
+jobs:
+  publish:
+    runs-on: windows-latest # action can only be run on windows
+    steps:
+      - uses: vedantmgoyal2009/winget-releaser@latest
+        with:
+          identifier: wez.wezterm
+          version-regex: '.*'
+          installers-regex: '\.exe$'
+          token: ${{ secrets.WINGET_TOKEN }}


wez:
Thanks for this!

Please don't take this personally, but I don't feel comfortable running this action directly with my PAT. I feel more comfortable forking it and referencing that fork.

That said: I took a look through the action and amazed by the amount of code that it takes to produce a PR like https://github.com/microsoft/winget-pkgs/pull/64366/files

It feels like, for wezterm at least, that a relatively simple shell script that uses the `gh` CLI could be used to create those three files and create a PR.  Have you considered something like that in the past?

I'd feel much more comfortable having that simpler logic here in the wezterm repo as it reduces concerns about the PAT "escaping" and the code would be much easier to follow.

What do you think?

wez:
> Publish releases using a PAT and not GitHub Actions.

I'm not sure what this means; today I make releases by pushing a tag and leaving it up to my GH actions to create the actual release as a pre-release/draft.
Later, I manually set the release as released.

Is that compatible with this?

wez:
What do think about 40d346b58fe5359fbec3b1b0e6708721b0a4b82a which a first pass over integrating a winget upload more closely to the release asset upload?

vedantmgoyal2009:
I'm closing this pull request in favor of the script.

russellbanks:
I would still argue that the winget releaser action is preferable because it will require less effort on your part to maintain (new manifest versions, etc). It makes more sense to use the yamlcreate script rather than hardcode most of the values. Also, there are toggleable features with winget releaser, like only keeping the latest version, etc

wez:
https://github.com/microsoft/winget-pkgs/pull/68991 seems to have been submitted correctly

PR #2221, opened 2022-07-04
Add catppuccin colorschemes

vihu:
Notes
----
Adding the catppuccin colorscheme with all its variants.

The contents of the .toml files were found [here](https://github.com/catppuccin/wezterm/blob/main/catppuccin.lua).

Caveats
----
- Catppuccin also defines styles for tab_bar, visual bell etc. I have only included the colors here to stay consistent with the existing scheme.
- Some of the color choices are arguable, but I remained consistent with the variants of catppuccin regardless.

### DIFF
--- a/assets/colors/catppuccin_frappe.toml
+++ b/assets/colors/catppuccin_frappe.toml
@@ -1,0 +1,11 @@
+[colors]
+foreground = "#c6d0f5"
+background = "#303446"
+cursor_bg = "#f2d5cf"
+cursor_border = "#f2d5cf"
+cursor_fg = "#232634"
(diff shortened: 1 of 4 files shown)


wez:
wezterm's color scheme's are synced from https://github.com/mbadolato/iTerm2-Color-Schemes
If you could submit these color schemes over there, then we can sync them back over here!

vihu:
> wezterm's color scheme's are synced from https://github.com/mbadolato/iTerm2-Color-Schemes If you could submit these color schemes over there, then we can sync them back over here!

Certainly, thanks for pointing in the right direction.

PR #2233, opened 2022-07-09
win32: fix encoding of delete and backspace keys

kreudom:
When using the win32 keyboard encoding, the delete and backspace keys
are encoded with the wrong character codes compared to the native
terminal.

This causes issues in multiple terminal applications.

For example, when entering a command in the plain command line,
ctrl + backspace does not delete the previous word in wezterm and instead
functions like a regular backspace. Ctrl + delete removes the previous word
in wezterm, while it should do nothing.

In clink, delete, ctrl + delete and ctrl + backspace are all detected as the same
key and remove the previous word in the default configuration.

To debug this, I used the [ConEchoKey](https://github.com/microsoft/terminal/blob/main/src/tools/echokey/main.cpp) tool and compared the readings to
wezterm's implementation. The readings from the native terminal were
the following:

```
DEL	Down: 1 Repeat: 1 KeyCode: 0x2e ScanCode: 0x53 Char: \0 (0x0) KeyState: 0x120
CTRL + DEL	Down: 1 Repeat: 1 KeyCode: 0x2e ScanCode: 0x53 Char: \0 (0x0) KeyState: 0x128
ALT + DEL	Down: 1 Repeat: 1 KeyCode: 0x2e ScanCode: 0x53 Char: \0 (0x0) KeyState: 0x122
SHIFT + DEL	Down: 1 Repeat: 1 KeyCode: 0x2e ScanCode: 0x53 Char: \0 (0x0) KeyState: 0x130
CTRL + SHIFT + DEL	Down: 1 Repeat: 1 KeyCode: 0x2e ScanCode: 0x53 Char: \0 (0x0) KeyState: 0x138
CTRL + ALT + DEL	skipped
SHIFT + ALT + DEL	Down: 1 Repeat: 1 KeyCode: 0x2e ScanCode: 0x53 Char: \0 (0x0) KeyState: 0x132
CTRL + SHIFT + ALT + DEL	Down: 1 Repeat: 1 KeyCode: 0x2e ScanCode: 0x53 Char: \0 (0x0) KeyState: 0x13a

BACK	Down: 1 Repeat: 1 KeyCode: 0x8 ScanCode: 0xe Char: \b (0x8) KeyState: 0x20
CTRL + BACK	Down: 1 Repeat: 1 KeyCode: 0x8 ScanCode: 0xe Char: ⌂  (0x7f) KeyState: 0x28
ALT + BACK	Down: 1 Repeat: 1 KeyCode: 0x8 ScanCode: 0xe Char: \b (0x8) KeyState: 0x22
SHIFT + BACK	Down: 1 Repeat: 1 KeyCode: 0x8 ScanCode: 0xe Char: \b (0x8) KeyState: 0x30
CTRL + SHIFT + BACK	Down: 1 Repeat: 1 KeyCode: 0x8 ScanCode: 0xe Char: \0 (0x0) KeyState: 0x38
CTRL + ALT + BACK	Down: 1 Repeat: 1 KeyCode: 0x8 ScanCode: 0xe Char: \0 (0x0) KeyState: 0x2a
SHIFT + ALT + BACK	Down: 1 Repeat: 1 KeyCode: 0x8 ScanCode: 0xe Char: \b (0x8) KeyState: 0x32
CTRL + SHIFT + ALT + BACK	Down: 1 Repeat: 1 KeyCode: 0x8 ScanCode: 0xe Char: \0 (0x0) KeyState: 0x3a
```

This is different from wezterm's encoding in the following aspects:

* The delete key is always encoded with character code 0 instead of 0x7f that wezterm uses.
* The backspace key is encoded as 0x8 only when the CTRL modifier is not used.
  If the control modifier is used, the character code 0x7f or 0x0 is used, 
  depending on if other modifiers are used as well.

The proposed changes try to replicate this behaviour.

With this fix, clink and others now behave as expected.

I tried to keep the code readable, but let me know if you have any objections.

### DIFF
--- a/wezterm-input-types/src/lib.rs
+++ b/wezterm-input-types/src/lib.rs
@@ -1222,13 +1222,33 @@ impl KeyEvent {
         match &self.key {
             KeyCode::Composed(_) => None,
             KeyCode::Char(c) => {
+                let c = match *c {
+                    // Delete key is transmitted as 0x0
+                    '\x7f' => '\x00',
+                    // Backspace key is transmitted as 0x8, 0x7f or 0x0
+                    '\x08' => {
+                        if self.modifiers.contains(Modifiers::CTRL) {
+                            if self.modifiers.contains(Modifiers::ALT)
+                                || self.modifiers.contains(Modifiers::SHIFT)
+                            {
+                                '\x00'
+                            } else {
+                                '\x7f'
+                            }
+                        } else {
+                            '\x08'
+                        }
+                    }
+                    _ => *c,
+                };
+
                 let c = if self.modifiers.contains(Modifiers::CTRL) {
                     // Ensure that we rewrite the unicode value to the ASCII CTRL
                     // equivalent value.
                     // <https://github.com/microsoft/terminal/issues/13134>
-                    ctrl_mapping(*c).unwrap_or(*c)
+                    ctrl_mapping(c).unwrap_or(c)
                 } else {
-                    *c
+                    c
                 };
                 let uni = c as u32;
 


wez:
Thank you! The code looks readable and reasonable. Does ConEchoKey now output the same info in wezterm as conhost?

kreudom:
Yes, i checked all cases above in wezterm with the fix applied and they were equal besides a missing NUMLOCK_ON (0x20) flag in the key state. But i assume that difference is negligible for now.

wez:
Thanks!

wez:
Since you have conechokey handy, would you mind also taking a peek at https://github.com/wez/wezterm/issues/2196 ?

Issue #2234, opened 2022-07-10
lua: add is_tty_raw_mode function


Funami580:
**Is your feature request related to a problem? Please describe.**
I would like to have a function to check whether the current tty for the pane is in raw mode or not, so that I can send the original PageUp sequence if I'm using e.g. `less` or some other pager or otherwise just scroll the terminal lines up or down. 

**Describe the solution you'd like**
Some function like `pane.is_tty_raw_mode()` or similar, so that I can use it like this:
```lua
{key="PageUp", mods="", action=wezterm.action_callback(function(window, pane)
  if pane.is_tty_raw_mode() then
    window:perform_action(wezterm.action{SendKey={key="PageUp", mods=""}}, pane)
  else
    window:perform_action(wezterm.action{ScrollByPage=-1}, pane)
  end
end)}
```

**Describe alternatives you've considered**
Use two different keybindings.

**Additional context**
I'm not sure if this will work as I imagine it, but I hope it does.

wez:
I don't believe that you can test for raw mode in that way, but we have `Pane::is_alt_screen_active()` that could be exposed to lua using something similar to https://github.com/wez/wezterm/blob/main/wezterm-gui/src/scripting/pane.rs#L37 that would probably work as a close enough approximation for what you want to do.  Want to try hacking a PR for that?

Funami580:
Thanks, using `Pane::is_alt_screen_active()` seems to work! I still don't really understand how it is determined whether it is an "alternative screen", but I'm going to create a PR soon.

wez:
Thanks!

PR #2236, opened 2022-07-10
lua: add pane:is_alt_screen_active()

Funami580:
refs: #2234

The docs still need refinement, since nobody really knows what an alternative screen is, when it is activated etc. I can't help with that, though.


docs/config/lua/pane/is_alt_screen_active.md
```diff
@@ -0,0 +1,5 @@
+# `pane:is_alt_screen_active()`
+
+*Since: nightly builds only*
+
+Returns whether the alternative screen is active for the pane.
```
wez, line 5:
```suggestion
Returns whether the alternate screen is active for the pane.
```

Funami580: Done.

wez:
Thanks!

PR #2240, opened 2022-07-10
color-types: add missing license file

davide125:
This crate is missing the license file, but the MIT license requires it to be present. I've verified that `cargo package` picks it up as expected.

### DIFF
--- a/color-types/LICENSE.md
+++ b/color-types/LICENSE.md
@@ -1,0 +1,19 @@
+Copyright (c) 2018-Present Wez Furlong
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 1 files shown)


color-types/LICENSE.md
```diff
@@ -0,0 +1,59 @@
...
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+
+Portions of this crate were derived from the bidiref reference implementation
```
wez, line 21: This looks like it is the license from my bidi crate, as this section is talking about that stuff.
I think what I'd prefer to do for this crate is set `license-file = "../LICENSE.md"` in its `Cargo.toml` to explicitly pick up the more general license file that applies to the repo

davide125: Ah yes, sorry, I just copied that one assuming it'd be a standard MIT text, but obviously it isn't. We can reference the top-level one, but that one includes stuff like
```
WezTerm bundles `JetBrains Mono`, `Noto Color Emoji` and `Roboto` fonts.
Those are distributed under the terms of the OFL 1.1, the text of which
can be found in the assets/fonts directory.
```
which I don't think applies here either.

davide125:
Updated to use the generic MIT text for now.

wez:
Thanks!

PR #2268, opened 2022-07-15
x11/wayland: always attempt to use the portal for appearance

vimpostor:
This fixes a weird problem where sometimes wezterm would disregard the very first XDG desktop portal notification.
Please see the commit message for more details, basically we have to read the appearance at startup from the same connection that we read the Lua API appearance from, otherwise we get an inconsistency in internal state.

### DIFF
--- a/window/src/os/wayland/connection.rs
+++ b/window/src/os/wayland/connection.rs
@@ -7,7 +7,7 @@ use crate::os::wayland::output::OutputHandler;
 use crate::os::x11::keyboard::Keyboard;
 use crate::screen::{ScreenInfo, Screens};
 use crate::spawn::*;
-use crate::{Connection, ScreenRect, WindowEvent};
+use crate::{Appearance, Connection, ScreenRect, WindowEvent};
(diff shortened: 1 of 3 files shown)


wez:
what I mean by putting it in both branches is putting it into the get_appearance method of both the x11 and wayland connection implementations

vimpostor:
Thanks, I have updated the PR.
There was no `get_appearance()` for Wayland connections yet, I hope the `Appearance::Light` fallback is correct.

vimpostor:
If you want, I can do the following to avoid the code duplication:
- Add a blocking `get_appearance()` to xdg-desktop-portal that returns an `Option`
- In x11 and wayland connection, call that blocking function and use `unwrap_or_else()` to provide x11 or wayland specific alternatives

Obviously, the code duplication right now is only about 3 lines, so it doesn't really matter that much, so it's your call.

wez:
This is fine, thanks!
re: duplication, it's right on the threshold and I think it's acceptable in this case.

Issue #2103, opened 2022-06-09
Dolphin: Open in terminal should open the current directory.


Quackdoc:
**Is your feature request related to a problem? Please describe.**

In Linux and windows, you can use a context menu to open terminals from inside file managers. on desktops etc. it would be nice to have this functionality work in wezterm.

**Describe the solution you'd like**

current behavior just opens wezterm as if it was opened normally, instead of opening at the location specified by the menu

Quackdoc:
I am using dolphin, I think the issue stems from wezterm not opening in the directory it's being called from. 

say you desend into into another directoy most terminals if you launch them via another terminal emulator from that directory will open in said directory. 

Ie in konsole, if I launch `foot` from `/usr/lib` foot will open in `/usr/lib` in contrast, wezterm will still open the default directory, in linux's case usually, `$HOME`

Quackdoc:
I believe this is how most things implement various "open terminal here" functionality

wez:
I see.  wezterm chooses the home dir by default (actual logic is shown here: https://wezfurlong.org/wezterm/config/lua/config/default_cwd.html) for consistency across platforms, as some GUI environments will spawn the terminal with a surprising and undesirable location by default.

In the Nautilus integration, it knows to tell wezterm to use a specific directory.
I think there should probably also be a Dolphin integration that does something similar.

wez:
Thanks to @Anomalocaridid, this should be fixed now in `main`.

It typically takes about an hour before fixes are available as nightly builds for all platforms.  Linux builds are the fastest to build and are often available within about 20 minutes.  Windows and macOS builds take a bit longer.

Please take a few moments to try out the fix and let me know how that works out. You can find the nightly downloads for your system in the [wezterm installation docs](https://wezfurlong.org/wezterm/installation.html).

If you prefer to use packages provided by your distribution or package manager of choice and don't want to replace that with a nightly download, keep in mind that you can download portable packages (eg: a `.dmg` file on macOS, a `.zip` file on Windows and an `.AppImage` file on Linux) that can be run without permanently installing or replacing an existing package, and can then simply be deleted once you no longer need them. 

If you are eager and can [build from source](https://wezfurlong.org/wezterm/install/source.html) then you may be able to try this out more quickly.

reidrac:
I tried it with Thunar (XFCE) and I didn't see any change, but I don't know if I'm testing it correctly or if the fix doesn't apply to XFCE.

[...]

PR #2271, opened 2022-07-17
Change `Exec` command in desktop file to `wezterm start --`

Anomalocaridid:
Some systems rely on the default terminal emulator's desktop entry for things like launching terminal applications with an application launcher or having a resource opener open a file with a terminal application. Currently, setting WezTerm as the default terminal emulator without overriding its desktop file breaks functionality like this because its desktop entry uses `wezterm` instead of `wezterm start` in its `Exec` field, which just terminates with an error when it is passed a command to run.


wez:
Interesting!  What do you think about also adding in something like `--cwd .` in there?
I'm wondering if that might resolve issues like https://github.com/wez/wezterm/issues/2103, assuming those other applications are resolving how to run wezterm via the .desktop file

Anomalocaridid:
I was able to replicate the problem with Dolphin in #2103 and I appear to have fixed it by setting the `Exec` field in WezTerm's desktop entry to `wezterm start --cwd . --`.

wez:
Thanks!

vimpostor:
FYI I think this commit broke `mailto:` links when using `wezterm` as default terminal emulator with a terminal email client.

For example when `mutt` is configured as the default email program, executing `xdg-open 'mailto:test@test.com'` prints the following error:

```
kf.kio.core: Terminal "wezterm start --cwd . --" not found, service "Mutt"
kf.kio.gui: "Terminal „wezterm start --cwd . --“ beim Versuch, „~/.local/share/applications/mutt.desktop“ auszuführen, nicht gefunden"
```

From a quick glance at the XDG spec it looks like it isn't a wezterm bug though, the `Exec=` line in the desktop file is allowed to be a whole command line, so I guess I should report the bug at KDE instead.

Issue #2253, opened 2022-07-13
Consistent Lua code formatting with StyLua


azzamsa:
Hi.

What if we format all the Lua code snippets in the documentation using [StyLua](https://github.com/JohnnyMorganz/StyLua)?
It keeps the code consistent and more readable.

However, I don't think we can feed him with a markdown file. We need to do it by hand.

wez:
Provided:
* It's easy to build/install (looks like a Rust program, so I have high hopes)
* It can be made to run over the docs without breaking them
* It's actually a style I like(!)

then I'd be interested to see that integrated.
I don't personally have bandwidth to work on that myself.

I have zero interest in manually feeding each snippet through a formatter: it's too much effort.

PR #2273, opened 2022-07-17
docs: reformat lua code

azzamsa:
I build an app called [Gelatyx](https://github.com/azzamsa/gelatyx) for this purpose.  It worked well.  However, StyLua find some invalid code such;

```bash
[31merror parsing: error occurred while creating ast: unexpected token `(`. (starting from line 1, character 9 and ending on line 1, character 10)
additional information: expected function name[0m
[31merror parsing: error occurred while creating ast: unexpected token `(`. (starting from line 3, character 9 and ending on line 3, character 10)
additional information: expected function name[0m
[32mdocs/config/lua/ExecDomain.md[0m is formatted
```

It says the code is missing function name: https://github.com/wez/wezterm/blob/2b17a64d740023efed4d596640346a931b2a29a3/docs/config/lua/ExecDomain.md?plain=1#L37

Here is the full list of files that has invalid code:
[result.txt](https://github.com/wez/wezterm/files/9127839/result.txt)

Usage;

```bash
$ gelatyx lua --file docs/**/*.md --check
```

fixes:  #2253

### DIFF
--- a/docs/config/files.md
+++ b/docs/config/files.md
@@ -78,8 +78,7 @@ degree of flexibility.   The script is expected to return a configuration
 table, so a basic empty configuration file will look like this:
 
 ```lua
-return {
-}
+return {}
 ```
 
 Throughout these docs you'll find configuration fragments that demonstrate
@@ -87,16 +86,16 @@ configuration and that look something like this:
 
 ```lua
 return {
-  color_scheme = "Batman",
+  color_scheme = 'Batman',
 }
 ```
 
 and perhaps another one like this:
 
 ```lua
-local wezterm = require 'wezterm';
+local wezterm = require 'wezterm'
 return {
-  font = wezterm.font("JetBrains Mono"),
+  font = wezterm.font 'JetBrains Mono',
 }
 ```
 
@@ -104,10 +103,10 @@ If you wanted to use both of these in the same file, you would merge them togeth
 like this:
 
 ```lua
-local wezterm = require 'wezterm';
+local wezterm = require 'wezterm'
 return {
-  font = wezterm.font("JetBrains Mono"),
-  color_scheme = "Batman",
+  font = wezterm.font 'JetBrains Mono',
+  color_scheme = 'Batman',
 }
 ```
 
--- a/docs/config/key-tables.md
+++ b/docs/config/key-tables.md
@@ -22,31 +22,39 @@ local wezterm = require 'wezterm'
 local act = wezterm.action
 
 -- Show which key table is active in the status area
-wezterm.on("update-right-status", function(window, pane)
+wezterm.on('update-right-status', function(window, pane)
   local name = window:active_key_table()
   if name then
-    name = "TABLE: " .. name
+    name = 'TABLE: ' .. name
   end
-  window:set_right_status(name or "")
+  window:set_right_status(name or '')
 end)
 
 return {
-  leader = {key="Space", mods="CTRL|SHIFT"},
+  leader = { key = 'Space', mods = 'CTRL|SHIFT' },
   keys = {
     -- CTRL+SHIFT+Space, followed by 'r' will put us in resize-pane
     -- mode until we cancel that mode.
-    {key="r", mods="LEADER", action=act.ActivateKeyTable{
-      name="resize_pane",
-      one_shot=false,
-    }},
+    {
+      key = 'r',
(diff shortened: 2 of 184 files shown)


wez:
Thanks for this!

Seeing what it changes, I hope we can configure some of the opinions of sylua.

A number of the choices in the docs and the set of modules and types in wezterm were carefully selected to work with lua's syntax to make things reasonably expressive without excessive punctuation.

The things that stand out to me as undesirable in the styling on my first pass through are:

* Indent depth is different than my preference; for example `docs/config/lua/config/foreground_text_hsb.md` in this PR should ideally only add the trailing `,` on the closing `}` and leave the whitespace alone
* `local wezterm = require 'wezterm';` vs `local wezterm = require("wezterm")`.  Not a deal breaker, but the parens are not needed and less punctuation makes the file feel cleaner and less complexer.
* `local new_pane = pane:split{}` vs `local new_pane = pane:split({})` (the latter is awful IMO)

Aside from controlling the styling, for this to stick I think we'd need to fold this tool into the doc build and CI:

* Teach `ci/build-docs.sh` to run gelatyx
* Augment CI pages workflow to install gelatyx, and then verify that `git status` is clean after running it to encourage keeping the docs correctly formatted.

azzamsa:
I reformated the docs using Gelatyx v0.1.4 with this command and configuration.

```bash
$ gelatyx lua --file docs/**/*.md --language-config config.toml
```

config.toml

```toml
column_width = 120
line_endings = "Unix"
indent_type = "Spaces"
indent_width = 2
quote_style = "AutoPreferDouble"
call_parentheses = "NoSingleString"
```

Config reference: https://github.com/JohnnyMorganz/StyLua#options

docs/config/files.md
```diff
@@ -94,19 +93,19 @@ return {
...
 
 If you wanted to use both of these in the same file, you would merge them together
 like this:
 
 ```lua
-local wezterm = require 'wezterm';
+local wezterm = require "wezterm"
 return {
-  font = wezterm.font("JetBrains Mono"),
+  font = wezterm.font "JetBrains Mono",
```
wez, line 108: At the risk of sounding mercurial, this is one case where I would lean towards having parens

docs/config/key-tables.md
```diff
@@ -31,22 +31,26 @@ wezterm.on("update-right-status", function(window, pane)
...
     -- mode until we press some other key or until 1 second (1000ms)
     -- of time elapses
-    {key="a", mods="LEADER", action=act.ActivateKeyTable{
-      name="activate_pane",
-      timeout_milliseconds=1000,
-    }},
+    {
+      key = "a",
+      mods = "LEADER",
+      action = act.ActivateKeyTable({
```
wez, line 49: and these `act` cases don't need parens. What are the chances that we can control the formatting to be this specific?

wez (commented):
Thanks! I think this looks better, but it's still not quite how I like.

Can we try:

`call_parentheses = "None"` and `column_width = 78`?

wez:
Can we also set `quote_style = "AutoPreferSingle"`

azzamsa:
Should I override the current commit with new config, or is it better to create a new commit?

wez:
I think it's fine to force-push here so we can see the diff.

In terms of merging this, I was thinking that I don't trust the GH UI to let me see everything in a large diff, and what I'd like to see is a new separate PR that adds the style config and call to `gelatyx` in `ci/build-docs.sh` so that I can run and inspect locally before committing the actual reformatted docs.  That would mean updating `.github/workflows/pages.yml` to install gelatyx and stylua in order for the CI to continue to work.

azzamsa:
I have updated the ci and added the config.

wez:
I split the docs changes out from your change to the build scripts, then did some work to address the errors in the doc pages; that was a bit frustrating because the error output is lacking in context, and formatting seems non-deterministic. I filed two issues for that:
* https://github.com/azzamsa/gelatyx/issues/1
* https://github.com/azzamsa/gelatyx/issues/2

I've pushed the end result of the formatting in e0ea0f46a81ff5e2dd5297e32ecee52550b56c06

Thanks for working on this!

azzamsa:
Thank you so much for merging this PR. Very glad to be part of wezterm contributor.

I will keep maintaining Gelatyx. Yes, I have replied to both of your issue.

PR #2286, opened 2022-07-20
zsh  p10k and fast syntax highlight base16 preview for docs

digitallyserviced:
the palette is great and all but not the greatest for showing the color usage in an actual use case. figured it would be nice to show the scheme in a real usage scenario

Screenshot_2022-07-20-02-33-01_1920x1080

adds an extra asciipreview with a dump from a tmux pane with p10k prompt and output of `fast-theme -t base16` which is the `zsh-users/F-sy-h` syntax highlighted

### DIFF
--- a/ci/generate-docs.py
+++ b/ci/generate-docs.py
@@ -174,6 +174,17 @@ def screen_shot_table(scheme):
 
     return base64.b64encode(f"{header}\n{data}\n".encode("UTF-8")).decode("UTF-8")
 
+def prompt_preview(scheme, screen) -> str:
+    header = {
+        "version": 2,
+        "width": 80,
+        "height": 24,
+        "title": scheme["name"],
+    }
+    header = json.dumps(header, sort_keys=True)
+    data = json.dumps([0.0, "o", screen])
+
+    return base64.b64encode(f"{header}\n{data}\n".encode("UTF-8")).decode("UTF-8")
 
 class GenColorScheme(object):
     def __init__(self, title, dirname, index=None):
@@ -197,6 +208,9 @@ class GenColorScheme(object):
             scheme_filename = f"{self.dirname}/{scheme_prefix}/index.md"
             os.makedirs(os.path.dirname(scheme_filename), exist_ok=True)
             children.append(Page(scheme_prefix, scheme_filename))
+            ansifile = open(f"{self.dirname}/zsh-fsyh-base16.ansi", "r")
+            ansistr = ansifile.read()
+            ansistr = ansistr.replace("\n","\r\n")
 
             with open(scheme_filename, "w") as idx:
 
@@ -205,11 +219,13 @@ class GenColorScheme(object):
                     idx.write(f"# {title}\n")
 
                     data = screen_shot_table(scheme)
+                    prompt_data = prompt_preview(scheme, ansistr)
                     ident = scheme["ident"]
 
                     idx.write(
                         f"""
-<div id="{ident}-player"></div>
+<div id="{ident}-palette-player"></div>
+<div id="{ident}-prompt-player"></div>
 
 <style>
 {scheme["css"]}
@@ -219,7 +235,13 @@ class GenColorScheme(object):
 window.addEventListener('load', function () {{
     AsciinemaPlayer.create(
         'data:text/plain;base64,{data}',
-        document.getElementById('{ident}-player'), {{
+        document.getElementById('{ident}-palette-player'), {{
+        theme: "{ident}",
+        autoPlay: true,
+    }});
+    AsciinemaPlayer.create(
+        'data:text/plain;base64,{prompt_data}',
+        document.getElementById('{ident}-prompt-player'), {{
         theme: "{ident}",
         autoPlay: true,
     }});
--- a/docs/colorschemes/zsh-fsyh-base16.ansi
+++ b/docs/colorschemes/zsh-fsyh-base16.ansi
@@ -1,0 +1,23 @@
+[37m░▒▓[38;5;232m[47m  [37m[44m[38;5;254m  [1m[38;5;255m~[0m[38;5;254m[44m/[38;5;250mDoc[38;5;254m/[1m[38;5;255mcoolors[0m[38;5;254m[44m [34m[43m[30m   master !5 ?13 [33m[49m[39m [34mfast-theme[39m [33m-t[39m [91mbase16
+[90m# simple fn with loop
+[39m_fmt_tmux_string[35m()[1m[33m{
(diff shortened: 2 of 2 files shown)


wez:
I pushed the workflow fix as 19b8efa46a69946f43c5a514a15ad32f4d33957e; thanks!

For the main thrust of this PR: I think having multiple previews at once for the same scheme is probably overkill. What I've been idly thinking about was having the asciicast that we play cycle through a handful of different screens every few seconds. Having them animate by default is probably irritating so I would propose:

* The initial screen shows the current screen shot table, paused, but the last line says "Press play to see more previews"
* We have one or two other representative screenshots added to the ascii cast but separate say 3 seconds for the first one, then 10 or maybe 30 seconds apart for any remaining items.

The timing is easy to play with: the `0.0` in the `data = json.dumps([0.0, "o", screen])` is the time (relative to start of playback) at which that output should be shown.

What do you think?

digitallyserviced:
Ahhh [#41ee60c](https://github.com/wez/wezterm/commit/41ee60c97a228664db19751e3f9b1a87941516d0;) lol no wonder it was bitching at me about cherry pickin's lol

I like it. Having them stacked on each other did seem like a bit too much but it opened a dialogue...

I like the idea of having one more ansi cap for previews... DO you have any sugg's on what would be best? 

- A common utility `ls --color`? Maybe with grc{,-rs}(colorizer)?
- THe p10k config wizard where it shows multiple p rompts on screen? 
- Multiple prompt themes/utils separated by a few lines (starship, p10k, spaceship, zinc, ohmyposh)? 
- A utility with high screen refreshes that works off the default color scheme that we just trim down to 1 or 2 refreshes?
- Neovim with a base16 or 8/4-bit theme? 
- A super busy bigger full-window preview of a tmux session and all those in one?

colorschemeusagepreviews

That is wezterm with terafox scheme, all panes are using it in base16 indexed coloring...

wez:
My preference is that we don't exceed the current 80x24 for a given preview; we can't resize the asciicast dynamically and I think larger screens can be overwhelming and overly information dense.

I think showing vim/nvim may be slightly disingenuous unless it is running with no defaults/no color schemes loaded, because I think almost everyone has their own thoughts on which colors go where, and I don't think there is a "standard" "just use the terminal colors" scheme.  My own personal vim color scheme takes care to use the terminal palette based on the typical ANSI color hues, but it is again my preference.

https://terminal.sexy/ has a number of templates that I think are potentially interesting; looks like the sources for those are in https://github.com/stayradiated/terminal.sexy/tree/master/templates/vim/vivify but they'd need some processing to colorize them, which gets into the same subjective colorization issue as vim color schemes.

Similarly, I don't think things that target base16 for their output are good for non-base16: base16 doesn't use all 16 of its colors in terminal schemes, so you're losing out on a number of palette entries and won't be able to see the full range of the scheme's palette.

I think the most useful previews will be those that show the palette without "promising" how a particular app might look, with a few that show colors in apps that everyone uses but for which very few people customize the colors.

digitallyserviced:
Sorry I wrote a memoir hopefully they publish my book here...

>Similarly, I don't think things that target base16 for their output are good for non-base16: base16 doesn't use all 16 of its colors in terminal schemes, so you're losing out on a number of palette entries and won't be able to see the full range of the scheme's palette.

Just so I am clear and we are on the same page... When I say base16, I am not referring to the base16 scheme sets... I am refering to the base 16 colors using the 8 named ansi colors+bright, or the 0-15 indexed ansicolors (using this tool.. the iso 8 color, or 256-color ansicolors subset) [shell color picker](https://github.com/digitallyserviced/dotfiles/blob/main/.local/bin/shcolor) 

> I think the most useful previews will be those that show the palette without "promising" how a particular app might look, with a few that show colors in apps that everyone uses but for which very few people customize the colors.

I agree, I never want to gurantee or promise people stuff and now be married to making it work/look that way but the reason I am using base 16 or mentioning that in the preview is that they are indeed always guaranteed to be from the terminal's specified color scheme and is actually promised to look that way given the theme being truely base16 (0-15idx or ansinamed). The 16-232 indexes are 6*6*6 rgb that gradiate and are not guranteed to be part of the base 16, nor are they even generated or relate to the base 16. The 232-254 are greyscale. While the color schemes allow for setting the indexed colors they are seldom used, but when we say the base16 color scheme will look X with Y color scheme, it will actually look X when used with Y color scheme (sans opacity or hsb configs).

>[terminal.sexy](https://terminal.sexy/) has a number of templates that I think are potentially interesting; looks like the sources for those are in [https://github.com/stayradiated/terminal.sexy/tree/master/templates/vim/vivify](https://github.com/stayradiated/terminal.sexy/tree/master/templates/vim/vivify?rgh-link-date=2022-07-20T16%3A36%3A40Z) but they'd need some processing to colorize them, which gets into the same subjective colorization issue as vim color schemes.

We could actually use those quite easily using a base 16 type scheme / inline source editor.  And/or we could capture the coloring info using `bat --syntax base16{,-256}`. 

`bat --list-themes` you will see there is base16 and base16-256. Remember those are not related to the base16 theme set. They are only referring to using the original 8-bit ansicolors. From the list themes you can see those two syntax themes use the color scheme from the terminal, when the other named styles are completely different since they specify true color 8*8*8. Either way, that means that you could even include `bat` as part of the workflow to generate the output we need given a few source docs to have it style?

> My preference is that we don't exceed the current 80x24 for a given preview; 

When mentioning this I was thinking of a separate page that expanded to show more content at one shot/glance/viewport.

> we can't resize the asciicast dynamically and I think larger screens can be overwhelming and overly information dense.

It also would not be too difficult to allow there be dynamic changes to asciinema player... We could offer a toolbar/dropdown or something that can change the preview content. This would also allow us to lighten and trim the size of the color scheme previews. 

Currently you are generating base64 data for every style. We could update that to only encode the scheme color info, then the actual preview content could just be injected once for each diff content piece or loaded via ajax. We could also make it so that the preview is not loaded until the content scrolls into view or the content to preview is chosen .


>I think showing vim/nvim may be slightly disingenuous unless it is running with no defaults/no color schemes loaded, because I think almost everyone has their own thoughts on which colors go where, 

I guess I feel that way about those palette bars / color scripts that only print out the base 16 colors for BG of the other colors overlaid on top of them. That seems quite disingenuous because all those other colors have nothng to do with the color scheme usually. 

> and I don't think there is a "standard" "just use the terminal colors" scheme. 

There actually is a scheme that does this. There is also one called Xresources that uses the 16 Xresources colors you can define which would work the same way almost...

Either way I am open to anything. Just my thoughts.

Issue #2173, opened 2022-06-25
Mouse: ScrollUp and ScrollDown events


Funami580:
**Is your feature request related to a problem? Please describe.**
I would like to use `Ctrl+MouseWheelScroll` to increase/decrease the font size.

**Describe the solution you'd like**
Some binding in the config would be nice, e.g.
```
mouse_bindings = {
    {
      event={ScrollUp},
      mods="CTRL",
      action="IncreaseFontSize",
    },
}
```

**Describe alternatives you've considered**
Just use a keyboard binding.

wez:
It's not documented, but button should recognize wheel events:
https://github.com/wez/wezterm/blob/main/term/src/input.rs#L21-L22

Have you tried something like this?

```lua
mouse_bindings = {
  {
     event={Down={streak=1, button=WheelUp={1}}},
     mods="CTRL",
     action=wezterm.action.IncreaseFontSize,
   }
}
```

it might need some finesse to make it usable: I think both the `streak` and the delta value in the `WheelUp` might make it difficult to match.

Funami580:
I just tried
```lua
mouse_bindings = {
  {
    event={Down={streak=1, button={WheelUp=1}}},
    mods="CTRL",
    action=wezterm.action.IncreaseFontSize,
  },
},
```
but nothing happened.

PR #2296, opened 2022-07-23
lua: WheelUp and WheelDown support for mouse_bindings

Funami580:
refs: #2173

Allows you to do things like
```lua

return {
  mouse_bindings = {
    {
      event="WheelUp",
      mods="CTRL",
      action=wezterm.scroll_callback(function(window, pane, delta)
        window:perform_action("IncreaseFontSize", pane)
      end),
    },
    {
      event="WheelDown",
      mods="CTRL",
      action=wezterm.scroll_callback(function(window, pane, delta)
        window:perform_action("DecreaseFontSize", pane)
      end),
    },
  },
}
```

### DIFF
--- a/wezterm-gui/src/scripting/guiwin.rs
+++ b/wezterm-gui/src/scripting/guiwin.rs
@@ -133,6 +133,15 @@ impl UserData for GuiWin {
                 Ok(text)
             },
         );
+        methods.add_async_method("current_event", |lua, this, _: ()| async move {
+            let (tx, rx) = smol::channel::bounded(1);
+            this.window
+                .notify(TermWindowNotif::Apply(Box::new(move |term_window| {
+                    tx.try_send(term_window.current_event.to_dynamic()).ok();
+                })));
+            let result = rx.recv().await.map_err(mlua::Error::external)?;
+            luahelper::dynamic_to_lua_value(lua, result)
+        });
         methods.add_method(
             "perform_action",
             |_, this, (assignment, pane): (KeyAssignment, PaneObject)| {
--- a/wezterm-gui/src/termwindow/mouseevent.rs
+++ b/wezterm-gui/src/termwindow/mouseevent.rs
@@ -14,7 +14,8 @@ use std::ops::Sub;
 use std::rc::Rc;
 use std::sync::Arc;
 use std::time::Duration;
-use wezterm_term::input::MouseEventKind as TMEK;
+use wezterm_dynamic::ToDynamic;
+use wezterm_term::input::{MouseButton, MouseEventKind as TMEK};
 use wezterm_term::{ClickPosition, LastMouseClick, StableRowIndex};
 
 impl super::TermWindow {
@@ -754,7 +755,11 @@ impl super::TermWindow {
                     None
                 }
             }
-            WMEK::VertWheel(amount) if !pane.is_mouse_grabbed() && !pane.is_alt_screen_active() => {
+            WMEK::VertWheel(amount)
+                if !pane.is_mouse_grabbed()
+                    && !pane.is_alt_screen_active()
+                    && event.modifiers.is_empty() =>
+            {
                 // adjust viewport
                 let dims = pane.get_dimensions();
                 let position = self
@@ -765,7 +770,18 @@ impl super::TermWindow {
                 context.invalidate();
                 return;
             }
-            WMEK::VertWheel(_) | WMEK::HorzWheel(_) => None,
+            WMEK::VertWheel(amount) => Some(match *amount {
+                0 => return,
+                1.. => MouseEventTrigger::Down {
+                    streak: 1,
+                    button: MouseButton::WheelUp(*amount as usize),
+                },
+                _ => MouseEventTrigger::Down {
+                    streak: 1,
+                    button: MouseButton::WheelDown(-amount as usize),
+                },
+            }),
+            WMEK::HorzWheel(_) => None,
         };
 
         if allow_action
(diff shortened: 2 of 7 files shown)


wez:
Thanks for taking a crack at this!

I think that the way that the `delta` value is passed around is a bit awkward, because it requires coordination across a few layers, and the introduction of a new callback helper.

As an alternative suggestion:
* For both key and mouse events, use `to_dynamic` to "serialize" the event into a `wezterm_dynamic::Value` when they are received and stash the event into a new `TermWindow` field: `current_event: Option<Value>`.  They can do that at the top of the respective functions before they do anything else, so that the `current_event` always reflects the most recent event.
* Add a lua method to `GuiWin` named `current_event` that will return that value using `dynamic_to_lua_value`
* Your new wheel triggers can schedule a regular window event like the others that accept `(window, pane)`.
* The user can then access the event information using `window:current_event()` if they need access to the event and its delta field

The advantage is that we won't need to carry that event information across multiple layers, and that the same technique will generalize to other types of events if/when they are added in the future, without needing more special cases.

What do you think?

Funami580:
Sounds good.
It now prints this huge log when I print it using
```lua

return {
  mouse_bindings = {
    {
      event="WheelUp",
      mods="CTRL",
      action=wezterm.action_callback(function(window, pane)
        wezterm:log_info(window:current_event()["Down"]["button"]["WheelUp"])
        window:perform_action("IncreaseFontSize", pane)
      end),
    },
  },
}
```
```
13:50:30.783  INFO   logging > lua: {
    "GLOBAL": Error getting tostring: error converting Lua nil to function,
    "action": Error getting tostring: error converting Lua nil to function,
    "action_callback": function,
    "add_to_config_reload_watch_list": function,
    "background_child_process": function,
    "battery_info": function,
    "color": {
        "extract_colors_from_image": function,
        "from_hsla": function,
        "get_builtin_schemes": function,
        "get_default_colors": function,
        "gradient": function,
        "load_base16_scheme": function,
        "load_scheme": function,
        "load_terminal_sexy_scheme": function,
        "parse": function,
        "save_scheme": function,
    },
    "column_width": function,
    "config_dir": "/home/username/.config/wezterm",
    "config_file": "/home/username/.config/wezterm/wezterm.lua",
    "default_wsl_domains": function,
    "emit": function,
    "enumerate_ssh_hosts": function,
    "exec_domain": function,
    "executable_dir": "/home/username/wezterm/target/release",
    "font": function,
    "font_with_fallback": function,
    "format": function,
    "get_builtin_color_schemes": function,
    "glob": function,
    "gradient_colors": function,
    "gui": {
        "get_appearance": function,
        "gui_window_for_mux_window": function,
        "screens": function,
    },
    "home_dir": "/home/username",
    "hostname": function,
    "json_encode": function,
    "json_parse": function,
    "log_error": function,
    "log_info": function,
    "log_warn": function,
    "mux": {
        "all_windows": function,
        "get_active_workspace": function,
        "get_pane": function,
        "get_tab": function,
        "get_window": function,
        "get_workspace_names": function,
        "set_active_workspace": function,
        "spawn_window": function,
    },
    "nerdfonts": Error getting tostring: error converting Lua nil to function,
    "on": function,
    "open_with": function,
    "pad_left": function,
    "pad_right": function,
    "permute_any_mods": function,
    "permute_any_or_no_mods": function,
    "procinfo": {
        "current_working_dir_for_pid": function,
        "executable_path_for_pid": function,
        "get_info_for_pid": function,
        "pid": function,
    },
    "read_dir": function,
    "reload_configuration": function,
    "run_child_process": function,
    "running_under_wsl": function,
    "shell_join_args": function,
    "shell_quote_arg": function,
    "shell_split": function,
    "sleep_ms": function,
    "split_by_newlines": function,
    "strftime": function,
    "strftime_utc": function,
    "target_triple": "x86_64-unknown-linux-gnu",
    "time": {
        "call_after": function,
        "now": function,
        "parse": function,
        "parse_rfc3339": function,
    },
    "truncate_left": function,
    "truncate_right": function,
    "utf16_to_utf8": function,
    "version": "20220727-134013-77549537",
} 5
```
Is this to be expected? Also, I haven’t been able to implement ToDynamic for `KeyEvent`, so I only implemented the `current_event` functionality for mouse events for now.

wezterm-gui/src/scripting/guiwin.rs
```diff
@@ -133,6 +133,15 @@ impl UserData for GuiWin {
                 Ok(text)
             },
         );
+        methods.add_async_method("current_event", |lua, this, _: ()| async move {
+            let (tx, rx) = smol::channel::bounded(1);
+            this.window
+                .notify(TermWindowNotif::Apply(Box::new(move |term_window| {
+                    tx.try_send(term_window.current_event.to_dynamic()).ok();
```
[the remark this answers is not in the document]
Funami580: Then you’d have to do
```rs
dynamic_to_lua_value(lua, result.to_dynamic())
```
so I’m not sure, how that would improve things?

Funami580:
> the mystery of the log output: you have `wezterm:log_info(...)` which is syntactic sugar for `wezterm.log_info(wezterm, ...)`, which is why it is logging the `wezterm` module before your other arguments. You can fix that by changing the `:` to `.`:
> 
> ```lua
> wezterm
> ```

Oh, yeah that makes sense, thanks!

wezterm-gui/src/termwindow/mouseevent.rs
```diff
@@ -785,10 +802,29 @@ impl super::TermWindow {
                     }
                 }
 
-                if let Some(action) = self
-                    .input_map
-                    .lookup_mouse(event_trigger_type.clone(), modifiers)
-                {
+                match event_trigger_type {
```
wez, line 805:
```suggestion
                // normalize delta and streak to make mouse assignment
                // easier to wrangle
                match event_trigger_type {
```

wez (commented):
OK, I think this looks good! Thanks!
Could you write up some docs to go with it?

wezterm-gui/src/termwindow/mouseevent.rs
Funami580: Done.

Funami580:
>  Could you write up some docs to go with it?

Sure, done.

wez:
Thanks!

PR #2307, opened 2022-07-26
termwiz: allow to disable default image formats

jirutka:
The image crate includes support for many image formats by default,
including PNM, TGA, DDS, OpenEXR and other little-used formats.
Termwiz is not an image editor, is it?

All the default image features are still included by default, for
backward compatibility. However, the consumers of this crate can exclude
them by disabling default-features.

### DIFF
--- a/termwiz/Cargo.toml
+++ b/termwiz/Cargo.toml
@@ -21,7 +21,7 @@ filedescriptor = { version="0.8", path = "../filedescriptor" }
 fixedbitset = "0.4"
 fnv = {version="1.0", optional=true}
 hex = "0.4"
-image = {version="0.24", optional=true}
+image = {version="0.24", default-features=false, optional=true}
 lazy_static = "1.4"
 libc = "0.2"
 log = "0.4"
@@ -45,6 +45,7 @@ wezterm-color-types = { path = "../color-types", version="0.1" }
 wezterm-dynamic = { path = "../wezterm-dynamic" }
 
 [features]
+default = ["image?/default"]
 widgets = ["cassowary", "fnv"]
 use_serde = ["serde", "wezterm-color-types/use_serde"]
 use_image = ["image"]


wez:
I understand that you think it is not necessary to include these, but what are you trying to optimize for?
Each time a configuration like this is introduced, the complexity of the system increases, and complexity in the build system is one of the most painful kinds of complexity in software engineering.
My default disposition is to push back on it unless there is a truly compelling upside.

jirutka:
> but what are you trying to optimize for?

Remove unnecessary (and maybe unwanted) transitive dependencies to reduce unwanted binary bloat, attack vectors etc. The problem is that Cargo doesn’t provide an option to disable features in transitive dependencies (without patching them).

> Each time a configuration like this is introduced, the complexity of the system increases, and complexity in the build system is one of the most painful kinds of complexity in software engineering.

Complexity also increases with each third-party package in the dependency graph.

A better solution would be to just disable _default-features_ for _image_ crate and instruct developers to declare _termwiz_ crate together with _image_ crate with only the features they need. I assumed that you wouldn’t accept such soluton because of backward compatibility.

My main/original goal is to remove unnecessary image formats from WezTerm, but it turned out to be non-trivial, so I decided to start with this for now.

wez:
I appreciate your effort, but I don't want this.

There needs to be a concrete reason that has clear benefit that outweighs the complexity of making more of this stuff optional, and there isn't one here.  I'm sure you're going to propose other similar changes that will only increase the complexity further.

The closest practical reason is around reducing binary size and that is not something I am going to bias towards in 2022; this software is not intended to run on constrained embedded systems and the binary size just isn't that big in the grand scheme of things.

jirutka:
> clear benefit that outweighs the complexity of making more of this stuff optional, and there isn't one here.

Security and complexity of dependencies are not your concern…?

> this software is not intended to run on constrained embedded systems 

termwiz is released on crates.io and used even outside of WezTerm.

> …reducing binary size and that is not something I am going to bias towards in 2022

There are many people who don’t have access to broadband internet even in 2022. A few hundred kilobytes may not look significant by itself, but when you add it up, you suddenly get megabytes of unnecessary bloat that the user has to download each time when updating the software.

wez:
I'm closing this because of the reasons I mentioned above.
For others who may wonder why I didn't reply to the latest comment, please see: https://github.com/wez/wezterm/pull/2305#issuecomment-1195507887

PR #2308, opened 2022-07-26
Add link to download sha256 checksum

yashpalgoyal1304:
Continuation of https://github.com/wez/wezterm/issues/2114

This is more like a reminder... as:
* i dont know the specifics, and we can go back and forth on this, 
* but i guess it would be much much faster if u do it alone too, as you know all the details of it

that's why i did not write this for all operating systems too....

destination:
* https://wezfurlong.org/wezterm/install/windows.html
* https://github.com/wez/wezterm/tree/main/docs/install

### DIFF
--- a/docs/install/windows.markdown
+++ b/docs/install/windows.markdown
@@ -10,7 +10,7 @@ You can download a setup.exe style installer to guide the installation
 yourself (no special privileges required).
 
 <a href="{{ windows_exe_stable }}" class="btn">Windows (setup.exe)</a><a href="{{ windows_exe_stable }}.sha256">Setup checksum</a><br/>
-<a href="{{ windows_exe_nightly }}" class="btn">Nightly Windows (setup.exe)</a><a href="{{  }}">Nightly setup checksum</a>
+<a href="{{ windows_exe_nightly }}" class="btn">Nightly Windows (setup.exe)</a><a href="{{ windows_exe_nightly }}.sha256">Nightly setup checksum</a>
 
 WezTerm is available in a setup.exe style installer; the installer is produced
 with Inno Setup and will install wezterm to your program files directory and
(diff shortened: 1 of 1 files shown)


docs/install/windows.markdown
```diff
@@ -9,8 +9,8 @@ You can download a setup.exe style installer to guide the installation
 (requires admin privileges) or a simple zip file and manage the files for
 yourself (no special privileges required).
 
-<a href="{{ windows_exe_stable }}" class="btn">Windows (setup.exe)</a>
-<a href="{{ windows_exe_nightly }}" class="btn">Nightly Windows (setup.exe)</a>
+<a href="{{ windows_exe_stable }}" class="btn">Windows (setup.exe)</a><a href="{{  }}">Setup checksum</a><br/>
```
wez, line 12: This is how you could refer to one of those new files:

```suggestion
<a href="{{ windows_exe_stable }}" class="btn">Windows (setup.exe)</a><a href="{{ windows_exe_stable }}.sha256">Setup checksum</a><br/>
```

However:
* Very few people ever verify hashes so it doesn't warrant being made into a big button here
* There are no stable releases with hashes yet, so putting that here now would just give a broken link

I think it is more reasonable to have a separate section below that links to the hashes and has instructions for verifying them.

I'm also skeptical that folks will do this on Windows; I don't think they even have the tools installed to this by default.  The other OSs seem more reasonable to target.

yashpalgoyal1304: > I think it is more reasonable to have a separate section below that links to the hashes and has instructions for verifying them.

* yeah, i was thinking exactly same. like some paragraph mentioning smth like: "go to here for getting the sha256 checksum" would be better.

> I'm also skeptical that folks will do this on Windows; 

😂 on point, 

> I don't think they even have the tools installed to this by default.

i guess it does have some CLI tool for that, but 'ts nowhere as lean as GNU coreutils' `sha256sum` or `cksum` (v9.0+)
i am on windows, and doing this before i discovered GNU coreutils via MSYS2 was very involving task - open the tool's GUI, drag n drop files, open the checksum file and copy paste in the tool gui  etc etc. Starting to use GNU coreutils via MSYS2 was the best of the best thing on computer i have learnt/done in years.

wez (changes requested):
(This needs changes as outlined above before it is ready to review or merge)

yashpalgoyal1304:
oops, wrong commit message....

feel free to squash these commits if you feel so

wez:
Can you build the docs and capture a screenshot of how that changed page looks?
I think it will look a bit odd!

These are the steps used by the CI: https://github.com/wez/wezterm/blob/main/.github/workflows/pages.yml#L36-L51

Install prerequisites:

```
cargo install mdbook --no-default-features --features search --vers "^0.4" --locked
cargo install mdbook-linkcheck --locked
cargo install mdbook-mermaid --locked
cargo install gelatyx --locked
```

Then to build:

```
./ci/build-docs.sh
```

You can run them locally:

```
mdbook serve docs
```

PR #2327, opened 2022-07-30
Accept empty table as None

aznhe21:
This PR allows `wezterm.actions.CopyMode{SetSelectionMode={}}` to work.

This feature was introduced in 957cc592a6e4974efbefefd98aed904c206677ad but was removed at some point.

### DIFF
--- a/wezterm-dynamic/src/fromdynamic.rs
+++ b/wezterm-dynamic/src/fromdynamic.rs
@@ -85,6 +85,7 @@ impl<T: FromDynamic> FromDynamic for Option<T> {
     fn from_dynamic(value: &Value, options: FromDynamicOptions) -> Result<Self, Error> {
         match value {
             Value::Null => Ok(None),
+            Value::Object(table) if table.is_empty() => Ok(None),
             value => Ok(Some(T::from_dynamic(value, options)?)),
         }
     }


wez:
I'm going to have to think about this one; the reason I took this out again was that there were some ambiguities and things that broke with the last implementation.

Meanwhile, I think you may appreciate: https://github.com/wez/wezterm/commit/10121790df10fbc40896f567046b6fcd54c42bfb

aznhe21:
`CopyMode{SetSelectionMode={}}` can definitively clear the selection mode even if the SelectionMode is Line or Block (which is not possible in 1012179). This makes it possible to copy and then remain in copy mode, as shown below.

```lua
{
  key='y',
  mods='NONE',
  action=act.Multiple{
    act.CopyTo('PrimarySelection'),
    act.ClearSelection,
    -- clear the selection mode, but remain in copy mode
    act.CopyMode{SetSelectionMode={}},
  },
},
```

I understand that this PR is ambiguous. Instead of merging this PR, it might be a good idea to add a new action (like CopyMode('ClearSelectionMode')).

wez:
Yeah, I think adding a `ClearSelectionMode` is the cleanest way to express this; want to re-spin this PR to do that?

aznhe21:
Thank you! I am closing this PR as I submitted a new PR #2352.

PR #2340, opened 2022-08-01
bidi: update license tag to match the actual license

davide125:
According to the README and LICENSE files, this crate contains data and code from Unicode, so update the crate license tag accordingly. This was raised as part of the Fedora packaging review in https://bugzilla.redhat.com/show_bug.cgi?id=2105831


wez:
Thanks!

PR #2341, opened 2022-08-01
Ensure parse_first_as_vec groups ST with OSC sequence

markbt:
When `parse_first_as_vec` is parsing an OSC sequence termined by an ST encoded at `ESC \`, we must ensure we process the `\` and return the parser to its ground state before returning, so that the caller gets a complete set of actions and the offset returned describes a valid span of the input.

Fixes https://github.com/markbt/streampager/issues/57

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5018,7 +5018,7 @@ dependencies = [
 
 [[package]]
 name = "vtparse"
-version = "0.6.1"
+version = "0.6.2"
 dependencies = [
  "k9",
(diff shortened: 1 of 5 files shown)


markbt:
Not sure if the version bump is the way you want to handle the new public method, so feel free to change it.

wez:
Thanks! Looks good!

wez:
I published new versions of vtparse and termwiz (and a couple of deps) to crates.io. Thanks again!

PR #2349, opened 2022-08-02
[termwiz] add impl-default feature to winapi dep to fix the build

mitrandir77:
Without the feature the build fails with:
```

    --> src\escape\apc.rs:377:57
     |
377  |         let mut memory_info = MEMORY_BASIC_INFORMATION::default();
     |                                                         ^^^^^^^ function or associated item not found in `MEMORY_BASIC_INFORMATION`
     |
```


wez:
Thanks!

PR #2352, opened 2022-08-03
Add new action `CopyMode('ClearSelectionMode')`

aznhe21:
This PR adds new action `CopyMode('ClearSelectionMode')`. It allows definitively clear the selection mode even if the SelectionMode is Line or Block. This makes it possible to copy and then remain in copy mode, as shown below.

```lua
{
  key='y',
  mods='NONE',
  action=act.Multiple{
    act.CopyTo('PrimarySelection'),
    act.ClearSelection,
    -- clear the selection mode, but remain in copy mode
    act.CopyMode{'ClearSelectionMode'},
  },
},
```


wez:
Thanks!

PR #2366, opened 2022-08-07
Improve get-deps script for Gentoo Linux

omarandlorraine:
There's some difficulty running the `get-deps` script on Gentoo Linux. I've got two commits here that fix the problems I've seen so far.

1) there are several ebuilds called `git`; we need to select the right one. (I hope I got it right! I've assumed we need `dev-vcs/git`, which is the version control system)

2) installing a package on Gentoo really means recompiling it, so installing a lot of packages will take a long time. Therefore I've tweaked the script to not reinstall the dependencies that we've already got.

### DIFF
--- a/get-deps
+++ b/get-deps
@@ -227,7 +227,7 @@ gentoo_deps() {
     'cmake' \
     'fontconfig' \
     'openssl' \
-    'git' \
+    'dev-vcs/git' \
     'libX11' \
     'libxkbcommon' \
     'pkgconf' \
--- a/get-deps
+++ b/get-deps
@@ -223,7 +223,7 @@ gentoo_deps() {
   portageq envvar USE | xargs -n 1 | grep '^X$' \
   || (echo 'X is not found in USE flags' && exit 1)
   EMERGE="$SUDO emerge"
-  $EMERGE --select \
+  for pkg in \
     'cmake' \
     'fontconfig' \
     'openssl' \
@@ -239,6 +239,9 @@ gentoo_deps() {
     'xcb-util-image' \
     'xcb-util-keysyms' \
     'xcb-util-wm'
+  do
+	  equery l "$pkg" > /dev/null || $EMERGE --select $pkg
+  done
 }
 
 fallback_method() {


get-deps
```diff
@@ -223,22 +223,10 @@ gentoo_deps() {
...
-    'pkgconf' \
-    'python' \
-    'perl' \
-    'wayland' \
-    'mesa' \
-    'xcb-util' \
-    'xcb-util-image' \
-    'xcb-util-keysyms' \
-    'xcb-util-wm'
+  for i in 'cmake' 'fontconfig' 'openssl' 'dev-vcs/git' 'libX11' 'libxkbcommon' 'pkgconf' 'python' 'perl' 'wayland' 'mesa' 'xcb-util' 'xcb-util-image' 'xcb-util-keysyms' 'xcb-util-wm'
```
wez, line 226: For the sake of easier diffs in the future, please preserve the one-line-per-package format that we had before:

```
  for i in \
      'cmake' \
      'fontconfig' \
```

get-deps
```diff
@@ -223,22 +223,10 @@ gentoo_deps() {
...
+  do
+	  equery l "$i" > /dev/null || emerge --select $i
```
wez, line 228: please use `$EMERGE` rather than `emerge` so that `sudo` is used when appropriate:

```suggestion
	  equery l "$i" > /dev/null || $EMERGE --select $i
```

get-deps
wez, line 226: Let's call this `$pkg` so the inside of the loop is more readable

```suggestion
  for pkg in 'cmake' 'fontconfig' 'openssl' 'dev-vcs/git' 'libX11' 'libxkbcommon' 'pkgconf' 'python' 'perl' 'wayland' 'mesa' 'xcb-util' 'xcb-util-image' 'xcb-util-keysyms' 'xcb-util-wm'
```

wez (changes requested):
Thanks for diving in!
I have a couple of suggestions!

omarandlorraine:
Those are good suggestions. So what i plan to do is patch them up, rewrite the history a little and then push --force. I am on mobile at the moment but I can have a look in a couple of hours. Sounds okay?

get-deps
omarandlorraine: good idea

omarandlorraine:
How does this look

wez:
Thanks!

PR #2380, opened 2022-08-09
Add a configuration flag to close a tab with the middle button click

### LINKED ISSUE #2350: Way too easy to close accidentaly a tab using middle mouse key

Linux Wayland

Gnome/Mutter

### WezTerm version

20220802-215653-1df1f166

### Describe the bug

I use a lot the middle mouse to paste text in the console, and if the cursor is accidentally in the tab bar, the tab is closed without any warning or confirmation. I tried the patch below but it didn't help;
```
diff --git a/wezterm-gui/src/termwindow/mouseevent.rs b/wezterm-gui/src/termwindow/mouseevent.rs
index 4d5f303c1..310f8a469 100644
--- a/wezterm-gui/src/termwindow/mouseevent.rs
+++ b/wezterm-gui/src/termwindow/mouseevent.rs
@@ -408,7 +408,7 @@ impl super::TermWindow {
             },
             WMEK::Press(MousePress::Middle) => match item {
                 TabBarItem::Tab { tab_idx, .. } => {
-                    self.close_tab_idx(tab_idx).ok();
+                    self.close_specific_tab(tab_idx, true);
                 }
                 TabBarItem::NewTabButton { .. } | TabBarItem::None => {}
             },
```
At the end, I removed the whole ``self.close_..`` line

### To Reproduce

Use a lot the middle mouse button..

### Configuration

no config

### Expected Behavior

_No response_

### Logs

_No response_

### Anything else?

The problem appeared after https://github.com/wez/wezterm/commit/d19db3c87b9412dbe25cb5e029c737de7a7f96ca

yaslama:
fix #2350

### DIFF
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -358,6 +358,10 @@ pub struct Config {
     #[dynamic(default)]
     pub hide_tab_bar_if_only_one_tab: bool,
 
+    /// When true (the default), a middle click in a tab title close the tab.
+    #[dynamic(default = "default_true")]
+    pub middle_click_close_tab: bool,
+
     #[dynamic(default)]
     pub enable_scroll_bar: bool,
 
--- a/docs/config/lua/config/middle_click_close_tab.md
+++ b/docs/config/lua/config/middle_click_close_tab.md
@@ -1,0 +1,4 @@
+# `middle_click_close_tab = true`
+
+When `middle_click_close_tab = true`, clicking with the middle button on the
+title of a tab will cause that tab to be closed.
--- a/wezterm-gui/src/termwindow/mouseevent.rs
+++ b/wezterm-gui/src/termwindow/mouseevent.rs
@@ -408,7 +408,9 @@ impl super::TermWindow {
             },
             WMEK::Press(MousePress::Middle) => match item {
                 TabBarItem::Tab { tab_idx, .. } => {
-                    self.close_tab_idx(tab_idx).ok();
+                    if self.config.middle_click_close_tab {
+                        self.close_tab_idx(tab_idx).ok();
+                    }
                 }
                 TabBarItem::NewTabButton { .. }
                 | TabBarItem::None


bew:
I think this is too restrictive, it would be better to allow mouse events on tabs be configurable.
For example it should be possible to have middle click to duplicate the tab.
@wez will probably help you with the design if you want to try to do it (:

yaslama:
> I think this is too restrictive, it would be better to allow mouse events on tabs be configurable. For example it should be possible to have middle click to duplicate the tab. @wez will probably help you with the design if you want to try to do it (:

You describe here another improvement which is to be able to do something else when clicking on the middle button.
This PR goal is to be able to **disable** any effect when clicking on middle button in the tab bar because the middle button is used to paste text and it's very easy to middle click by mistake in the tab bar instead of the terminal itself and any action can be problematic. cf issue #2350.
I agree that a more generic mechanism can be added to handle keyboard and mouse events in the tab bar, but in the context of this specific issue/pr, « _le mieux est le mortel ennemi du bien_ ». @wez What do you think?

wez:
Thanks for taking a crack at this!

This is a very targeted solution and while it is certainly very easy to add it, it, together with other similarly targeted solutions, results in increased complexity and makes it harder to maintain and evolve.  The mouse handling stuff is already suffering from this.  I'm hesitant to accept the initial approach in this PR as it would essentially be immediate tech debt.

I would much prefer to solve this in a similar way to how bindings were introduced to the various overlays: we use `key_tables` for those and I think there is room for `mouse_tables` (name is illustrative rather than ideal), so a hypothetical config might look like:

```lua
local act = wezterm.act

return {
   mouse_tables = {
       -- probably want `active_tab` and `inactive_tab` as separately bindable things
       tab = {
           {
               event = { Down = { streak = 1, button = 'Middle' } },
               mods = 'NONE',
               action = act.Nop,
           },
       }
   }
}
```

yaslama:
Thanks @wez for your answer, but in fact, in order to fix #2350, I think that a prompt when closing a tab is better.

PR #2386, opened 2022-08-09
Add a configuration option to prompt confirmation when closing tab

### LINKED ISSUE #2350: Way too easy to close accidentaly a tab using middle mouse key

yaslama:
fix #2350

Replace #2380

### DIFF
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -359,6 +359,9 @@ pub struct Config {
     pub hide_tab_bar_if_only_one_tab: bool,
 
     #[dynamic(default)]
+    pub tab_close_confirmation: TabCloseConfirmation,
+
+    #[dynamic(default)]
     pub enable_scroll_bar: bool,
 
     #[dynamic(try_from = "crate::units::PixelUnit", default = "default_half_cell")]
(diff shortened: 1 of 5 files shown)


wez:
I pushed f2f2b8389 which is a subset of this; just the bits that make middle clicking trigger the prompt according to the established rules for closing tabs with confirmation.

I don't want to add another confirmation option for tabs that bypasses that logic because it makes things even more complex than they already are, diverges the behavior around window and tab closing, and the implementation in this PR has that at the wrong layer: it forces the use of a global configuration rather than the config scoped to the window which makes things a bit more difficult to manage.

What I'd recommend for you is that you set `skip_close_confirmation_for_processes_named = {}` so that you always get a prompt when you close a tab using the mouse.

yaslama:
Thanks a  lot!! This fixes the issue for me.

PR #2402, opened 2022-08-13
Cache portal appearance value

vimpostor:
This improves the startup time by about $n\cdot 2ms$, where $n$ is the number of times the user calls `get_appearance()` in their config.

For example when the user uses the documented `scheme_for_appearance(wezterm.gui.get_appearance())` snippet in their config, wezterm would make one additional unnecessary dbus query at startup.

Instead of making redundant duplicate dbus queries, we now cache the portal setting.

Regarding the implementation, I had to do some slight code changes to the surrounding code structure, because `get_appearance()` in `os::xdg_desktop_portal` was implemented without `&self`, but now with the caching behaviour we obviously need sideeffects in this function.

I looked at the rest of the codebase and noticed that you like to use interior mutability with `RefCell` in quite a lot of places, so I made the same approach using `RefCell` for the caching here.

Let me know if you dislike this implementation and like to store the cached appearance in a different way, I can make it happen! :)

### DIFF
--- a/window/src/os/xdg_desktop_portal.rs
+++ b/window/src/os/xdg_desktop_portal.rs
@@ -6,6 +6,8 @@ use crate::{Appearance, Connection, ConnectionOps};
 use anyhow::Context;
 use futures_util::stream::StreamExt;
 use std::collections::HashMap;
+use std::sync::Mutex;
+use std::time::Instant;
 use zbus::dbus_proxy;
 use zvariant::OwnedValue;
 
@@ -24,6 +26,22 @@ trait PortalSettings {
 
     #[dbus_proxy(signal)]
     fn SettingChanged(&self, namespace: &str, key: &str, value: OwnedValue) -> Result<()>;
+}
+
+struct State {
+    appearance: Option<Appearance>,
+    subscribe_running: bool,
+    last_update: Instant,
+}
+
+lazy_static::lazy_static! {
+  static ref STATE: Mutex<State> = Mutex::new(
+          State {
+              appearance: None,
+              subscribe_running: false,
+              last_update: Instant::now(),
+          }
+   );
 }
 
 pub async fn read_setting(namespace: &str, key: &str) -> anyhow::Result<OwnedValue> {
@@ -48,8 +66,50 @@ fn value_to_appearance(value: OwnedValue) -> anyhow::Result<Appearance> {
 }
 
 pub async fn get_appearance() -> anyhow::Result<Appearance> {
-    let value = read_setting("org.freedesktop.appearance", "color-scheme").await?;
-    value_to_appearance(value)
+    let mut state = STATE.lock().unwrap();
+    if state.appearance.is_some()
+        && (state.subscribe_running || state.last_update.elapsed().as_secs() < 1)
+    {
+        Ok(state.appearance.unwrap())
+    } else {
+        let value = read_setting("org.freedesktop.appearance", "color-scheme").await?;
+        let appearance = value_to_appearance(value)?;
+        state.appearance = Some(appearance);
+        state.last_update = Instant::now();
+        Ok(appearance)
+    }
+}
+
+pub async fn run_signal_loop(stream: &mut SettingChangedStream<'_>) -> Result<(), anyhow::Error> {
+    // query appearance again as it might have changed without us knowing
+    if let Ok(value) =
+        value_to_appearance(read_setting("org.freedesktop.appearance", "color-scheme").await?)
+    {
+        let mut state = STATE.lock().unwrap();
+        if state.appearance != Some(value) {
+            state.appearance = Some(value);
+            state.last_update = Instant::now();
+            drop(state);
+            let conn = Connection::get().ok_or_else(|| anyhow::anyhow!("connection is dead"))?;
+            conn.advise_of_appearance_change(value);
+        }
+    }
+
+    while let Some(signal) = stream.next().await {
+        let args = signal.args()?;
+        if args.namespace == "org.freedesktop.appearance" && args.key == "color-scheme" {
+            if let Ok(appearance) = value_to_appearance(args.value) {
+                let mut state = STATE.lock().unwrap();
+                state.appearance = Some(appearance);
+                state.last_update = Instant::now();
+                drop(state);
+                let conn =
+                    Connection::get().ok_or_else(|| anyhow::anyhow!("connection is dead"))?;
+                conn.advise_of_appearance_change(appearance);
+            }
+        }
+    }
+    Result::<(), anyhow::Error>::Ok(())
 }
 
 pub fn subscribe() {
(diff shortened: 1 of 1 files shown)


wez:
Thanks for taking this on!
I would tackle this slightly differently: it should be possible to resolve this entirely within `xdg_desktop_portal.rs` and have less fan-out into the rest of the code.

* Introduce two pieces of global state (which could both be in a struct):
  * A flag that indicates wheher the subscribe function is currently running
  * A copy of the appearance value most recently received by the the subscribe function

```rust
struct State {
   subscribe_running: bool,
   appearance: Appearance,
}

lazy_static::lazy_static! {
  static ref STATE: Mutex<State> = Mutex::new(
          State {
              subscribe_running: false,
              appearance: Appearance::Light
          }
   );
}
```

* Change subscribe so that:
  * before it processes the stream, it calls `read_setting` to populate the initial cached appearance value and sets `subscribe_running = true`.
  * updates the cached value just prior to calling `conn.advise_of_appearance_change`
  * Guarantees to set `subscribe_running = false` when it returns
     * (I'd suggest adding an intermediate function that calls the real implementation and then sets it to false when it returns)  
 
* Change `get_appearance` so that:
  * If `subscribe_running`, return the cached value
  * otherwise, call `read_setting`

vimpostor:
Ah I see. I initially tried implementing it with a static variable too, but then avoided it because of the necessary mutex.

Anyway I have implemented your suggested changes now, except for a minor difference: Your proposed order of change would not actually cache the value, because `get_appearance()` is called before the subscription loop starts.

I have now implemented it in such a way that `get_appearance()` is always cached, not just if the loop is already running. This should also be more robust if some other code starts querying appearance in the future.

I also use one Option<Appearance> variable instead of two separate variables, because it makes for more elegant pattern matching, but I can revert the latter if you don't like it.

window/src/os/xdg_desktop_portal.rs
```diff
@@ -48,8 +61,20 @@ fn value_to_appearance(value: OwnedValue) -> anyhow::Result<Appearance> {
...
+                    state.appearance = Some(a);
+                    Ok(a)
+                }
+                err => err,
+            }
+        }
+    }
 }
 
 pub fn subscribe() {
```
wez, line 80: there's a race condition in this implementation: if the appearance is changed between our first call to `get_appearance` and before we have set up a subscription, we will not be aware of the change because `get_appearance` is not aware of whether the subscribe function is running and will blindly continue to use its cached value.

To resolve this race I would suggest adding a call to `read_setting` immediately before the `while` loop below (in between the `receive_SettingChanged` and the `while` loop).

If the returned value is different from the cache value, then `conn.advise_of_appearance_change(appearance);` should be called to ensure that the updated value is observed.

window/src/os/xdg_desktop_portal.rs
```diff
@@ -26,6 +27,18 @@ trait PortalSettings {
     fn SettingChanged(&self, namespace: &str, key: &str, value: OwnedValue) -> Result<()>;
 }
 
+struct State {
+    appearance: Option<Appearance>,
```
wez, line 31: I'm "worried" about edge cases and also about `get_appearance` having surprising behavior if `subscribe` is never called. Prior to this PR `get_appearance` was easy to reason about, but right now there are some implementation details that aren't called out; in particular, it has no way to invalidate a stale value if `subscribe` isn't called, or if the thread has stopped for whatever reason.

What I'd suggest is:

* Track an `Instant` here to show when STATE.appearance was last assigned
* Track a boolean here to show whether the subscribe thread is running

In `get_appearance`, if subscribe is running then the cached appearance value can be read without fear.

Otherwise, if `Instant::elapsed` is within a reasonable threshold (say, 1 second?), then the cached value can be returned, otherwise, call `read_settings` and update the cached value and the time it was assigned.

In the `subscribe` thread, be sure to update the time each time the cached value is updated as well.

That way we can reasonably cache the appearance in the brief window before the subscribe function is able to be set up and called, and also stand a chance of still behaving reasonably if something weird happens to the subscribe thread for whatever reason.

window/src/os/xdg_desktop_portal.rs
vimpostor: I will implement your other suggested change later today, but I see a problem with the above idea.
If we call `read_setting` again before the while loop (in addition to the `read_setting` in `get_apperance`), then we have at least one duplicated dbus call again. Now in this case it wouldn't actually block startup because it would run in a separate thread, but it still sounds rather backwards to me since we could easily solve this in another way.

For example why can't we just solve the race condition by subscribing before calling the first `get_appearance`?

wez: get_appearance is often called as part of loading the config which happens before the gui subsystem is started up; subscribe cannot be called until after the config is processed at least once.

vimpostor:
I have implemented all of your above concerns, but I still kinda dislike the fact that we keep making one redundant dbus query right before the while loop. This second redundant dbus query is now in a `promise::spawn::spawn()`, so it does not block, but it still itches me a bit.

Do you have any idea how we can get rid of that second dbus query while also avoiding a race condition?

window/src/os/xdg_desktop_portal.rs
```diff
@@ -48,8 +66,52 @@ fn value_to_appearance(value: OwnedValue) -> anyhow::Result<Appearance> {
...
+        let value = read_setting("org.freedesktop.appearance", "color-scheme").await?;
+        match value_to_appearance(value) {
+            Ok(a) => {
+                state.appearance = Some(a);
+                state.last_update = Instant::now();
+                Ok(a)
+            }
+            err => err,
+        }
+    }
```
wez, line 84:
```suggestion
        let appearance = value_to_appearance(value)?;
        state.appearance = Some(appearance);
        state.last_update = Instant::now();
        Ok(appearance)
    }
```

prefer to have less nesting/braces!

wez (commented):
Thanks! One last suggestion inline!
I think it's fine to have that extra read_settings call in the subscribe thread: it's the only way to guarantee that we can't miss that change event and it costs us nothing in practice.

window/src/os/xdg_desktop_portal.rs
vimpostor: Thanks, the `?` is indeed much more elegant here.

vimpostor:
Thanks for taking the time to review!

I also noticed a deadlock as I forgot to unlock the mutex before calling `advise_of_appearance_change`.
The diff for the fix luckily was very simple:

```diff
diff --git a/window/src/os/xdg_desktop_portal.rs b/window/src/os/xdg_desktop_portal.rs
index 5d3a9a87b..06aead7c7 100644
--- a/window/src/os/xdg_desktop_portal.rs
+++ b/window/src/os/xdg_desktop_portal.rs
@@ -93,6 +89,7 @@ pub async fn run_signal_loop(stream: &mut SettingChangedStream<'_>) -> Result<()
         if state.appearance != Some(value) {
             state.appearance = Some(value);
             state.last_update = Instant::now();
+            drop(state);
             let conn = Connection::get().ok_or_else(|| anyhow::anyhow!("connection is dead"))?;
             conn.advise_of_appearance_change(value);
         }
@@ -105,6 +102,7 @@ pub async fn run_signal_loop(stream: &mut SettingChangedStream<'_>) -> Result<()
                 let mut state = STATE.lock().unwrap();
                 state.appearance = Some(appearance);
                 state.last_update = Instant::now();
+                drop(state);
                 let conn =
                     Connection::get().ok_or_else(|| anyhow::anyhow!("connection is dead"))?;
                 conn.advise_of_appearance_change(appearance);
```

wez:
Thanks!

PR #2418, opened 2022-08-17
Add missing license files

davide125:
This adds missing license files to all crates that are explicitly MIT licensed in Cargo.toml and don't already have one. Note that there are more crates in this repository that don't set an explicit license (e.g. `wezterm-gui`) -- those are ignored here, but should probably be fixed in a separate PR as well.


wez:
Thanks!

PR #2435, opened 2022-08-20
Enable control key in macOS IME

yuys13:
This PR enables the control key in the macOS IME.

When using the IME, Ctrl +h is expected to behave the same as a backspace.
As shown in the following image.
expected_ctrlh

However, currently it behaves as shown in the following image.
actual_ctrlh

Similarly, Ctrl+m is expected to behave the same as Enter.
As in the following image.
expected_ctrlm

Currently, it behaves like the following image.
actual_ctrlm

So I modified it to behave as expected.
As far as I know, with this PR fix, it behaves the same as wezterm on Windows and X11.

However, I only use Japanese IMEs and have only tested the behavior with some Japanese IMEs. Therefore, I did not know if the problem occurs in other language IMEs.
I also read the source code, but could not determine if the exclusion of the control key was intentional or not.

Please let me know what you think about this PR. I would like to help in resolving this issue.

### DIFF
--- a/window/src/os/macos/window.rs
+++ b/window/src/os/macos/window.rs
@@ -2233,7 +2233,8 @@ impl WindowView {
         // Also respect `send_composed_key_when_(left|right)_alt_is_pressed` configs
         // when `use_ime` is true.
         let forward_to_ime = {
-            if modifiers.is_empty() || modifiers == Modifiers::SHIFT {
+            if modifiers.is_empty() || modifiers == Modifiers::SHIFT || modifiers == Modifiers::CTRL
+            {
                 true
             } else if only_left_alt && !send_composed_key_when_left_alt_is_pressed {
                 false


yuys13:
@wez What do you think about this pull request?

wez:
Sorry for the delay in responding to this; I wanted to take a deeper look at the surrounding code and didn't get the chance until now.  I've pushed an alternative implementation of this change to `main` and I would appreciate it if you could try it out and let me know how it works for you!

yuys13:
I built the `main` branch and tried it out. It works as expected for me!
Not only do `CTRL-H` and `CTRL-M` work, but `CTRL-N`, `CTRL-P`, etc. also work.
Thanks for creating a really cool terminal!!!

wez:
Note that this has been effectively reverted in `main`, but has been replaced by an option to allow you to opt-in to this if desired:

https://wezfurlong.org/wezterm/config/lua/config/macos_forward_to_ime_modifier_mask.html

PR #2439, opened 2022-08-21
Add `fig` to shells for shell completion

grant0417:
Adds support for generating [Fig](https://github.com/withfig/autocomplete) completions using [`clap_complete_fig`](https://github.com/clap-rs/clap/tree/HEAD/clap_complete_fig).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -539,6 +539,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "clap_complete_fig"
+version = "3.2.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


wez:
Thanks!

fedeci:
Hey @wez, would you be open to have a GitHub workflow running whenever a new wezterm version is released and updating the spec automatically?

wez:
I like the idea of it, but because it runs at release time, those sorts of flows need me to test/iterating on them and the last one I did took up most of a day, and I don't really have that amount of time to burn. I'm also entirely uninterested in fixing up stuff like https://github.com/withfig/autocomplete/pull/1439/files#r950861287 that might crop up due to weirdness in the definition/automation.

I'm inclined to say: let's see how many times it needs to be updated for fig in the next 6 months and then revisit to see if it is worth automating.

PR #2480, opened 2022-09-04
font: fix the function signature for log_buffer_message

liushuyu:
This pull request fixes the function signature for log_buffer_message.

This will fix the build on non-x86 architectures (where `c_char` is an `u8` instead of an `i8`).


wez:
Thanks!

PR #2491, opened 2022-09-06
colors: fix base16 cursor having the same bg and fg

unrelentingtech:
…has anyone been using these themes? :)

---

and I'm wondering if these themes should also set the `indexed` colors 16-21 like [base16-shell](https://github.com/chriskempson/base16-shell) does:

I really don't know anyone who would really care about always preserving the original indexed palette but :shrug:


wez:
Heh, thanks for this!

> and I'm wondering if these themes should also set the `indexed` colors 16-21 like [base16-shell](https://github.com/chriskempson/base16-shell) does:

Oh, I didn't realize that it did that. Yeah, I think we should do that on import too; want to include that in this PR?

unrelentingtech:
Done

PR #2492, opened 2022-09-06
wayland: only cancel key repeat when the *held* key was released, fixes #2452

### LINKED ISSUE #2452: Key repeat gets stuck after pressing two keys in fast succession

Linux Wayland

Sway

### WezTerm version

20220823-103207-53bc8711

### Describe the bug

The key repeat is stuck when pressing two keys in quick succession resulting in the second key printing only one character instead of continuously printing.

### To Reproduce

Go into a terminal. Press the key `A` on your keyboard and hold it. Release it and quickly press the letter `B` on the keyboard and hold it. `b` is only printed once, even though it should be continuously be printed.

### Configuration

no config

### Expected Behavior

Every key that is pressed and held should be repeated, even when keys are typed in quick succession.

### Logs

There seem to be no logs created from this action

### Anything else?

_No response_

unrelentingtech:


wez:
Looks good, thanks!

PR #2493, opened 2022-09-06
ci: correctly set the tag for release description

orhun:
The latest release have a release description as follows:

```md
See https://wezfurlong.org/wezterm/changelog.html#$name for the changelog
```

This PR sets the correct tag for this message instead of `$name`.

### Additional context

Quotes prevent parameter expansion while using [here-doc](https://en.wikipedia.org/wiki/Here_document#Unix_shells):

Without quotes:

```
a=0
cat <<EOF
$a
EOF

0
```

With quotes:

```
a=0
cat <<'EOF'
$a
EOF

$a
```

From `man bash`:

> If any part of word is quoted, the delimiter is the result of quote removal on word, and the lines in the here-document  are  not  expanded.
If word is unquoted, all lines of the here-document are subjected to parameter expansion, command substitution, and arithmetic expansion

### DIFF
--- a/ci/create-release.sh
+++ b/ci/create-release.sh
@@ -2,7 +2,7 @@
 set -x
 name="$1"
 
-notes=$(cat <<'EOT'
+notes=$(cat <<EOT
 See https://wezfurlong.org/wezterm/changelog.html#$name for the changelog
 
 If you're looking for nightly downloads or more detailed installation instructions:


orhun:
This is a regression in https://github.com/wez/wezterm/commit/ee436437b4294724c950602057ddd2a4aaaa7a55

PR #2495, opened 2022-09-06
Make Lua dependencies in procinfo crate optional and fix bug in parsing sysctl KERN_PROCARGS2

mikayla-maki:
Hi Wez,

I'm working on building the integrated terminal in [Zed](https://zed.dev/) and wanted to use your procinfo crate to show the executable in the title. To do so, we had to make the lua dependencies optional and found a bug: the `argv` Vec was always empty. When I dumped the buffer for the process spawned via `sleep 5` I got this buffer:

```
[2, 0, 0, 0, 47, 98, 105, 110, 47, 115, 108, 101, 101, 112, 0, 0, 0, 0, 0, 0, 115, 108, 101, 101, 112, 0, 53, 0,]
 ^ argc      ^exe_path                                         ^ !!            ^ argv[0]                   ^ argv[1]
```

Note the `!!`, for some reason there were trailing 0s between the exe_path and the `argv` fields. This seems to be a common problem with the KERN_PROCARGS2 format, note these two corroborating examples of I found googling around:

* Here's an example from [Chromium Crashpad](https://chromium.googlesource.com/crashpad/crashpad/+/360e441c53ab4191a6fd2472cc57c3343a2f6944/util/posix/process_util_mac.cc), note the `args.find_first_not_of('\0', nul_pos);` call on line 82 that trims out these trailing 0s
* Here's an example from [GNU Emacs](https://lists.gnu.org/archive/html/bug-gnu-emacs/2021-05/msg01652.html) which has two lines near the bottom of the diff which also manage this issue:
```
+      start += strlen (start) + 1; /* skip executable name and any '\0's */
+      while ((start - args < len) && ! *start) start++;
```

I implemented a preliminary solution to this by copying and modifying the code from `consume_cstr()` to move the pointer past the trailing zeroes in the buffer. That said, I've never worked with raw pointers and unsafe code like this, and am unsure if I did it correctly.


procinfo/src/macos.rs
```diff
@@ -178,6 +178,11 @@ impl LocalProcessInfo {
 
             let exe_path = consume_cstr(&mut ptr)?.into();
 
+            // For some reason, sysctl sometimes puts null bytes
+            // between the executable field and the argv field
+            let not_nul = ptr.iter().position(|&c| c != 0)?;
```
wez, line 183: perhaps this logic should be swept up into the bottom of `consume_cstr` above?

It would be great if you could encapsulate the example you shared in the description of the PR into a unit test for this; it would make the PR a bit bigger, but it feels worthwhile to have at least two tests that show this working for data that uses 1x NUL and 2x NUL separators.

I think factoring out a function like: `parse_exe_and_argv_sysctl(buf: &[u8]) -> Option<(PathBuf, Vec<String>)>` would make this more easily testable; we'd pass the buffer slice that we compute on line 167 to this function and return the result. Lines 169 through 191 would be the body of this new function.

mikayla-maki: That sounds like a good change, I'll make that and get back to this.

mikayla-maki:
Tests added! Lmk what you think

procinfo/src/macos.rs
```diff
@@ -227,3 +198,97 @@ impl LocalProcessInfo {
...
+        let buf = vec![
+            2, 0, 0, 0, 47, 98, 105, 110, 47, 115, 108, 101, 101, 112, 0, 115, 108, 101, 101, 112,
+            0, 53, 0,
+        ];
+
+        let (exe_path, argv) = parse_exe_and_argv_sysctl(buf).unwrap();
+
+        assert_eq!(exe_path, Path::new("/bin/sleep").to_path_buf());
+        assert_eq!(argv[0], "sleep".to_string());
+        assert_eq!(argv[1], "5".to_string());
```
wez, line 276: nit: as written, the test wouldn't fail if we somehow ended up with more than 2 elements in the array. Re-phrasing to assert the whole array is a bit more robust; something like this:

```suggestion
        assert_eq!(argv, vec![
             "sleep".to_string(),
             "5".to_string()
         ]);
```

(same for the other tests!)

procinfo/src/macos.rs
```diff
@@ -227,3 +198,97 @@ impl LocalProcessInfo {
...
+    let argc: c_int = unsafe { std::ptr::read(ptr.as_ptr() as *const c_int) };
+    ptr = &ptr[std::mem::size_of::<c_int>()..];
+
+    fn consume_cstr(ptr: &mut &[u8]) -> Option<String> {
+        // Trim out trailing nulls
+        let not_nul = ptr.iter().position(|&c| c != 0)?;
+        *ptr = ptr.get(not_nul..)?;
+
+        // Then parse to the end of the null terminated string
+        let nul = ptr.iter().position(|&c| c == 0)?;
```
wez, line 226: From a safety perspective, the biggest concern with the pointer manipulation is whether we walk outside of the buffer.  I think this code is OK because of the way that the slice manipulations are written, but it would be good to add a test case where we have multiple trailing zeroes at the end of the buffer to assert that we don't somehow walk off the very end

wez: also one where there are no trailing zeroes!

mikayla-maki:
Tests extended!

procinfo/src/macos.rs
```diff
@@ -227,3 +198,121 @@ impl LocalProcessInfo {
...
+
+    let mut ptr = &buf[0..buf.len()];
+
+    let argc: c_int = unsafe { std::ptr::read(ptr.as_ptr() as *const c_int) };
+    ptr = &ptr[std::mem::size_of::<c_int>()..];
+
+    fn consume_cstr(ptr: &mut &[u8]) -> Option<String> {
+        // Trim out trailing nulls
+        let not_nul = ptr.iter().position(|&c| c != 0)?;
+        *ptr = ptr.get(not_nul..)?;
```
wez, line 223: This comment doesn't match what's happening here: it's skipping leading NULs.  I think this is a bit counterintuitive: I think this function is a bit easier to reason about if we know that `ptr` points to the start of a c-string and then updates ptr to point to the start of the next one when it returns.

I think moving the nul-skipping to the bottom of this function would satisfy that!

wez (commented):
Many thanks for this!
I have one last comment and then I think we can merge this!

procinfo/src/macos.rs
mikayla-maki: This required me to add a branch to make sure we haven't run off the end, added!

mikayla-maki:
Done! I also added some comments for why the rearrangement is safe as it took me a few tries to get it right. Thanks for the advice on the tests!

wez:
Thank you!

PR #2509, opened 2022-09-12
Added installation links to the README.

HamishPoole:
Added installation links back, to cater for folk may not be up for reading documentation.

`powershell.exe -Command "curl.exe https://api.github.com/repos/wez/wezterm/releases/latest | ForEach-Object {if ((Get-Variable -name PSItem).Value -match 'https://github.com/wez/wezterm/releases/.*setup.exe(?!.{2,})'){curl.exe -OjL (Get-Variable -name Matches).Value[0]}}" `

Also had a look at above as a built-in alternative to winget for cmd.exe.  winget was far simpler.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,16 @@ User facing docs and guide at: https://wezfurlong.org/wezterm/
 
 https://wezfurlong.org/wezterm/installation.html
 
+#### Windows
+```winget install wez.wezterm```
+
+#### macOS
+```brew tap wez/wezterm && brew install --cask wez/wezterm/wezterm```
+
+#### Linux:
+Binaries, packages, & portable distributions available in the installation guide.
+
+
 ## Getting help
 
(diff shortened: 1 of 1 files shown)


wez:
Thanks for this, but it's making assumptions that people use eg: winget and brew, which may turn folks away if they don't use them (I don't use them, for example).

Installation is a lot more nuanced than just those one liners and I prefer that people do actually read the docs!

HamishPoole:
> Installation is a lot more nuanced than just those one liners and I prefer that people do actually read the docs!

image

I agree totally.  In an ideal world, that's how it should be.
 
The issue is.. some (many) people will go to pretty extreme lengths to avoid documentation at the best of times.

You almost need a safety valve to capture those folk who don't click through, which was what kept me up at night.  Hence the cmd.exe script with curl.exe.

wez:
Again, thanks, but I don't want this!

PR #2510, opened 2022-09-12
Trivial example revision of default_key_tables()

choplin:


docs/config/lua/wezterm.gui/default_key_tables.md
```diff
@@ -17,7 +17,7 @@ local act = wezterm.action
 
 local copy_mode = nil
 if wezterm.gui then
-  copy_mode = wezterm.gui.default_key_tables().copy_mode
+  local copy_mode = wezterm.gui.default_key_tables().copy_mode
```
wez, line 20: `copy_mode` is already declared as local above

choplin: Ah, correct. Sorry for the misreading.

PR #2516, opened 2022-09-13
Use SVG icon

LitoMore:
### Preview

|Before|After|
|:-:|:-:|
|||

The actual size is 128x128. I doubled its size here for a better preview. Their visual difference is especially noticeable on Retina display monitors.

BTW, I noticed this icon problem while working on https://github.com/simple-icons/simple-icons/pull/7774.


wez:
Thanks!

PR #2538, opened 2022-09-21
Add flag to force fallback rendering

muirdm:
Hi Wez - wanted some early feedback if I'm on the right track here.

I think the sgr fallback you were referring to offline is when we don't have `self.get_capability::<cap::SetAttributes>`. However, the "else" branch does an ExitAttributeMode AKA sgr0=\E[m\017 (also contains the problematic \017). So, I added a flag that short circuits all the capability checks in there. How close am I?

### DIFF
--- a/termwiz/src/caps/mod.rs
+++ b/termwiz/src/caps/mod.rs
@@ -107,6 +107,13 @@ builder! {
 
         /// Whether mouse support is present and should be used
         mouse_reporting: Option<bool>,
+
+        /// When true, rather than using the terminfo `sgr` or `sgr0` entries,
+        /// assume that the terminal is ANSI/ECMA-48 compliant for the
+        /// common SGR attributes of bold, dim, reverse, underline, blink,
+        /// invisible and reset, and directly emit those sequences.
+        /// This can improve rendered text compatibility with pagers.
+        force_terminfo_render_to_use_ansi_sgr: Option<bool>,
     }
 }
 
@@ -154,6 +161,7 @@ pub struct Capabilities {
     terminfo_db: Option<terminfo::Database>,
     bracketed_paste: bool,
     mouse_reporting: bool,
+    force_terminfo_render_to_use_ansi_sgr: bool,
 }
 
 impl Capabilities {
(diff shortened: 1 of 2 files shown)


termwiz/src/render/terminfo.rs
```diff
@@ -50,13 +50,17 @@ impl TerminfoRenderer {
     #[cfg_attr(feature = "cargo-clippy", allow(clippy::cognitive_complexity))]
     fn flush_pending_attr<W: RenderTty + Write>(&mut self, out: &mut W) -> Result<()> {
         macro_rules! attr_on {
-            ($cap:ident, $sgr:expr) => {
-                if let Some(attr) = self.get_capability::<cap::$cap>() {
+            ($cap:ident, $sgr:expr) => {{
+                let mut cap = self.get_capability::<cap::$cap>();
+                if self.caps.force_terminfo_sgr_fallback() {
+                    cap = None;
+                }
```
wez, line 57: Minor stylistic nit: I would phrase it this way around to avoid the less-idiomatic use of `mut cap`; same suggestion for the code around `cap::SetAttributes` below:

```suggestion
                let cap = if self.caps.force_terminfo_sgr_fallback() {
                   None
                } else {
                    self.get_capability::<cap::$cap>()
                }
```

termwiz/src/caps/mod.rs
```diff
@@ -107,6 +107,8 @@ builder! {
 
         /// Whether mouse support is present and should be used
         mouse_reporting: Option<bool>,
+
+        force_terminfo_fallback_rendering: Option<bool>,
```
wez, line 111: Naming wise, how about this:

```suggestion
        /// When true, rather than using the terminfo `sgr` or `sgr0` entries,
        /// assume that the terminal is ANSI/ECMA-48 compliant for the
        /// common SGR attributes of bold, dim, reverse, underline, blink,
        /// invisible and reset, and directly emit those sequences.
        force_terminfo_render_to_use_ansi_sgr: Option<bool>,
```

wez (commented):
Yeah, that looks right. Couple of comments inline!

termwiz/src/caps/mod.rs
muirdm: Added one extra sentence about why you might want to use it.

wez:
Thanks; published termwiz-0.18.0 with this change

PR #2558, opened 2022-09-27
Doc improvement - open the last line in vim

sayanarijit:
Sorry for the auto formatter changes.


wez:
Thanks for your time on this, but I don't want this change:

* The formatting changes go against how I like to organize and see the docs
* The vim jump is subjective and different people will want different behavior, and I don't want these docs to be bumped by each person with a different preference

Thanks again!

PR #2578, opened 2022-10-03
feat: support Natilus 4.0 API

GabMus:
As explained in the comments, the Nautilus 4.0 API just passes files without the window object (which was unused anyway). This small fix keeps compatibility with the 3.0 API and checks the number of arguments.

In addition, `require_version` raises an error since version 4.0 has already been required by Nautilus itself, so I just removed it.

Would be really nice if you could make this count towards hacktoberfest as well!

### DIFF
--- a/assets/wezterm-nautilus.py
+++ b/assets/wezterm-nautilus.py
@@ -17,7 +17,6 @@
 import os.path
 
 from gi import require_version
-require_version('Nautilus', '3.0')
 from gi.repository import Nautilus, GObject, Gio, GLib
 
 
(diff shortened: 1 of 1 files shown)


wez:
Thanks for this!

> Would be really nice if you could make this count towards hacktoberfest as well!

I'm not sure what that entails: I don't have any relationship with hacktoberfest

GabMus:
Thanks for the reply, and happy to contribute!

As for hacktoberfest, you'd have to register as a maintainer on their website, assuming it's something you are fine with. If not, don't worry about it, I would still have made this contribution regardless.

wez:
I'm not going to sign up to maintain another project :-)
But thank you for for this!

PR #2600, opened 2022-10-07
Expose applying an AttributeChange on CellAttributes

groves:
This moves Surface.change_attribute to CellAttributes.apply. 

I'm currently recreating change_attribute in my application to be able to update a CellAttributes as the terminal stream is parsed. This lets me call CellAttributes.apply instead of duplicating: https://github.com/groves/ate/commit/3e121a65e4c9dd9240edc6154890d356cdbf1c8f

I believe the existing tests in surface/mod.rs exercise this pretty well, but I'm happy to add anything that seems like it'd be a good addition.

Thanks for creating such a splendid library!


termwiz/src/cell.rs
```diff
@@ -532,6 +532,42 @@ impl CellAttributes {
             .map(|fat| fat.underline_color)
             .unwrap_or(ColorAttribute::Default)
     }
+
+    pub fn apply(&mut self, change: &AttributeChange) {
```
wez, line 536: nit: naming wise, apply is very generic, so I'd suggest renaming to something like:

```suggestion
    pub fn apply_change(&mut self, change: &AttributeChange) {
```

wez:
Thanks for this! The change sounds reasonable; just a minor naming suggestion!

wez:
Thanks!

PR #2601, opened 2022-10-07
Allow cursor visibility to be changed by widgets

groves:
This adds visibility to CursorShapeAndPosition to let it be modified along with shape, coords, and color.


wez:
Looks good, thanks!

PR #2608, opened 2022-10-09
Add a State parameter to Ui that's passed to all widget calls

groves:
I'm building a terminal pager using terwiz and its widget API. It's been a huge boost to the whole project. I have been struggling with modifying shared state from the widgets though.

In the pager, there's a document widget that displays the current page and a search widget that lets you type in what you want to see. They need to share some state to coordinate on what's shown. I created a context struct with the shared state and use a ton of Rcs and RefCells to let me get to it from the widgets: https://github.com/groves/ate/blob/6879bf4b1bb8edc5857182b4d7689b795ad8731d/src/ui.rs#L84 I'm new to Rust, but that's easily the grossest Rust code I've written so far.

Poking about for ways to expose state to Rust UIs, I found [this proposal for Xilem](https://raphlinus.github.io/rust/gui/2022/05/07/ui-architecture.html). The tl;dr on that is that you pass around a state object to your UI calls and that lets you share state while still letting the UI toolkit maintain ownership of the widgets.

This PR is the simplest way I could come up with to play with that idea. I used it to rework my state sharing in the pager app, and the sheer reduction in RefCells makes me feel like it's a better approach: https://github.com/groves/ate/compare/main...widget-ui-state#diff-87619baf236177a7a9fc38150961151b7583b6f2f082ade75915f3ac4ca49c41R37

I'm not tied to this approach at all. I opened a PR more to open a discussion than to truly advocate for this change. Would love to know if this seems worth pursuing or if I should try something else. Thanks!

### DIFF
--- a/termwiz/src/widgets/mod.rs
+++ b/termwiz/src/widgets/mod.rs
@@ -53,16 +53,16 @@ pub struct UpdateArgs<'a> {
 
 /// Implementing the `Widget` trait allows for defining a potentially
 /// interactive component in a UI layout.
-pub trait Widget {
+pub trait Widget<State> {
     /// Draw the widget to the RenderArgs::surface, and optionally
     /// update RenderArgs::cursor to reflect the cursor position and
     /// display attributes.
-    fn render(&mut self, args: &mut RenderArgs);
+    fn render(&mut self, args: &mut RenderArgs, state: &mut State);
 
     /// Override this to have your widget specify its layout constraints.
     /// You may wish to have your widget constructor receive a `Constraints`
     /// instance to make this more easily configurable in more generic widgets.
-    fn get_size_constraints(&self) -> layout::Constraints {
+    fn get_size_constraints(&self, _state: &State) -> layout::Constraints {
         Default::default()
     }
 
@@ -70,7 +70,12 @@ pub trait Widget {
     /// other widget events.
     /// Return `true` if your widget handled the event, or `false` to allow
     /// the event to propagate to the widget parent.
-    fn process_event(&mut self, _event: &WidgetEvent, _args: &mut UpdateArgs) -> bool {
+    fn process_event(
+        &mut self,
+        _event: &WidgetEvent,
+        _args: &mut UpdateArgs,
+        _state: &mut State,
+    ) -> bool {
         false
     }
 }
@@ -137,11 +142,11 @@ impl Default for WidgetId {
     }
 }
 
-struct RenderData<'widget> {
+struct RenderData<'widget, State> {
     surface: Surface,
     cursor: CursorShapeAndPosition,
     coordinates: ParentRelativeCoords,
-    widget: Box<dyn Widget + 'widget>,
+    widget: Box<dyn Widget<State> + 'widget>,
 }
 
 #[derive(Default)]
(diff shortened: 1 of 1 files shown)


wez:
Hmm, the way I've been looking at this, the state would live in the impl of the Widget:

```rust
struct SharedState {
   something: bool,
}

struct DocumentWidget {
   state: Arc<SharedState>,
}

struct SearchWidget {
   state: Arc<SharedState>,
}

impl Widget for DocumentWidget {
   fn render(&mut self, args: &mut RenderArgs) {
       let something = self.state.something;
       ...
   }
}
```

Would that help in your scenario?

I'm not hugely in favor of making `Widget` generic over some other type as I think that has potential to make a lot of things more awkward if different `State` types need to be mixed in an application, which may make it harder to share widget implementations from other crates.

Something that I think is missing is the ability to see into the Widget if all you have is a `dyn Widget`; not sure if you're running up against that, but a solution for that is the [downcast](https://docs.rs/downcast/latest/downcast/) crate; wezterm uses that for `Pane` in the internal multiplexer:

https://github.com/wez/wezterm/blob/5feadb347eb1f6fe45c534c5186187e537b714a2/mux/src/pane.rs#L188-L189
https://github.com/wez/wezterm/blob/5feadb347eb1f6fe45c534c5186187e537b714a2/mux/src/pane.rs#L363-L364

and then you could runtime cast to eg: `DocumentWidget` or `SearchWidget` using something like:

https://github.com/wez/wezterm/blob/5feadb347eb1f6fe45c534c5186187e537b714a2/wezterm-client/src/client.rs#L157-L163

What do you think?

groves:
> Hmm, the way I've been looking at this, the state would live in the impl of the Widget. Would that help in your scenario?

That was how I started. Since I needed to change the state in response to user input in `process_event`, I needed to make the state on the widget mutable. As far as I could tell, the only way to do that in something sharable like `Rc` or `Arc` is by putting a `Cell` or a `RefCell` inside it. That meant scattering `borrow` and `borrow_mut` calls all over the place. Cutting out the compiler from checking my borrows made me nervous and felt gross overall.

That said, I'm new to Rust, so I may be missing a better way to do it!
 
> I'm not hugely in favor of making `Widget` generic over some other type as I think that has potential to make a lot of things more awkward if different `State` types need to be mixed in an application, which may make it harder to share widget implementations from other crates.

That makes sense. The way Xilem proposes to handle this is by having optional "adapters" at each level of the widget tree. If a particular widget needs only a subset of the data state, you provide an adapter function from the data state the parent uses to the data state the child uses. 

Thinking of a using list widget provided by a library, it seems like you'd need to write an adapter from your application's data state to extract the vec the list expects to render and modify. If that explanation's unclear, I'm happy to take a swing at adding it to this PR.

> Something that I think is missing is the ability to see into the Widget if all you have is a `dyn Widget`

This may be me not knowing Rust well again, but as far as I can tell, there's no way to get a reference to a widget once you've added it to `Ui`. `Ui` owns the widgets, which makes sense to me. It doesn't have any accessors that I can see.

That said, I haven't had a need yet to get back to widgets after adding them, at least not with this external state approach. The widgets only need to make a call into the state and the state can be responsible for any cross-widget concerns.

PR #2631, opened 2022-10-15
termwiz: hardcode widths of certain codepoints

jbenden:
Hi,

This patch fixes a tiny typographical rendering bug involving dashes (em dash) that are partially overwritten by the next character. The actual problem is that these characters are actually width of two(2), but report as one(1). This patches hard-codes a solution for these characters.

Best regards,
-Joe

### DIFF
--- a/termwiz/src/widechar_width.rs
+++ b/termwiz/src/widechar_width.rs
@@ -1481,6 +1481,9 @@ impl WcWidth {
             return Self::Two;
         }
         if in_table(&AMBIGUOUS_TABLE, c) {
+            if c == 0x2014 || c == 0x2015 {
+                return Self::Two;
+            }
             return Self::Ambiguous;
         }
         if in_table(&UNASSIGNED_TABLE, c) {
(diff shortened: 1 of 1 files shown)


wez:
Thanks for diving in, however, both of those codepoints are defined as single-cell width in the context of a terminal, so changing the width here is not the correct thing to do: it will cause misalignment in various TUI programs.

```
; wezterm ls-fonts --codepoints 2014,2015
LeftToRight
 0 —    \u{2014}     x_adv=8.3125 cells=1  glyph=emdash,278  wezterm.font("Operator Mono SSm Lig", {weight="DemiLight", stretch="Normal", style="Normal"})
                                      /Users/wez/.fonts/OperatorMonoSSmLig-Medium.otf, FontDirs
 3 ―    \u{2015}     x_adv=9.815340909090908 cells=1  glyph=110  wezterm.font(".Apple SD Gothic NeoI", {weight="Regular", stretch="Normal", style="Normal"})
                                      /System/Library/Fonts/AppleSDGothicNeo.ttc index=1 variation=0, CoreText
```

On my system the glyph for U+2014 is found in my monospace font and renders within the bounds of the cell, but U+2015 comes from a proportional font and overflows the cell bounds.

This isn't a wezterm bug; having a proper monospace font with those glyphs will resolve it.

jbenden:
Thanks! I hope this helps others, too!
-Joe

PR #2636, opened 2022-10-16
Fix background cover algorithm

xiaopengli89:
Before: 
image
After:
image


wez:
Thanks for this! Sorry it has taken a while to get to reviewing it!

PR #2661, opened 2022-10-22
Fix relative CWD path given on command line being wrongly

exactly-one-kas:
This allows `wezterm start --cwd .` to work when another terminal window is already open in another directory

This has worked before, but it might be that the server wasn't spawned automatically?


wez:
Thanks!

PR #2670, opened 2022-10-26
Support passing the cmdline via "-e" to wezterm

vimpostor:
The easiest way to get this working with `clap` was to implement this using a dummy `-e` option.

This might look hacky at first, but it seems to be working quite reliably, for example the following command will still do the expected thing:

```bash
wezterm start -e echo --position 10,10
```

That is, it will echo "--position 10,10" and **not** parse it as a wezterm argument.

In the second commit I remove the "--" from the desktop file, which should be a better default, given how it improves the compatibility.


wezterm-gui-subcommands/src/lib.rs
```diff
@@ -163,6 +163,18 @@ pub struct StartCommand {
...
+    /// Dummy argument that consumes "-e" and does nothing.
+    /// This is meant as a compatibility layer for supporting the
+    /// widely adopted standard of passing the command to execute
+    /// to the terminal via a "-e" option.
+    /// This works because we then treat the remaining cmdline as
+    /// trailing options, that will automatically be parsed via the
+    /// existing "prog" option.
+    /// This option exists only as a fallback. It is recommended to pass
+    /// the command as a normal trailing command instead if possible.
+    #[arg(long = "command", short = 'e', hide = true)]
```
wez, line 175: Do we also need `--command`? I'd prefer not to use it for this if we don't need it.

```suggestion
    #[arg(short = 'e', hide = true)]
```

wezterm-gui-subcommands/src/lib.rs
```diff
@@ -163,6 +163,18 @@ pub struct StartCommand {
...
+    pub cmd: bool,
```
wez, line 176:
```suggestion
    _cmd: bool,
```

wezterm-gui/src/main.rs
```diff
@@ -139,6 +139,7 @@ async fn async_run_ssh(opts: SshCommand) -> anyhow::Result<()> {
         always_new_process: true,
         class: opts.class,
         cwd: None,
+        cmd: false,
```
wez, line 142: maybe replace with `.. Default::default()` construction, so we can avoid referencing the dummy arg?

wezterm-gui-subcommands/src/lib.rs
vimpostor: I agree with you.

wezterm-gui-subcommands/src/lib.rs
vimpostor: Removing the `pub` won't be possible, as neither the `Default::default()` nor the explicit initialization in wezterm-gui/src/main.rs will work then.

wez:
Thanks!

PR #2681, opened 2022-10-30
Elaborate error message for unsupported height/width values for Color and Gradient

KevinSilvester:
The current background implementation of  `Color` and `Gradient` doesn't support `Cover` or `Contain` for its height and width values.

I couldn't find any mention of this in the docs and the only indication of this fact is in a vague error message logged onto the screen when `wezterm` is launched through the cli.

I'm not sure whether this feature is in development or not, so all I did was elaborate on the existing error message and added a small notice for this fact in the background docs with an example.

This should hopefully minimize the headache for anyone who couldn't get non-`File` background sources working. 🙂


wez:
Thanks! I took just the error messaging and refined it a bit!

PR #2698, opened 2022-11-02
Add left arrow before first right-status cell

aaronkollasch:
Adds a powerline left arrow before the first cell of the right-status bar in `window:set_right_status.md`.

This prevents the first cell from having a square border if the tab_bar's background color is different from the right-status cell's background.


wez:
Thanks for this, but I don't think this is the best place for this: the example is based on something from one of my configs to illustrate what is possible, rather than to be used as a library.

I do want to put some effort into bootstrapping a way to build and share modules for wezterm configs and when that materializes, this sort of contribution would fit better in that.

I'm going to close this. Thanks again!

PR #2714, opened 2022-11-06
fixes flow chart and typo in config file docs

JakeHL:
Super lightweight PR that fixes the flow chart skipping the "can it be loaded" stage for "$XDG_CONFIG_HOME environment set" stage.

Also fixes a typo `uses` should be `users`.


wez:
Thanks!

Issue #1180, opened 2021-09-28
Fancy tab/title bar/frame


wez:
This issue is tracking the work needed to provide a Chrome/Firefox style integrated tab/titlebar.

The intent is for it to be usable in place of the existing tab bar, potentially replacing it completely in the future; a config option will allow choosing between the two while things are worked out.

There are some considerations that need to be made for how to translate the existing monospace/cell based tab and right-status configurations to the new thing.  I'll punt those for now

The custom titlebar has a number of platform specific considerations that are architecturally a bit awkward:

* On Windows systems, https://docs.microsoft.com/en-us/windows/win32/dwm/customframe explains how to make a custom frame
* On macOS, iTerm2 uses https://github.com/gnachman/iTerm2/tree/master/ThirdParty/PSMTabBarControl which has a lot of code to read and understand. I don't want to (and can't really) directly use it, but can draw from how it works
* On X11 we can simply disable window decorations and then draw everything ourselves
* On Wayland we handle this in the existing custom frame stuff
* On Windows, the minimize/maximize/close buttons are on the right hand size of the window are in a different order from those on macOS, which are on the left side of the window
* On Windows, the hit testing stuff has has callouts from the window proc

Anyway, with a design that accommodates the above, we could have a nice looking tab/title bar with new tab button that is more visually appealing that the simpler cell based tab bar.

sdrik:
Hi,

Now that the fancy tab bar is the default, I'm wondering if the old tab bar will still be maintained.
Because if that wasn't the case, it would totally ruin my setup.

`use_fancy_tab_bar = false` connected to a local tmux
![Capture d’écran de 2021-12-28 12-07-04](https://user-images.githubusercontent.com/623981/147560778-1d5a1e92-8f91-42d6-9de7-db1edbbb14b9.png)
_(the bottom bar comes from wezterm, the bar just above comes from tmux)_

`use_fancy_tab_bar = false` connected to a remote tmux
![Capture d’écran de 2021-12-28 12-35-20](https://user-images.githubusercontent.com/623981/147562304-8d5366de-b5ba-47e4-bd52-21a39251a566.png)
_(the bottom bar comes from wezterm, the bar just above comes from tmux)_

`use_fancy_tab_bar = true` connected to a local tmux
![Capture d’écran de 2021-12-28 11-51-06](https://user-images.githubusercontent.com/623981/147560901-6b8c5d95-bd01-49d8-8179-6ae3672221a4.png)
[...]

PR #2722, opened 2022-11-07
Integrated titlebar

yuraiz:
I made simple version of chrome-like integrated titlebar #1180

With fancy tab bar:
Screenshot from 2022-11-07 21-30-38

Without fancy tab bar
Screenshot from 2022-11-07 21-30-28

Video recording with buttons in action

[Kooha-2022-11-07-21-51-16.webm](https://user-images.githubusercontent.com/63008755/200391085-48f5179c-32c7-48d7-b7b6-790b7ef5ec0f.webm)

I have linux (with gnome), macos and windows, so I think I can make it for those platforms, but I don't have freebsd

**Todo:**
- [x] use two icons for maximize button
- [x] use custom config colors (currently used colors from the new tab button)
- [x] use different button styles (windows | macos | gnome | ??)
- [x] use the current system layout (left | right and hide useless buttons)
- [x] finish macos style
- [ ] Other suggestions

### DIFF
--- a/config/src/color.rs
+++ b/config/src/color.rs
@@ -534,11 +534,12 @@ fn default_active_tab() -> TabBarColor {
 }
 
 // Colors for window buttons
-#[cfg(not(windows))]
+#[cfg(not(any(windows, target_os = "linux")))]
 fn default_window_button_color() -> TabBarColor {
     default_inactive_tab()
 }
-#[cfg(not(windows))]
+
+#[cfg(not(any(windows, target_os = "linux")))]
 fn default_window_button_hover_color() -> TabBarColor {
     default_inactive_tab_hover()
 }
@@ -591,6 +592,25 @@ fn default_window_button_hover_color() -> TabBarColor {
 fn default_window_close_hover_color() -> TabBarColor {
     TabBarColor {
         bg_color: (0xFF, 0x00, 0x00).into(),
+        fg_color: (0xFF, 0xFF, 0xFF).into(),
+        ..TabBarColor::default()
+    }
+}
+
+// Linux dependend colors for window buttons
+#[cfg(target_os = "linux")]
+fn default_window_button_color() -> TabBarColor {
+    TabBarColor {
+        bg_color: (0x45, 0x45, 0x45).into(),
+        fg_color: (0xFF, 0xFF, 0xFF).into(),
+        ..TabBarColor::default()
+    }
+}
+
+#[cfg(target_os = "linux")]
+fn default_window_button_hover_color() -> TabBarColor {
+    TabBarColor {
+        bg_color: (0x4F, 0x4F, 0x4F).into(),
         fg_color: (0xFF, 0xFF, 0xFF).into(),
         ..TabBarColor::default()
     }
--- a/config/src/color.rs
+++ b/config/src/color.rs
@@ -372,24 +372,6 @@ pub struct TabBarColors {
 
     #[dynamic(default)]
     pub inactive_tab_edge_hover: Option<RgbaColor>,
-
-    #[dynamic(default)]
-    pub window_hide: Option<TabBarColor>,
-
-    #[dynamic(default)]
-    pub window_hide_hover: Option<TabBarColor>,
-
-    #[dynamic(default)]
-    pub window_maximize: Option<TabBarColor>,
-
-    #[dynamic(default)]
-    pub window_maximize_hover: Option<TabBarColor>,
-
-    #[dynamic(default)]
-    pub window_close: Option<TabBarColor>,
-
-    #[dynamic(default)]
-    pub window_close_hover: Option<TabBarColor>,
 }
 
 impl TabBarColors {
@@ -433,42 +415,6 @@ impl TabBarColors {
             .unwrap_or_else(default_inactive_tab_edge_hover)
     }
 
-    pub fn window_hide(&self) -> TabBarColor {
-        self.window_hide
-            .clone()
-            .unwrap_or_else(default_window_hide_color)
-    }
-
-    pub fn window_hide_hover(&self) -> TabBarColor {
-        self.window_hide_hover
-            .clone()
-            .unwrap_or_else(default_window_hide_hover_color)
-    }
-
-    pub fn window_maximize(&self) -> TabBarColor {
-        self.window_maximize
-            .clone()
-            .unwrap_or_else(default_window_maximize_color)
-    }
-
-    pub fn window_maximize_hover(&self) -> TabBarColor {
-        self.window_maximize_hover
-            .clone()
-            .unwrap_or_else(default_window_maximize_hover_color)
-    }
-
-    pub fn window_close(&self) -> TabBarColor {
-        self.window_close
-            .clone()
-            .unwrap_or_else(default_window_close_color)
-    }
-
-    pub fn window_close_hover(&self) -> TabBarColor {
-        self.window_close_hover
-            .clone()
-            .unwrap_or_else(default_window_close_hover_color)
-    }
-
     pub fn overlay_with(&self, other: &Self) -> Self {
         macro_rules! overlay {
             ($name:ident) => {
@@ -488,12 +434,35 @@ impl TabBarColors {
             inactive_tab_edge_hover: overlay!(inactive_tab_edge_hover),
             new_tab: overlay!(new_tab),
             new_tab_hover: overlay!(new_tab_hover),
-            window_hide: overlay!(window_hide),
-            window_hide_hover: overlay!(window_hide_hover),
-            window_maximize: overlay!(window_maximize),
-            window_maximize_hover: overlay!(window_maximize_hover),
-            window_close: overlay!(window_close),
-            window_close_hover: overlay!(window_close_hover),
+        }
+    }
+}
+
+#[derive(Default, Debug, Clone, PartialEq, FromDynamic, ToDynamic)]
+#[dynamic(try_from = "String")]
+pub enum IntegratedTitleButtonColor {
+    #[default]
+    Auto,
+    Custom(RgbaColor),
+}
+
+impl Into<String> for IntegratedTitleButtonColor {
+    fn into(self) -> String {
+        match self {
+            Self::Auto => "auto".to_string(),
+            Self::Custom(color) => color.into(),
+        }
+    }
+}
+
+impl TryFrom<String> for IntegratedTitleButtonColor {
+    type Error = <RgbaColor as TryFrom<String>>::Error;
+
+    fn try_from(value: String) -> Result<Self, Self::Error> {
+        if value.to_lowercase() == "auto" {
+            Ok(Self::Auto)
+        } else {
+            Ok(Self::Custom(RgbaColor::try_from(value)?))
         }
     }
 }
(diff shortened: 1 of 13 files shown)


YuraIz:
That how titlebar for windows looks

image

image

image

image

YuraIz:
Default colors for windows

https://user-images.githubusercontent.com/63008755/200615149-60ce975a-c31b-4253-9114-05ea2ed65f5f.mp4

YuraIz:
Libadwaita styled buttons

image

image

with rounded corners extension:
image

YuraIz:
Button placement at left

Screenshot from 2022-11-09 11-13-33

hosxy:
It looks beautiful，and will it support wayland ?

YuraIz:
> will it support wayland?

Yes, it just works as "RESIZE" option but with buttons
you can try it already

for some reason "RESIZE" doesn't support double click to maximize and resize on top corners with wayland

currently it doesn't work correctly with xorg)
image

wez:
Just wanted to say: Thanks for taking this on!

I'm excited to see it completed and review it once you're ready for feedback; there are a couple of stylistic things (eg: a few hard coded pixel sizes) that I'd like to see cleaned up in the final version, but I'd rather encourage you to get things functionally complete before focusing at that level!

YuraIz:
x11 version has strange bug (now both in "resize" and "fancy" variants):
image

looks like it draws window decorations behind the window

YuraIz:
Also both wayland and x11 modes doesn't support maximize on double click action, but it supported on windows

YuraIz:
For macos I use system buttons, that's how it looks
Screen Shot 2022-11-13 at 11 32 41 PM

But they doesn't work whell with `use_fancy_tab_bar = false`
Screen Shot 2022-11-13 at 11 33 56 PM

YuraIz:
> Also both wayland and x11 modes doesn't support maximize on double click action, but it supported on windows

Macos also doesn't support that, I think that we need to handle it in the mouseevent.rs somehow

YuraIz:
I think I can add "double click to maximize" function later

wezterm-gui/src/tabbar.rs
```diff
@@ -179,6 +182,141 @@ impl TabBarState {
...
+            window::FancyWindowDecorations::default()
+        } else {
+            config.fancy_window_decorations.clone()
+        };
+
+        let button_order = if is_left {
+            ['X', '.', '-']
+        } else {
+            ['.', '-', 'X']
+        };
```
wez, line 262: As a stylistic nit, using characters for these is prone to typos and the need for a catch-all in the match statement below.

I'd suggest using a little enum for this:

```rust
enum Button {
   Hide,
   Maximize,
   Close,
}

let button_order = if is_left {
   [Button::Close, Button::Hide, Button::Maximize]
} else {
    [Button::Hide, Button::Maximize, Button::Close]
};
```

and then using that below; it would also mean that you wouldn't need the comments alongside those cases to explain that `'.'` means hide and so on.

wezterm-gui/src/termwindow/render.rs
```diff
@@ -145,6 +145,281 @@ const PLUS_BUTTON: &[Poly] = &[
...
+    is_maximized: bool,
+    font: &Rc<LoadedFont>,
+    metrics: &RenderMetrics,
+    colors: &TabBarColors,
+    style: window::FancyWindowDecorationsStyle,
+) -> Element {
+    use window::FancyWindowDecorationsStyle as Style;
+
+    #[cfg(windows)]
+    let style = Style::Windows;
```
wez, line 292: Can we remove this? In general, I'd love to avoid as many system-dependent overrides like this in the main body of code as possible.

wezterm-gui/src/termwindow/render.rs
```diff
@@ -145,6 +145,281 @@ const PLUS_BUTTON: &[Poly] = &[
...
+                _ => 18.0,
+            };
+            let scale = 72.0 / 96.0;
+
+            element
+                .zindex(1)
+                .vertical_align(VerticalAlign::Middle)
+                .padding(BoxDimension {
+                    left: Dimension::Points(left_padding * scale),
+                    right: Dimension::Points(18. * scale),
```
wez, line 353: these hard coded numbers should also be moved into the OS parameters struct

wezterm-gui/src/termwindow/render.rs
```diff
@@ -817,6 +1104,15 @@ impl super::TermWindow {
             match item.item {
                 TabBarItem::LeftStatus => left_status.push(item_to_elem(item)),
                 TabBarItem::None | TabBarItem::RightStatus => right_eles.push(item_to_elem(item)),
+                TabBarItem::WindowHideButton
+                | TabBarItem::WindowMaximizeButton
+                | TabBarItem::WindowCloseButton => {
+                    if self.config.fancy_window_decorations.is_left && cfg!(not(windows)) {
```
wez, line 1110: can we eliminate the `cfg!` portion of this from here and keep this purely in terms of the data we have available via `self.config` and the OS parameters?

wezterm-gui/src/termwindow/render.rs
```diff
@@ -899,12 +1195,31 @@ impl super::TermWindow {
             );
         }
 
+        let window_buttons_at_left = self.config.window_decorations
+            == window::WindowDecorations::FANCY
+            && (self.config.fancy_window_decorations.is_left
+                || (cfg!(target_os = "macos")) && cfg!(not(windows)));
+
+        let left_padding = if window_buttons_at_left {
+            if cfg!(target_os = "macos") {
```
wez, line 1204: can we make this something that is set in OS parameters instead of `cfg!` logic in here?

wezterm-input-types/src/lib.rs
```diff
@@ -1275,6 +1275,7 @@ bitflags! {
     pub struct WindowDecorations: u8 {
         const TITLE = 1;
         const RESIZE = 2;
+        const FANCY = 4;
```
wez, line 1278: I wonder if we should rename `FANCY` to `INTEGRATED_BUTTONS` so that is a bit more descriptive of the function?

wezterm-input-types/src/lib.rs
```diff
@@ -1305,6 +1308,84 @@ impl Default for WindowDecorations {
...
 }
 
+#[derive(Debug, FromDynamic, ToDynamic, Clone)]
+pub struct FancyWindowDecorations {
+    #[dynamic(default)]
+    pub is_left: bool,
+    #[dynamic(default)]
+    pub remove_hide_button: bool,
+    #[dynamic(default)]
+    pub remove_maximize_button: bool,
```
wez, line 1318: What do you think about changing this config struct and breaking it apart into three things?

```lua
integrated_title_buttons = "HIDE|MAXIMIZE|CLOSE" 
```

OR: allow the user to control the order: 

```lua
integrated_title_buttons = {"Hide", "Maximize", "Close"}
```

Alignment as a separate option:

```lua
integrated_title_button_alignment = "Left" -- or "Right"
```

And a separate option for the style:

```lua
integrated_title_button_style = "Windows"
```

Rationale: these are a bit more flexible for the user, and there is a gotcha with default-initializing fields of structs in the lua config: when you omit fields, the `dynamic(default)` attribute takes the natural Default impl from the type of the field, but when you have struct-level Defaults for the containing struct, those are not used.  That can lead to discrepancies in the actual value used when the user wants to override just a single field.
Breaking things out makes it easier for the user, and makes it easier for a future online config editing UI to edit the individual values.

window/src/os/windows/window.rs
```diff
@@ -1012,12 +1014,29 @@ unsafe fn wm_nccalcsize(hwnd: HWND, _msg: UINT, wparam: WPARAM, lparam: LPARAM)
...
+) -> euclid::Point2D<isize, wezterm_input_types::PixelUnit> {
+    let mut client_rect = RECT::default();
+    GetClientRect(hwnd, &mut client_rect);
+
+    let scale = GetDpiForWindow(hwnd) as f64 / 96.0;
+
+    let mut coords = mouse_coords(0);
+
+    coords.x = client_rect.right as isize - (60.0 * scale) as isize;
+    coords.y = 5;
```
wez, line 1028: Can these hardcoded numbers be extracted from window/theme metrics? Take a look at the `get_os_parameters` function impl in this file:

https://github.com/wez/wezterm/blob/main/window/src/os/windows/window.rs#L826

wezterm-gui/src/tabbar.rs
```diff
@@ -373,6 +567,12 @@ impl TabBarState {
             line.insert_cell(x, black_cell.clone(), title_width, SEQ_ZERO);
         }
 
+        #[cfg(not(target_os = "macos"))]
```
wez, line 570: I'd love to eliminate as many of these `cfg` attributes from the bulk of the code as possible. Can we simply trust that the config is set appropriately by either the defaults or the user?

wez (commented):
Many thanks for this! It's a big PR and there are a lot of OS-specific details, and I envision that we'll go through a couple more rounds of review before we're done.

My comments on this round are stylistic and in a couple of categories:

* Avoiding hard-coding system specific values or using system-specific code in the main part of the application. Ideally the system specific stuff is moved to the OS parameters struct in the `window` crate and the rest can be reasonable defaults in the `config` crate, then the main body of the code can act based purely on the configured values.  That makes things a bit more testable and potentially allows eg: setting Windows style decorations on a mac and testing how that looks without having to go to a Windows machine.

* Breaking apart some of the config options partly with the above in mind, and partly because I think it will make it easier for users

wezterm-gui/src/tabbar.rs
YuraIz: I've found that titlebar buttons on macos behaves different than on windows and linux, for example when you hover buttons they react as single item, so I think it's better to use native macos buttons and somehow ignore decorations settings

window/src/os/windows/window.rs
YuraIz: That's just center of the maximize button to hover/click on it, it will be better if I understood how to transform screen relative coordinates to client area relative coordinates. I tried to substract client_rect's values but it looks that it has 5px shift for window resizing area.

Also I use hardcoded coorinated for minimize button, but if I correctly can handle cursor position I will be able to set static variable to inform wm_nchittest that cursor pointing onto maximize button

YuraIz: Also I need ability to correctly handle gnome styled circle buttons, because currently they behave like squares

config/src/color.rs
```diff
@@ -473,19 +533,120 @@ fn default_active_tab() -> TabBarColor {
...
+fn default_window_close_color() -> TabBarColor {
+    default_window_button_color()
+}
+
+#[cfg(not(windows))]
+fn default_window_close_hover_color() -> TabBarColor {
+    default_window_button_hover_color()
+}
+
+// Windows dependend colors for window buttons
```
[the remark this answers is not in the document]
YuraIz: I want to remove this colors from config and use black|white with transparency dependent of the tabbar color.

Also buttons looks diffrent on diffrent apps, for example the ms teams and the windows terminal (apps from microsoft) have vertically expanded buttons, but the chrome has not

About the accent color: it's used for the default titlebar when enabled, also it's used by the chrome. but it's ignored by most of windows 11 apps like the explorer, the settings and the windows terminal. so we don't need to support it. It's mostly used by winui for buttons, switches, indicators, etc. but it has changed brightness for better contrast depend on theme

Also on windows 11 app recolor the border color to any color that you want, for example to the tabbar color (like the telegram desktop)

wezterm-input-types/src/lib.rs
YuraIz: )
image

We'll be able to add other button types like that chrome's strange thing
image


but this doesn't work yet for not fancy tabbar, because it reserves place only for three buttons
image
image

YuraIz: BTW what do you think about case insentive values? Would you like to make that behavior by default for enums?
image

wezterm-input-types/src/lib.rs
YuraIz: We can name it also CSD (client-side decoration) it's shorten to write and that's how it regurarly named: https://en.wikipedia.org/wiki/Client-side_decoration

but wayland version already draws client side decoration

config/src/color.rs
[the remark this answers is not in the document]
YuraIz: Proof that you can use any color that you want)

image

YuraIz:
Now you can make some faces on windows too
image
image

But snap layouts will only work if maximize button has default style and position
image

Also I've fixed snapping with mouse
image

YuraIz:
Now snap layouts will work for any button

https://user-images.githubusercontent.com/63008755/202559729-279a5687-a9b2-4b77-a2cf-513a71c4d52c.mp4

wezterm-gui/src/customglyph.rs
```diff
@@ -222,11 +223,13 @@ impl PolyStyle {
...
             }
 
-            PolyStyle::Outline | PolyStyle::OutlineHeavy => {
+            PolyStyle::OutlineThin | PolyStyle::Outline | PolyStyle::OutlineHeavy => {
                 let mut stroke = Stroke::default();
                 stroke.width = width;
                 if self == PolyStyle::OutlineHeavy {
                     stroke.width *= 3.0; // NOTE: Using 2.0, the difference is almost invisible
+                } else if self == PolyStyle::OutlineThin {
+                    stroke.width = 1.2;
```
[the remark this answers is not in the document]
YuraIz: For some reason 1px is too thin

here's 1px outline, native icons, 1.2px outline in comparsion
Screenshot from 2022-12-26 21-05-41

YuraIz:
> What is the status for this?

Currently I have exams at my university, so my activity is freezed

I like how the windows version looks and works, but I need to finish a macos version with ability to test different styles

I have an idea to add macos only "Native" style that will be a default value for macos

zolrath:
Looks great!

Is the windows tab going to be themeable like the linux tab shown?

The latest Windows Terminal uses the terminal background color for the tab as well
image

immorBen:
I can't love this function more, may I know how can I implement this on my PC?

eugenesvk:
Thanks for the PR, is there any chance to have a config that would allow exactly 0 padding from any side (especially the top, but also left)? Basically, I'd love to have the full replica of the current setup where the bar has no padding, but the pure beauty of tabs, but with proportional fonts instead of the current monospaced ones (and maybe some extra styling)?

wez:
Terminals are built around a monospace model. There is no sane way to map it to proportional fonts.

eugenesvk:
I don't mean the whole terminal, just the tabs in the titlebar.
Currently proportional fonts is the benefit of moving from text to "gui" titlebar, but the downside is you waste space to padding (which is unfortunately a very common design flaw in many tabbed apps)

eugenesvk:
Oh, if corner-radius/paddings are configurable, then we're golden, I can set those to 0 (and remove the ×) to have the current setup with better fonts!

yuraiz:
Anyway I don't plannig to change tabbar style in that PR, it's already big

wez:
Many thanks for this!

I rebased this and merged it into main. I tidied up a couple of things, but there are still some outstanding items to follow up on:

* Move hard-coded sizes into an OS parameters struct and try to populate them from available system info, or if not, make those params configurable
* macOS native button positioning could do with some finesse. I've made it work OK in fancy tab bar mode, but retro tabs are too short for it at the default font size. Not sure of best solution to this right now.
* Docs!

I'm going to close this PR because your work so far has been integrated.
Thanks again!

pwsandoval:
This icons can they be customized?

Screenshot_1a

PR #2729, opened 2022-11-11
Enable the user to get lines from a pane via cli

micmine:
This feature exists in tmux [implementation](https://github.com/tmux/tmux/blob/master/cmd-capture-pane.c]) and i also wanted to use it in wezterm.

### DIFF
--- a/codec/src/lib.rs
+++ b/codec/src/lib.rs
@@ -466,6 +466,8 @@ pdu! {
     MovePaneToNewTab: 48,
     MovePaneToNewTabResponse: 49,
     ActivatePaneDirection: 50,
+    GetRenderebleLines: 51,
+    GetRenderableLinesResponse: 52,
 }
 
 impl Pdu {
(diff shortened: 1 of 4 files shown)


wez:
Thanks for this, but I don't think this will behave how you want; `GetLines::lines` are `StableRowIndex` values which are offsets from the first ever row of scrollback.  When more than `scrollback_lines` are received, `StableRowIndex = 0` will no longer be a resolvable line in the way that you want.

To work without surprises, I think you'll want to retrieve the `RenderableDimensions` so that you can reason about the range of lines to using either `scrollback_top` or `physical_top` as the starting index, then adding the number of lines to it. `RenderableDimensions` would also allow you to make it easy to select either the viewport  (`viewport_rows`) or full scrollback (`scrollback_rows`) with convenient flags.

There isn't currently a way to return just `RenderableDimensions` in the rpc/codec stuff, so you'd need to add that functionality!

Something else that I think would be great to consider around this allowing the user to select from:
* physical lines (what you're currently retrieving)
* logical lines

And:
* Getting just text (eg: https://wezfurlong.org/wezterm/config/lua/window/get_selection_text_for_pane.html)
* Getting the text with escapes  (eg: https://wezfurlong.org/wezterm/config/lua/window/get_selection_escapes_for_pane.html)

micmine:
Thanks for looking at my pr. I have tried to implement your suggestion. The history of the output also looks fine.

wez:
Sorry; I'd forgotten about this, and when I sat down to write up the review I figured I may was well just write the code, so I just pushed a version of `wezterm cli get-text`. I'd appreciate your feedback on it!

Issue #2667, opened 2022-10-25
Installer does not work on arm64 windows 


xeysz:
### What Operating System(s) are you seeing this problem on?

Windows

### WezTerm version

20221023-205047-43f2265e

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

Installer fails with message "This program can only be installed on versions of Windows designed for the following processor architectures: x64"

### To Reproduce

Download installer and run as administrator

### Configuration

no config

### Expected Behavior

x64 Inno Setup should work on arm64 targets under emulation

### Anything else?

Windows on arm64 will [emulate x64](https://blogs.windows.com/windows-insider/2020/12/10/introducing-x64-emulation-in-preview-for-windows-10-on-arm-pcs-to-the-windows-insider-program/) if x64 binaries are used. The zipfile (portable) version of wezterm works fine (as far as I can tell), but the installer will refuse to run. That means users can't use winget to install or update and have to manually do both.

A native arm64 build would work but that might be more effort than getting Inno Setup to run where [x64 emulation is supported](https://stackoverflow.com/questions/73707415/can-inno-setup-detect-windows11-on-arm64-hardware-which-can-emulate-x64).

I haven't tested on native windows arm64 installations (like a Surface Pro X), but I have access to one if that helps testing.

wez:
Blocked on GitHub Actions not supporting arm64: https://github.com/actions/runner-images/issues/768

wez:
I'll accept a PR that allows the inno setup stuff to run on arm64 and allow emulation to kick in, but I have no way to validate it

xeysz:
I have a patch that works for x64 emulation on arm64. I will make sure this doesn't break native x64 and that it properly works or reports errors on older versions of windows and will send a PR your way.

PR #2746, opened 2022-11-15
Allow windows installer to run on arm64 emulating x64

xeysz:
Sorry this took so long, but being new to InnoSetup I had to test different builds to make sure I understood its behavior.

The main goal is to allow people to install wezterm on windows using the installer for x64 when running on arm64 machines (virtual or physical, doesn't matter). This addresses #2667 

Windows running on arm64 supports [emulating x64 since build 21277](https://blogs.windows.com/windows-insider/2020/12/10/introducing-x64-emulation-in-preview-for-windows-10-on-arm-pcs-to-the-windows-insider-program/) which means that even if we don't have a native arm64 build yet, we can use the existing installer to achieve that.

This is the validation matrix tested:

- Column headers are build versions and each result is "expected / measured", e.g. 🟢 / 🔴  means expected result was pass (green circle) and measured result was fail (red circle). NA means it was not tested.
- I only ran x86 tests against 17763 since that's the minimum supported version by wezterm, but there was no point in testing other builds because InnoSetup checks `ArchitecturesAllowed` before it reaches the `InitializeSetup` event. And according to [wikipedia](https://en.wikipedia.org/wiki/List_of_Microsoft_Windows_versions), starting with build 22000 windows no longer supports (or even offers) 32 bit images.
- I could not find an arm64 image to test 21277 or before, but 22000 worked and my goal was to make sure this didn't break the original x64 installer.

| arch     |   17134    |   17763   |   21277  |  22000  |
|----------|:----------:|:---------:|:--------:|:-------:|
| x86      |   🔴 / NA  |  🔴 / 🔴  |  🔴 / NA | 🔴 / NA |
| x64      |   🔴 / 🔴  |  🟢 / 🟢  |  🟢 / 🟢 | 🟢 / 🟢 |
| arm64    |   🔴 / NA  |  🔴 / NA  |  🟢 / NA | 🟢 / 🟢 |

### DIFF
--- a/ci/windows-installer.iss
+++ b/ci/windows-installer.iss
@@ -10,8 +10,8 @@
 
 [Setup]
 AppId={{BCF6F0DA-5B9A-408D-8562-F680AE6E1EAF}
-ArchitecturesAllowed=x64
-ArchitecturesInstallIn64BitMode=x64
+ArchitecturesAllowed=x64 arm64
+ArchitecturesInstallIn64BitMode=x64 arm64
 AppName={#MyAppName}
 AppVersion={#MyAppVersion}
 ;AppVerName={#MyAppName} {#MyAppVersion}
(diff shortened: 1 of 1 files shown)


wez:
Thank you for this!

PR #2762, opened 2022-11-20
Add bindings for Home and End in copy_mode

cunha:
Seemed like a reasonable addition. Let me know if additional changes to the docs or other improvements are needed.


wez:
Thanks! Please add entries for these to the table in the docs: https://github.com/wez/wezterm/blob/main/docs/copymode.md

wez:
Thanks!

PR #2789, opened 2022-11-24
Add line height scale for tab bar

xiaopengli89:
The retro tab bar looks too small and hard to interact with, I added a 1.5x scale and it looks better.

1.0x:
截屏2022-11-24 15 29 44
1.5x:
截屏2022-11-24 15 28 09

### DIFF
--- a/wezterm-gui/src/termwindow/render.rs
+++ b/wezterm-gui/src/termwindow/render.rs
@@ -145,6 +145,8 @@ const PLUS_BUTTON: &[Poly] = &[
     },
 ];
 
+const TAB_BAR_LINE_HEIGHT_SCALE: f32 = 1.5;
+
 /// The data that we associate with a line; we use this to cache it shape hash
 #[derive(Debug)]
 pub struct CachedLineState {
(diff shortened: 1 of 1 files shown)


wez (changes requested):
Sorry for the long delay in looking at this; it's been a bit of a crazy end of year.
Thanks for diving in!

My knee-jerk reaction to this is: you should probably use the fancy tab bar if you want to do this.

Here's the rationale: the retro tab bar is considered to be part of the terminal grid and subject to the same font and font metrics; that's kinda the point for the retro style.

There are (unfortunately) a number of places in the rest of the code that are sensitive to the tab bar when it comes to selection and mouse coordinates that I'm not sure are covered by this PR.

Lastly, hard coding a fixed scale in the code is not the way to go: you can see from the reactions on your PR that there is at least one negative reaction so if we were to do something like this, it would need to have a new configuration option and default to the existing 1.0 scale.

Thinking ahead: I'd like to see the tab bar stuff refactored so that there is just one set of logic in the current fancy tab bar code that can also be set to a retro style, and then retire the current retro logic.

With that in mind, I don't really want to add more special cases to the retro tab bar logic until that refactoring has taken place, as it just makes things more difficult to maintain and refactor in the mean time.

PR #2825, opened 2022-12-02
Documented closing WSL Panes without confirmation

clawsl:
[related issue](https://github.com/wez/wezterm/issues/2824#issuecomment-13352971750)

### DIFF
--- a/docs/config/lua/config/skip_close_confirmation_for_processes_named.md
+++ b/docs/config/lua/config/skip_close_confirmation_for_processes_named.md
@@ -49,7 +49,7 @@ return {
 ```
 
 For closing WSL Terminals on Windows without confirmation, the (non default) processes `wslhost.exe`, `wsl.exe`,
-`conhost.exe` could be added to the list. **Please keep in mind this can be a dangerous setting. It could result in unsaved files (e.g. files open in a VIM session) inside the WSL terminal to be lost.**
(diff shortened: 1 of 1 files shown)


docs/config/lua/config/skip_close_confirmation_for_processes_named.md
```diff
@@ -50,6 +50,28 @@ return {
 
 *Since: 20220101-133340-7edc5b5a*
 
+For closing WSL Terminals on Windows without confirmation, the (non default) processes `wslhost.exe`, `wsl.exe`,
```
wez, line 53: 1. The placement of this paragraph relative to the *Since* text above is incorrect; I think you probably want this text to go above it, as the *Since* stuff applies to the `mux-is-process-stateful` event mentioned below.

2.  If you're running `vim` inside the WSL instance in one of those tabs, does your suggested configuration cause it to be killed without prompting? If so, I don't want to make that suggestion in the docs as it can result in lost work in those programs. It would need to point out that it is a "dangerous" setting to employ.

clawsl: I tested vim and it closes without warning. Therefore, I included a section to point out the potential dangers as well as moving the _Since ..._ paragraph down.

wez:
I've decided not to include this in the docs; I don't think giving potentially dangerous advice is good because folks copy and paste blindly.
I also want to avoid the docs becoming a series of overly specific workarounds to issues in individual setups, because that makes the docs longer and less generally useful.

PR #2828, opened 2022-12-02
Remove `Source` entry in Debian control file

### LINKED ISSUE #2827: [Minor] Debian/Ubuntu package contains invalid `Source` field in control file

Linux X11

xmonad

### WezTerm version

20221119-145034-49b9839f

### Describe the bug

It is very minor :)

On Debian (and I believe also Ubuntu), there is a utility called [debsecan](https://packages.debian.org/bullseye/debsecan) that examines the packages installed on a system and generates a list of vulnerabilities.
When I run it on my system with the `wezterm` package installed, I get:
```
$ debsecan
(...)
Traceback (most recent call last):
  File "/usr/bin/debsecan", line 1370, in <module>
    rate_system(target, options, fetch_data(options, config), history)
  File "/usr/bin/debsecan", line 1305, in rate_system
    raise SyntaxError(('package %s references '
SyntaxError: package wezterm references invalid source package 'https://wezfurlong.org/wezterm/'
```
Indeed, according to the Debian documentation, the `Source` field in the [control file for a binary package](https://www.debian.org/doc/debian-policy/ch-controlfields.html#binary-package-control-files-debian-control) is [supposed to contain the related source package](https://www.debian.org/doc/debian-policy/ch-controlfields.html#s-f-source), not a URL. To be noted that this field is optional for a binary package.

### To Reproduce

Install the `wezterm` and `debsecan` packages on a Debian or Ubuntu system, then run `debsecan`.

### Configuration

N/A

### Expected Behavior

`debsecan` without reporting a syntax error for `wezterm`.

### Logs

_No response_

### Anything else?

I did not try the nightly version because AFAICT this has not changed.

cgrigis:
Fixes #2827

### DIFF
--- a/ci/deploy.sh
+++ b/ci/deploy.sh
@@ -237,6 +237,7 @@ EOF
         mv pkg/debian/prerm pkg/debian/DEBIAN/prerm
         chmod 0755 pkg/debian/DEBIAN/prerm
         mv pkg/debian/control pkg/debian/DEBIAN/control
+        sed -i '/^Source:/d' pkg/debian/DEBIAN/control  # The `Source:` field needs to be valid in a binary package
         echo $deps | sed -e 's/shlibs:Depends=/Depends: /' >> pkg/debian/DEBIAN/control
         cat pkg/debian/DEBIAN/control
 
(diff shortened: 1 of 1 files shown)


wez:
Looks like Source is necessary for the package to build:

```
dpkg-shlibdeps: error: syntax error in debian/control at line 12: first block lacks a Source field
```

cgrigis:
Indeed, good point.
Trying a different approach to the fix...

cgrigis:
That failure is not related to my change, is it? :thinking:

wez:
Looks good, thanks!

PR #2836, opened 2022-12-05
Clarify 'font_rules' docs and fix example.

bldur:


docs/config/lua/config/font_rules.md
```diff
@@ -7,7 +7,7 @@ fonts that you'd like to use.  The default is to auto-generate reasonable
 italic and bold styling based on the [font](font.md) configuration.
 
 If you do specify `font_rules`, they are applied in the order that they are
-specified in the configuration file, stopping with the first matching rule:
+specified, serving all non-negative partial matches.
```
jankatins, line 10: I've no clue what this sentence means: what is "serving", what is a "partial match" :-)

docs/config/lua/config/font_rules.md
```diff
@@ -45,48 +45,53 @@ return {
...
-Here's an example from my configuration file;
+Here's an example from a configuration file;
 
 ```lua
 
 -- A helper function for my fallback fonts
 function font_with_fallback(name, params)
-  local names = { name, 'Noto Color Emoji', 'JetBrains Mono' }
+  local names = { name, 'Noto Color Emoji', 'Symbols Nerd Font Mono' }
```
jankatins, line 55: Isn't the symbols nerd font happen anyways?

docs/config/lua/config/font_rules.md
```diff
@@ -45,48 +45,53 @@ return {
...
       intensity = 'Half',
-      font = font_with_fallback 'Operator Mono SSm Lig Light',
+      font = font_with_fallback('VictorMono', { weight = "ExtraLight", style = "Normal" }),
     },
   },
+    font_shaper = "Harfbuzz",
+    -- https://github.com/tonsky/FiraCode/wiki/How-to-enable-stylistic-sets
+    -- Victor Mono ss01-07 https://github.com/rubjo/victor-mono
+    harfbuzz_features = { 'cv13', 'cv18', 'cv31', 'kern', 'liga', 'clig', 'calt' },
+    -- ss04 is for victor as cv13 is 0 for fira that change $ with ss04
```
jankatins, line 93: This and the next line are unclear to me.

docs/config/lua/config/font_rules.md
bldur: If you match italic = true, as the first font_rules, it will match italic = true, and all intensities = ' ' unless defined to allow negative match. Tat is, partial meant to mean a larger set of styles the rule is applied to. Italics = true, is a partial match for all intensities.

docs/config/lua/config/font_rules.md
bldur: https://github.com/ryanoasis/nerd-fonts/tree/498ebf372bfc84731df6fe9411c909b4e97b1ec0/patched-fonts/NerdFontsSymbolsOnly/complete

That depends, for my setup I only have this, and only want to use Mono.

docs/config/lua/config/font_rules.md
bldur: Can be removed, as it is harfbuzz_features tuning and links to the fonts and some notes about the stylistic sets.
Becomes clear if you follow the Victor Mono and Fira Code links, and anyone new and wanting to use it would do that.

docs/config/lua/config/font_rules.md
jankatins: So you want to make clearer what "stopping with the first matching rule" means in practise?

My suggestion would be something like:

```
... specified in the configuration file, stopping with the first matching rule. For that reason, it is recommended to specify rules from most specific to least specific, as e.g. a earlier rule which only specifies `italic = true` will not allow a rule with ` italic = true, intensity = "Normal"` to ever match.
```

Or do I still misunderstand your intend? I still do not understand what `defined to allow negative match` means.

bldur: That's exactly right, happy with your proposal. As clarity involves what is assumed about the data and rule structure. Assuming that it first tries to match with the most rules first is one of those, for the strength and specificity not mentioned.

docs/config/lua/config/font_rules.md
bldur: It works for me, unless I have it Mono here, the icons become "dull".

docs/config/lua/config/font_rules.md
bldur: Ideally, the notes can  be removed and harfbuzz_features accepted on a per rule basis, or merged.

bldur: Nice to have an example that works with open fonts.
Both Fira Code and Victor Mono providing the Italics are OFL-1.1 licenses.

Specificing Symbols Nerd Font Mono changes icons for me, any need to have builtin at all by default?

Last commit is completely workable font example.
https://ibb.co/NW896gp

docs/config/lua/config/font_rules.md
```diff
@@ -45,48 +48,56 @@ return {
...
 function font_with_fallback(name, params)
-  local names = { name, 'Noto Color Emoji', 'JetBrains Mono' }
+  local names = { name, 'Noto Color Emoji', 'Symbols Nerd Font Mono' }
   return wezterm.font_with_fallback(names, params)
 end
 
 return {
-  font = font_with_fallback 'Operator Mono SSm Lig Medium',
+  font = font_with_fallback('FiraCode', {weight = 'Light'}),
+  font_size = 8.0,
```
wez, line 64: Let's take `font_size` out of this page: this value is much much smaller than the default font size, and will harm readability/accessibility for a number of users. Additionally, `font_size` is an independent concept from `font_rules`, so it doesn't need to be in this example.

docs/config/lua/config/font_rules.md
```diff
@@ -45,48 +48,56 @@ return {
...
   font_rules = {
     -- Select a fancy italic font for italic text
     {
       italic = true,
-      font = font_with_fallback 'Operator Mono SSm Lig Medium Italic',
+      intensity = "Normal",
+      font = font_with_fallback('VictorMono', {
+        weight = 'Regular', italic = true }),
```
wez, line 71: I'd recommend installing the pre-requisites for building the docs, so that this can get automatically formatted.

```
cargo install mdbook mdbook-linkcheck mdbook-mermaid gelatyx
```

then run:

```
ci/build-docs.sh
```

it should auto-format the lua in these examples.

docs/config/lua/config/font_rules.md
wez: `Noto Color Emoji` , `JetBrains Mono` and `Symbols Nerd Font Mono` are always automatically appended to the list of fonts, whether you use `wezterm.font` or `wezterm.font_with_fallback`, so this example could be simplified by removing these.

This helper function is more useful when you have fonts that are not part of the automatic fallback that you wish to reference from multiple `font_rules`.

docs/config/lua/config/font_rules.md
```diff
@@ -7,7 +7,10 @@ fonts that you'd like to use.  The default is to auto-generate reasonable
 italic and bold styling based on the [font](font.md) configuration.
 
 If you do specify `font_rules`, they are applied in the order that they are
-specified in the configuration file, stopping with the first matching rule:
+specified in the configuration file, stopping with the first matching rule.
+For that reason, it is recommended to specify rules from most specific to
```
wez, line 11:
```suggestion
For that reason, you must specify rules from most specific to
```

wez (commented):
Thanks for this!

I have some comments!

docs/config/lua/config/font_rules.md
```diff
@@ -45,46 +48,72 @@ return {
...
-      font = font_with_fallback 'Operator Mono SSm Lig Light',
+      italic = false,
+      font = wezterm.font {
+        family = 'VictorMono',
+        weight = 'ExtraLight',
+        harfbuzz_features = { 'ss04' },
+      },
     },
   },
 }
```
bldur, line 119: Hope this is sufficently simple and a reasonably relevant example.

docs/config/lua/config/font_rules.md
```diff
@@ -45,46 +48,72 @@ return {
...
-end
-
 return {
-  font = font_with_fallback 'Operator Mono SSm Lig Medium',
+  font = wezterm.font_with_fallback {
+    { family = 'FiraCode', weight = 'Light' },
+    'Terminus',
+  },
+  harfbuzz_features = {
+    'cv13',
```
wez, line 61: It would be good to explain what these do in a comment, as this is the kind of thing that looks mysterious and folks will just copy because they're not sure if they need it or not, and then it will end up get copied around and used in situations where it doesn't make sense.

Alternatively, moving this into line 57 above to make it actually apply only to FiraCode would help to reinforce that it belongs with that font:

```lua
   { family = 'FiraCode', weight = 'Light',
        -- adjust presentation of FiraCode
        -- see: https://github.com/tonsky/FiraCode/wiki/How-to-enable-stylistic-sets
        harfbuzz_features= {
        'cv13',
```

bldur: Pushed changes with more comments addressing this, your suggestion actually doesn't apply to bold typeface Fira Code in that case, due to weight = 'Light'.
           Made that clear as well, and have checked that this indeed does work as expected, and in ways promotes a desire to tweak the weightings, and perhaps complete the micromanagement for other things like underline and strikethrough, but think that is going way too far with the example.

And leaving Terminus fallback with comments for coherency and also showing another way to set font.

bldur: There is another thing,   'ss04' changes zero for Victor Mono, and changes $ for Fira Code.

bldur: https://www.codingfont.com/blog/coding-font-with-italic-and-programming-ligature/

According to here, it covers two out of three open fonts with ligatures, and one of them being one of three best with true italics or cursive italics. So going over board here should be a mute point, as many may find wezterm for naming them in an example.

bldur: Added comment on ss04 and link to Victor.
If weight up slightly, looks like this.
https://ibb.co/DMnggpQ

wez: Hmm, I think that we should take the `harfbuzz_features` part out of this example.  The rationale is: it is insufficiently explained here, and too complicated to explain here without dominating what would otherwise be a pretty straightforward example of using `font_rules`.

Rather than draw questions or feeling that `font_rules` are too complex, I think dropping the stylistic sets from this initial example makes sense.

docs/config/lua/config/font_rules.md
```diff
@@ -45,46 +48,79 @@ return {
...
-  local names = { name, 'Noto Color Emoji', 'JetBrains Mono' }
-  return wezterm.font_with_fallback(names, params)
-end
-
 return {
-  font = font_with_fallback 'Operator Mono SSm Lig Medium',
+  font = wezterm.font_with_fallback {
+    { family = 'FiraCode', weight = 'Light' },
+    -- Fallback that looks nice in small font sizes
+    'Terminus',
```
wez, line 59: in a similar spirit to my comment about removing `harfbuzz_features` from this example, let's also remove `Terminus` from here, as I think it steals focus from demonstrating the `font_rules`; `Terminus` being listed here doesn't affect the the `FiraCode` as the base font vs. `VictorMono` for italics that is the main benefit of spelling out these `font_rules`.

bldur: As it currently is, it is a workable example, and that removing the stylistic sets and giving a clue that it works in small font sizes, removing Terminus that shows a different way of defining or adding fonts with wezterm.font_with_fallback, and having removed font_size from this in the past as well.

The example could very well define a few random fonts like Times New Roman and have the bare minimum without any tips and tricks, especially not a lua function to shorten the code with style and way of doing the structure that is incomplete for dealing with it as a string instead of a table.

Removed the harfbuzz_features, an entirely synthetic example mixing JetbrainsMono and VictorMono is my proposal.

wez: I appreciate the effort you've put into this so far, but it has reached the point where it is taking a lot more time and effort to work through than I'd like to spend on what should be a simple doc fix.

I'm going to make a pass over this page myself to clarify it.

bldur: > I'm going to make a pass over this page myself to clarify it.

In the meantime this is an improvement for new users, which I still am.

PR #2843, opened 2022-12-06
Removing the ls-fonts output, and making my example complete as used …

bldur:
To make the example complete instead of minimal, and my vote to remove the output from ls-fonts. #2836
Not the best of unixporn, but for the particular desires, and practicality of not confusing 0Øoø8& at small sizes.


bldur:
And aehm, it should be a nerd-snipe page, so maybe mention more the combining of cursive italic fonts with monospaced fonts for regular text.

PR #2848, opened 2022-12-07
changelog: chase username change

valpackett:
as going to `github.com/unrelentingtech` right now requires an extra click to navigate to my page :)

I've also added a [mailmap](https://git-scm.com/docs/gitmailmap) with an entry for myself, if you're ok with that


wez:
Sounds good; thanks!

Issue #2756, opened 2022-11-18
WebGPU front_end


wez:
I just pushed preliminary support for using [wgpu.rs](https://wgpu.rs/) to power an rendering front end in wezterm.

You can enable it by adding `front_end = "WebGpu"` to your configuration.

When using WebGPU, you will have access to a variety of different rendering backends, depending on your system:

* Vulkan
* Metal
* DirectX 12 (or 11)
* OpenGL
* Cpu

The default behavior is to select the best supported low-power (integrated) GPU available to your system.

You can explicitly select a discrete GPU if you wish by setting `webgpu_power_preference = "HighPerformance"`.

You can explicit request a fallback/software based backend by setting `webgpu_force_fallback_adapater = true`.

You can list the available GPUs using the debug overlay; call the `wezterm.gui.enumerate_gpus()` function:

![image](https://user-images.githubusercontent.com/117777/202768357-fec5a586-7250-4428-98e0-dce76ad5e2de.png)

You can then force the use of one of those if you wish:

```lua
return {
   webgpu_preferred_adapter={
      backend="Vulkan",
      device_type="Cpu",
      name="llvmpipe (LLVM 14.0.0, 256 bits)"
   }
}
```

This is a new backend and likely has some issues that are waiting to be discovered: I'd love to hear about them!

luisiacc:
using the Dx12 on nvidia makes the font looks weird:

with it:
<img width="419" alt="image" src="https://user-images.githubusercontent.com/31720261/202788417-f9e5e36e-a8cd-4341-87b2-1689ce9fdb23.png">


without it:
<img width="479" alt="image" src="https://user-images.githubusercontent.com/31720261/202788604-5d7ff4f7-21bf-4e5d-ba5c-bd87a835a27d.png">

wez:
@luisiacc: I'm seeing the same thing here on linux with nvidia/vulkan.  It is reminiscent of this earlier issue:
* https://github.com/wez/wezterm/issues/1025

I'm continuing to dig into this.

Suyashtnt:
![image](https://user-images.githubusercontent.com/45307955/202843113-265e1a9b-1df1-48e0-8ae1-7ae6e06d6e6b.png)
It seems to just not want to create the adapter at all.

Config
```lua
local wezterm = require "wezterm"

return {
	color_scheme = "Catppuccin Mocha",
	font = wezterm.font_with_fallback {
[...]

PR #2887, opened 2022-12-24
wayland: set scale factor without wegl_surface (for WebGpu)

valpackett:
This fixes wrong scale with the WebGpu backend. Related: #2756

### DIFF
--- a/window/src/os/wayland/window.rs
+++ b/window/src/os/wayland/window.rs
@@ -682,36 +682,36 @@ impl WaylandWindowInner {
                         // assume no.
                         live_resizing: false,
                     });
+                    // Avoid blurring by matching the scaling factor of the
+                    // compositor; if it is going to double the size then
+                    // we render at double the size anyway and tell it that
+                    // the buffer is already doubled.
+                    // Take care to detach the current buffer (managed by EGL),
+                    // so that the compositor doesn't get annoyed by it not
+                    // having dimensions that match the scale.
+                    // The wegl_surface.resize won't take effect until
+                    // we paint later on.
+                    // We do this only if the scale has actually changed,
+                    // otherwise interactive window resize will keep removing
+                    // the window contents!
                     if let Some(wegl_surface) = self.wegl_surface.as_mut() {
                         wegl_surface.resize(pixel_width, pixel_height, 0, 0);
-                        // Avoid blurring by matching the scaling factor of the
-                        // compositor; if it is going to double the size then
-                        // we render at double the size anyway and tell it that
-                        // the buffer is already doubled.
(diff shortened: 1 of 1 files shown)


window/src/os/wayland/window.rs
```diff
@@ -713,6 +713,9 @@ impl WaylandWindowInner {
                                 self.surface_factor = factor;
                             }
                         }
+                    } else if self.surface_factor != factor {
+                        // no wegl_surface for WebGpu, still have to set this
+                        self.surface.set_buffer_scale(factor);
```
wez, line 718: I think the same buffer replacement stuff that is done above is likely also needed here, as well as setting `self.surface_factor = factor` for consistency.

Perhaps the logic should be something like this:

```rust
                        if let Some(wegl_surface) = self.wegl_surface.as_mut() {
                             wegl_surface.resize(pixel_width, pixel_height, 0, 0);
                        }
                        
                        if self.surface_factor != factor {
                            let wayland_conn = Connection::get().unwrap().wayland();
                            let mut pool = wayland_conn.mem_pool.borrow_mut();
                            // Make a "fake" buffer with the right dimensions, as
                            // simply detaching the buffer can cause wlroots-derived
                            // compositors consider the window to be unconfigured.
                            if let Ok((_bytes, buffer)) = pool.buffer(
                                factor,
                                factor,
                                factor * 4,
                                wayland_client::protocol::wl_shm::Format::Argb8888,
                            ) {
                                self.surface.attach(Some(&buffer), 0, 0);
                                self.surface.set_buffer_scale(factor);
                                self.surface_factor = factor;
                            }
                        }
```

can you try that out both with and without webgpu?

valpackett: Makes sense. Though I just started testing on wlroots hidpi and actually nothing works, everything I've tried results in wezterm getting disconnected due to an `error in client communication`. :/

wez: This fell off my radar over the holidays, just catching back up now.
Is this PR good to merge? It wasn't clear whether you had issues related to this or not

valpackett: I think it's an improvement either way as it works on gnome, but I haven't managed to fix the error on wlroots :/

wez: Ok, thanks!
For debugging your wlroots issue, have you tried running with WAYLAND_DEBUG=1 set in the environment to see if that reveals anything?

valpackett: Of course! that only explains *when* something went wrong, not what. IIRC wlroots didn't log anything, so the disconnect was entirely unexplained. I'll need to come back to this later if no one else does..

PR #2899, opened 2022-12-27
Fix typo

Omochice:
Thank you for great product.

I found this typo when write my setting.

I hope it is not my fault...


wez:
Looks good, thanks!

PR #2911, opened 2023-01-02
Improve Copy Mode key documentation

proski:
Add the key to enter Copy Mode to the table. It may be repetitive, but it makes the table a useful cheatsheet.

Document `y`, it was undocumented.

Capitalize all keys with modifiers for consistency with other documentation and within the table.


wez:
Thanks for this!

Regarding capitalizing keys with modifiers, I'm a bit hesitant because of:
* https://github.com/wez/wezterm/issues/1906

which means that, when *defining* key assignments, the case of the key and the corresponding modifier keys is important and should not always be uppercased.  In particular, the `ALT-f` and `ALT-b` keys in copy mode are explicitly for lowercase keys, and I think it might be confusing to show them as `ALT-F` which implies an uppercase `f` key.

I'm not opposed to coming up with a convention for the docs that could be documented somewhere.  Something like:

> When referencing a key combination, these docs will always use the complete list of modifier keys followed by the key cap from a US keyboard layout shown as an upper case letter for visual clarity.  The key is shown in uppercase but unless `SHIFT` is listed, then the key is assumed to be the unshifted or lowercase form.

I think this would be less confusing (and perhaps not need a paragraph of explanation) if we could use the [`<kbd>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/kbd) html element to clearly indicate when we're talking about key caps rather than configuration.

Would you mind taking a look at adopting the `kbd` element for this?

proski:
I converted the key combinations to use `kbd` tags. Unfortunately, it makes the markdown file less readable as a plain text, but it's probably not a big deal, considering that the documentation is available online.

There is a long standing tradition to write `Ctrl-C` rather than `Ctrl-c`. Also, `CTRL-C` is uncommon. So the conversion to `kbd` tags is an opportunity to ditch some local conventions that don't help with readability. I used mixed-case modifiers and `+` for key combinations.

The rendered version is available at https://github.com/proski/wezterm/blob/document-y/docs/copymode.md

PR #2928, opened 2023-01-08
x11: implement active screen detection

NBonaparte:
I noticed this was unimplemented so I took a shot at it.

### DIFF
--- a/window/src/os/x11/connection.rs
+++ b/window/src/os/x11/connection.rs
@@ -5,7 +5,7 @@ use crate::os::x11::xsettings::*;
 use crate::os::Connection;
 use crate::screen::{ScreenInfo, Screens};
 use crate::spawn::*;
-use crate::{Appearance, DeadKeyStatus, ScreenRect, ScreenPoint};
+use crate::{Appearance, DeadKeyStatus, ScreenPoint, ScreenRect};
 use anyhow::{anyhow, bail, Context as _};
 use mio::event::Source;
 use mio::unix::SourceFd;
@@ -239,7 +239,9 @@ impl ConnectionOps for XConnection {
             .ok_or_else(|| anyhow::anyhow!("no screens were found"))?
             .clone();
 
-        let active = self.screen_from_focused_window(&by_name).unwrap_or_else(|_| main.clone());
+        let active = self
+            .screen_from_focused_window(&by_name)
+            .unwrap_or_else(|_| main.clone());
 
         Ok(Screens {
             main,
@@ -795,16 +797,22 @@ impl XConnection {
         future
     }
 
-    fn screen_from_focused_window(&self, by_name: &HashMap<String,ScreenInfo>) -> anyhow::Result<ScreenInfo> {
+    fn screen_from_focused_window(
+        &self,
+        by_name: &HashMap<String, ScreenInfo>,
+    ) -> anyhow::Result<ScreenInfo> {
         let focused = self
             .send_and_wait_request(&xcb::x::GetInputFocus {})
             .context("querying focused window")?;
         let geom = self
-            .send_and_wait_request(&xcb::x::GetGeometry {
-                drawable: xcb::x::Drawable::Window(focused.focus()),
+            .send_and_wait_request(&xcb::x::TranslateCoordinates {
+                src_window: focused.focus(),
+                dst_window: self.root,
+                src_x: 0,
+                src_y: 0,
             })
-            .context("querying geometry")?;
-        let window_origin = ScreenPoint::new(geom.x().into(), geom.y().into());
+            .context("querying root coordinates")?;
+        let window_origin = ScreenPoint::new(geom.dst_x().into(), geom.dst_y().into());
         Ok(by_name
             .values()
             .find(|screen| screen.rect.contains(window_origin))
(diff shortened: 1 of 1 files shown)


window/src/os/x11/connection.rs
```diff
@@ -239,12 +239,22 @@ impl ConnectionOps for XConnection {
             .ok_or_else(|| anyhow::anyhow!("no screens were found"))?
             .clone();
 
-        // We don't yet know how to determine the active screen,
-        // so assume the main screen.
-        // TODO: find focused window and resolve it!
-        // Maybe something like <https://stackoverflow.com/a/43666928/149111>
-        // but ported to Rust?
-        let active = main.clone();
+        self.conn.send_request(&xcb::x::GrabServer {});
```
wez, line 242: I was going to suggest that we need to ensure that we ungrab the server in all of the error paths below (each `?` is potentially an early return!), to avoid locking things up (see: https://tronche.com/gui/x/xlib/window-and-session-manager/XGrabServer.html), but I don't think we need to grab it at all here.

In the stackoverflow link someone suggested grabbing the server while walking the tree of windows to prevent the focus changing out from under the client, but I don't think that matters here: we're only looking at a single window, so grabbing just means the user's focus change happens a fraction of a second later.

So let's remove the grab/ungrab logic to streamline and simplify a little.

```suggestion
```

window/src/os/x11/connection.rs
```diff
@@ -239,12 +239,22 @@ impl ConnectionOps for XConnection {
...
+                drawable: xcb::x::Drawable::Window(focused.focus()),
+            })
+            .context("querying geometry")?;
+        let window_origin = ScreenPoint::new(geom.x().into(), geom.y().into());
+        let active = by_name
+            .values()
+            .find(|screen| screen.rect.contains(window_origin))
+            .ok_or_else(|| anyhow::anyhow!("active window is not in any screen"))?
+            .clone();
+        self.conn.send_request(&xcb::x::UngrabServer {});
```
wez, line 257:
```suggestion
```

window/src/os/x11/connection.rs
```diff
@@ -239,12 +239,22 @@ impl ConnectionOps for XConnection {
...
+            .send_and_wait_request(&xcb::x::GetGeometry {
+                drawable: xcb::x::Drawable::Window(focused.focus()),
+            })
+            .context("querying geometry")?;
+        let window_origin = ScreenPoint::new(geom.x().into(), geom.y().into());
+        let active = by_name
+            .values()
+            .find(|screen| screen.rect.contains(window_origin))
+            .ok_or_else(|| anyhow::anyhow!("active window is not in any screen"))?
+            .clone();
```
wez, line 256: I think it's technically possible that no windows have focus, so perhaps this bit should look something like:

```rust
let active = self.screen_from_focused_window(&by_name).unwrap_or_else(|| main.clone());
```

with the heavy lifting moved into a helper method:

```rust
fn screen_from_focused_window(&self, by_name: &HashMap<String,ScreenInfo>) -> anyhow::Result<ScreenInfo> {
       let focused = self
            .send_and_wait_request(&xcb::x::GetInputFocus {})
            .context("querying focused window")?;
        let geom = self
            .send_and_wait_request(&xcb::x::GetGeometry {
                drawable: xcb::x::Drawable::Window(focused.focus()),
            })
            .context("querying geometry")?;
        let window_origin = ScreenPoint::new(geom.x().into(), geom.y().into());
        Ok(by_name
            .values()
            .find(|screen| screen.rect.contains(window_origin))
            .ok_or_else(|| anyhow::anyhow!("active window is not in any screen"))?
            .clone())
}
```

wez (changes requested):
Thanks for taking this on! I have a couple of suggestions; could you try them out?

NBonaparte:
I've tested this on bspwm, which works as intended, and on i3, which seems to have an odd geometry implementation. It returns `x: 0, y: 0` for every window (resulting in the left-most screen being selected) and x: -1, y: -1 if no window is selected.

wez:
regarding i3: if you run `xwininfo` and/or `xprop` on windows that are not actually at `0,0`, are there any other properties that look like they have the correct geometry/positioning info?

I think i3 might be doing something a bit weird, but it's worth double checking to see if we can be more robust and avoid having people report this as an issue over and over again :)

NBonaparte:
Good point. I found `xwininfo` had the correct coordinates by using `xcb_translate_coordinates` ([the answer to this stackoverflow question also corroborates this](https://stackoverflow.com/questions/64319785/x11-xcb-window-information-not-up-to-date)), so I've switched to it instead.

wez:
Oh, interesting! I wonder if this bit of code should also be switched over:
https://github.com/wez/wezterm/blob/main/window/src/os/x11/window.rs#L303-L308

NBonaparte:
I also just realized that using the top left corner is not a great idea, since the window could be floating over the edge of the screen. 
If we have a setup like this, with the window floating over the top of two monitors, none of the corners or the center would give a good indication.
```
                 ┌──────────┐
                 │          │
┌────────────────┼──────────┼──────────────────┐
│                └──────┼───┘                  │
│                       │                      │
│                       │                      │
│                       │                      │
│                       │                      │
│                       │                      │
└───────────────────────┴──────────────────────┘
```
Computing the area of the intersection of the window with each monitor might give more accurate results.

wez:
Yeah, going by area sounds good. I feel like I wrote some code that did that once before, but I can't see it in the repo now, so I'm not sure if I dreamed that up, or whether it got refactored away at some point.

The `Rect` type should let you do something like `rect1.intersection(rect2).area()` and you should be able to collect and `sort_by` that to find the biggest intersection.

wez:
Looks good, thanks!

PR #2959, opened 2023-01-15
x11: ignore pointer focus events

NBonaparte:
X11 generates `Pointer` focus events, which are irrelevant to our use case and cause spurious `FocusIn/Out` events if the pointer is located in the wezterm window. The only relevant details seem to be `Ancestor` and `Nonlinear`. I've tested this with bspwm and i3.

Protocol doc: https://www.x.org/releases/current/doc/xproto/x11protocol.html#events:FocusIn

### DIFF
--- a/window/src/os/x11/window.rs
+++ b/window/src/os/x11/window.rs
@@ -585,12 +585,18 @@ impl XWindowInner {
                     self.verify_focus = true;
                 }
             }
-            Event::X(xcb::x::Event::FocusIn(_)) => {
-                self.focus_changed(true);
-            }
-            Event::X(xcb::x::Event::FocusOut(_)) => {
-                self.focus_changed(false);
-            }
+            Event::X(xcb::x::Event::FocusIn(e)) => match e.detail() {
+                xcb::x::NotifyDetail::Pointer => (),
+                _ => {
+                    self.focus_changed(true);
+                }
+            },
+            Event::X(xcb::x::Event::FocusOut(e)) => match e.detail() {
+                xcb::x::NotifyDetail::Pointer => (),
+                _ => {
+                    self.focus_changed(false);
+                }
+            },
             Event::X(xcb::x::Event::LeaveNotify(_)) => {
                 self.events.dispatch(WindowEvent::MouseLeave);
             }


wez:
I'm not sure about this: what issue are you trying to solve here?
The `focus_changed` method should only propagate a change if the focus state actually changes.

The fear I have is that by ignoring some of the events we'll miss an actual legitimate focus event.
I tried to follow the different cases in the protocol doc but it's a lot of state to track and there's too much text to make it easy to grasp.

NBonaparte:
Here's a log showing the pointer issue; hopefully this is clearer. My mouse is hovering over the wezterm window, and I am switching focus to a window on a different monitor, and then to another window on the 2nd monitor (with the mouse still hovering over the wezterm window):
```
(startup)      00:54:16.881  INFO   window::os::x11::window > FocusIn: FocusInEvent { response_type: 9, detail: Nonlinear, sequence: 178, event: Window { res_id: 33554435 }, mode: Normal, pad: 3 }
(2nd window)   00:54:26.920  INFO   window::os::x11::window > FocusOut: FocusInEvent { response_type: 10, detail: Ancestor, sequence: 402, event: Window { res_id: 33554435 }, mode: Grab, pad: 3 }
               00:54:26.958  INFO   window::os::x11::window > FocusOut: FocusInEvent { response_type: 10, detail: Ancestor, sequence: 404, event: Window { res_id: 33554435 }, mode: WhileGrabbed, pad: 3 }
               00:54:26.971  INFO   window::os::x11::window > FocusOut: FocusInEvent { response_type: 10, detail: Pointer, sequence: 404, event: Window { res_id: 33554435 }, mode: WhileGrabbed, pad: 3 }
               00:54:27.079  INFO   window::os::x11::window > FocusOut: FocusInEvent { response_type: 10, detail: Pointer, sequence: 404, event: Window { res_id: 33554435 }, mode: Ungrab, pad: 3 }
(3rd window)   00:54:28.931  INFO   window::os::x11::window > FocusIn: FocusInEvent { response_type: 9, detail: Pointer, sequence: 404, event: Window { res_id: 33554435 }, mode: Grab, pad: 3 }
               00:54:29.010  INFO   window::os::x11::window > FocusIn: FocusInEvent { response_type: 9, detail: Pointer, sequence: 413, event: Window { res_id: 33554435 }, mode: WhileGrabbed, pad: 3 }
               00:54:29.010  INFO   window::os::x11::window > FocusOut: FocusInEvent { response_type: 10, detail: Pointer, sequence: 413, event: Window { res_id: 33554435 }, mode: WhileGrabbed, pad: 3 }
               00:54:29.139  INFO   window::os::x11::window > FocusOut: FocusInEvent { response_type: 10, detail: Pointer, sequence: 432, event: Window { res_id: 33554435 }, mode: Ungrab, pad: 3 }
```
The wezterm window isn't actually being focused in the last few events, but it still receives `FocusIn` because the mouse is hovering over it.
In each of the cases in the doc, the pointer window P seems to be neither A nor B, so it should not matter if we ignore it since we will receive an actual (non-`Pointer`, seems to be `Ancestor`/`Nonlinear`) event when actually focusing/unfocusing it.

If you're unsure about ignoring the other details (`Inferior`, `Virtual`, etc.) we can just ignore the `Pointer` ones.

wez:
I think having the code exclude Pointer more closely matches the intent and is easier to understand.

I'm still a bit worried about edge cases.  Here's an earlier issue around focus:
* https://github.com/wez/wezterm/issues/1437

I can't find it now, but I think there was also an issue around dispatching wheel events when we hadn't sent a focus-in escape sequence to applications that were listening to mouse and focus reporting.  I'm worried that this might break that.

For example, if I have window A focused, then move my mouse over window B without focusing it, but then use the scroll wheel; does it successfully scroll the vim running in B with `set mouse=a` enabled? Likewise for `kak` and `nvim`, without generating bells or other weird artifacts?

NBonaparte:
AFAIK, all events are passed through on X11, or rather, all focus and mouse events are independent of each other.

Mouse events should be sent regardless of the window being focused. I've checked all button types (1-5) and nvim scrolling works as well.

wez:
Thanks!

PR #2969, opened 2023-01-16
docs: use new gelatyx `file-list` option

azzamsa:
Is this needs to be updated too?

- https://github.com/wez/wezterm/blob/a2e027dff4c8c6f640f604eda2b70050dc3f2898/ci/update-derived-files.sh#L12

### DIFF
--- a/ci/build-docs.sh
+++ b/ci/build-docs.sh
@@ -1,9 +1,11 @@
 #!/bin/bash
 
-tracked_markdown=$(git ls-tree -r HEAD --name-only docs | egrep '\.(markdown|md)$')
+tracked_markdown=$(mktemp)
+trap "rm ${tracked_markdown}" "EXIT"
+git ls-tree -r HEAD --name-only docs | egrep '\.(markdown|md)$' > $tracked_markdown
 
-gelatyx lua --file $tracked_markdown --language-config ci/stylua.toml
-gelatyx lua --file $tracked_markdown --language-config ci/stylua.toml --check || exit 1
+gelatyx --language lua --file-list $tracked_markdown --language-config ci/stylua.toml
+gelatyx --language lua --file-list $tracked_markdown --language-config ci/stylua.toml --check || exit 1
 
 set -x
 
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -53,7 +53,7 @@ jobs:
           (test -x ~/.cargo/bin/mdbook-mermaid || (cd && cargo install mdbook-mermaid --locked))
       - name: Install gelatyx
         run: |
-          (test -x ~/.cargo/bin/gelatyx || (cd && cargo install gelatyx --locked))
+          (test -x ~/.cargo/bin/gelatyx || (cd && cargo install gelatyx --version "^0.2" --locked))
       - name: Build
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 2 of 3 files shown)


ci/build-docs.sh
```diff
@@ -1,9 +1,9 @@
 #!/bin/bash
 
-tracked_markdown=$(git ls-tree -r HEAD --name-only docs | egrep '\.(markdown|md)$')
+git ls-tree -r HEAD --name-only docs | egrep '\.(markdown|md)$' > tracked_markdown
 
-gelatyx lua --file $tracked_markdown --language-config ci/stylua.toml
-gelatyx lua --file $tracked_markdown --language-config ci/stylua.toml --check || exit 1
+gelatyx --language lua --file-list tracked_markdown --language-config ci/stylua.toml
+gelatyx --language lua --file-list tracked_markdown --language-config ci/stylua.toml --check
```
wez, line 6: I think the exit is still needed here, as the script isn't running with `-e`:

```suggestion
gelatyx --language lua --file-list tracked_markdown --language-config ci/stylua.toml --check || exit 1
```

ci/build-docs.sh
```diff
@@ -1,9 +1,9 @@
 #!/bin/bash
 
-tracked_markdown=$(git ls-tree -r HEAD --name-only docs | egrep '\.(markdown|md)$')
+git ls-tree -r HEAD --name-only docs | egrep '\.(markdown|md)$' > tracked_markdown
```
wez, line 3: Let's use something like:

```bash
tracked_markdown=$(mktemp)
trap "EXIT" "rm ${tracked_markdown}"
```

and then use `$tracked_markdown` for the filename, so that this list is out of the way of the repo

wez (commented):
Thanks!

I think the github workflows also need to be updated to reference the new version:

I think the cache key needs to be bumped:
https://github.com/wez/wezterm/blob/main/.github/workflows/verify-pages.yml#L30
to build the latest version-it's probably wise to explicitly list the version there.

For https://github.com/wez/wezterm/blob/main/.github/workflows/pages.yml#L44, it looks like the cache key changes when the workflow file changes, so I think changing that file to explicitly set the version of gelatyx would do the job.

wez:
> * # gelatyx doesn't understand the file include mechanism

Oh, yeah! If you could teach it handle the includes, that would be great!

azzamsa:
- https://github.com/wez/wezterm/actions/runs/3987672445/jobs/6837806149#step:6:1047

Is this related to my changes?

wez:
No, it's unrelated. If you rebase, hopefully 2b1f152e90ad646955724df6ba119eb98da5e0ba will allow it to run correctly

azzamsa:
Works like a charm.

ci/build-docs.sh
```diff
@@ -1,9 +1,11 @@
 #!/bin/bash
 
-tracked_markdown=$(git ls-tree -r HEAD --name-only docs | egrep '\.(markdown|md)$')
+tracked_markdown=$(mktemp)
+git ls-tree -r HEAD --name-only docs | egrep '\.(markdown|md)$' > $tracked_markdown
+trap "EXIT" "rm ${tracked_markdown}"
```
wez, line 5: `trap` works like: `trap "thing to evaluate" list of signals`

```suggestion
trap "rm ${tracked_markdown}" "EXIT"
git ls-tree -r HEAD --name-only docs | egrep '\.(markdown|md)$' > $tracked_markdown
```

wez:
Looks great, many thanks!

Issue #618, opened 2021-04-02
Hide mouse cursor when typing


gwww:
Hide the mouse cursor if no movement and type a character in wezterm.

gwww:
I have tested and works as expected. Thank you!

SuperSandro2000:
Can this be turned off? I always loose my cursor when it is hidden.

gwww:
I'm finding this functionality buggy. I have not been able to pinpoint it but what happens is that the cursor turns off and will not turn back on until I click outside of wezterm window. Moving the mouse, inside or outside, of wezterm does not turn it on. Clicking inside of wezterm does not turn it on. Only clicking outside of wezterm will get it to turn back on.

This happens multiple times per day but I have found "the" sequence to make it always happen.

wez:
@gwww are you on macOS?

gwww:
Yes, macOS, 11.1

SuperSandro2000:
I also noticed that when selecting text through tmux the cursor disappears and reappears every time I hold shift and move my cursor.

gwww:
I'm using the nightly with this fix and it is working properly so far.

ram02z:
Has this been implemented on wayland?

wez:
@ram02z not yet. Wayland support is difficult because it is incompatible with my nvidia hardware.  If you're in a position to have a go at implementing this feature on Wayland, I'd be happy to take a PR for it!

PR #2977, opened 2023-01-18
wayland: hide mouse cursor when typing

jmbaur:
Fixes wayland not hiding the cursor when typing from issue #618.

### DIFF
--- a/window/src/os/wayland/pointer.rs
+++ b/window/src/os/wayland/pointer.rs
@@ -340,11 +340,16 @@ impl PointerDispatcher {
             .insert(surface.as_ref().id(), Arc::clone(pending));
     }
 
-    pub fn set_cursor(&self, name: &str, serial: Option<u32>) {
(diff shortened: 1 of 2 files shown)


wez:
Looks good, thank you!

hgaiser:
I seem to be getting issues with my cursor hiding and not reappearing. It might be related to neovim (or the way neovim captures mouse events), as it only seems to happen there (but not always).

Steps to reproduce:

1. Open a file in neovim with `set mouse=a`.
1. Click with the mouse inside the neovim window.
1. Press Shift or some other key, the cursor should be hidden.
1. Move the cursor, it doesn't reappear until it leaves the window.

I will test it in X11 to see if the same issue occurs there.

sandikata:
This issue is still there, even with latest commits.

wez:
This is a pull request, not a bug.
If you have an issue, file an issue: https://github.com/wez/wezterm/issues/new?assignees=&labels=bug&template=bug.yml

PR #3006, opened 2023-01-25
Add lua methods to access position and pane in the direction

alfa07:
These methods allow to implement in Lua moving focus to adjacent windows (aka ActivateWindowDirection). Not sure if this is the right implementation - so I will open an issue to discuss and will use this PR as an illustration.


wez (changes requested):
Thanks for diving in!

Please split this into two PRs; one for the pane direction stuff which I think looks mostly fine, and the other for the window positioning stuff which will need some more discussion and fleshing out to keep the build working on all platforms!

alfa07:
I have split the `get_pane_direction` into #3011 

This one I guess requires implementation for window systems other than MacOS. Any suggestions how to test/proceed?

wez:
For Windows, you can use [GetWindowRect](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getwindowrect) to get the position.  There are a couple of places that call that already that you can probably copypasta.

For X11, there is a GetGeometry request that yields a [GetGeometryReply](https://docs.rs/xcb/latest/xcb/x/struct.GetGeometryReply.html). You can find an example of that in the x11/window.rs file. Note that on some window managers it might be necessary to translate the result to get the correct coordinates.  There's an example of that at the bottom of the x11/connection.rs file. I'd suggest skipping that complexity to start and we can see if we need to add it in later.

For Wayland I'd suggest always returning `0,0`.

If you don't have linux and windows machines available, I'm fine with you iterating by pushing to this PR to see if things compile; I think that complexity of these methods is low enough that that will probably be fine.

window/src/lib.rs
```diff
@@ -270,6 +270,14 @@ pub trait WindowOps {
     /// click before forwarding the event (Windows).
     fn set_window_drag_position(&self, _coords: ScreenPoint) {}
 
+    /// Returns the location of the window on the screen.
+    /// The coordinates are of the top left pixel of the
+    /// client area.
+    ///
+    /// This is only implemented on backends that allow
+    /// windows to move themselves (not Wayland).
+    fn get_window_position(&self) -> Future<ScreenPoint>;
```
wez, line 279: I wonder, should this be `get_window_geometry` instead, and return a `ScreenRect`; since the implementation on all platforms will have the window rectangle, it feels like it might simplify some cases for callers in the future to return it all, especially because right now the window size is tracked in TermWindow based on WindowEvents.

wez: That way, the wayland implementation could return the dimensions but leave the position at 0,0 and this method would be half-usable.  Not super convinced that that is the best idea though.

alfa07: I guess the current API choice favors `get_position` as we have `set_position` and `get_dimensions` calls

alfa07:
Looks like freebsd-12 failure in CI is unrelated to the commit

alfa07:
I have tested that round-tripping coordinates through `get_position`, `set_position` works and does not change position of the window (on macos, ubuntu-x11, and windows):
```
> wezterm.mux.all_windows()[1]:gui_window():get_position()
{
    "x": 722,
    "y": 259,
}
> wezterm.mux.all_windows()[1]:gui_window():set_position(722, 259)
> wezterm.mux.all_windows()[1]:gui_window():get_position()
{
    "x": 722,
    "y": 259,
}
```

wez:
The code looks good!

Following up on my earlier comment, can you try your round tripping test on X11 running i3?
It may be "doing something" to the reported coordinates (See discussion in https://github.com/wez/wezterm/pull/2928#issuecomment-1376801633).
If it doesn't return what we expect, 5d5efa04fc9df991674aaa0839f10f01e38e1c6e shows how to translate the coordinates.

Then all that's left is adding a doc page for this new method!

alfa07:
@wez given I test it on i3 is this ready to merge (with all the recent wayland changes)?

Lockszmith-GH:
Was this PR abandoned? Is there another solution for `get_position` ?

Issue #3007, opened 2023-01-25
Implement ActivatePaneOrWindowDirection


alfa07:
**Is your feature request related to a problem? Please describe.**
Similar to `ActivatePaneDirection` it would be awesome to have `ActivatePaneOrWindowDirection` - it is very handy when using multiple monitors and could be a killer feature in these situations.

**Describe the solution you'd like**
One option is to implement a bare minimum in wezterm and allow users to take it from there. See #3006 and lua config attached below.
The bare minimum:
1. `gui_window:get_position()` - allows finding window relative to the current
2. `tab:get_pane_direction(direction)` - gets pane in the given direction in the current tab or returns nil if there is none

**Describe alternatives you've considered**
I've tried to implement it in lua without changes to wezterm but it turns out we are not exposing enough API to do it.

**Additional context**
Here is screen [recording](https://a.cl.ly/p9uL2NRb) of me using `Ctrl-hjkl` to navigate between windows.

And here is config:
```lua
local wezterm = require 'wezterm'
local act = wezterm.action
local os = require("os")

local function active_tab(window)
  for _, t in pairs(window:tabs_with_info()) do
    if t.is_active then
      return t.tab
    end
  end
end

local function find_window_direction(window, direction)
  local wx = {}
  local p0 = window:get_position()
  local sz0 = window:get_dimensions()
  for _, w in pairs(wezterm.mux.all_windows()) do
    local x = 0
    local x0 = 0
    local p = w:gui_window():get_position()
    local sz = w:gui_window():get_dimensions()
    if direction == 'Left' then
      x = p.x + sz.pixel_width
      x0 = p0.x
    elseif direction == 'Right' then
      x = p.x
      x0 = p0.x + sz0.pixel_width
    elseif direction == 'Up' then
      x = p.y + sz.pixel_height
      x0 = p0.y
    elseif direction == 'Down' then
      x = p.y
      x0 = p0.y + sz0.pixel_height
    end
    if (direction == 'Left' or direction == 'Up') and x < x0 then
      wx[#wx + 1] = { x0 - x, w }
    elseif (direction == 'Right' or direction == 'Down') and x > x0 then
      wx[#wx + 1] = { x - x0, w }
    end
  end
  table.sort(wx, function(a, b) return a[1] < b[1] end)
  if #wx > 0 then
    return wx[1][2]
  end
  return nil
end

-- Nvim navigator implementation
[...]

PR #3011, opened 2023-01-26
Add get_pane_direction method 

alfa07:
It is useful for implementing custom navigation in lua. For example jumping between windows.

See also #3006 the PR it is split from and issue #3007 describing idea in more detail.

### DIFF
--- a/docs/config/lua/pane/get_pane_direction.md
+++ b/docs/config/lua/pane/get_pane_direction.md
@@ -1,0 +1,18 @@
+# `pane:get_pane_direction(direction)`
+
+*Since: nightly builds only*
+
+Returns pane adjacent to the given *pane* in the direction *direction*.
+
+Valid values for *direction* are:
+
+* `"Left"`
+* `"Right"`
+* `"Up"`
+* `"Down"`
+* `"Prev"`
+* `"Next"`
+
+See [ActivatePaneDirection](../keyassignment/ActivatePaneDirection.md) for more information
+about how panes are selected given direction.
+
--- a/lua-api-crates/mux/src/tab.rs
+++ b/lua-api-crates/mux/src/tab.rs
@@ -1,3 +1,5 @@
+use config::keyassignment::PaneDirection;
+
 use super::*;
 use luahelper::to_lua;
 use std::sync::Arc;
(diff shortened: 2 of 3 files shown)


wez:
The CI failures were fixed by 36519f0d90e1875fb4b3f11f6cbf94c7d716ef78 ; please rebase!

alfa07:
@wez looks like CI is broken?

jbeich:
`wayland-sys` failed due to missing dependencies: `pkgconf` + `wayland`. Caused by the following:
> env PATH=$HOME/.cargo/bin:$PATH bash get-deps
> Updating FreeBSD repository catalogue...
> FreeBSD repository is up to date.
> All repositories are up to date.
> pkg: No packages available to install matching 'mesa-devel' have been found in the repositories

`mesa-devel` [failed to build](https://pkg-status.freebsd.org/beefy16/data/131amd64-default/8881966b49a8/logs/mesa-devel-23.0.b.740.log) after GitHub hiccup (checksum mismatch as gzip compression ratio changed). Also, `mesa-devel` shouldn't be used by CI as it doesn't provide any headers in order to co-exist with `mesa-dri`.

alfa07:
hmm, does not look like macos failures are related to the PR either

jbeich:
Forgot to mention, `mesa-devel` will reapper after a new package build. This is completely automatic as builds start on Tue/Thu/Sat/Sun at 01:00 UTC but maybe skipped if the previous one didn't finish yet. FreeBSD always builds packages as a set to avoid ABI mismatches due to rolling nature with total ~35k packages and tons of build options if building manually.

wez:
@jbeich oh, should https://github.com/wez/wezterm/blob/main/get-deps#L205 be switched out for something else? That script is used by both humans and CI, FWIW

wez:
@alfa07 this is looking good, thanks! One thing left to do is add docs for this new method:

Create `docs/config/lua/pane/get_pane_direction.md`; you can base the new page off of `docs/config/lua/pane/get_semantic_zones.md`.

If you want to test the doc build, you need to install a couple of utilities:

```
cargo install --vers "^0.4" mdbook
cargo install mdbook-linkcheck
cargo install mdbook-mermaid
cargo install gelatyx --version "^0.2"
```

Then you can run the build like this:

```
./ci/build-docs.sh
```

and then open `gh_pages/html/index.html` in your browser.

jbeich:
@wez, `mesa-devel` should be simply dropped but `bsd_deps` likely has more cruft. For example,
```
# List runtime dependencsies
$ pkg rquery %dn wezterm
xcb-util-image
xcb-util
libxkbcommon
libxcb
libX11
fontconfig
libssh2 # due to LIBSSH2_SYS_USE_PKG_CONFIG=1
wayland

# List build dependencies (also, omit docs and duplicates)
$ make -V BUILD_DEPENDS WITHOUT=DOCS | sed 'y/ /\n/' | sed '/\.pc/d; /^$/d; s,.*:[^/]*/,,'
rust
pkgconf
```

alfa07:
@wez Do you think I need to change anything?

wez:
Looks good, thank you!

PR #3053, opened 2023-02-04
Detect ALT key for mouse events on Windows

mbikovitsky:
This adds detection of ALT key presses for mouse events, which fixes the [default mouse bindings](https://wezfurlong.org/wezterm/config/mouse.html) that use this key as a modifier.

The code is basically lifted from Microsoft's own example for handling this case: <https://learn.microsoft.com/en-us/windows/win32/learnwin32/keyboard-input#keyboard-state>.


wez:
Thanks!

PR #3054, opened 2023-02-04
Add “New Window” to the dock menu on macOS

dahlia:
This patch adds a dock menu *New Window* on macOS, which mimics Terminal.app and iTerm.  It looks like below:

“New Window” showing up on WezTerm's dock menu

(You might wonder when this menu would be helpful…  It's the only way I'm aware of to spawn a new window without switching macOS Spaces when there are multiple Spaces and your current Space has no WezTerm window but there are any Spaces having WezTerm windows.)

It's my first time to contribute here, so there probably are several misplaced parts.  Please let me know so that I can adjust them!

### DIFF
--- a/window/src/os/macos/app.rs
+++ b/window/src/os/macos/app.rs
@@ -1,10 +1,12 @@
 use crate::connection::ConnectionOps;
 use crate::macos::menu::RepresentedItem;
 use crate::macos::{nsstring, nsstring_to_str};
+use crate::menu::{Menu, MenuItem};
 use crate::{ApplicationEvent, Connection};
 use cocoa::appkit::{NSApp, NSApplicationTerminateReply};
 use cocoa::base::id;
 use cocoa::foundation::NSInteger;
+use config::keyassignment::KeyAssignment;
 use objc::declare::ClassDecl;
 use objc::rc::StrongPtr;
 use objc::runtime::{Class, Object, Sel};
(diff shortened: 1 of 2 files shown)


window/src/os/macos/app.rs
```diff
@@ -120,6 +129,26 @@ extern "C" fn application_open_file(
     }
 }
 
+extern "C" fn application_dock_menu(
+    self_: &mut Object,
+    _sel: Sel,
+    _app: *mut Object,
+) -> *mut Object {
+    unsafe {
+        let dock_menu = NSMenu::alloc(nil)
```
wez, line 138: Please implement this using the higher level menu wrappers found in:
https://github.com/wez/wezterm/blob/main/window/src/os/macos/menu.rs

You should be able to do something like:

```rust
let dock_menu = Menu::new_with_title("");

let new_window = MenuItem::new_with(
    "New Window",
    Some(sel!(weztermPerformKeyAssignment:)),
    "");
new_window.set_represented_item(
    RepresentedItem::KeyAssignment(KeyAssignment::SpawnWindow));
dock_menu.add_item(new_window);
dock_menu.autorelease()
```

and then you can delete the `wezterm_spawn_window` function.

Note: I think you may need to tweak `Menu::autorelease` to have the same signature as the underlying https://docs.rs/objc/latest/objc/rc/struct.StrongPtr.html#method.autorelease method that it calls for this to work

wez (changes requested):
Thanks for this!

I'd like to do something similar to the more general menubar stuff, but focused on the dock (and for Windows in the future: a try icon), so it is desirable to target the slightly higher level set of functions for this, which make it a little easier to set up a menu and assign actions

window/src/os/macos/app.rs
dahlia: I amended the commit so that it uses the existing wrapper API instead of direct use of `cocoa::appkit::{NSMenu, NSMenuItem}` which requires to be `unsafe`.

wez:
Looks good, thanks!

PR #3074, opened 2023-02-08
palette: add CTRL-p / CTRL-n key bindings to move up/down

abusch:
Add `ctrl-p`/`ctrl-n` key bindings to move up and down in the command palette.


abusch:
Some tests are failing due to rate-limiting I think.

wez:
Thanks!

PR #3078, opened 2023-02-09
Fix missing longopt in cargo command

acook:
Cargo requires the `--feature` longopt in order to enable `vendored-fonts` and other feature options, otherwise the following error will occur:

```
error: Found argument 'vendored-fonts' which wasn't expected, or isn't valid in this context

Usage: cargo build [OPTIONS]

For more information try '--help'
```


wez:
Thanks!

PR #3089, opened 2023-02-09
ci: bump alpine version

panekj:


wez:
You'll need to git rm the old and git add the new workflow files after running this script, then amend those changes into this PR!

panekj:
> You'll need to git rm the old and git add the new workflow files after running this script, then amend those changes into this PR!

Forgot about that, will update PR in a bit.

ci/generate-workflows.py
```diff
@@ -944,7 +944,9 @@ def tag(self):
     Target(container="fedora:36"),
     Target(container="fedora:37"),
     Target(container="fedora:38"),
-    Target(container="alpine:3.15"),
+    Target(container="alpine:3.16"),
+    Target(container="alpine:3.17"),
+    Target(container="alpine:3.18"),
```
wez, line 949: I only want to build one version for alpine; I don't think anyone other than you uses it and building so many versions clogs up CI, making it less useful for the systems that most people do use.

Which version are we going to go with?

panekj: I can remove it altogether since I have to build wezterm on different platforms myself (due to incompatibility of mux server) anyway, so I can do it on Alpine as well

PR #3094, opened 2023-02-11
Allow negative number for cli get-text line

micmine:
The cli options of GetText --> start_line, end_line
Where not able to call with a negative number.


wez:
Thanks!

PR #3136, opened 2023-02-19
Add blur non-opaque windows on macos(macOS 10.10+)

Gkirito:
based on [NSVisualEffectView](https://developer.apple.com/documentation/appkit/nsvisualeffectview)

And this PR changed cocoa lib to[ Gkirito/core-foundation-rs](https://github.com/Gkirito/core-foundation-rs/tree/feature/add-NSVisualEffectMaterial), So I don't recommend merging it temporarily.

I opened a PR [Add NSVisualEffectView](https://github.com/servo/core-foundation-rs/pull/538) for cocoa lib([servo/core-foundation-rs](https://github.com/servo/core-foundation-rs)).When this PR([Add NSVisualEffectView](https://github.com/servo/core-foundation-rs/pull/538)) is merged, I will revert cocoa lib to the official repo and upgrade it.

### DIFF
--- a/window/Cargo.toml
+++ b/window/Cargo.toml
@@ -85,7 +85,7 @@ futures-util = "0.3"
 futures-lite = "1.12"
 
 [target.'cfg(target_os="macos")'.dependencies]
-cocoa = "0.20"
+cocoa = { git = "https://github.com/Gkirito/core-foundation-rs.git", subdir = "cocoa", branch = "feature/add-NSVisualEffectMaterial" }
 objc = "0.2"
 clipboard_macos = "0.1"
 core-foundation = "0.7"
--- a/window/Cargo.toml
+++ b/window/Cargo.toml
@@ -85,7 +85,7 @@ futures-util = "0.3"
 futures-lite = "1.12"
 
 [target.'cfg(target_os="macos")'.dependencies]
-cocoa = { git = "https://github.com/Gkirito/core-foundation-rs.git", subdir = "cocoa", branch = "feature/add-NSVisualEffectMaterial" }
+cocoa = { git = "https://github.com/servo/core-foundation-rs.git", subdir = "cocoa" }
 objc = "0.2"
 clipboard_macos = "0.1"
 core-foundation = "0.7"
--- a/window/Cargo.toml
+++ b/window/Cargo.toml
@@ -85,7 +85,7 @@ futures-util = "0.3"
 futures-lite = "1.12"
 
 [target.'cfg(target_os="macos")'.dependencies]
-cocoa = { git = "https://github.com/servo/core-foundation-rs.git" }
+cocoa = { git = "https://github.com/servo/core-foundation-rs.git" } # switch to release once 0.24.2 or later is out
 objc = "0.2"
 clipboard_macos = "0.1"
 core-foundation = "0.7"
--- a/window/Cargo.toml
+++ b/window/Cargo.toml
@@ -85,7 +85,7 @@ futures-util = "0.3"
 futures-lite = "1.12"
 
 [target.'cfg(target_os="macos")'.dependencies]
-cocoa = { git = "https://github.com/servo/core-foundation-rs.git", subdir = "cocoa" }
+cocoa = { git = "https://github.com/servo/core-foundation-rs.git" }
 objc = "0.2"
 clipboard_macos = "0.1"
 core-foundation = "0.7"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -623,13 +623,13 @@ dependencies = [
 [[package]]
 name = "cocoa"
 version = "0.24.1"
-source = "git+https://github.com/Gkirito/core-foundation-rs.git?branch=feature/add-NSVisualEffectMaterial#8bbdd5fd4ce925d3e0a7835df1fff67049e6d480"
+source = "git+https://github.com/servo/core-foundation-rs.git#65eccd9e792e777c4f41031ccd7400690d9de7ce"
 dependencies = [
  "bitflags",
  "block",
  "cocoa-foundation",
- "core-foundation 0.9.3 (git+https://github.com/Gkirito/core-foundation-rs.git?branch=feature/add-NSVisualEffectMaterial)",
- "core-graphics 0.22.3 (git+https://github.com/Gkirito/core-foundation-rs.git?branch=feature/add-NSVisualEffectMaterial)",
+ "core-foundation 0.9.3 (git+https://github.com/servo/core-foundation-rs.git)",
+ "core-graphics 0.22.3 (git+https://github.com/servo/core-foundation-rs.git)",
  "foreign-types",
  "libc",
  "objc",
@@ -638,12 +638,12 @@ dependencies = [
 [[package]]
 name = "cocoa-foundation"
 version = "0.1.1"
-source = "git+https://github.com/Gkirito/core-foundation-rs.git?branch=feature/add-NSVisualEffectMaterial#8bbdd5fd4ce925d3e0a7835df1fff67049e6d480"
+source = "git+https://github.com/servo/core-foundation-rs.git#65eccd9e792e777c4f41031ccd7400690d9de7ce"
 dependencies = [
  "bitflags",
  "block",
- "core-foundation 0.9.3 (git+https://github.com/Gkirito/core-foundation-rs.git?branch=feature/add-NSVisualEffectMaterial)",
- "core-graphics-types 0.1.1 (git+https://github.com/Gkirito/core-foundation-rs.git?branch=feature/add-NSVisualEffectMaterial)",
+ "core-foundation 0.9.3 (git+https://github.com/servo/core-foundation-rs.git)",
+ "core-graphics-types 0.1.1 (git+https://github.com/servo/core-foundation-rs.git)",
  "foreign-types",
  "libc",
  "objc",
@@ -801,9 +801,9 @@ dependencies = [
 [[package]]
 name = "core-foundation"
 version = "0.9.3"
-source = "git+https://github.com/Gkirito/core-foundation-rs.git?branch=feature/add-NSVisualEffectMaterial#8bbdd5fd4ce925d3e0a7835df1fff67049e6d480"
+source = "git+https://github.com/servo/core-foundation-rs.git#65eccd9e792e777c4f41031ccd7400690d9de7ce"
 dependencies = [
- "core-foundation-sys 0.8.3 (git+https://github.com/Gkirito/core-foundation-rs.git?branch=feature/add-NSVisualEffectMaterial)",
+ "core-foundation-sys 0.8.3 (git+https://github.com/servo/core-foundation-rs.git)",
  "libc",
 ]
 
@@ -822,7 +822,7 @@ checksum = "5827cebf4670468b8772dd191856768aedcb1b0278a04f989f7766351917b9dc"
 [[package]]
 name = "core-foundation-sys"
 version = "0.8.3"
-source = "git+https://github.com/Gkirito/core-foundation-rs.git?branch=feature/add-NSVisualEffectMaterial#8bbdd5fd4ce925d3e0a7835df1fff67049e6d480"
+source = "git+https://github.com/servo/core-foundation-rs.git#65eccd9e792e777c4f41031ccd7400690d9de7ce"
 
 [[package]]
 name = "core-graphics"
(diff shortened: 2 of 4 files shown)


Gkirito:
hello @wez,
That PR(https://github.com/servo/core-foundation-rs/pull/538) was closed.
I revert the cocoa lib to the official repo and upgrade it.

wez:
Thanks for this!
Can you try it with `front_end = "WebGpu"` to see if it is compatible? That causes wezterm to more directly use Metal.

Gkirito:
> Thanks for this! Can you try it with `front_end = "WebGpu"` to see if it is compatible? That causes wezterm to more directly use Metal.

> 

OK, This display effect use `front_end = "WebGpu"`
wezterm.lua
``` lua
--!/usr/bin/lua
return {
  color_scheme = 'DoomOne',
  font = wezterm.font_with_fallback {
    {
      family = 'JetBrains Mono',
      weight = 'Regular',
      harfbuzz_features = { 'calt=0', 'clig=0', 'liga=0' },
    },
    { family = 'MesloLGS NF', weight = 'Regular' },
  },
  front_end = "WebGpu",
  font_size = 13.5,
  -- Because the NSVisualEffect view is behind the view of which layers you want to draw, the window_background_opactity must be zero
  initial_cols = 150,
  initial_rows = 45,
  window_decorations = "TITLE|MACOS_FORCE_ENABLE_SHADOW|MACOS_NS_VISUAL_EFFECT_MATERIAL_BLUR"
}
```

Gkirito:
I try it with `front_end = "OpenGL"` or `front_end = "Software"`
image-20230221p2SPZFff@2x

I thought it maybe is not compatible with `OpenGL` or `Software`
In the code, I sorted these views. The `blurred_view` is behind the `view` which layers you want to draw

``` rust
let _: () = msg_send![
    *background_view,
    addSubview: blurred_view
    positioned: NSWindowOrderingMode::NSWindowBelow
    relativeTo: 0
];

let _: () = msg_send![
    *background_view,
    addSubview: *view
    positioned: NSWindowOrderingMode::NSWindowAbove
    relativeTo: 0
];
```

Gkirito:
Hello @wez, Do I need to fix this incompatible?  Or just use this feature on WebGpu. 👻

Gkirito:
Hello @wez, Do you think I need to change anything?
I'm very much looking forward to this feature! 👻
Could you review this PR?
Thanks so much!

BIKA-C:
@Gkirito , Hi, how did you add these system menubar entries on the top (like `shell`, `edit`, `view`..)

Gkirito:
> @Gkirito , Hi, how did you add these system menubar entries on the top (like `shell`, `edit`, `view`..)

Are you use a stable version? Menubar was added by this [commit](https://github.com/wez/wezterm/commit/b224aa1b5677fe4145143dfdf38a91a7270922ec).
You can use the nightly version.
``` bash
brew install wezterm-nightly
```

BIKA-C:
Thanks! can’t wait for it to be merged.
osx visualeffect function uses coregraphic. so it should be opengl independent.

Gkirito:
> Thanks for this! It looks good with webgpu, but no text is visible when not using webgpu. webgpu is not enabled by default, so I think this would be confusing. Please fix that and we can get this merged!

Okay, I will do my best to fix that

Gkirito:
Hello @wez ,
I fixed it.
Using `Software`
image-20230318uT1YCD0V@2x

And using `OpenGL`
image-20230318DVZ1suEm@2x

window/Cargo.toml
```diff
@@ -85,7 +85,7 @@ futures-util = "0.3"
 futures-lite = "1.12"
 
 [target.'cfg(target_os="macos")'.dependencies]
-cocoa = "0.20"
+cocoa = { git = "https://github.com/servo/core-foundation-rs.git" }
```
wez, line 88:
```suggestion
cocoa = { git = "https://github.com/servo/core-foundation-rs.git" } # switch to release once 0.24.2 or later is out
```

wez (changes requested):
This is looking good!
There's another issue to resolve though; here's the test scenario:

Save this config as `/tmp/3136.lua`:

```lua
return {
  color_scheme = 'DoomOne',
  front_end = "WebGpu",
  font_size = 13.5,
  window_background_opacity = 0.3,
  -- window_decorations = "TITLE|MACOS_FORCE_ENABLE_SHADOW|MACOS_NS_VISUAL_EFFECT_MATERIAL_BLUR"
}
```

Then, while wezterm is still running, edit the config to remove the `--` comment from the `window_decorations` line and then save the config file.

What should happen is that the window should change to match the same style that would have been used if the window was created with those window decorations.

The same should be true for the reverse: when started with those window_decorations active, editing the config to comment out the `window_decorations` line should undo the blur effect.

Gkirito:
Ok, I will check this, and fix it

Gkirito:
> This is looking good! There's another issue to resolve though; here's the test scenario:
> 
> Save this config as `/tmp/3136.lua`:
> 
> ```lua
> local wezterm = require 'wezterm'
> return {
>   color_scheme = 'DoomOne',
>   front_end = "WebGpu",
>   font_size = 13.5,
>   window_background_opacity = 0.3,
>   -- window_decorations = "TITLE|MACOS_FORCE_ENABLE_SHADOW|MACOS_NS_VISUAL_EFFECT_MATERIAL_BLUR"
> }
> ```
> 
> Then, while wezterm is still running, edit the config to remove the `--` comment from the `window_decorations` line and then save the config file.
> 
> What should happen is that the window should change to match the same style that would have been used if the window was created with those window decorations.
> 
> The same should be true for the reverse: when started with those window_decorations active, editing the config to comment out the `window_decorations` line should undo the blur effect.

It's done
Add the new function `update_window_background_blur` to do that

wez:
This configuration no longer respects the opacity setting:

```lua
return {
  color_scheme = 'DoomOne',
 window_background_opacity = 0.3,
}
```

Gkirito:
> This configuration no longer respects the opacity setting:
> 
> ```lua
> local wezterm = require 'wezterm'
> return {
>   color_scheme = 'DoomOne',
>  window_background_opacity = 0.3,
> }
> ```

I think it still has some effect. 
This pic will show the difference between `window_background_opacity` is 0.8 and `window_background_opacity` is 0.1

`window_background_opacity` = 0.8
image-202303205JnFRPCt@2x

`window_background_opacity` = 0.1
image-20230320gF51jf1N@2x

wez:
image

The window on the left is wezterm without your patches applied. The window on the right is this PR.

Both are running this configuration, which doesn't enable the material effect:

```lua
return {
  color_scheme = 'DoomOne',
 window_background_opacity = 0.3,
}
```

both versions should look the same.

Gkirito:
Maybe I have mistaken something. 
Let me check again.

Gkirito:
Hello @wez,
Because of NSVisualEffectView's limitations, I can't apply `Opaque = false` on this view, and setting this view `State = Inactive` will make this opaque.
So I use another way.
- Enable `MACOS_NS_VISUAL_EFFECT_MATERIAL_BLUR`:
```
  | - Window(background_view)
  |    | - blur_view 
  |    |    | - draw_view
```
- Disable `MACOS_NS_VISUAL_EFFECT_MATERIAL_BLUR`:
```
  | - Window(background_view)
  |   | - draw_view
```

I add three views to this way, the background view is transparent, and always exists.
The blur_view will remove or add by the config
And the draw_view will be the subview of blur_view or background_view.

BIKA-C:
i am a rust noob. here is a hack for workaround. `CGSSetWindowBackgroundBlurRadius` can blur all non fully transparent pixels for a window. This is done by core graphics. You need to set
```objc
window.backgroundColor = [[NSColor clearColor] colorWithAlphaComponent:0.01];
window.opaque = NO;

```

where the alpha value and colour are the configured values for background colour and opacity respectively.

`CGSSetWindowBackgroundBlurRadius` is unpublished api. To use it:
```objc

extern OSStatus CGSSetWindowBackgroundBlurRadius(CGSConnection connection, NSInteger windowNumber, int radius);
extern void* CGSDefaultConnectionForThread();


CGSSetWindowBackgroundBlurRadius(CGSDefaultConnectionForThread(),  nsWindow.windowNumber, 100);
```

Gkirito:
Hello @BIKA-C, 
I found it must add `CGSPrivate.h` header according to your plan and the unpublished API may not be easy to implement.
I will finish this pr first and then try your plan, If it finishes, I will open another pr.
Thanks so mush!

wez:
I'd be interested to see what this PR looks like using CGSSetWindowBackgroundBlurRadius; it seems like it would be a much simpler change that way.

Gkirito:
> I'd be interested to see what this PR looks like using CGSSetWindowBackgroundBlurRadius; it seems like it would be a much simpler change that way.

But It will make it impossible to publish the wezterm on the app store.
Like this [gist](https://gist.github.com/Serjip/c24853dc06e613ab9574)
> The use of non-public APIs can lead to a poor user experience should these APIs change in the future and is therefore not permitted

And it risks that it will suddenly be unavailable in future macOS versions.

I think we can keep both implementations and switch by config.

wez:
FWIW, I'm pretty sure wezterm already cannot be published on the app store, and I have no interest in publishing it there.

I'd prefer to optimize for simpler and easier to follow code, so if `CGSSetWindowBackgroundBlurRadius` is significantly easier to follow and test, then I'm fine with adopting that.

With that in mind, I'm not enthusiastic about adopting both techniques right now: it will just make the code even more complex.

Gkirito:
> I'd prefer to optimize for simpler and easier to follow code, so if CGSSetWindowBackgroundBlurRadius is significantly easier to follow and test, then I'm fine with adopting that.
>
> With that in mind, I'm not enthusiastic about adopting both techniques right now: it will just make the code even more complex.

So do I.
You means right, It will make the code more complex.
Okay, I will try `CGSSetWindowBackgroundBlurRadius` and open another pr.
Also, many thanks @BIKA-C 's plan.
Maybe we can collaborate on a new pr. There may be something I don't know about macOS development

BIKA-C:
@Gkirito 
image

I made a sample app if that's helpful. Here is the code, you can compile it with 
```shell
clang main.m -framework Appkit -o main
```

```objc
#import <AppKit/AppKit.h>
#import <AppKit/NSApplication.h>

extern OSStatus CGSSetWindowBackgroundBlurRadius(void* connection, NSInteger windowNumber, int radius);
extern void*    CGSDefaultConnectionForThread();

@interface AppDelegate : NSObject <NSApplicationDelegate>

@property (nonatomic, assign) NSWindow*  window;

@end

@implementation AppDelegate

- (BOOL)applicationShouldTerminateAfterLastWindowClosed:(NSApplication*)sender
{
    return true;
}

- (void)makeBlurWithRadius:(int)raidus
{
    self.window.opaque = false;
    CGSSetWindowBackgroundBlurRadius(CGSDefaultConnectionForThread(), self.window.windowNumber, raidus);
}

- (void)makeBlurWithRadius:(int)raidus color:(NSColor*)color
{
    self.window.backgroundColor = color;
    [self makeBlurWithRadius:raidus];
}

- (void)applicationDidFinishLaunching:(NSNotification*)aNotification
{
    // Create the window
    self.window = [[NSWindow alloc]
        initWithContentRect:NSMakeRect(0, 0, 400, 200)
                  styleMask:NSWindowStyleMaskTitled | NSWindowStyleMaskClosable
                            | NSWindowStyleMaskMiniaturizable | NSWindowStyleMaskResizable
                    backing:NSBackingStoreBuffered
                      defer:NO];
    [self.window setTitle:@"My App"];
    [self.window makeKeyAndOrderFront:NSApp];
    [self.window setTitlebarAppearsTransparent:YES];
    [self.window center];

    // blur
    [self makeBlurWithRadius:10 color:[[NSColor greenColor] colorWithAlphaComponent:0.4]];

    // NSVisualEffectView* blur = [[NSVisualEffectView alloc] initWithFrame:[[self.window contentView]
    // bounds]]; [blur setBlendingMode:NSVisualEffectBlendingModeBehindWindow]; [blur
    // setState:NSVisualEffectStateActive]; [blur
    // setMaterial:NSVisualEffectMaterialUnderWindowBackground]; [blur
    // setAutoresizingMask:NSViewWidthSizable | NSViewHeightSizable]; [self.window setContentView:blur];
}
@end

int main(int argc, const char* argv[])
{
    @autoreleasepool {
        AppDelegate*   delegate    = [[AppDelegate alloc] init];
        NSApplication* application = [NSApplication sharedApplication];
        [application setDelegate:delegate];
        [application run];
    }
}

```

Gkirito:
Wow! Thanks.
Btw, Is there a limit to the number of values that can be assigned to `radius`? If so, what is the maximum number?

BIKA-C:
Not so sure, tried with 1000000000. It works, but I can't tell the difference with 500

Gkirito:
@BIKA-C done, the new pr is  #3344

wez:
Docs: https://wezfurlong.org/wezterm/config/lua/config/macos_window_background_blur.html

idr4n:
Thank you for this guys 🙏🏻,this is really a nice feature that so far only iTerm had in Mac.

PR #3146, opened 2023-02-23
Improve XDG configuration search paths

jmbaur:
In addition to XDG_CONFIG_HOME, use XDG_CONFIG_DIRS (https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) to append paths for searching for configuration files.

### DIFF
--- a/config/src/lib.rs
+++ b/config/src/lib.rs
@@ -61,7 +61,7 @@ type ErrorCallback = fn(&str);
 
 lazy_static! {
     pub static ref HOME_DIR: PathBuf = dirs_next::home_dir().expect("can't find HOME dir");
-    pub static ref CONFIG_DIR: PathBuf = xdg_config_home();
+    pub static ref CONFIG_DIRS: Vec<PathBuf> = config_dirs();
     pub static ref RUNTIME_DIR: PathBuf = compute_runtime_dir().unwrap();
     static ref CONFIG: Configuration = Configuration::new();
     static ref CONFIG_FILE_OVERRIDE: Mutex<Option<PathBuf>> = Mutex::new(None);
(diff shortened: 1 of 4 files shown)


env-bootstrap/src/lib.rs
```diff
@@ -73,9 +73,11 @@ pub fn fixup_appimage() {
             clean_wezterm_config_env();
         }
 
-        if config::CONFIG_DIR.starts_with(append_extra_file_name_suffix(&appimage, ".config")) {
-            std::env::remove_var("XDG_CONFIG_HOME");
-            clean_wezterm_config_env();
+        for dir in config::CONFIG_DIRS.iter() {
+            if dir.starts_with(append_extra_file_name_suffix(&appimage, ".config")) {
+                std::env::remove_var("XDG_CONFIG_HOME");
```
wez, line 78: The purpose of this block was to check to see if XDG_CONFIG_HOME is controlled/manipulated by appimage. I don't think it makes sense to unset XDG_CONFIG_HOME here if a secondary directory under XDG_CONFIG_DIRS happens to match this criteria.

Perhaps the `for dir in CONFIG__DIRS` would be better off actually checking `std::env::var('XDG_CONFIG_HOME`)` directly?

config/src/lib.rs
```diff
@@ -337,6 +337,16 @@ fn xdg_config_home() -> PathBuf {
     }
 }
 
+fn xdg_config_dirs() -> Vec<PathBuf> {
+    let mut dirs = vec![xdg_config_home()];
+
+    if let Ok(d) = std::env::var("XDG_CONFIG_DIRS") {
+        dirs.extend(d.split(":").map(|s| PathBuf::from(s).join("wezterm")));
```
wez, line 344: Should probably use https://doc.rust-lang.org/std/env/fn.split_paths.html here rather than hard-coding the `:` splitting?

Also, consider making this XDG_CONFIG_DIRS bit guarded by `#[cfg(unix)]`

wez (commented):
Seems broadly OK; just a couple of comments!

env-bootstrap/src/lib.rs
jmbaur: That makes sense. My new changes should reflect this!

config/src/lib.rs
jmbaur: `split_paths` makes much more sense, thanks!

I applied the guard for using XDG_CONFIG_DIRS, although it does appear to be a thing on windows as well, maybe just unset by default?

wez:
Thanks!

PR #3213, opened 2023-03-06
Filter figterm from stateful check

grant0417:
Fig uses figterm a pseudo terminal for a lot of functionality, it runs between the shell and terminal. Unfortunately it is typically named `<shell> (figterm)`, which prevents the statuful check from passing.

Not sure if this is the best implementation, I personally would prefer to add `figterm` to the `default_stateless_process_list` but as a regex searching for `(figterm)` in the process name. This would be a bigger change but also up for doing. 

Also for some more context this is also done similarly in VSCode where they filter `(figterm)` out from the process name, but they do it at the process API level, I didn't want to change the `procinfo` impl however. 

https://github.com/microsoft/vscode/blob/728ee42999e6f07c32539f3dd4b09be1e89c8556/src/vs/platform/terminal/node/terminalProcess.ts#L408

### DIFF
--- a/mux/src/localpane.rs
+++ b/mux/src/localpane.rs
@@ -509,6 +509,10 @@ impl Pane for LocalPane {
             });
 
             fn default_stateful_check(proc_list: &LocalProcessInfo) -> bool {
+                // Fig uses `figterm` a pseudo terminal for a lot of functionality, it runs between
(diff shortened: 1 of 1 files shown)


wez:
Thanks! I think this looks reasonable and just needs a comment in the code

wez:
Thanks!

PR #3248, opened 2023-03-14
fix: Use cargo workspace dependecies for common crate dependecies

jalil-salame:
For example, `anyhow` is used in 32 crates and `log` is used in 26, this should make updating the required version of those dependencies much easier.


jalil-salame:
Force pushed fix to root Cargo.toml

wez:
Thanks, but I'm not sure I want this. The crates in the list are crates that rarely change.  eg: anyhow is always `1.0`, and it is less to type and easier to see at a glance than specifying a workspace dependency for it.

Factoring the version number out means that there is an additional level of indirection when looking to see what is in use.  I don't mind changing multiple files when a version bump happens; actively looking at those files can help find other issues or opportunities to make improvements in those moments.

jalil-salame:
I would tell you that this improves compilation times if it did, but as far as I know it is a purely aesthetic thing.

Thanks for taking the time to review this.

PR #3249, opened 2023-03-15
Run `cargo clippy --fix`

jalil-salame:
Categorized fixes by type, see git log.

`clippy` should probably be run on CI, there were a lot of unnecessary clones/inneficiencies that clippy caught.


wez:
Thanks for diving in, but this PR is too big to review. I'm pleased to see that you have different commits under the covers. I would be grateful if you could submit the high-value fixes as individual PRs when you are ready, rather than a giant PR like this.

For me, high-value fixes are where there is an obvious and objectively good reduction in boilerplate (that default derivation for enums commit seems good!), or a significant performance gain.  Low value changes are moving code around for no benefit other than the opinion of clippy.

I have limited time and I don't really want to spend it wading through low-value clippy lints, so I would appreciate it if you could be very selective with submitting PRs for this, when you are ready. Pick the top 3 and wait for those to get merged before the next 3 and so on.

Some of the files swept up in this PR are for non-critical things, like syncing color schemes.  Lint fixes there are much lower value than changes to the main part of the terminal.

Thanks!

PR #3254, opened 2023-03-15
fix(clippy): Remove unnecessary clone/to_string calls

jalil-salame:
# Purpose

Implement clippy fixes that relate to allocations, by removing calls to `clone`, `to_string` and `format!` when unnecessary.

Sometimes `clone` is removed in favor of copy (`*my_var`), which does not reduce allocations (as clone would not have allocated anyways), but it makes it clear that no allocation is performed.

## Concerns

A lot of calls to `to_string` are passed directly to `format!`, ie. `format!("{}", e.to_string())`, is this intended?

## Tests

All tests pass locally (Linux).

### DIFF
--- a/lua-api-crates/json/src/lib.rs
+++ b/lua-api-crates/json/src/lib.rs
@@ -150,7 +150,7 @@ fn lua_value_to_json_value(value: LuaValue, visited: &mut HashSet<usize>) -> mlu
                         mlua::Error::FromLuaConversionError {
                             from: lua_type,
                             to: "value",
-                            message: Some(format!("while processing {key:?}: {}", e.to_string())),
+                            message: Some(format!("while processing {key:?}: {e}")),
                         }
                     })?;
                     obj.insert(key, value);
--- a/lua-api-crates/share-data/src/lib.rs
+++ b/lua-api-crates/share-data/src/lib.rs
@@ -222,7 +222,7 @@ fn lua_value_to_gvalue_impl(value: LuaValue, visited: &mut HashSet<usize>) -> ml
                         mlua::Error::FromLuaConversionError {
                             from: lua_type,
                             to: "value",
-                            message: Some(format!("while processing {key:?}: {}", e.to_string())),
+                            message: Some(format!("while processing {key:?}: {e}")),
                         }
                     })?;
                     obj.insert(key, value);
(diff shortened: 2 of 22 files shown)


lua-api-crates/json/src/lib.rs
```diff
@@ -150,7 +150,7 @@ fn lua_value_to_json_value(value: LuaValue, visited: &mut HashSet<usize>) -> mlu
                         mlua::Error::FromLuaConversionError {
                             from: lua_type,
                             to: "value",
-                            message: Some(format!("while processing {key:?}: {}", e.to_string())),
+                            message: Some(format!("while processing {key:?}: {}", e)),
```
wez, line 153:
```suggestion
                            message: Some(format!("while processing {key:?}: {e}")),
```

lua-api-crates/share-data/src/lib.rs
```diff
@@ -222,7 +222,7 @@ fn lua_value_to_gvalue_impl(value: LuaValue, visited: &mut HashSet<usize>) -> ml
                         mlua::Error::FromLuaConversionError {
                             from: lua_type,
                             to: "value",
-                            message: Some(format!("while processing {key:?}: {}", e.to_string())),
+                            message: Some(format!("while processing {key:?}: {}", e)),
```
wez, line 225:
```suggestion
                            message: Some(format!("while processing {key:?}: {e}")),
```

luahelper/src/lib.rs
```diff
@@ -229,7 +229,7 @@ fn lua_value_to_dynamic_impl(
                         mlua::Error::FromLuaConversionError {
                             from: lua_type,
                             to: "value",
-                            message: Some(format!("while processing {key:?}: {}", e.to_string())),
+                            message: Some(format!("while processing {key:?}: {}", e)),
```
wez, line 232:
```suggestion
                            message: Some(format!("while processing {key:?}: {e}")),
```

wez (commented):
I think this looks good, thanks!
Couple of minor suggestions inline!

lua-api-crates/json/src/lib.rs
jalil-salame: force-pushed fix

jalil-salame:
Rebased on top of main and fixed comments, ready for review again.

wez:
Thanks!

PR #3272, opened 2023-03-18
fix(clippy): Don't use single character strings

jalil-salame:
Should improve performance as it passes a 16-bit codepoint instead of a 64-bit - 128-bit string slice, but it doesn't?

I see ~4% worse performance on my machine (Arch Linux, Intel i5 8th gen Processor).

Can I rely on the benchmarks?

### DIFF
--- a/bidi/generate/src/main.rs
+++ b/bidi/generate/src/main.rs
@@ -19,7 +19,7 @@ fn gen_class() -> anyhow::Result<()> {
     impl Entry {
         fn parse(line: &str) -> anyhow::Result<Option<Self>> {
             let line = line.trim();
-            if line.starts_with("#") || line.is_empty() {
+            if line.starts_with('#') || line.is_empty() {
                 return Ok(None);
             }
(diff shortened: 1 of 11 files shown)


wez:
This is a great example of theory vs. practice.
Yes, in theory, it is better/faster to use single chars for this stuff.
But: you won't notice the difference unless the code is being called frequently.
Almost none of the changed files here have a high cardinality of calls; most of these are (essentially) one-time calls, or things that happen only in test suites, or batch utilities for synchronizing code.

Depending on the nature of your benchmark, 4% sounds a bit too low of a variance to be statistically significant; I don't know if you can draw conclusions about that difference, but also, considering the magnitude of the difference, if it was instead 4% *better*, it's still not a convincingly large gain.

jalil-salame:
I was just running the benchmarks with `cargo criterion`, so I was asking about the quality of those benches (they are not in use for CI so...).

I am ready for this to be closed otherwise.

PR #3273, opened 2023-03-18
fix(clippy): Use `faster` methods on Iterators

jalil-salame:
- Use `find` instead of `filter(..).next()`
- Use `any` instead of `position(..).is_some()/.is_none()`
- Use `first/next` instead of `get(0)/nth(0)`
- Prefer `for` loops over `while let` loops on iterators

May improve performance.


wez:
Thanks!

PR #3274, opened 2023-03-18
fix(clippy): Remove closures where possible

jalil-salame:
- Removes closures and function calls for types that implement default:

  ```rust
  // Change
  let _my_str = get_str().unwrap_or(String::new);
  // To
  let _my_str = get_str().unwrap_or_default();
  ```

- Uses the `.cloned()/.copied()` methods where possible
- Use function pointer instead of simple closure

May improve performace as closures generate more code, and this might unlock some inlining opportunities.

### DIFF
--- a/tabout/src/lib.rs
+++ b/tabout/src/lib.rs
@@ -94,7 +94,7 @@ pub fn tabulate_output<S: std::string::ToString, W: std::io::Write>(
 
     for row in &display_rows {
         for (idx, col) in row.iter().enumerate() {
-            let max_width = col_widths.get(idx).cloned().unwrap_or_else(|| col.len());
+            let max_width = col_widths.get(idx).cloned().unwrap_or(col.len());
             let alignment = columns
                 .get(idx)
                 .map(|c| c.alignment)
@@ -200,7 +200,7 @@ pub fn tabulate_for_terminal(
 
     for row in rows {
         for (idx, col) in row.iter().enumerate() {
-            let max_width = col_widths.get(idx).cloned().unwrap_or_else(|| col.len());
+            let max_width = col_widths.get(idx).cloned().unwrap_or(col.len());
             let alignment = columns
                 .get(idx)
                 .map(|c| c.alignment)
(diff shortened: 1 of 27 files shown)


tabout/src/lib.rs
```diff
@@ -94,7 +94,7 @@ pub fn tabulate_output<S: std::string::ToString, W: std::io::Write>(
 
     for row in &display_rows {
         for (idx, col) in row.iter().enumerate() {
-            let max_width = col_widths.get(idx).cloned().unwrap_or_else(|| col.len());
+            let max_width = col_widths.get(idx).cloned().unwrap_or(col.len());
```
wez, line 97: Note that for these, we're now unconditionally calling `col.len()`.
Previously, it would only be called in the fallback case.
I don't think this matters in this file, but in general, that eager computation is undesirable.

wezterm-font/src/db.rs
```diff
@@ -22,7 +22,7 @@ impl FontDatabase {
         for parsed in font_info {
             self.by_full_name
                 .entry(parsed.names().full_name.clone())
-                .or_insert_with(|| vec![])
+                .or_insert_with(std::vec::Vec::new)
```
wez, line 25:
```suggestion
                .or_insert_with(Vec::new)
```

wezterm-font/src/ftwrap.rs
```diff
@@ -279,7 +279,7 @@ impl Face {
 
             names
                 .entry(sfnt_name.name_id as u32)
-                .or_insert_with(|| vec![])
+                .or_insert_with(std::vec::Vec::new)
```
wez, line 282:
```suggestion
                .or_insert_with(Vec::new)
```

tabout/src/lib.rs
jalil-salame: As far as I can tell `col` is a `String` thus `col.len()` is just reading a property and cheaper than calling a function conditionally.

Will keep in mind in the future though.

wezterm-font/src/db.rs
jalil-salame: Force pushed fix.

wez:
Thanks!

PR #3278, opened 2023-03-19
mux: Don't add excessively long lines to the logical lines

### LINKED ISSUE #3177: Newlines added to text copied from Windows full-screen applications

Windows

_No response_

### WezTerm version

wezterm 20230226-223328-0ac1cbc4

### Describe the bug

When text is selected by mouse in certain Windows programs (Far Manager, Midnight Commander for Windows, vi from OpenWatcom 1.9), sometimes it is copied to the clipboard with extra newlines at the end. The way to trigger the issue is to select multi-line text with mouse. After that, even single words selected by mouse have newlines at the end.

I know that the newlines are in the clipboard by pasting text in another program such as Notepad.

I've seen up to 30 newlines appended, but sometimes it's just one or two. After selecting several short single-line snippets, the issue goes away.

Some full-screen Windows programs are not affected. For instance, vim from msys2 doesn't appear to be affected. Programs run remotely over SSH are not affected either.

### To Reproduce

Install OpenWatcom 1.9
Start wezterm
Run OpenWatcom vi: `C:\WATCOM\binnt\vi.exe`
Select several lines by mouse
Select one word e.g. `no_name` by mouse
Open notepad
Paste
Most of the time, at least one newline would appear after the `no_name`

### Configuration

No config

### Expected Behavior

`no_name` is pasted without any newlines at the end

### Logs

```
Debug Overlay
wezterm version: 20230226-223328-0ac1cbc4 x86_64-pc-windows-msvc
Window Environment: Windows
OpenGL version: Intel(R) UHD Graphics 4.5.0 - Build 31.0.101.3358
Press ESC or CTRL-D to exit
>
```

### Anything else?

My guess is that there is some Windows-specific code for adding newlines to the text added to the clipboard (perhaps for certain kind of Windows programs), and the data used by that code is not cleared when a new text is selected.

proski:
The behavior of impl_get_logical_lines_via_get_lines() was inconsistent when all physical lines appeared to be one wrapped logical line.

The combined logical line would be appended to the result if and only if it exceeded MAX_LOGICAL_LINE_LEN (1024) characters.

Having an extra logical line in the result would lead to an extra newline being added to the text copied to clipboard.

Fixes #3177.

### DIFF
--- a/mux/src/pane.rs
+++ b/mux/src/pane.rs
@@ -508,9 +508,9 @@ pub fn impl_get_logical_lines_via_get_lines<P: Pane + ?Sized>(
                 });
             }
             Some(prior) => {
-                if prior.logical.last_cell_was_wrapped()
-                    && prior.logical.len() <= MAX_LOGICAL_LINE_LEN
-                {
+                if prior.logical.len() > MAX_LOGICAL_LINE_LEN {
+                    break;
+                } else if prior.logical.last_cell_was_wrapped() {
                     let seqno = prior.logical.current_seqno().max(line.current_seqno());
                     prior.logical.set_last_cell_was_wrapped(false, seqno);
                     prior.logical.append_line(line.clone(), seqno);


proski:
I don't feel very strong about this PR, but it fixes the issue #3177 that bites me all the time. So I'd rather have a PR so we can discuss the issue in more technical terms.

When text is selected on screen, `selection_text()` (`wezterm-gui/src/termwindow/selection.rs`) requests logical lines containing the selection. Even if the selection spans one line on screen, two logical lines can be reported, leading to a newline being added to the selected text.

The code for converting screen contents to logical lines interprets some lines as "wrapped". I don't fully understand the mechanism, but some full-screen Windows programs make every line "wrapped".

As a result, the logical line keeps growing until one of the following conditions happens:

* There are no more screen lines - the logical line is discarded.
* The logical line length exceeds 1024 characters - the logical line is preserved.

I don't see any reason to have different behavior. But the 1024 character limit is arbitrary. If we increase it, all excessive logical lines would be discarded. But somebody with a large monitor and small font would still hit the bug one day.

So this change makes all excessive logical lines to be discarded, whether they reach the end of the screen or the size limit.

I still don't understand why the code cares about logical lines beyond the selection - maybe we just need another function that would stop at the end of the selected text.

wez:
Thanks for digging into this!

The purpose of the limit is to constrain the CPU cost for things like `hyperlink_rules` in the case where a really big json object has been output to the terminal. There could be several MB of text on that single logical line that spans much more than the viewport of the display.

Can you help me understand the problem scenario? I think you are saying that the output looks something like this:

```
TEXT<SPACES><WRAP, BUT NO NEWLINE>
TEXT<SPACES><WRAP, BUT NO NEWLINE>
TEXT<SPACES>
```

and that all three of those physical lines end up being rolled into a single logical line.
For sufficiently long runs of these, the max logical line stuff then forces in a newline
where there shouldn't be one.

Is my understanding correct?

If so, I think the fix for this might be better as either:

* At the time of receiving the line and applying it to the terminal model, if we know that we are talking to conpty and the end of the line is a space, skip calling `set_last_cell_was_wrapped`:

https://github.com/wez/wezterm/blob/c1e9bcba5fd924161cd68694698537fe2d0e41ae/term/src/terminalstate/performer.rs#L145-L156

That would "fix" the data at its source, and might help in some other situations where things look janky on windows: for example, I'm pretty sure I've seen some odd resize/reflow behavior on windows where it looks like where falsely wrapped together.

I'm not at a Windows box right now, but I think the logic in there would need to be changed look something like this:

```rust
let mark_wrapped = !self.is_conpty || screen.line(y).visible_cells().last().map(|cell| cell.str() != " ").unwrap_or(false);
if mark_wrapped {
   screen.line_mut(y).set_last_cell_was_wrapped(true, seqno);
}
```

* Another option, which I'm a bit less fond of, would be to have a config option to adjust selection to work in terms of only physical lines. I think this might be a more difficult change with more fan-out, and it would be difficult to know when to automatically apply this, so I think of this as a "just make the pain stop" rather than "make it nice" kind of solution.

wez:
```diff
diff --git a/term/src/terminalstate/performer.rs b/term/src/terminalstate/performer.rs
index 714383cd9..be7494692 100644
--- a/term/src/terminalstate/performer.rs
+++ b/term/src/terminalstate/performer.rs
@@ -149,9 +149,19 @@ impl<'a> Performer<'a> {
                 // resized.
                 {
                     let y = self.cursor.y;
+                    let is_conpty = self.state.enable_conpty_quirks;
                     let screen = self.screen_mut();
                     let y = screen.phys_row(y);
-                    screen.line_mut(y).set_last_cell_was_wrapped(true, seqno);
+                    let should_mark_wrapped = !is_conpty
+                        || screen
+                            .line_mut(y)
+                            .visible_cells()
+                            .last()
+                            .map(|cell| cell.str() != " ")
+                            .unwrap_or(false);
+                    if should_mark_wrapped {
+                        screen.line_mut(y).set_last_cell_was_wrapped(true, seqno);
+                    }
                 }
                 self.new_line(true);
             }
```

This is the sort of change I meant. I can't reproduce the issue you reported because I don't have any of the programs you mentioned, so it's difficult for me to evaluate.

proski:
@wez: thank you for your quick response!

I've tested your patch. It's definitely "stopping the bleeding" in my real life scenarios, i.e. copying and pasting reasonably formatted code in Midnight Commander for Windows. As an extra bonus, copying and pasting multiple lines doesn't add trailing spaces anymore. I haven't noticed any breakage.

I'm still seeing those extra newlines if I fill the whole screen with printable characters and attempt copying text close to the top (3rd line from the top would do). Also, vi from OpenWatcom 1.9 is still affected because it draws a scrollbar on the right (so lame). So we might need to revisit the issue to limit line joining to the actual selection. That would likely require new, selection-aware code for line joining.

wez:
Thanks for the feedback!

I think that adding an option to control wrapping might be in order, somewhat similar to https://wezfurlong.org/wezterm/config/lua/config/canonicalize_pasted_newlines.html but have it process the data at the time at which it is output.

The patch I suggested above does something simple based on whitespace, but it could be augmented to consider non-ascii (eg: graphics) characters as something that shouldn't be wrappable, based on a config option.

Do you think you could capture a session recording from one of these apps outputting text that is broken in this way?

* Launch wezterm.  If possible, please use the default terminal size of 80x24 cells as it helps to keep things smaller and easier to manage.
* Inside that terminal run `wezterm record` to start a recording session; it will spawn a shell/prompt.
* Run through your reproduction steps with eg: watcom vi.  If you want to demonstrate different apps, please make a separate recording for each.  It is helpful for me if you leave the offending screen idle for ~5-10 seconds before you quit the application.
* Then type `exit` to exit the shell session and return to the prompt where you started recording.
* You should see a message like:

```
*** Finished recording to /var/tmp/wezterm-recording-sF6B3u.cast.txt
```

* Attach the file that it produced to this issue.

The file is an asciicast (compatible with https://asciinema.org/) and can also be replayed using `wezterm replay`.

The terminal recording allows me to replicate what is being sent to the terminal without requiring me to install the same applications as you and replicate your configuration for everything.

proski:
Here's the recording of the Watcom vi. There was no wezterm configuration during recording, everything is default. Version: `20230320-124340-559cb7b0`.
[wezterm-recording-kkwL7g.cast.txt](https://github.com/wez/wezterm/files/11025067/wezterm-recording-kkwL7g.cast.txt)

You can see that I'm selecting "no_name" in the title and pasting it. The `no_nam;101;1;0;1` must be an artifact of `wezterm record`, I'm getting `no_name` when not recording, But in both cases I'm getting two newlines that must be caused by `impl_get_logical_lines_via_get_lines()` overrunning the 1024 character limit twice when processing lines after the selection.

Let's consider use cases:

Selecting text on one line should never cause any trailing newlines. Even if it's a screen filled with text e.g. a JSON without newlines. But I'm OK if we tolerate that issue for the wall of text and mitigate common situations for now.

If the selection spans multiple lines, we cannot have a perfect behavior that would suit all use cases. Wrapping can occur on whitespace. Conversely, lines filling the screen to the end can be logically separate. But we can make the best guess and let the user add or remove newlines if the guess was wrong.

Support for ignoring various frames, scrollbars seems rather unimportant to me. Maybe I'm biased because I'm not using editors with those fancy features.

wez:
I've committed a first pass at handling this better. I think it may still be necessary to add some configuration around this, because I'm not convinced that my baked-on logic is universally good.

I'd appreciate your feedback!

proski:
I've been using the nightly version with that fix for several days. That's where I see the expected behavior:

* Internal editor of Midnight Commander for Windows
* Internal viewer of Midnight Commander for Windows

That accounts for the majority of the cases when I want to copy and paste text on Windows, so it's a significant improvement for me.

That's where I see the old behavior:

* Copying text from the panels Midnight Commander for Windows (e.g. filenames).
* Copying small snippets of text from a very long output line (e.g. a compiler flag). That happens even when using ssh to a Linux system.

I don't feel I understand all the implications to comment on the code change (dc36fe18f42ad5af8559ad5b4f04d9212142465d).

However, I believe that it would be beneficial to split the code for converting lines to logical lines. One version might be suitable for URL highlighting and similar tasks where the selection doesn't exist or doesn't matter. Another version should be for selection, where processing outside the selection would be removed, but processing inside the selection would not have any size limits. If the user selects a lot of text, it's because all that text is needed, so the terminal should do its best to process it properly, even if it's slow or memory consuming.

proski:
This PR is not serving any purpose. The issue has been resolved in the current code in a different way. There are still cases when newline are added when copying text from a very long line, but the fix should be different. The issue #3177 would be a better place for further discussion.

Issue #3300, opened 2023-03-20
new window opening is delayed on X11 with multiple displays


duhdugg:
### What Operating System(s) are you seeing this problem on?

Linux X11

### Which Wayland compositor or X11 Window manager(s) are you using?

bspwm 0.9.10

### WezTerm version

20230320-124340-559cb7b0

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

I am getting delayed window spawns when launching new windows only while connected to my docking station which has three external displays.

When not connected to my docking station, new windows open instantaneously.
When connected to my docking station, there is a delay of about 0.5 seconds before the window opens. Sometimes it takes up to 1 second before the window opens. The log I'm attaching shows 529 milliseconds between `TermWindow::new_window` being called and the screens being logged within the `resolve_geometry` function in `window/src/connection.rs`.

### To Reproduce

This should be reproducible on any X11 system with multiple displays, as I believe the issue is due to the nature of how `xcb::Connection` works. The more screens you have connected to the system, the more you're calling `XConnection::send_and_wait_request`, and more time is ultimately spent waiting on `xcb::Connection::wait_for_reply` to return.

### Configuration

I have tested that the issue persists with no config.

### Expected Behavior

Instant window opening is ideal.

### Logs

```
16:46:48.163  TRACE  wezterm_gui::termwindow            > TermWindow::new_window called with mux_wind
ow_id 0 TerminalSize { rows: 24, cols: 80, pixel_width: 800, pixel_height: 480, dpi: 96 } Dimensions 
{ pixel_width: 800, pixel_height: 480, dpi: 96 }
16:46:48.164  TRACE  wezterm_gui::termwindow            > RequestedWindowGeometry { width: Pixels(800
.0), height: Pixels(480.0), x: None, y: None, origin: ScreenCoordinateSystem }
16:46:48.262  TRACE  async_io::driver                   > block_on()
16:46:48.262  TRACE  async_io::driver                   > block_on: sleep until notification
16:46:48.692  TRACE  window::connection                 > Screens { main: ScreenInfo { name: "DP-2-2"
, rect: Rect(2560x1440 at (0, 0)), scale: 1.0, max_fps: Some(60) }, active: ScreenInfo { name: "DP-2-
[...]

PR #3301, opened 2023-03-20
new config option: resolve_new_window_geometry_bounds_from_screens

duhdugg:
addresses #3300

adds a new config option: `resolve_new_window_geometry_bounds_from_screens`

The default is true to maintain existing behavior. When set to false, the `bounds` variable within `resolve_geometry` is set to `euclid::rect(0, 0, 65535, 65535)`, which is consistent with the behavior that happens when the result of `screens()` is an `Err`.

With this set to false, new windows open instantly on X11 even while connected to three external displays. The tradeoff is that the `ResolvedGeometry` fields are based on these fallback values rather than the `ScreenRect` you would get from the default behavior, meaning window positioning will likely behave differently, but this is a non-issue with tiling window managers.

### DIFF
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -653,6 +653,9 @@ pub struct Config {
     #[dynamic(default = "default_true")]
     pub adjust_window_size_when_changing_font_size: bool,
 
+    #[dynamic(default = "default_true")]
+    pub resolve_new_window_geometry_bounds_from_screens: bool,
+
     #[dynamic(default)]
     pub use_resize_increments: bool,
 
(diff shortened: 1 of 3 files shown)


wez:
Slowness here is an issue with your drivers/installation. See https://github.com/wez/wezterm/issues/2073

I don't want to add an option to skirt this issue; in aggregate, this sort of things makes my life incrementally more and more complex and the real solution is to resolve the configuration problem with your system.

duhdugg:
That's OK, and I totally understand. I think it's more of an upstream driver problem, but I'm not confident I can get the right information to the people working on that. This is great software, and I'm happy with maintaining this patch privately for my own niche use case.

wez:
Closing this for the reasons mentioned above!

PR #3303, opened 2023-03-21
fix: wezterm cli activate-pane-direction now repaints

### LINKED ISSUE #2879: Window doesn't re-render gui when 'cli activate-pane-direction' is called

Windows

_No response_

### WezTerm version

wezterm 20221219-220659-19b01261

### Describe the bug

Wezterm's gui doesn't seem to recognize that `wezterm cli activate-pane-direction` was called until something causes the gui to visually update. I ran into this trying to get this Neovim plugin working https://github.com/numToStr/Navigator.nvim/issues/20. If there is some recommended way to fix this or add something to the plugin's [provided wezterm snippet](https://github.com/numToStr/Navigator.nvim/wiki/WezTerm-Integration) to make it re-render properly I couldn't find it in Wezterm's docs.

### To Reproduce

Running `wezterm cli activate-pane-direction Right; sleep 5` moves the pane focus but you can't visually see a difference until the 5 second sleep is over. 

### Configuration

no config

### Expected Behavior

I would expect Wezterm to visually display the fact that it has successfully changed the active pane

### Logs

```
Debug Overlay
wezterm version: 20221219-220659-19b01261
OpenGL version: NVIDIA GeForce RTX 3080/PCIe/SSE2 4.5.0 NVIDIA 517.48
Press ESC or CTRL-D to exit
```

### Anything else?

_No response_

ir-ae:
Closes #2879
I have no idea if this change is good or not but from testing it out it seems to fix the issue I was having (the activated pane not changing highlight) as well as other strange behavior around pane content overflowing into other panes when calling activate-pane-direction with one pane zoomed


mux/src/tab.rs
```diff
@@ -1417,6 +1417,8 @@ impl TabInner {
         if let Some(panel_idx) = self.get_pane_direction(direction, false) {
             self.set_active_idx(panel_idx);
         }
+        let mux = Mux::get();
+        mux.notify(MuxNotification::WindowInvalidated(self.id));
```
wez, line 1421: `WindowInvalidated` is for a window, but here we are a tab.
This likely just so happens to work because the ids for both of these start at 0.

What you may want to try instead is the `PaneOutput` notification, and emit it for both the prior active and the newly active pane ids.

wez (changes requested):
Thanks for diving in!

I think the general idea sounds fine, but the current implementation is a bit of a fluke!

I have a suggestion for you to try!

mux/src/tab.rs
ir-ae: I'm having trouble getting `PaneOutput` to work correctly. It does fix the issue of only working for the initial window, however it results in this error each time `wezterm cli activate-pane-direction` is called:
 ```
 22:47:57.991  ERROR  wezterm_mux_server_impl::local > encoding PDU to client: writing pdu data buffer: An existing connection was forcibly closed by the remote host. (os error 10054)
 ```
 I'm really not sure what's causing it, nor how to fix it as it seems to happen regardless of how I call notify with `PaneOutput`.

wez: Can you share the latest version of your code?

ir-ae: I've pushed my changes, although my latest attempt was to try to send `PaneOutput` to every pane.
Some other things I've tried include: 
- using `panel_idx` (not sure if 'pane' and 'panel' are different in this context)
- using `self.iter_panes().iter().find(|pane| pane.is_active)`.
- hard coding a value (just to check)

They each result in the same error.

wez: Can you try this instead?:

```diff
diff --git a/mux/src/tab.rs b/mux/src/tab.rs
index 89a256878..4b5922395 100644
--- a/mux/src/tab.rs
+++ b/mux/src/tab.rs
@@ -1416,6 +1416,11 @@ impl TabInner {
         }
         if let Some(panel_idx) = self.get_pane_direction(direction, false) {
             self.set_active_idx(panel_idx);
+
+            let mux = Mux::get();
+            if let Some(window_id) = mux.window_containing_tab(self.id) {
+                mux.notify(MuxNotification::WindowInvalidated(window_id));
+            }
         }
     }
```

wez: FWIW, I think `panel_idx` may be a typo. I don't remember writing that!

ir-ae: That works!

wez:
Thanks for confirming!

PR #3308, opened 2023-03-21
Update wezterm-linuxbrew.rb.template

davidmartos96:
Hello, I encountered some issues when upgrading to latest version with Homebrew because AppImage for Ubuntu 18.04 doesn't exist anymore (https://github.com/wez/wezterm/releases). 

I believe this would be the required change is that correct? Or should 20.04 be used? 22.04 is LTS as well.


davidmartos96:
Also, I imagine another tag would be necessary in order to trigger the bot on https://github.com/wez/homebrew-wezterm-linuxbrew

PR #3309, opened 2023-03-21
Drop duplicate entry

SuperSandro2000:
Already in the New section


wez:
I added this entry to tell you that the key assignment changed, in case you didn't read or realize that it was also new.

PR #3321, opened 2023-03-22
Fix some documentation pages

bew:
I'm basically adding `` ` `` around page titles when relevant.

Also changing many pages to have an h1 header, otherwise mkdocs generates a h1 header from the file name which is usually not what you want I think

If you have other ideas of large small changes to do, I can easily do them #vim-all-the-things :smiley:


bew:
(to be squashed & merged, the individual commits are not very interesting)

wez:
Thank you!

PR #3327, opened 2023-03-22
Ci toolchain: Do not use actions-rs/toolchain

jalil-salame:
[actions-rs/toolchain](https://github.com/actions-rs/toolchain) is **unsupported**:

- Has not recieved updates since November 2020 (~2.5 years)
- It uses Node.js 12 and GitHub will stop supporting it Summer this year(?) see this [article](https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/).

[dtolnay/rust-toolchain](https://github.com/dtolnay/rust-toolchain) is actively supported and its mostly a 1-1 replacement, the differences are:

- Uses the minimal profile always, so no need to specify it.
- There is no need to override the toolchain.

I also removed some things:

- None of the generated actions use `rustfmt` so I removed the component
- The toolchain is always stable, so I specified it in the action itself instead of in the action parameters.

### Follow up questions

I am in the process of planning a follow up PR to run less CI workflows if one fails and I have some questions about it:

1. Is this of any interest to you? I don't want you to review something you are not interested in.
2. Is it okay not to run `_continous` workflows if the non continous one fails?
3. Is it okay to run one Linux workflow first (I am thinking Ubuntu22.04), and skip all other Linux workflows if it fails? It would still run the MacOS/Windows workflows.
4. Is it okay to run `cargo check` instead of `cargo build` in the non `_tag`/`_continous` workflows? Should be much faster than a full compile/optimize run.

### DIFF
--- a/.github/workflows/gen_alpine3.15.yml
+++ b/.github/workflows/gen_alpine3.15.yml
@@ -56,12 +56,7 @@ jobs:
         with:
           submodules: "recursive"
       - name: "Install Rust"
-        uses: actions-rs/toolchain@v1
-        with:
-          profile: "minimal"
-          toolchain: "stable"
-          override: true
-          components: "rustfmt"
+        uses: dtolnay/rust-toolchain@stable
         env:
           ACTIONS_ALLOW_UNSECURE_COMMANDS: "true"
       - name: "Cache cargo"
--- a/.github/workflows/gen_alpine3.15.yml
+++ b/.github/workflows/gen_alpine3.15.yml
@@ -57,8 +57,6 @@ jobs:
           submodules: "recursive"
       - name: "Install Rust"
         uses: dtolnay/rust-toolchain@stable
-        env:
-          ACTIONS_ALLOW_UNSECURE_COMMANDS: "true"
       - name: "Cache cargo"
         uses: actions/cache@v3
         with:
--- a/ci/generate-workflows.py
+++ b/ci/generate-workflows.py
@@ -325,21 +325,16 @@ ln -s /usr/local/git/bin/git /usr/local/bin/git""",
 
         return steps
 
-    def install_rust(self, cache=True):
+    def install_rust(self, cache=True, toolchain="stable"):
         salt = "2"
         key_prefix = f"{self.name}-{self.rust_target}-{salt}-${{{{ runner.os }}}}-${{{{ hashFiles('**/Cargo.lock') }}}}"
-        params = {
-            "profile": "minimal",
-            "toolchain": "stable",
-            "override": True,
-            "components": "rustfmt",
-        }
+        params = dict()
         if self.rust_target:
             params["target"] = self.rust_target
         steps = [
             ActionStep(
                 name="Install Rust",
-                action="actions-rs/toolchain@v1",
+                action=f"dtolnay/rust-toolchain@{toolchain}",
                 params=params,
                 env={"ACTIONS_ALLOW_UNSECURE_COMMANDS": "true"},
             ),
--- a/ci/generate-workflows.py
+++ b/ci/generate-workflows.py
@@ -338,6 +338,25 @@ ln -s /usr/local/git/bin/git /usr/local/bin/git""",
                 params=params,
             ),
         ]
+        if "centos7" in self.name:
+            steps = [
+                RunStep(
+                    name="Install Rustup",
+                    run = """
+if ! command -v rustup &>/dev/null; then
+  curl --proto '=https' --tlsv1.2 --retry 10 -fsSL "https://sh.rustup.rs" | sh -s -- --default-toolchain none -y
+  echo "${CARGO_HOME:-$HOME/.cargo}/bin" >> $GITHUB_PATH
+fi
+"""
+                ),
+                RunStep(
+                    name="Setup Toolchain",
+                    run =f"""
+rustup toolchain install {toolchain} --profile minimal --no-self-update
+rustup default {toolchain}
+"""
+                ),
+            ]
         if "macos" in self.name:
             steps += [
                 RunStep(
--- a/ci/generate-workflows.py
+++ b/ci/generate-workflows.py
@@ -331,15 +331,10 @@ ln -s /usr/local/git/bin/git /usr/local/bin/git""",
         params = dict()
         if self.rust_target:
             params["target"] = self.rust_target
-        steps = [
-            ActionStep(
-                name="Install Rust",
-                action=f"dtolnay/rust-toolchain@{toolchain}",
-                params=params,
-            ),
-        ]
+        steps = []
+        # Manually setup rust toolchain in CentOS7 curl is too old for the action
         if "centos7" in self.name:
-            steps = [
+            steps += [
                 RunStep(
                     name="Install Rustup",
                     run = """
@@ -355,6 +350,14 @@ fi
 rustup toolchain install {toolchain} --profile minimal --no-self-update
 rustup default {toolchain}
 """
+                ),
+            ]
+        else:
+            steps += [
+                ActionStep(
+                    name="Install Rust",
+                    action=f"dtolnay/rust-toolchain@{toolchain}",
+                    params=params,
                 ),
             ]
         if "macos" in self.name:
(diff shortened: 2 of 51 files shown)


wez:
Thanks for this!

> planning a follow up PR to run less CI workflows if one fails and I have some questions about it:
> 
> 1. Is this of any interest to you? I don't want you to review something you are not interested in.

In general, I do prefer that all of the builds run, because the main value I derive from them is running on all the different platforms that I cannot test locally.  If one fails I don't want it to stop the others from trying and failing; they may fail differently, or they may succeed.  Either way: it is valuable signal for my next steps.

In particular: I strongly dislike the fail-fast strategy that is the default on GH actions. That makes sense when trying to minimize infrastructure cost, but my goal with the CI is to maximize the value of the time of the human that is making changes.

> 2. Is it okay not to run `_continous` workflows if the non continous one fails?

The continuous builds in this repo are triggered on push or by schedule.  The push builds should always run, as that is the main way that I catch things I may have overlooked before I push.

The scheduled builds are a bit redundant these days: they were from a time where the nightly builds used to only run on schedule and not on push.

The scheduled builds do occasionally surface problems with issues that may arise from a new rust compiler release, for example.

> 3. Is it okay to run one Linux workflow first (I am thinking Ubuntu22.04), and skip all other Linux workflows if it fails? It would still run the MacOS/Windows workflows.

No (because of the reasons above) but one thing I have been considering recently, because GH actions queues have been really clogged,  is that I'd like the pages, a linux build + mac + windows to get priority over the rest of the linux distros.

I think that would mean changing the workflows for the other linux builds to use workflow_dispatch and then adding a "deferred linux dispatcher" workflow that essentially did something like:

* sleep 10 minutes
* dispatch the other linux workflows

I don't know if that will work very well, and it could end up being more frustrating always waiting an extra 10 minutes for them in cases where the queues work well.

Another alternative option is to mark a number of the linux platforms as "tier 2" and switch them from building on push, and instead have them run on a schedule, but again: that would just make the latency for getting a build worse in most cases.

I've yet to be convinced that this sort of optimization would actually be better either for me personally, or for users that may be waiting on a fix in the nightly.

> 4. Is it okay to run `cargo check` instead of `cargo build` in the non `_tag`/`_continous` workflows? Should be much faster than a full compile/optimize run.

The build does need to run. From experience, I've seen `check` succeed when `build` fails.

.github/workflows/gen_alpine3.15.yml
```diff
@@ -56,12 +56,7 @@ jobs:
...
       - name: "Install Rust"
-        uses: actions-rs/toolchain@v1
-        with:
-          profile: "minimal"
-          toolchain: "stable"
-          override: true
-          components: "rustfmt"
+        uses: dtolnay/rust-toolchain@stable
         env:
           ACTIONS_ALLOW_UNSECURE_COMMANDS: "true"
```
wez, line 61: Do you know if this is still needed with this action? Can you try removing it?

wez:
I edited the security settings to allow dtolnay/rust-toolchain to run; I think you'll need to update this PR for them to trigger again.

.github/workflows/gen_alpine3.15.yml
jalil-salame: I have no idea, but I saw it being used without it, I will push a fix and you can try to run the ci.

jalil-salame: Seeing that CentOS failed to run curl, it might not be as simpler as removing ALLOW_UNSAFE_COMMANDS, I will recheck the actions used by dtonlay and see how he does it.

jalil-salame: Seeing that the rest of the builds have succeeded I think you are right.

Would you like me to report the issue with CentOS 7 upstream?

If I remember you were planning on dropping the CI for some Linux OSs, so I could also wait for that.

jalil-salame:
May I bring [`workflow_run`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run) to your attention?

You can trigger an action based on whether another action is `requested`, `in_progress` or `completed`. Setting all but one Linux distro to have `workflow_run` set to `in_progress`, might achive exactly what you were looking for, it is actually more work to set them to fail fast.

The only problem I see is with PRs like this one, which change the workflows, as I understand it, GH actions trigger on each matching `on` trigger, so the only options I see are:

1. Set the paths of the delayed distros to only trigger the action when the `.yml` file changes.

   Will clog the CI when you regenerate the actions.

2. Remove the `on.push`/`on.pull_request` triggers on the delayed distros.

   Will not trigger the CI if only the action for that specific distro is changed.

.github/workflows/gen_alpine3.15.yml
wez: I have users running on centos7, so I can't just drop support for that build.
Realistically, centos7 is not going to update the version of curl just for that command line flag, as it is essentially frozen except for critical security issues until it reaches its EOL date.

I think another option, that is honestly quite reasonable, is that we just manually run the rustup installer for centos7 in our centos7 workflow.  It's not much more than a one-liner anyway.

wez:
I haven't had a chance to think on the workflow_run thing, but in the meantime, if you're eager to hack on CI/github actions related things, here are a couple of suggestions:

* I believe that I've resolved the issue with `cargo nextest run`, so migrating the test runner to use that would be nice

* Adopting https://github.com/marketplace/actions/rust-cache and https://github.com/marketplace/actions/cargo-install (the latter to install nextest) might streamline some things

* Splitting the macOS build into separate arm and intel builds, and then adding a stage at the end to lipo the two builds together would:
   * allow publishing separate arm and intel builds, resulting in smaller downloads for more savvy users on macos
   * cut the build time for the universal binary in ~half if the two workflows happen to run in parallel

    Relevant portions of the build:
    https://github.com/wez/wezterm/blob/main/ci/generate-workflows.py#L384-L394

    https://github.com/wez/wezterm/blob/main/ci/deploy.sh#L44-L49

.github/workflows/gen_alpine3.15.yml
jalil-salame: Do you mean to modify the `yml` file or the python script? I am assuming it's the `yml` file.

jalil-salame:
Running nextest 3 times on main did not show any flaky tests!

Should I work on those improvements within this PR or make a separate one?

wez:
> Should I work on those improvements within this PR or make a separate one?

Separate please!

wez:
I think the changes for centos7 appear to be working, and just need to be integrated back into `ci/generate-workflows.py` in order to merge this PR!

.github/workflows/gen_alpine3.15.yml
jalil-salame: I patched the python file to generate the CI, now it manually downloads and installs rustup.

ci/generate-workflows.py
```diff
@@ -325,25 +325,38 @@ def install_git(self):
...
         steps = [
             ActionStep(
                 name="Install Rust",
-                action="actions-rs/toolchain@v1",
+                action=f"dtolnay/rust-toolchain@{toolchain}",
                 params=params,
-                env={"ACTIONS_ALLOW_UNSECURE_COMMANDS": "true"},
             ),
         ]
+        if "centos7" in self.name:
```
wez, line 341: I think this conditional should have an else and that the ActionStep from above should go in the else?

jalil-salame: Addressed in latest commit, I don't think it looks much better...

wez: it's not about how it looks so much as not trying to run the action that will fail on centos7.

Did you re-run this and amend in the workflows? GH is only showing me this file as being change in the latest update to the PR?

jalil-salame: There is no change to the actions when rerun, it's a very tiny thing, thanks Python, but originally the `if` replaces steps `=`, instead of appending to them `+=`, the change is semantically the same...

wez: Ah, subtle. I didn't notice it was replacing, which is why I asked for the `if` to be moved

wez:
Thank you for working on this!

PR #3336, opened 2023-03-23
Update shell-integration.md

derekthecool:
Add missing line to powershell starship OSC 7 integration. The variable for $ansi_escape was missing.

I have tested this on windows 11 and proved that it works with this addition.


wez:
Thanks!

PR #3339, opened 2023-03-23
Fix leak of Pane struct

mbikovitsky:
AFAICT this is a cyclic dependency between `read_from_pane_pty` and `parse_buffered_data`:

1. `read_from_pane_pty` exits when it gets an EOF on the pane reader.
2. `parse_buffered_data` exits when `read_from_pane_pty`  signals it to exit.
3. But, `parse_buffered_data` holds an Arc for the pane.
4. So `read_from_pane_pty` will never get an EOF, since the pane is being kept alive by `parse_buffered_data`, which is waiting for `read_from_pane_pty` to get an EOF.

It may be that this is a leak only on Windows. The symptoms I observed were dangling `OpenConsole.exe` processes after panes were closed. _Maybe_ it's because `Pane` owns the `OpenConsole.exe` process? And as long as this process is alive then the PTY is technically alive. I haven't dug deeper, so this is just conjecture.

The fix here kinda reverts the state to before commit 9e67ad7. Except instead of passing `PaneId`s around, we're passing a `Weak` pointer.

### DIFF
--- a/mux/src/lib.rs
+++ b/mux/src/lib.rs
@@ -20,7 +20,7 @@ use std::collections::HashMap;
 use std::convert::TryInto;
 use std::io::{Read, Write};
 use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
-use std::sync::Arc;
+use std::sync::{Arc, Weak};
 use std::thread;
 use std::time::{Duration, Instant};
 use termwiz::escape::csi::{DecPrivateMode, DecPrivateModeCode, Device, Mode};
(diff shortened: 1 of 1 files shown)


wez:
Thanks for this!

Yeah, I think this is likely Windows specific because I haven't noticed anything like this on mac or linux and I've been running that change for months now, while I've only fired up windows once or twice in the time period.

I think this looks like a reasonable change: it's conceptually unfortunate to have to "pay the cost" of upgrading the weak pointer every time, but the cost of that is dwarfed by all the other work that we need to do in those functions anyway, so it sounds good!

PR #3341, opened 2023-03-24
ci: Use rust-cache instead of the general cache action

jalil-salame:
## Blockers

- [x] Merge #3327 so that conflicts can be resolved in this smaller PR.
- [x] Allow `Swatinem/rust-cache@v2` action in CI

## Reasoning

Cache rust artifacts using [`Swatinem/rust-cache@v2`](https://github.com/marketplace/actions/rust-cache) as discussed in #3327


jalil-salame:
It seems neither `termwiz`, nor `wezterm_ssh` actions use caching, is this on purpose or should I add it as part of the PR?

- wezterm_ssh: https://github.com/wez/wezterm/blob/main/.github/workflows/wezterm_ssh.yml#L25-L61
- termwiz: https://github.com/wez/wezterm/blob/main/.github/workflows/termwiz.yml#L27-L45

wez:
I've enabled the cache action in the CI settings

> should I add it as part of the PR?

Yes please!

jalil-salame:
Fedora 35 is failing to fetch some URLS with HTTP: 404, the CI might need to be rerun.

.github/workflows/gen_windows_tag.yml
```diff
@@ -20,12 +20,8 @@ jobs:
...
           target: "x86_64-pc-windows-msvc"
       - name: "Cache cargo"
-        uses: actions/cache@v3
+        uses: Swatinem/rust-cache@v2
         with:
-          path: |
-            ~/.cargo/registry
-            ~/.cargo/git
-            target
           key: "windows-x86_64-pc-windows-msvc-2-${{ runner.os }}-${{ hashFiles('**/Cargo.lock') }}-cargo"
```
wez, line 25: it may be worth revisiting the `hashFiles('**/Cargo.lock')` portion of this.  Right now any change to the deps in the Cargo.lock will invalidate the whole cache, but it feels desirable to reuse some of the previous cache if only one or two deps are different

jalil-salame: I should remove it but the action itself hashes all Cargo.lock/Cargo.toml files https://github.com/Swatinem/rust-cache#cache-details so it makes little difference

wez: Then removing just makes things simpler on this end, so that's fine :)

jalil-salame: Solved in latest commit!

wez:
Thank you again!

PR #3342, opened 2023-03-24
ci: Use nextest

jalil-salame:
## Blockers

- [x] Merge #3327 , it has conflicts with this PR.
- [x] Merge #3341 , it has conflicts with this PR.
- [x] Fix OpenSUSE failing `wezterm-ssh` tests see [this comment](https://github.com/wez/wezterm/pull/3342#issuecomment-1483164632)
- [x] Allow `baptiste0928/cargo-install@v2` in CI.

## Reasoning

As discussed in #3327 , nextest no longer has problems with flaky tests and can be used to speed up CI.

_The more I work with the generate-workflows.py, the more I get the urge to RiiR_

### DIFF
--- a/ci/generate-workflows.py
+++ b/ci/generate-workflows.py
@@ -437,7 +437,7 @@ cargo build --all --release""",
         ]
 
     def test_all_release(self):
-        run = "cargo nextest run --all --release"
+        run = "cargo nextest run --all --release --no-fail-fast"
         if "macos" in self.name:
             run += " --target=x86_64-apple-darwin"
         if self.name == "centos7":
--- a/ci/generate-workflows.py
+++ b/ci/generate-workflows.py
@@ -919,7 +919,7 @@ TARGETS = [
     Target(container="fedora:36"),
     Target(container="fedora:37"),
     Target(container="alpine:3.15"),
-    Target(name="opensuse_leap", container="registry.opensuse.org/opensuse/leap:15.3"),
+    Target(name="opensuse_leap", container="registry.opensuse.org/opensuse/leap:15.4"),
     Target(
         name="opensuse_tumbleweed",
         container="registry.opensuse.org/opensuse/tumbleweed",
--- a/ci/generate-workflows.py
+++ b/ci/generate-workflows.py
@@ -143,7 +143,7 @@ class CheckoutStep(ActionStep):
 
 class InstallCrateStep(ActionStep):
     def __init__(self, crate: str, key: str, version=None):
-        params = {"crate": crate, "key": key}
+        params = {"crate": crate, "cache-key": key}
         if version is not None:
             params["version"] = version
         super().__init__(
--- a/ci/generate-workflows.py
+++ b/ci/generate-workflows.py
@@ -141,6 +141,18 @@ class CheckoutStep(ActionStep):
         super().__init__(name, action="actions/checkout@v3", params=params)
 
 
+class InstallCrateStep(ActionStep):
+    def __init__(self, crate: str, version=None):
+        params = {"crate": crate}
+        if version is not None:
+            params["version"] = version
+        super().__init__(
+            f"Install {crate} from Cargo",
+            action="baptiste0928/cargo-install@v2",
+            params=params
+        )
+
+
 class Job(object):
     def __init__(self, runs_on, container=None, steps=None, env=None):
         self.runs_on = runs_on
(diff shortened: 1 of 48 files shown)


jalil-salame:
## Problems

Seems like `nextest` has problems with GLIBC on CentOS 7 & 8, and maybe also alpine? It might be because `cargo-install` is caching incompatible versions.

OpenSUSE Leap fails with a SIGSEV on `wezterm_ssh` which might be because it's flaky there.

It seems that `nextest` enables `--fail-fast` by default when on CI, I will turn that off.

## Solutions

Locally I have reverted CentOS 7 & 8 and Alpine to `cargo-test`, I might try setting the caching options before applying that though.

OpenSUSE Leap might need some more work to be done on the flaky tests, that or enabling `--retries=2`

jalil-salame:
On the other hand, I am seeing ~1-3 minutes less spend on testing compared to #3327 :confetti_ball:

ci/generate-workflows.py
```diff
@@ -136,6 +136,18 @@ def __init__(self, name="checkout repo", submodules=True):
         super().__init__(name, action="actions/checkout@v3", params=params)
 
 
+class InstallCrateStep(ActionStep):
+    def __init__(self, crate: str, version=None):
+        params = {"crate": crate}
```
wez, line 141: It's best to always use `--locked` with `cargo install`.  nextest explicitly doesn't support installing without `--locked`.

If you pass the container name in here, you can use it to set the cache key and that should resolve the glibc issue which arises from falsely sharing cache artifacts.

```suggestion
        params = {"crate": crate, "args": "--locked", "cache-key": container_name}
```

wez:
> The more I work with the generate-workflows.py, the more I get the urge to RiiR

I'm kinda open to it, but I don't know if it is really necessary :-)

ci/generate-workflows.py
jalil-salame: Pretty sure the install action defaults to locked, will double check tomorrow, thanks!

wez: https://nexte.st/book/installing-from-source.html?highlight=cache#using-a-cached-install-in-ci makes a point of using the old, deprecated `locked: true` flag, which was replaced by `args: --locked`.

jalil-salame: `baptiste0928/cargo-install@v1` doesn't use `--locked` by default

We are using `v2` here though, which does `--locked` by default, I think merging #3341 might fix caching as it automatically keys by OS, I think it's better to wait for that to be merged before tinkering with this.

jalil-salame:
> I'm kinda open to it, but I don't know if it is really necessary :-)

It's not necessary but I am tempted... I was thinking of leveraging `serde_yaml` to write the actions; `yv` seems so fragile... I thought of using python-yaml to rewrite the script, but then you need a `virtualenv` and `requirements.txt` and that seems like another headache, if we already have a rust toolchain then installing `serde_yaml` is trivial.

ci/generate-workflows.py
jalil-salame: I will set them explicitely then, and maybe file an issue about containers

jalil-salame: Used `name` instead of `container_name`

jalil-salame: Looking at the CI runs I can confirm that it is not necessary to specify `--locked`

jalil-salame:
OpenSUSE Leap still has problems with `wezterm-ssh`:

SIGSEGV [   1.039s] wezterm-ssh::lib e2e::sftp::remove_file_should_fail_if_path_to_directory
SIGSEGV [   0.822s] wezterm-ssh::lib e2e::sftp::rename_should_support_singular_file
SIGSEGV [   1.085s] wezterm-ssh::lib e2e::sftp::symlink_should_succeed_even_if_path_missing

jalil-salame:
Rebased on top of `main` for easier merging.

Current PR problems:

- OpenSUSE Leap has Flaky tests
- OpenSUSE Leap doesn't have tar installed(?) and can't restore the cache.
- Fedora 35 had problems with its mirrors

Issues for future PRs:

- MacOS doesn't use caching (maybe address in PR handling the split of ARM and x86_64 MacOS)
- Cache is overfull (80GB/10GB)
  - Each build takes up ~750MB with 17 different targets this ends up being ~12.5GB which means that a full rebuild causing all actions to fire will overwhelm the cache, maybe consider removing caching from some Linux distros.

wez:
re: failures for opensuse leap, most likely they are due to the openssl issue mentioned in:
f56069e279dc957fe7d83d7265f8bc8dbf9f10ef

which should have been resolved by:
https://github.com/wez/libssh-rs/commit/44a08196acd1a2277ec7ebce2a47618775f3c59c

That hinges on this condition:
https://github.com/sfackler/rust-openssl/blob/4ff734fe4c5a22f7346b7b3c47ece4c4c1c01817/openssl-sys/src/lib.rs#L122

if opensuse has a version of openssl that doesn't support `OPENSSL_INIT_NO_ATEXIT` then the mitigation will be ineffective.  That should be active for openssl 110 and later (https://github.com/sfackler/rust-openssl/blob/4ff734fe4c5a22f7346b7b3c47ece4c4c1c01817/openssl-sys/build/cfgs.rs#L73-L74)

Looking at pkgs.org, it looks like leap 15.3 has `openssl-1.1.1d`, while 15.4 has `openssl-1.1.1l`.

AFAICT, both of those should be acceptable.  I wonder if there might be an alternative openssl implementation installed there? Or if there is something weird about the version on 15.3?

I would suggest bumping the leap container from 15.3 to 15.4 to see if that helps, and if not, the way I traced the crash previously was:

* `ulimit -S -c unlimited` - allows the segv to drop core file(s)
* `cargo nextest run e2e` - trigger the segv
* `coredumpctl debug` - launches gdb on the latest coredump

If you find yourself in a position to try that (eg: using a docker leap container locally) you could extract the backtrace to see if it is also in openssl (libcrypto in particular), and maybe see what the deps are.

Another option, if you just want to get this PR merged, is to skip using `nextest` and just use regular `cargo test` for the leap build until someone has a chance to look into this further.

jalil-salame:
That was the issue! Thanks!

wez:
Thank you again!!

PR #3344, opened 2023-03-24
feat(config & window): use CGSSetWindowBackgroundBlurRadius to set ba…

Gkirito:
…ckground blur on macos

### DIFF
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -485,7 +485,7 @@ pub struct Config {
 
     /// Only works on MacOS
     #[dynamic(default)]
-    pub window_background_blur: i64,
+    pub macos_window_background_blur: i64,
 
     /// Specifies the alpha value to use when rendering the background
     /// of the window.  The background is taken either from the
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -483,6 +483,10 @@ pub struct Config {
     #[dynamic(default)]
     pub background: Vec<BackgroundLayer>,
 
+    /// Only works on MacOS
+    #[dynamic(default)]
+    pub window_background_blur: i64,
+
     /// Specifies the alpha value to use when rendering the background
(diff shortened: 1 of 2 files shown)


Gkirito:
#3136 
It is easier to set blur using `CGSSetWindowBackgroundBlurRadius`.

Gkirito:
Hello, @wez Could you please review it again?

Gkirito:
use this config file
``` lua
return {
  color_scheme = 'DoomOne',
  window_background_opacity = 0.8,

-- To disable the blur, set the value to `0`.
  window_background_blur = 60,
  window_decorations = "TITLE|MACOS_FORCE_ENABLE_SHADOW"
}

```

config/src/config.rs
```diff
@@ -483,6 +483,10 @@ pub struct Config {
     #[dynamic(default)]
     pub background: Vec<BackgroundLayer>,
 
+    /// Only works on MacOS
+    #[dynamic(default)]
+    pub window_background_blur: i64,
```
wez, line 488: naming wise: if this can only every apply to macos, then it would be preferable to rename it to `macos_window_background_blur`.

If it is feasible to also make this same setting work with the same semantics on Windows, which I believe has a similar concept, then keeping the name more general is reasonable.

wez (commented):
Yeah, this looks much simpler!
I haven't tried to run this yet, but it looks clean.
I have one comment on naming

config/src/config.rs
Gkirito: OK, I rename it to `macos_window_background_blur`
And on Windows seems to need to use the gaussian blur effect

wez:
Thanks! rebased and pushed locally as 1b24ff1abf3ee6ecfcb5728d6478d758fa377bd8 and 63676c30d23711bb698bab5a719ab728e8e6e4da

PR #3358, opened 2023-03-24
Use install action instead of manually calling `cargo install`

jalil-salame:
As part of #3342 I found that `cargo install` was being called directly to install `cargo-fuzz`, this is a smaller PR that fixes just that.

The result should be cached thus making the termwiz CI faster.


wez:
Thanks!

PR #3367, opened 2023-03-26
CI: Delay less important Linux Distros' workflows

### LINKED ISSUE #3360: ci: Prefer certain builds
## Problem

Current CI is very expansive, so runners may pick up builds in an inconvenient order.

## Proposed solution

Use [`workflow_run`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run):

1. Select Linux Targets to be priorized.
2. Add yaml to non priorized Linux Targets, referencing priorized targets:

   ```yaml
   on:
     workflow_run:
       workflows: ["ubuntu_build", "centos_build"]
       type: ["in_progress"]
   ```

This will make it so that those Linux targets wait until any of the priorized targets are running before starting.

It might still clog the CI, but its worth a try.

## Discussion questions

- Which distros should be priorized?
  - `opensuse_tumbleweed`, `alpine` and `ubuntu` seem to be the fastest workflows (by ~3-4 minutes)
- Should we instead use `on_completed`? This might ensure that the other CI workers are running.
- `workflow_run` doesn't support `path` triggers, only branch triggers, my suggestion is to add a `push.paths = .github/workflows/this_workflow.yaml` trigger, this will clog the CI if the workflows are updated, but it will ensure they are triggered if the priorized workflows aren't triggered.

jalil-salame:
# Changes

Closes #3360

Priorize workflows:

- Windows
- MacOS
- Ubuntu 22.02

Every other target specific workflow, ie. CenOS 7, will run after any of the
above mentioned targets' workflow finishes.

## Reasoning

Ubuntu 22.04 covers a high percentage of the Linux testing needed by Wezterm,
delaying other workflows is therefore undesireable.

See issue #3360 for a more in depth discussion

### Changeable

Windows and MacOS might trigger every other Linux Distro's workflow to run even
if any of their trigger paths don't change, fixing this is a bit more involved
but might be desireable.

### Not Changeable

As it waits for the workflows to complete the minimum time it takes for all the
workflows to complete will be higher.

### DIFF
--- a/ci/generate-workflows.py
+++ b/ci/generate-workflows.py
@@ -177,7 +177,7 @@ class Target(object):
         continuous_only=False,
         app_image=False,
         is_tag=False,
-        priorize=False,
+        hi_pri=False,
     ):
         if not name:
             if container:
@@ -193,7 +193,7 @@ class Target(object):
         self.app_image = app_image
         self.env = {}
         self.is_tag = is_tag
-        self.priorize = priorize
+        self.hi_pri = hi_pri
 
     def render_env(self, f, depth=0):
         self.global_env()
@@ -903,7 +903,7 @@ cargo build --all --release""",
 
 TARGETS = [
     Target(container="ubuntu:20.04", continuous_only=True, app_image=True),
-    Target(container="ubuntu:22.04", continuous_only=True, priorize=True),
+    Target(container="ubuntu:22.04", continuous_only=True, hi_pri=True),
     # debian 8's wayland libraries are too old for wayland-client
     # Target(container="debian:8.11", continuous_only=True, bootstrap_git=True),
     # harfbuzz's C++ is too new for debian 9's toolchain
@@ -915,7 +915,7 @@ TARGETS = [
     ),
     Target(name="centos8", container="quay.io/centos/centos:stream8"),
     Target(name="centos9", container="quay.io/centos/centos:stream9"),
-    Target(name="macos", os="macos-11", priorize=True),
+    Target(name="macos", os="macos-11", hi_pri=True),
     # https://fedoraproject.org/wiki/End_of_life?rd=LifeCycle/EOL
     Target(container="fedora:35"),
     Target(container="fedora:36"),
@@ -930,11 +930,11 @@ TARGETS = [
         name="windows",
         os="windows-latest",
         rust_target="x86_64-pc-windows-msvc",
-        priorize=True
+        hi_pri=True
     ),
 ]
 
-PRIORITY_TARGETS = [target.name for target in TARGETS if target.priorize]
+PRIORITY_TARGETS = [target.name for target in TARGETS if target.hi_pri]
 
 
 def generate_actions(namer, jobber, trigger, is_continuous, is_tag=False):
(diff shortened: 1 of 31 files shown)


wez:
Hmm, perhaps for the purposes of testing the delay in this PR, you should remove the push paths triggers for the lower priority workflows? Then when we see that the concept worked, you can restore those before we merge?

jalil-salame:
I'll give that a try

jalil-salame:
I'll come back to this tomorrow I can't tell what I'm doing anymore. I think GH reruns workflows based on the diff of the whole PR but I'm to tired to investigate.

jalil-salame:
Marking it as a draft until testing on #3369 is done

jalil-salame:
Reading the documentation it seems you can't test it on a PR, see #3372 , I could try testing it in a fork though.

ci/generate-workflows.py
```diff
@@ -177,6 +177,7 @@ def __init__(
         continuous_only=False,
         app_image=False,
         is_tag=False,
+        priorize=False,
```
wez, line 180: spelling here. This is actually quite awkward to spell/type, so how about we rename it to `hi_pri` instead? :-)

```suggestion
        hi_pri=False,
```

jalil-salame: Sure

ci/generate-workflows.py
```diff
@@ -953,23 +962,26 @@ def generate_actions(namer, jobber, trigger, is_continuous, is_tag=False):
...
+            elif "macos" in name:
+                trigger_paths += TRIGGER_PATHS_MAC
+            else:
+                trigger_paths += TRIGGER_PATHS_UNIX
+            if t.app_image:
+                trigger_paths += TRIGGER_PATHS_APPIMAGE
 
         trigger_paths = "- " + "\n      - ".join(yv(p) for p in sorted(trigger_paths))
         trigger_with_paths = trigger.replace("@PATHS@", trigger_paths)
+        on_workflow = "" if t.priorize or is_continuous or is_tag else f"  workflow_run:\n    types: [completed]\n    branches: [main]\n    workflows: {yv(PRIORITY_TARGETS)}\n"
```
wez, line 978: I'd be happy to have the prioritization stuff working for continuous as well

wez:
> Reading the documentation it seems you can't test it on a PR, see #3372 , ~I could try testing it in a fork though~. Scratch that, GH only allows one fork per project...

I'm willing to merge this so we can test with a followup PR/push and see if it worked out!

ci/generate-workflows.py
jalil-salame: I'll try that in a bit, the fix is more involved

jalil-salame:
You can test it by adding a comment to the ubuntu 22.04 yml, I'll fix all the review comments, hopefully today, if not tomorrow.

jalil-salame:
> I'm willing to merge this so we can test with a followup PR/push and see if it worked out!

I'm worried that it wont run on PRs, I'll draft it again until I've tested more edge cases, thanks GH.

jalil-salame:
Delayed runs don't fail the whole CI: https://github.com/jalil-salame/Test-GH-Ci/actions/runs/4526162470

ci/generate-workflows.py
jalil-salame: Reworked the `trigger` in order to implement it.

jalil-salame:
I think this approach is a bad idea, it doesn't show errors outside notifications:

- Test PR: https://github.com/jalil-salame/Test-GH-Ci/pull/1
- Test commit: https://github.com/jalil-salame/Test-GH-Ci/commit/e11b68957f259d587255a2bd7799b0ac107bf2eb

Both have failing workflows that are delayed by means of `workflow_run`. I think this needs to be reworked to not use `workflow_run`, I'll look into `workflow_call`...

PR #3369, opened 2023-03-26
TEST: Test #3367 by triggering the Ubuntu 22.04 workflow

jalil-salame:
- feat(ci): Delay less important targets
- chore: Regenerate workflows
- test: Trigger Ubuntu Workflow


jalil-salame:
Testing PR #3367 , will resume work on this tommorrow

jalil-salame:
It seems that GH triggers all workflows that are involved in a PR, that or the `workflow_run` doesn't work

PR #3370, opened 2023-03-26
hyperlink_rules.md: add link to the supported regex syntax

nbenitez:
I was using  `\c` in a regex and spent hours wondering why it was not working, the regex was working OK in a different program, and it was not producing any error from wezterm, it just didn't match as expected, until I found out the `\c` character class was not supported and instead I had to use the equivalent in ascii format `[:cntrl:]`

So add a link in the docs to the supported syntax so users have a way to avoid similar headaches  🙂 .


wez:
Thanks!

PR #3372, opened 2023-03-26
TEST workflow_run

jalil-salame:


jalil-salame:
@wez GH is telling me to test in production:

> Note: This event will only trigger a workflow run if the workflow file is on the default branch.

From https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run

wez:
I'm OK with that; the changes in the other PR look reasonable and I'm willing to give it a try. Worst case: we just revert it.

jalil-salame:
Tested it in my own Repo and the delay seems to be working: https://github.com/jalil-salame/Test-GH-Ci

wez:
Ah, it's interesting that the delayed action's results aren't visible on the originating commit:

image

How does a workflow failure in the delayed run manifest? Just in your notifications?

jalil-salame:
> Ah, it's interesting that the delayed action's results aren't visible on the originating commit:
> 
> image
> 
> How does a workflow failure in the delayed run manifest? Just in your notifications?
> 

I'll test a failing workflow

PR #3400, opened 2023-03-29
Add Homebrew tap for nightly build

plush-mesh:
Clarification: https://github.com/wez/wezterm/issues/3399#issuecomment-1489170090

Fixes https://github.com/wez/wezterm/issues/3399.


wez:
Thanks!

PR #3401, opened 2023-03-30
Fix minor typo

otezz:
As a new user,  I got confused as to why nothing happened when I created a config file named `wezterm.lua` when it should have been named `.wezterm.lua`


wez:
Thanks!

PR #3449, opened 2023-04-04
character select pgup/down, less jerky navigation, selected centered in list

digitallyserviced:
Figured this was simplest, to make the movements require a number, and just have a small number for arrows, big for pgup/down... Maybe later we can configure/optionify it.

### DIFF
--- a/wezterm-gui/src/termwindow/charselect.rs
+++ b/wezterm-gui/src/termwindow/charselect.rs
@@ -508,9 +508,9 @@ impl CharSelector {
         *self.top_row.borrow_mut() = 0;
     }
 
-    fn move_up(&self) {
+    fn move_up(&self, count: usize) {
         let mut row = self.selected_row.borrow_mut();
-        *row = row.saturating_sub(1);
+        *row = row.saturating_sub(count);
 
         let mut top_row = self.top_row.borrow_mut();
         if *row < *top_row {
@@ -518,7 +518,7 @@ impl CharSelector {
         }
     }
 
-    fn move_down(&self) {
+    fn move_down(&self, count:usize) {
         let max_rows_on_screen = *self.max_rows_on_screen.borrow();
         let limit = self
             .matches
@@ -526,12 +526,12 @@ impl CharSelector {
             .as_ref()
             .map(|m| m.matches.len())
             .unwrap_or_else(|| self.aliases.len())
-            .saturating_sub(1);
+            .saturating_sub(count);
         let mut row = self.selected_row.borrow_mut();
-        *row = row.saturating_add(1).min(limit);
+        *row = row.saturating_add(count).min(limit);
         let mut top_row = self.top_row.borrow_mut();
-        if *row + *top_row > max_rows_on_screen - 1 {
-            *top_row = row.saturating_sub(max_rows_on_screen - 1);
+        if *row + *top_row > max_rows_on_screen - count {
+            *top_row = row.saturating_sub(max_rows_on_screen - count);
         }
     }
 }
@@ -575,11 +575,17 @@ impl Modal for CharSelector {
                 self.selection.borrow_mut().clear();
                 self.updated_input();
             }
+            (KeyCode::PageUp, KeyModifiers::NONE) => {
+                self.move_up(6);
+            }
+            (KeyCode::PageDown, KeyModifiers::NONE) => {
+                self.move_down(6);
+            }
             (KeyCode::UpArrow, KeyModifiers::NONE) => {
-                self.move_up();
+                self.move_up(1);
             }
             (KeyCode::DownArrow, KeyModifiers::NONE) => {
-                self.move_down();
+                self.move_down(1);
             }
             (KeyCode::Char(c), KeyModifiers::NONE) | (KeyCode::Char(c), KeyModifiers::SHIFT) => {
                 // Type to add to the selection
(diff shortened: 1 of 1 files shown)


wezterm-gui/src/termwindow/charselect.rs
```diff
@@ -575,11 +575,17 @@ impl Modal for CharSelector {
                 self.selection.borrow_mut().clear();
                 self.updated_input();
             }
+            (KeyCode::PageUp, KeyModifiers::NONE) => {
+                self.move_up(6);
```
wez, line 579: we usually do a fully page at a time for these.  I'd suggest something like:

```suggestion
                self.move_up(*self.max_rows_on_screen.borrow());
```

digitallyserviced: I was thinking about that, but I seem to have a problem with the way the move_down handles the extra logic stuff... it would seem there is not much of a reason/indication that the list jumps/settles my selection at the bottom(but i know why it does in the code). THis is why i didnt want to play with that, and 6 seemed arbitrarily high enough jump.

digitallyserviced: [Peek 2023-04-04 13-36.webm](https://user-images.githubusercontent.com/1828125/229873600-73d63aac-c347-481b-ae63-3ce604f9783d.webm)

digitallyserviced:
[Peek 2023-04-04 13-36.webm](https://user-images.githubusercontent.com/1828125/229884766-9234ea5a-540c-463c-ac0d-cff0319b713a.webm)
I updated the up and down to be handled by a "wrap up" that both directions call, also keeps selection centered offset `max_rows/2`

wezterm-gui/src/termwindow/charselect.rs
```diff
@@ -508,32 +508,44 @@ impl CharSelector {
...
-        let mut row = self.selected_row.borrow_mut();
-        *row = row.saturating_add(1).min(limit);
-        let mut top_row = self.top_row.borrow_mut();
-        if *row + *top_row > max_rows_on_screen - 1 {
-            *top_row = row.saturating_sub(max_rows_on_screen - 1);
+            .unwrap_or_else(|| self.aliases.len());
+        {
+            let mut row = self.selected_row.borrow_mut();
+            let mut top_row = self.top_row.borrow_mut();
+            *row = row.min(limit-1);
```
wez, line 531:
```suggestion
            *row = row.min(limit.saturating_sub(1));
```

wezterm-gui/src/termwindow/charselect.rs
```diff
@@ -508,32 +508,44 @@ impl CharSelector {
...
+            if *row < *top_row {
+                *top_row = *row;
+            }
+            // *row = row.saturating_add(count).min(limit);
```
wez, line 535:
```suggestion
```

wezterm-gui/src/termwindow/charselect.rs
wez: Thanks for fixing up the jerky movement!

I think this page movement should either be a full page or a half page, rather than an arbitrary fixed number.  My expectation is for full page, but maybe full page minus 1 row would flow better?

wez (commented):
Don't forget to run `cargo +nightly fmt` to fixup code formatting!

digitallyserviced:
Yes, I agree now that I've had a little time with the new movement. Arbitrary is silly, but again I had gone with that because of the sooper weird movement jumps before, and I wanted to feel like i didnt "*overshoot*" ....

wezterm-gui/src/termwindow/charselect.rs
```diff
@@ -508,32 +508,54 @@ impl CharSelector {
...
     }
 
-    fn move_up(&self) {
-        let mut row = self.selected_row.borrow_mut();
-        *row = row.saturating_sub(1);
-
-        let mut top_row = self.top_row.borrow_mut();
-        if *row < *top_row {
-            *top_row = *row;
+    fn move_up(&self, count: usize) {
```
wez, line 511: Encoding special logic for `0` or `-1` is usually a good sign that it's time for an enum!

I'd suggest something like:

```rust
enum Move {
   Up(usize),
   Down(usize),
   PageUp,
   PageDown,
}
```

Then you can roll `move_up` and `move_down` into a single function:

```rust
fn move(&self, how: Move) {
   let page_size = *self.max_rows_on_screen.borrow();
   let current_row = *self.selected_row.borrow();
   let dest = match how {
       Move::Up(n) => current_row.saturating_sub(n),
       Move::PageUp => current_row.saturating_sub(page_size),
       Move::Down(n) => current_row.saturating_add(n),
       Move::PageDown => current_row.saturating_add(page_size),
   };
   *self.selected_row.borrow_mut() = dest;
   self.nav_selection();
}
```

digitallyserviced: lol, thats what i wanted to avoid... for now, also because i would probably roll up a few of your list/row selection gimmicks with the same enum/trait/interface.... thats how i at least handle stuff in Go... a single 

```go
func nav_selection(i int)*Object{}
```

that is used with +/- depending on direction... 

```
func moveUp(){nav_selection(-1)}
func moveDown(){nav_selection(1)}
```

wez:
Thanks!

Issue #3126, opened 2023-02-18
AMD  iGPU + WebGpu + Wayland renders only when mouse move over the window


Guara92:
### What Operating System(s) are you seeing this problem on?

Linux Wayland

### Which Wayland compositor or X11 Window manager(s) are you using?

Mutter

### WezTerm version

wezterm 20230217-120342-3b39aa55

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

I'm trying the `WebGpu` front end and is unusable on my system with wayland enabled, with wayland disabled runs really smooth.

```
> wezterm.gui.enumerate_gpus()
[
    {
        "backend": "Vulkan",
        "device": 5710,
        "device_type": "IntegratedGpu",
        "driver": "radv",
        "driver_info": "Mesa 23.1.0-devel",
        "name": "AMD Radeon Graphics (RADV GFX1036)",
        "vendor": 4098,
    },
    {
        "backend": "Vulkan",
        "device": 0,
        "device_type": "Cpu",
        "driver": "llvmpipe",
        "driver_info": "Mesa 23.1.0-devel (LLVM 15.0.7)",
        "name": "llvmpipe (LLVM 15.0.7, 256 bits)",
        "vendor": 65541,
    },
    {
        "backend": "Gl",
        "device": 0,
        "device_type": "Other",
        "name": "AMD Radeon Graphics (gfx1036, LLVM 15.0.7, DRM 3.49, 6.1.12-xm1.0e20221004.fc37.x86_64)",
        "vendor": 4098,
    },
]
```

### Configuration

```lua
enable_wayland = true,
front_end = "WebGpu",
```

wez:
Please share the the `OpenGL version` line from the debug overlay on your system

Guara92:
`OpenGL version: AMD Radeon Graphics (gfx1036, LLVM 15.0.7, DRM 3.49, 6.1.12-xm1.0e20221004.fc37.x86_64) 4.6 (Compatibility Profile) Mesa 23.1.0-devel`

colemickens:
I'm seeing the same, even when the debug overlay says its using WebGPU+Vulkan:

```
OpenGL version: WebGPU: name=AMD Radeon Graphics (RADV REMBRANDT), device_type=IntegratedGpu, backend=Vulkan, driver=radv, driver_info=Mesa 22.3.5, vendor=4098, device=5761
```

[...]

PR #3467, opened 2023-04-08
fix(wayland): only GL should wait for compositor

jokeyrhyme:
### Fixed

- only GL should wait for compositor, this should fix missing repaints when running with `front_end="WebGpu"` under wayland #3126


jokeyrhyme:
Okay, just dropped the experimental commits

jokeyrhyme:
Duplicate, already done https://github.com/wez/wezterm/commit/36d5307b801c4343065ad5e540ca420cc93d19e8

PR #3490, opened 2023-04-10
Update format-tab-title.md

asher-gh:


peng530:
The second example at line#106 also needs update.

asher-gh:
#3567

PR #3491, opened 2023-04-10
feat: resize pane from CLI

### LINKED ISSUE #3471: CLI subcommand to resize mux pane from command line
Ref: https://github.com/mrjones2014/smart-splits.nvim/pull/64

When only a single nvim window is open, we would like to be able to resize the mux window from within nvim.

Something like `wezterm cli resize-pane --pane-id [pane id] Left|Right|Up|Down [amount]` where `pane id` defaults to `$WEZTERM_PANE`.

There aren't really any I can think of.

**Additional context**
https://github.com/mrjones2014/smart-splits.nvim/pull/64

mrjones2014:
Fixes #3471


mrjones2014:
Got it working now -- @wez let me know if this looks like the right approach or not 🙏

mrjones2014:
@wez any chance you could take a look at this today or early next week? It should be a pretty quick review, hopefully 🙏

codec/src/lib.rs
```diff
@@ -29,6 +29,7 @@ use std::io::Cursor;
 use std::ops::Range;
 use std::path::PathBuf;
 use std::sync::Arc;
+use std::usize;
```
wez, line 32: This import is not needed; `usize` is always in scope by default.

```suggestion
```

codec/src/lib.rs
```diff
@@ -491,6 +492,7 @@ pdu! {
     EraseScrollbackRequest: 59,
     GetPaneDirection: 60,
     GetPaneDirectionResponse: 61,
+    AdjustPaneSize: 62,
```
wez, line 495: Please also increment the `CODEC_VERSION` which you can find in this file

wez (commented):
This looks right, thank you!

A couple of suggestions inline.

Please also add the docs for this. The process I'd suggest is to copy and adapt another page from the cli docs:

```console
$ cp docs/cli/cli/activate-pane-direction.md docs/cli/cli/adjust-pane-size.md
$ $EDITOR docs/cli/cli/adjust-pane-size.md
```

to edit the content appropriately.  Use `{{since('nightly')}}` instead of the date from that page.

The cli pages are set to include the help output from the command itself, so generate that via:

```console
$ cargo build
$ ./ci/update-derived-files.sh
```

That will also update the shell completions too.  Note that some other doc pages will likely appear in your git status; the next step will sort that out.

Then you can build and preview the docs; this also will re-format lua examples in the docs and check that the links are good:

```
$ ./ci/build-docs.sh serve
```

`serve` will run an http server and output the link to it once it has built the docs. It will monitor for changes and hot reload in your browser.  If you add new pages you will need to ctrl-c it and run it again to pick up the new structure.

You could also add a link to your new command to `docs/changelog.md` while you're in there!

mrjones2014:
Awesome, thanks! I think I've hit everything now.

docs/changelog.md
```diff
@@ -75,6 +75,7 @@ As features stabilize some brief notes about them will accumulate here.
 * [wezterm cli rename-workspace](cli/cli/rename-workspace.md). #2787
 * [wezterm.mux.rename_workspace](config/lua/wezterm.mux/rename_workspace.md). #2787
 * [wezterm cli get-pane-direction](cli/cli/get-pane-direction.md)
+* [wezterm cli adjust-pane-size](cli/cli/adjust-pane-size.md) #3471
```
wez, line 78: This is in the previous stable release's entries! Move it up to the first `New` section.
Hmm, it looks like you're not rebased on `main`, so I don't see the current entries here in this diff.

https://github.com/wez/wezterm/blob/main/docs/changelog.md?plain=1#L27

wez: When you rebase, be sure to re-run `./ci/update-derived-files.sh && ./ci/build-docs.sh`

mrjones2014: Ah, `./ci/build-docs.sh` is failing for me, which I think is causing some confusion. I rebased on `main`, and as far as I can tell, this is the `New` section under `Continuous/Nightly` which seems like the right place?

mrjones2014: Okay I think I've hit everything now. I had the [1Password Shell Plugin](https://github.com/1Password/shell-plugins) for `gh` enabled and it wasn't working within the script.

PR #3493, opened 2023-04-10
feat: add `nu.exe` to default `skip_close_confirmation_for_processes_named`

ErichDonGubler:
It's obvious that Nushell is intended to be skipped by default for close confirmation. However, on Windows, Nushell instances are named `nu.exe`, not `nu`, so they don't get skip-closed properly. Add `nu.exe` alongside `nu`, so Windows behaves as expected.


wez:
Thanks!

PR #3497, opened 2023-04-11
Do not wait indefinitely when being drip fed input.

### LINKED ISSUE #3466: WezTerm laggy when under CPU load and terminal output

Linux X11

xmonad 0.17.1

### WezTerm version

wezterm 20230312-204801-4afedd62

### Describe the bug

I have some embedded devices with serial ports connected to my workstation, which I monitor by running [tio](https://github.com/tio/tio) in a WezTerm terminal window. At the same time I frequently run CPU intensive build jobs.

I found that if one of the embedded devices begins sending debug logs to the serial port at the maximum speed (115200bps) while one of the build jobs is running, that causes all of my WezTerm terminal windows to freeze up for seconds at a time. Looking at `top` in another terminal emulator, I see that the `wezterm-gui` process is running at close to 100% CPU while this happens. My guess is that WezTerm becomes starved for CPU on account of the build jobs and cannot keep up with the output from the serial port.

I tried running the latest nightly but I ran into an unrelated issue that prevented wezterm from starting. I will update the bug if I manage to reproduce with the latest version.

### To Reproduce

1. Start a CPU intensive build job, e.g. build LLVM, Chromium or AOSP.
2. Run the following command in a terminal window to simulate a debug log:
```
sh -c 'while true; do date; sleep 0.001; done'
```
3. Wait a few seconds.
4. Observe that the date printed to the terminal window starts updating less frequently.

### Configuration

```
local act = wezterm.action

local keys = nil
if wezterm.gui then
  keys = wezterm.gui.default_keys()
  table.insert(
    keys,
    { key = 'Home', mods = 'SHIFT', action = act.ScrollToTop })
  table.insert(
    keys,
    { key = 'End', mods = 'SHIFT', action = act.ScrollToBottom })
end

return {
        scrollback_lines = 10000000,
        enable_scroll_bar = true,
        window_padding = {
    left = 0,
    right = 0,
    top = 0,
    bottom = 0,
  },
  keys = keys,
  font_size = 11.5,
}
```

### Expected Behavior

No lag in (at least) other terminal windows.

### Logs

_No response_

### Anything else?

_No response_

pcc:
If the client program is drip feeding us input at a rate faster than 1 character per mux_output_parser_coalesce_delay_ms (e.g. it is printing output read from a serial port faster than this), we would previously wait indefinitely for new input, because we would always poll with a new timeout [...]delay_ms in the future. In the worst case, this would create a busy loop that starves the other panes of an opportunity to read from their ptys, making all of them unresponsive. Address this by making a note of when the first character was read, and do not wait longer than [...]delay_ms after this time.

Fixes #3466.

### DIFF
--- a/mux/src/lib.rs
+++ b/mux/src/lib.rs
@@ -141,7 +141,8 @@ fn parse_buffered_data(pane: Weak<dyn Pane>, dead: &Arc<AtomicBool>, mut rx: Fil
     let mut actions = vec![];
     let mut hold = false;
     let mut action_size = 0;
-    let mut delay_ms = configuration().mux_output_parser_coalesce_delay_ms;
+    let mut delay = Duration::from_millis(configuration().mux_output_parser_coalesce_delay_ms);
+    let mut deadline = None;
 
     loop {
         match rx.read(&mut buf) {
@@ -194,13 +195,20 @@ fn parse_buffered_data(pane: Weak<dyn Pane>, dead: &Arc<AtomicBool>, mut rx: Fil
                     // that we coalesce a full "frame" from an unoptimized
                     // TUI program
                     if action_size < buf.len() {
-                        if delay_ms > 0 {
+                        let poll_delay = match deadline {
+                            None => {
+                               deadline.replace(Instant::now() + delay);
+                               Some(delay)
+                            },
+                            Some(target) => target.checked_duration_since(Instant::now()),
+                        };
+                        if poll_delay.is_some() {
                             let mut pfd = [pollfd {
                                 fd: rx.as_socket_descriptor(),
                                 events: POLLIN,
                                 revents: 0,
                             }];
-                            if let Ok(1) = poll(&mut pfd, Some(Duration::from_millis(delay_ms))) {
+                            if let Ok(1) = poll(&mut pfd, poll_delay) {
                                 // We can read now without blocking, so accumulate
                                 // more data into actions
                                 continue;
@@ -212,12 +220,13 @@ fn parse_buffered_data(pane: Weak<dyn Pane>, dead: &Arc<AtomicBool>, mut rx: Fil
                     }
 
                     send_actions_to_mux(&pane, &dead, std::mem::take(&mut actions));
+                    deadline = None;
                     action_size = 0;
                 }
 
                 let config = configuration();
                 buf.resize(config.mux_output_parser_buffer_size, 0);
-                delay_ms = config.mux_output_parser_coalesce_delay_ms;
+                delay = Duration::from_millis(config.mux_output_parser_coalesce_delay_ms);
             }
         }
     }


mux/src/lib.rs
```diff
@@ -194,13 +195,21 @@ fn parse_buffered_data(pane: Weak<dyn Pane>, dead: &Arc<AtomicBool>, mut rx: Fil
...
-                        if delay_ms > 0 {
+                        let next_update = match next_update {
+                            None => {
+                               next_update = Some(Instant::now() + delay);
+                               next_update.unwrap()
+                            },
+                            Some(next_update) => next_update,
+                        };
+                        let delay = next_update - Instant::now();
+                        if !delay.is_zero() {
```
wez, line 206: We need to avoid the `unwrap`. Also, Duration/Instant math can panic in some overflow situations, so I'd suggest using `checked_sub` to avoid that potential panic:

```suggestion
                        let poll_delay = match &deadline {
                            None => {
                               deadline.replace(Instant::now() + delay);
                               delay
                            },
                            Some(target) => target.checked_sub(Instant::now()).unwrap_or(Duration::ZERO)
                        };
                        if !poll_delay.is_zero() {                        
```

mux/src/lib.rs
```diff
@@ -194,13 +195,21 @@ fn parse_buffered_data(pane: Weak<dyn Pane>, dead: &Arc<AtomicBool>, mut rx: Fil
...
                             let mut pfd = [pollfd {
                                 fd: rx.as_socket_descriptor(),
                                 events: POLLIN,
                                 revents: 0,
                             }];
-                            if let Ok(1) = poll(&mut pfd, Some(Duration::from_millis(delay_ms))) {
+                            if let Ok(1) = poll(&mut pfd, Some(delay)) {
```
wez, line 212:
```suggestion
                            if let Ok(1) = poll(&mut pfd, Some(poll_delay)) {
```

mux/src/lib.rs
```diff
@@ -212,12 +221,13 @@ fn parse_buffered_data(pane: Weak<dyn Pane>, dead: &Arc<AtomicBool>, mut rx: Fil
                     }
 
                     send_actions_to_mux(&pane, &dead, std::mem::take(&mut actions));
+                    next_update = None;
```
wez, line 224:
```suggestion
                    deadline = None;
```

wez (commented):
Thanks! Good find!

The approach seems fine, just some stylistic suggestions:
* Avoid shadowing some of the local variables, which aids in clarity
* Avoid `unwrap`
* Defend against panicking for overflowing subtraction when doing Instant/Duration math

mux/src/lib.rs
```diff
@@ -141,7 +141,8 @@ fn parse_buffered_data(pane: Weak<dyn Pane>, dead: &Arc<AtomicBool>, mut rx: Fil
     let mut actions = vec![];
     let mut hold = false;
     let mut action_size = 0;
-    let mut delay_ms = configuration().mux_output_parser_coalesce_delay_ms;
+    let mut delay = Duration::from_millis(configuration().mux_output_parser_coalesce_delay_ms);
+    let mut next_update: Option<Instant> = None;
```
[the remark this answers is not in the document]
pcc: Done, also removed the type because it can be inferred (I'm still too used to C++ which can only infer from the initializer).

mux/src/lib.rs
pcc: Yes, that's what I was going for initially, but `checked_sub` takes a `Duration`, not an `Instant`. I looked again though and found the function `checked_duration_since` which does what we want.

It was a bit more concise to use `None` to represent an invalid `poll_delay`, rather than using zero, so that's what I went for in this change, but let me know if you prefer the other way.

wez:
Thank you!
I suspect that https://github.com/wez/wezterm/issues/837 is essentially the same issue; I was never able to reproduce it, but I think the symptoms sound similar to what you described

PR #3528, opened 2023-04-15
feat: add systembackdrop like acrylic and mica for windows os

### LINKED ISSUE #1614: Feature Request: Ability to blur non-opaque windows
No, purely new aesthetic feature.

Ability to toggle on and off a customizable blur on windows with some amount of opacity (i.e != 1.0).

The ability to enable a blur via a toggle would be good, but the ability to do so and also specify the radius or "amount" of blur would be awesome.

* [iTerm2](https://iterm2.com/documentation-preferences-profiles-window.html) (which I switched off of to wezterm a few months back, and have enjoyed the switch, this feature being the only main thing I've noticed that I used to use daily that isn't present):
    > Blur
    > If selected, the window background is blurred provided the background has some transparency. Selecting a large radius will blur the background more, but (especially on Retina displays) comes with a performance penalty.
* [alacritty](https://github.com/alacritty/alacritty/issues/972)
* mac native Terminal application

I tried looking through the docs/source to make sure I just wasn't missing it, and the only thing I seemed to notice was [this old previous wezterm issue](https://github.com/wez/wezterm/issues/626) which addresses transparency, but not blur, which is what this is requesting.

**Additional context**
* macOS operating system

* version:
```wezterm --version
wezterm 20220126-105230-0c6b4e26
```
* Not really relevant, but currently toggle transparency like so:

```

wezterm.on("toggle-opacity", function(window, pane)
  local overrides = window:get_config_overrides() or {}
  if not overrides.window_background_opacity then
    overrides.window_background_opacity = 0.8;
  else
    overrides.window_background_opacity = nil
  end
  window:set_config_overrides(overrides)
end)
```

kingavatar:
Added functionality which enables Setting SystemBackdrop for windows os like mica, Acrylic, tabbed. 

based on [SystemBackDrop](https://learn.microsoft.com/en-us/windows/win32/api/dwmapi/ne-dwmapi-dwm_systembackdrop_type) and [Vibe](https://github.com/pykeio/vibe/blob/main/src/dwm_win32.rs)

Adds a config option `win32_system_backdrop` with values `Auto`, `Disable`, `Acrylic`, `Mica`, `Tabbed` to configure the backdrop.
Also another config option `win32_acrylic_accent_color` for adding accent color to acrylic for windows build version less than `22621` 
resolves #1614

Screenshots

Acrylic:

acrylic

Mica:

mica

Tabbed(Mica Alt):

tabbed

### DIFF
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -22,9 +22,10 @@ use crate::unix::UnixDomain;
 use crate::wsl::WslDomain;
 use crate::{
     default_config_with_overrides_applied, default_one_point_oh, default_one_point_oh_f64,
-    default_true, GpuInfo, IntegratedTitleButtonColor, KeyMapPreference, LoadedConfig,
-    MouseEventTriggerMods, RgbaColor, SerialDomain, SystemBackdrop, WebGpuPowerPreference,
-    CONFIG_DIRS, CONFIG_FILE_OVERRIDE, CONFIG_OVERRIDES, CONFIG_SKIP, HOME_DIR,
+    default_true, default_win32_acrylic_accent_color, GpuInfo, IntegratedTitleButtonColor,
+    KeyMapPreference, LoadedConfig, MouseEventTriggerMods, RgbaColor, SerialDomain, SystemBackdrop,
+    WebGpuPowerPreference, CONFIG_DIRS, CONFIG_FILE_OVERRIDE, CONFIG_OVERRIDES, CONFIG_SKIP,
+    HOME_DIR,
 };
 use anyhow::Context;
 use luahelper::impl_lua_conversion_dynamic;
@@ -512,7 +513,10 @@ pub struct Config {
 
     /// Only works on Windows
     #[dynamic(default)]
-    pub windows_system_backdrop: SystemBackdrop,
+    pub win32_system_backdrop: SystemBackdrop,
+
+    #[dynamic(default = "default_win32_acrylic_accent_color")]
+    pub win32_acrylic_accent_color: RgbaColor,
 
     /// Specifies the alpha value to use when rendering the background
     /// of the window.  The background is taken either from the
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -23,8 +23,8 @@ use crate::wsl::WslDomain;
 use crate::{
     default_config_with_overrides_applied, default_one_point_oh, default_one_point_oh_f64,
     default_true, GpuInfo, IntegratedTitleButtonColor, KeyMapPreference, LoadedConfig,
-    MouseEventTriggerMods, RgbaColor, SerialDomain, WebGpuPowerPreference, CONFIG_DIRS,
-    CONFIG_FILE_OVERRIDE, CONFIG_OVERRIDES, CONFIG_SKIP, HOME_DIR,
+    MouseEventTriggerMods, RgbaColor, SerialDomain, SystemBackdrop, WebGpuPowerPreference,
+    CONFIG_DIRS, CONFIG_FILE_OVERRIDE, CONFIG_OVERRIDES, CONFIG_SKIP, HOME_DIR,
 };
 use anyhow::Context;
 use luahelper::impl_lua_conversion_dynamic;
@@ -509,6 +509,10 @@ pub struct Config {
     /// Only works on MacOS
     #[dynamic(default)]
     pub macos_window_background_blur: i64,
+
+    /// Only works on Windows
+    #[dynamic(default)]
+    pub windows_system_backdrop: SystemBackdrop,
 
     /// Specifies the alpha value to use when rendering the background
     /// of the window.  The background is taken either from the
--- a/window/src/os/windows/window.rs
+++ b/window/src/os/windows/window.rs
@@ -9,7 +9,7 @@ use crate::{
 };
 use anyhow::{bail, Context};
 use async_trait::async_trait;
-use config::{ConfigHandle, ImePreeditRendering};
+use config::{ConfigHandle, ImePreeditRendering, SystemBackdrop};
 use lazy_static::lazy_static;
 use promise::Future;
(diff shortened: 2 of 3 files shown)


window/src/os/windows/window.rs
```diff
@@ -1327,6 +1373,55 @@ fn apply_theme(hwnd: HWND) -> Option<LRESULT> {
...
+                SystemBackdrop::Tabbed => DWM_SYSTEMBACKDROP_TYPE::DWMSBT_TABBEDWINDOW,
+            };
+
+            // Apply Acrylic or Mica Backdrop
+            if *IS_WIN11_22H2 {
+                DwmSetWindowAttribute(
+                    hwnd,
+                    DWMWA_SYSTEMBACKDROP_TYPE,
+                    &pv_attribute as *const _ as _,
+                    4,
```
wez, line 1392: Can we get the compiler to compute the size for us?

```suggestion
                    std::mem::size_of_val(pv_attribute),
```

window/src/os/windows/window.rs
```diff
@@ -1327,6 +1373,55 @@ fn apply_theme(hwnd: HWND) -> Option<LRESULT> {
...
+                            cbData: std::mem::size_of_val(&policy),
+                        },
+                    );
+                };
+            } else if !*IS_WIN10 && !*IS_WIN11_22H2 {
+                // For build versions less than 22h2 but are
+                // still win11
+                if inner.config.windows_system_backdrop == SystemBackdrop::Mica {
+                    DwmSetWindowAttribute(hwnd, DWMWA_MICA_EFFECT, &1 as *const _ as _, 4);
+                }
```
wez, line 1422: I think we need something like this to handle the case where the user changes the config at runtime to disable mica.  Without this, I think that once mica is enabled, it will stick forever?

```suggestion
                let mica_enabled: u32 = if inner.config.windows_system_backdrop == SystemBackdrop::Mica { 1 } else { 0};
                 DwmSetWindowAttribute(hwnd, DWMWA_MICA_EFFECT, &mica_enbabled as *const _ as _, std::mem::size_of_val(mica_enabled));
```

window/src/os/windows/window.rs
```diff
@@ -1327,6 +1373,55 @@ fn apply_theme(hwnd: HWND) -> Option<LRESULT> {
...
+            // Apply Acrylic or Mica Backdrop
+            if *IS_WIN11_22H2 {
+                DwmSetWindowAttribute(
+                    hwnd,
+                    DWMWA_SYSTEMBACKDROP_TYPE,
+                    &pv_attribute as *const _ as _,
+                    4,
+                );
+            } else if inner.config.windows_system_backdrop == SystemBackdrop::Acrylic {
+                let colour: [u8; 4] = [40, 40, 40, 1]; // acrylic doesn't like to have 0 alpha
```
wez, line 1395: We should probably make this color configurable by adding a `win32_acrylic_accent_color` option (or other better name)

config/src/config.rs
```diff
@@ -510,6 +510,10 @@ pub struct Config {
     #[dynamic(default)]
     pub macos_window_background_blur: i64,
 
+    /// Only works on Windows
+    #[dynamic(default)]
+    pub windows_system_backdrop: SystemBackdrop,
```
wez, line 515: I'd lean towards calling this win32 rather than windows to disambiguate from generic non-OS windows options.

```suggestion
    pub win32_system_backdrop: SystemBackdrop,
```

wez (commented):
Thanks for this!
It looks good!

A couple of minor comments on the code!

wez:
Would you be OK with using your screenshots in the documentation for this?

kingavatar:
> Would you be OK with using your screenshots in the documentation for this?

Sure, I am okay with using the screenshots.

kingavatar:
Hi @wez, I have pushed the changes.

wez:
On my nvidia system, neither Mica nor Tabbed seem to have any effect. Do you have thoughts on why that is?

wez:
I'm on `Microsoft Windows [Version 10.0.22621.1555]`

kingavatar:
> On my nvidia system, neither Mica nor Tabbed seem to have any effect. Do you have thoughts on why that is?

Mica is supported on all versions 22000 or later. I am also on a Nvidia System and it is working for me 😅

Mica :
image

This is the config I have used, the only additional thing I have set is opacity, regardless title bar should change when backdrop is changed.
image

kingavatar:
@wez ,I have been reading the documentation and noticed that the same API method is being used to achieve the backdrop effect in both this [github issue](https://github.com/microsoft/windows-rs/issues/2189) and in [Vibe](https://github.com/pykeio/vibe/blob/main/src/dwm_win32.rs). 
However, an additional method to extend the margin is called in both cases. Based on this information, do you think the issue can be resolved using this approach?
```rust
	use winapi::um::dwmapi::DwmExtendFrameIntoClientArea;
        use winapi::um::uxtheme::MARGINS;
        DwmExtendFrameIntoClientArea(
            hwnd,
            &MARGINS {
                cxLeftWidth: -1,
                cxRightWidth: -1,
                cyTopHeight: -1,
                cyBottomHeight: -1,
            },
        );
```
According to the [documentation](https://learn.microsoft.com/en-us/windows/win32/api/dwmapi/nf-dwmapi-dwmextendframeintoclientarea) calling this function should provide a sheet of glass effect. I have pushed a new commit with the above change. Can you please verify if this resolves the issue? I have not observed any behavior changes on my end after calling this function.

wez:
I see; you need opacity = 0 for Mica or Tabbed to visibly have an effect, and you need a wallpaper that has contrasting regions in order to see that effect.

In addition, I had to turn off "Show accent color on title bars and windows borders" in the windows personalization settings to notice the effect of  `DwmExtendFrameIntoClientArea`.

wez:
I manually merged this.

Thank you!

PR #3547, opened 2023-04-17
Add nix flake

happenslol:
This PR adds a nix flake to the repository. I was looking to contribute to this project, and since I'm mainly using NixOS at the moment, that required me to at least set up a devshell to pull in the dependencies for this project. I generally find it easier to just set up a flake, since that also allows me to build the project and integrate it into my system configuration.

Adding a flake would have a few advantages, even for non-NixOS users:

* Being able to jump into a devshell with all required dependencies installed
* Being able to run the app just with `nix run github:wez/wezterm`
* For NixOS users, being able to depend on specific versions of the app more easily

The setup here mainly follows the packaged version of wezterm in nixpkgs, especially when it comes to the build. The flake itself is very minimal.

Please tell me if this is something you would consider including. I can understand if you don't want to have this be a part of the project, but I figured I would just open a PR since I already created the flake for developing myself :-)


wez:
Thanks for diving in!

I'm sorry to say that I don't want to add this to the repo: it appears to be mostly boilerplate from a system that I don't use and can't maintain, and that will drift out of date very quickly without someone actively maintaining it.

I would be more amenable to this if it wasn't adding more stuff at the top level of the repo, and if there was some kind of automation to validate that it continues to work and remain useful.  However, without someone who knows and uses nix to monitor and maintain it, I would be inclined to delete it if it does fall into disrepair as I don't have the bandwidth to learn and follow nix.

happenslol:
No worries, that's completely fair. While it would be possible to add scripts that automate some of the maintenance work, someone would of course need to maintain those scripts. So yeah, that doesn't really address the core issue.

Just a few notes to address the points you made:

* I would be open to maintaining this, but I can also understand that you would have to trust some random guy™ from github for this. I'm planning on working on a few other open issues and features for this project, so maybe I'll come back around to this once I've done that ;-)
* It's possible to stash away the flake configuration in a subdirectory, for example, neovim uses [a contrib directory](https://github.com/neovim/neovim/tree/master/contrib) for this.
* It's of course possible to add CI steps that ensure the flake still builds, but I can imagine that you would want to avoid CI blocking changes because of a build system foreign to you not agreeing with them.

Thanks for the response, and feel free to close this. I'll keep depending on my fork for now then.

gabyx:
@happenslol  This is very valuable! Do you intend to update this PR?

- Would actually be a Dockerfile which boots NixOS and builds this flake, be enough to put it into the CI to verify that it builds?

Emiller88:
I don't think there's anything keeping the flake from living in a different repo. Just use a `fetchGit`. 

Might be confusing if it's `happenslol/wezterm-flake` instead of just `wez/wezterm` but oh well 🤷🏻.

wez:
To clarify: I'm ok adding the flake here if we can move the boilerplate out of the top directory, and if there is an accompanying workflow that verifies that it continues to work.

It seems like between the folks here and in #4286 there are enough interested and capable folks around to troubleshoot this.

While I don't want anyone to feel like they are permanently committed to maintaining this, I would appreciate having a short list of 2-3 people that are willing to troubleshoot failures in the flake CI.

happenslol:
Good to know! I'll have time to update this and move it to a subfolder over the coming days. I'll probably just copy the approach that neovim is taking, I remember them having the flake in a subfolder as well.

I still think that it's important that flake CI does not block any merges, however.

wez:
Yeah, I'd like to have a separate GH actions workflow for this. It needs to fail if the flake stops working, otherwise the failure will be invisible.  There's no automation that depends on the status of the workflow, it's just me looking at it.

gabyx:
@happenslol : we should make sure we setup the shell such that we use the correct rust toolchain. I have the assume thats nightly because the rustftm tool only runs with nigthly toolchain.

Normally I use a file `rust-toolchain.toml` in the root:
```toml
[toolchain]
channel = "nightly"
components = [ "rustfmt", "rust-analyzer", "miri", "rust-docs", "clippy" ]
profile = "default"
```

and then use the following inside `flake.nix`:

```nix
  inputs = {
    # Nixpkgs
    nixpkgs.url = "github:nixos/nixpkgs/nixos-unstable";

    # You can access packages and modules from different nixpkgs revs
    # at the same time. Here's an working example:
    nixpkgsStable.url = "github:nixos/nixpkgs/nixos-23.05";
    # Also see the 'stable-packages' overlay at 'overlays/default.nix'.

    flake-utils.url = "github:numtide/flake-utils";
    rust-overlay = {
      url = "github:oxalica/rust-overlay";
      inputs = {
        nixpkgs.follows = "nixpkgs";
        flake-utils.follows = "flake-utils";
      };
    };
  };

  outputs = {
    self,
    nixpkgs,
    nixpkgsStable,
    flake-utils,
    rust-overlay,
    ...
  } @ inputs:
    flake-utils.lib.eachDefaultSystem
    # Creates an attribute map `{ devShells.<system>.default = ...}`
    # by calling this function:
    (
      system: let
        overlays = [(import rust-overlay)];

        # Import nixpkgs and load it into pkgs.
        pkgs = import nixpkgs {
          inherit system overlays;
        };

        rustToolchain = pkgs.pkgsBuildHost.rust-bin.fromRustupToolchainFile ./rust-toolchain.toml;

        # Things needed only at compile-time.
        nativeBuildInputs = with pkgs; [
          rustToolchain
          pkg-config
        ];

        # Things needed at runtime.
        buildInputs = with pkgs; [openssl llvmPackages_15.lldb];
      in
        with pkgs; {
          devShells.default = mkShell {
            inherit buildInputs nativeBuildInputs;
          };
        }
    );
}
```

@wez  what rust toolchain should one use to develop in this project?

gabyx:
@happenslol I updated the file above to reflect wez answer.

happenslol:
Alright, I've updated the branch to the latest upstream revision, and moved the flake to a subdirectory. I've also added the rust-overlay as you suggested in your code snippet, and set the versions for the toolchain.

I'm not too experienced with Github Actions, and I'm currently pretty busy, so I'm not sure if it makes sense for me to take care of that part as well. @gabyx: Do you think that's something you could work on? The flake builds as-is on my machine. You could open a PR for my branch, or I can give you collaborator permissions? Whatever you prefer.

gabyx:
@Emiller88 Happy if you could contribute in another PR. I am not so aware of the nix installer thing, I would probably have done it less proper in the sense that I would have uses a NixOs image ans just ran nix build… but thats probably not what we want. Better to go with the specific Github action to install it into ubuntu etc… I could tackle this with your help in a new PR if you like. Maybe you are just faster then me :)

gabyx:
Should we try to run a nix build on two systems for now? MacOs and Ubuntu?

Emiller88:
> Maybe you are just faster then me :)

Nah, just been through more pain than you 😆 

> if you're confident that just doing ubuntu will be high enough signal, that is "cheaper"

Agreed. There's going to be a pretty small number of people in the middle of the venn diagram of MacOS, Wezterm, and using Nix.

Emiller88:
CI is ready to go 👍🏻 https://github.com/Emiller88/wezterm/actions/runs/7122203276/job/19392833641?pr=1

futile:
:+1: from someone also wanting to take a stab at contributing, and who runs NixOS :) Super happy to see this, and looks good!

Smallest nit: I like replacing `forEachDefaultSystem` with an explicit list of `system`s, because that makes it much more obvious to the reader which systems are supported :) But really just a small detail!

nekowinston:
Just saw this PR, for anyone waiting for a nightly flake that offers a cache, I've been updating wezterm-nightly on my Nix User Repository https://github.com/nekowinston/nur for the last 10 months; it auto-updates daily, and caches are pushed to my cachix.

nix/flake.nix
```diff
@@ -0,0 +1,148 @@
...
+          xorg.xcbutilkeysyms
+          xorg.xcbutilwm # contains xcb-ewmh among others
+        ]
+        ++ lib.optionals stdenv.isDarwin [
+          Cocoa
+          CoreGraphics
+          Foundation
+          libiconv
+          UserNotifications
+        ];
```
smsegal, line 91: I tried building this flake on macOS - it doesn't work as it seems the `Cocoa`, `CoreGraphics` packages can't be found. 
In my own version of the flake that builds on macOS I have to provide them to `buildInputs` like so: 

```suggestion
 buildInputs = with pkgs;
        [
          fontconfig
          pkgs.zlib
        ]
        ++ lib.optionals stdenv.isLinux [
          xorg.libX11
          xorg.libxcb
          libxkbcommon
          openssl
          wayland
          xorg.xcbutil
          xorg.xcbutilimage
          xorg.xcbutilkeysyms
          xorg.xcbutilwm # contains xcb-ewmh among others
        ]
        ++ lib.optionals stdenv.isDarwin (with pkgs.darwin.apple_sdk.frameworks; [
          Cocoa
          CoreGraphics
          Foundation
          libiconv
          UserNotifications
        ]);
```

Note that I'm far from a nix expert, so I could easily be doing something wrong here.

happenslol: I haven't tried this on MacOS, just used the code from the previous packaged wezterm derivation from nixpkgs. Is the diff you posted working without any errors or warnings for you?

nekowinston: @happenslol yeah you need the additional `with pkgs.darwin.apple_sdk.frameworks;`, since nixpkgs passes them in at the top-level nix: https://github.com/NixOS/nixpkgs/blob/486f5597f2269680a9538bfd60922489ea5dffd9/pkgs/top-level/all-packages.nix#L3128-L3130

Instead of using `with`, it's better practice to explicitly `inherit` them:
```nix
let
  inherit (pkgs.darwin.apple_sdk.frameworks) Cocoa CoreGraphics Foundation UserNotifications;
in
```
Notice that `libiconv` isn't part of `pkgs.darwin.apple_sdk.frameworks`, but `pkgs`.

smsegal: I'm always trying to improve my nix - why is it considered better practice to inherit them instead of using `with`? A case of explicit is always better than implicit?

nekowinston: nix.dev has a best practice section about that: https://nix.dev/guides/best-practices#with-scopes
Essentially, yeah, it's about being explicit.

smsegal: thanks!

nix/flake.nix
```diff
@@ -0,0 +1,148 @@
...
+
+      libPath = lib.makeLibraryPath (with pkgs; [libGL vulkan-loader]);
+
+      rustPlatform = pkgs.makeRustPlatform {
+        cargo = pkgs.rust-bin.stable.latest.minimal;
+        rustc = pkgs.rust-bin.stable.latest.minimal;
+      };
+    in {
+      packages.default = rustPlatform.buildRustPackage rec {
+        inherit buildInputs nativeBuildInputs;
```
ThinkChaos, line 101: I think you're better of inlining these and using `inputsFrom = self.packages.${system}.default` in the devShell.

nix/flake.nix
```diff
@@ -0,0 +1,148 @@
...
+          installShellFiles
+          ncurses # tic for terminfo
+          pkg-config
+          python3
+
+          rust-bin.stable.latest.minimal
+          rust-bin.stable.latest.clippy
+
+          rust-bin.nightly.latest.rustfmt
+          rust-bin.nightly.latest.rust-analyzer
```
ThinkChaos, line 65: These (or at least most of these) shouldn't be needed for the package, only the dev shell. Inlining the build inputs in  the package and shell also has the benefit of making you think about what's really needed.

nix/flake.nix
```diff
@@ -0,0 +1,148 @@
...
+    self,
+    flake-utils,
+    rust-overlay,
+    nixpkgs,
+    freetype2,
+    harfbuzz,
+    libpng,
+    zlib,
+    ...
+  }:
```
ThinkChaos, line 46: This is mostly a style preference, but mentioning in case you don't know the pattern.  
You can use `{ self, ... }@inputs` so you don't need to spell out everything here, and it makes what's a flake input obvious below.

nix/flake.nix
```diff
@@ -0,0 +1,148 @@
...
+        ];
+
+      libPath = lib.makeLibraryPath (with pkgs; [libGL vulkan-loader]);
+
+      rustPlatform = pkgs.makeRustPlatform {
+        cargo = pkgs.rust-bin.stable.latest.minimal;
+        rustc = pkgs.rust-bin.stable.latest.minimal;
+      };
+    in {
+      packages.default = rustPlatform.buildRustPackage rec {
```
ThinkChaos, line 100: Putting the package definition in its own file is generally preferred, especially when you have a `nix` dir :)  
Then you can use `pkgs.callPackage`, which will also avoid using `with` and allow downstream users to override inputs easily.  
See [Nix Pills Chapter 13. Callpackage Design Pattern](https://nixos.org/guides/nix-pills/callpackage-design-pattern.html)

nix/flake.nix
```diff
@@ -0,0 +1,148 @@
...
+            $out/bin/wezterm-gui
+        '';
+      };
+
+      devShell = pkgs.mkShell {
+        name = "wezterm-shell";
+        inherit buildInputs nativeBuildInputs;
+
+        LD_LIBRARY_PATH = libPath;
+      };
```
ThinkChaos, line 146: Enabling backtraces is useful:
```suggestion
        RUST_BACKTRACE = 1;
      };
```

nix/flake.nix
```diff
@@ -0,0 +1,148 @@
...
+        cargo = pkgs.rust-bin.stable.latest.minimal;
+        rustc = pkgs.rust-bin.stable.latest.minimal;
+      };
+    in {
+      packages.default = rustPlatform.buildRustPackage rec {
+        inherit buildInputs nativeBuildInputs;
+
+        name = "wezterm";
+        src = ./..;
+        PKG_CONFIG_PATH = "${pkgs.openssl.dev}/lib/pkgconfig";
```
ThinkChaos, line 105: I don't think this is needed. At least in my Rust projects that depend on openssl, just having it in `buildInputs` is enough.

nix/flake.nix
```diff
@@ -0,0 +1,148 @@
...
+            --add-needed "${pkgs.vulkan-loader}/lib/libvulkan.so.1" \
+            $out/bin/wezterm-gui
+        '';
+      };
+
+      devShell = pkgs.mkShell {
+        name = "wezterm-shell";
+        inherit buildInputs nativeBuildInputs;
+
+        LD_LIBRARY_PATH = libPath;
```
ThinkChaos, line 145: Not convinced this is needed either. Having the libs in `buildInputs` is probably enough.

ThinkChaos (commented):
Thanks for working on this!  
I put a couple comments, don't take anything as a requirement, I'm just trying to give tips on stuff that stuck out to me during a quick read :)

happenslol:
> Thanks for working on this! I put a couple comments, don't take anything as a requirement, I'm just trying to give tips on stuff that stuck out to me during a quick read :)

Heya, thank you so much for your comments! I'm not super experienced on packaging in nix, so this is very appreciated. It might however delay getting this into a mergeable state. I hope that I'll have time to bring this up to speed over the coming few days, but I can't make any promises.

nix/flake.nix
```diff
@@ -0,0 +1,148 @@
...
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
+    flake-utils.url = "github:numtide/flake-utils";
+    rust-overlay = {
+      url = "github:oxalica/rust-overlay";
+      inputs = {
+        nixpkgs.follows = "nixpkgs";
+        flake-utils.follows = "flake-utils";
+      };
+    };
```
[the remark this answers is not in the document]
nekowinston: I'd advise keeping rust-overlay since the MSRV of wezterm can shift out of sync with the nixpkgs version. This has happened twice over the last 10 months on my NUR.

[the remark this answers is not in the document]
ThinkChaos: Good point, then I agree it seems worth it in this case :)

ThinkChaos:
Yeah no worries, just thought I'd give a couple tips cause learning this stuff takes time if you don't have anyone to show you around :)

gabyx:
@ThinkChaos Nice inputs! Reviews are always appreciated especially from more experienced folks! Thanks for taking the time!

anthr76:
This is great work! I plan on testing this on the branch until it's merged

ciarand:
I also just switched to this and it's great! One note: the nixpkgs version has the `terminfo` available as an installable derivation; it would be nice if this had that also. It looks relatively simple based on the nixpkgs version: https://github.com/NixOS/nixpkgs/blob/3dc440faeee9e889fe2d1b4d25ad0f430d449356/pkgs/applications/terminal-emulators/wezterm/default.nix#L130-L136

happenslol:
Just as a heads up to anyone following this: I'm currently very busy and won't get to this until the end of the month, sadly. Just so we're clear on what's left to do here: I'll incorporate the review comments and fix the build on mac (although someone else will have to test that), and that should be it for the flake part itself.

@wez Do you want me to incorporate the CI changes here already? They're not coming from me, so I'd have to cherrypick from @edmundmiller 's branch. I think doing it in 2 PRs would be better, but it's up to you.

deviantsemicolon:
I would love this because I can't use wezterm on hyprland right now(the issue was fixed in nightly), so this would be a great help

anthr76:
I think it can't hurt if we merge this and iterate upon it. Although since @wez  can't really review these someone else should be appointed reviews with codeowner or another similar mechanism.

wez:
Yes please merge the CI changes into this PR so that I can review both together!

nix/flake.nix
happenslol: Yup, this wasn't needed.

nix/flake.nix
happenslol: This was actually needed. I tried without it, and got missing shared libraries or no displays found in the case of libGL/vulkan.

nix/flake.nix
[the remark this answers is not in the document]
happenslol: @ThinkChaos  I'm interested in removing `flake-utils`, and I've read that article before and changed my own system flake to not use it anymore. I'm not 100% clear on what this would look like for this kind of project though. Is there an example I can look somewhere, or another repo, where something comparable to what I'm doing here is shown?

nix/flake.nix
happenslol: Moved to `devShell.buildInputs`

nix/flake.nix
smsegal: Looks like it's working! I was able to build this on my macbook with 

```
nix build "github:happenslol/wezterm/add-nix-flake?dir=nix"
```

Wezterm version reported as: 
```
$ ./result/bin/wezterm --version
wezterm e941399
```

nix/flake.nix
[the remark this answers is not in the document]
ThinkChaos: Even if you remove it, it'll still be an input because `rust-overlay` uses it. But if you prefer

But the example in the [Do we really need flake-utils?](https://ayats.org/blog/no-flake-utils/#do-we-really-need-flake-utils) section is pretty close. The idea is to move the equivalent of `eachDefaultSystem` into the output attrset so each attr calls it when needed:
`packages = forAllSystems (pkgs: { ... })` instead of `output = { ... }: eachDefaultSystem (pkgs: { packages = ...; })`.

I don't have any flake I can link to (most my stuff is on localhost ATM), but I usually use this function:
```nix
  eachSystem = f:
    lib.genAttrs
      lib.systems.flakeExposed
      (system: f system inputs.nixpkgs.legacyPackages.${system});
```

And then write `package = eachSystem (system: pkgs: { ... });`.

Anyways if you want to resolve this as is it's fine by me since `flake-utils` will be an input as long as we use `rust-overlay`, so up to you if you want to change it now. That info can always be used as a reference later or in other contexts :)

nix/flake.nix
ThinkChaos: I think there should be a way to get it to work, but not worth spending more time on it now IMO. You can resolve this thread.

nix/flake.nix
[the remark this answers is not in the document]
happenslol: Ah, I understand. I was a bit stuck on 2 question:

1. Where does the `system` come from inside `forAllSystems`? In your example you just pass it in, which the blog post doesn't do. I suppose the blog assumes that you'll do that if you need it :sweat_smile:
2. Wouldn't there be a lot of duplication if I had to repeat `buildInputs` and `nativeBuildInputs` inside `packages.${system}.default` and `devShell.${system}`? I suppose it would have been possible to just have a `makeBuildInputs` function and a `makeNativeBuildInputs` function, which is the called with `pkgs`, but that doesn't feel clean and I haven't ever seen that anywhere.
I suppose this one is solved by having an extra file for the package, using `callPackage` and then using `inputsFrom` as well as `packages` in the devshell to add extra packages just for the devshell.

Since we have the dependency anyway and this way of structuring the flake still seems to be very common (even if it's technically not the best way), I suppose I'll keep it like this for now. I'll move the package definition into a separate file (`default.nix`? `wezterm.nix`? I don't know what the convention is there) and then look how much more work would it be to do this.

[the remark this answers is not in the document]
nekowinston: 1. `system` is a function argument from [`lib.genAttrs`](https://noogle.dev/f/lib/genAttrs) in this case, `lib.systems.flakeExposed` is just a predefined list containing supported targets for Nix. I'm actually not a fan of using flakeExposed since it includes stuff like powerpc64le and 32bit, which often just don't work, but still show up in `nix flake show`.
2. `inputsFrom = pkgs.callPackage ./nix {};` or similar is the correct approach here IMO.

If you really want to avoid bringing in flake-utils, you could also use [fenix](https://github.com/nix-community/fenix) instead; As long as you don't use Rust-Analyzer from there, you can even set `inputs.rust-analyzer.follows = ""` there.
However, since this is intended as a nightly/development flake; I'd urge you to *not optimize for flake size* obsessively, and rather include the needed tools for smooth development.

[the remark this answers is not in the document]
happenslol: I understand. I'm always looking for reduced complexity, and in this case it seems like the more minimal solution would also be slightly more complex 😄 I'll settle with moving the package to a file, and keep `flake-utils`. Thanks for all the explanations!

wez:
Thank you!
Now, is there something that should be updated in the docs to help nix folks make use of this?

happenslol:
Ah, I wasn't done with all the review comments yet, but it can't hurt to already have this in main so that people can use it without relying on my fork. I'll make another PR addressing the rest of the comments.

And yeah, it would be great for discoverability if the docs mentioned that there is a flake people can use. Basically, anyone with nix installed can now run wezterm simply by doing `nix run github:wez/wezterm?dir=nix`, and specific versions can be pinned in user installations. Development on NixOS is also now possible by using the devshell with `nix develop ./nix` inside the repo.

I'm still busy so I don't know if I'm the right guy to take care of the documentation - @gabyx @nekowinston @ThinkChaos you all had a hand in this, would any of you be able to take care of that?

anthr76:
I will be happy to help as well! :)

raphaelchristin:
Hey I just got into the nixos world and have been using wezterm nightly for a while. It is great software and I would very much like to install it as a flake into my system. Could someone point me in the right direction? My system configuration is set up as a flake already, so I assume that just adding this repo in the inputs would be the first step, but I am unsure where to take it from there. Any links to relevant documentation would also be great as I could not find any. Thanks in advance and apologies if this is not the right place to ask.

sjcobb2022:
> Hey I just got into the nixos world and have been using wezterm nightly for a while. It is great software and I would very much like to install it as a flake into my system. Could someone point me in the right direction? My system configuration is set up as a flake already, so I assume that just adding this repo in the inputs would be the first step, but I am unsure where to take it from there. Any links to relevant documentation would also be great as I could not find any. Thanks in advance and apologies if this is not the right place to ask.

Hi there @raphaelchristin 

If you are using home-manager, you can do something like the following:

```nix
# flake.nix 
{
    inputs = {
      wezterm = {
        url = "github:wez/wezterm?dir=nix";
      };
    };
    # ... the rest of you flake
}


#some_other_file.nix that has access to inputs and pkgs.
{ inputs, pkgs, ...}: {
    environment.systemPackages = [   inputs.wezterm.packages.${pkgs.system}.default ]
}
```

If you are using home manager, still include your inputs, and then add it to your home manager config like so:

```nix
{ inputs, pkgs, ...}: {
    programs.wezterm = {
        enable = true;
        package = inputs.wezterm.packages.${pkgs.system}.default;
        # other config...
    }
}

```

raphaelchristin:
Thank you so much! This seems to be all I needed, and is much simpler than I thought. Thanks to everyone and especially @wez for the great software!

ppenguin:
> > Hey I just got into the nixos world and have been using wezterm nightly for a while. It is great software and I would very much like to install it as a flake into my system. Could someone point me in the right direction? My system configuration is set up as a flake already, so I assume that just adding this repo in the inputs would be the first step, but I am unsure where to take it from there. Any links to relevant documentation would also be great as I could not find any. Thanks in advance and apologies if this is not the right place to ask.
> 
> Hi there @raphaelchristin
> 
> If you are using home-manager, you can do something like the following:
> 
> ```nix
> # flake.nix 
> {
>   inputs = {
>     wezterm = {
>       url = "github:wez/wezterm?dir=nix";
>     };
>   };
>   # ... the rest of your flake
> }
> 
> 
> #some_other_file.nix that has access to inputs and pkgs.
> { inputs, pkgs, ...}: {
>   environment.systemPackages = [ inputs.wezterm.packages.${pkgs.system}.default ]
> }
> ```
> 
> If you are using home manager, still include your inputs, and then add it to your home manager config like so:
> 
> ```nix
> { inputs, pkgs, ...}: {
>   programs.wezterm = {
>       enable = true;
>       package = inputs.wezterm.packages.${pkgs.system}.default;
>       # other config...
>   }
> }
> ```

@sjcobb2022 Thanks! Just wanted to try out wezterm for the first time (NixOS/Hyprland) and was surprised it didn't work out of the box.

Just as a small extra/alternative: you could also put it in the overlay, so it affects all HM users on the system:
```
nixpkgs.overlays = [
   (final: prev {
      wezterm =
        if (lib.hasAttr "wezterm" inputs) then
          inputs.wezterm.packages.${prev.system}.default
        else
          prev.wezterm;
   })
]
```
I've been using this paradigm to quickly switch between standard `nixpkgs` and dedicated (upstream) inputs by just commenting in/out the relevant inputs.

I'll probably be PR'ing a `wezterm` update to `nixpkgs` if I'm fast enough and find time :wink: 
I've been

PR #3568, opened 2023-04-20
fixing an example

maxim-uvarov:
This comma produces an error in the config.

### DIFF
--- a/docs/config/lua/config/quick_select_patterns.md
+++ b/docs/config/lua/config/quick_select_patterns.md
@@ -13,7 +13,7 @@ This setting is a table listing out a set of regular expressions.
 config.quick_select_patterns = {
   -- match things that look like sha1 hashes
   -- (this is actually one of the default patterns)
-  '[0-9a-f]{7,40}',
+  '[0-9a-f]{7,40}'
 }
 ```
 


wez:
Can you share the error that you saw? Trailing commas should be totally fine in lua

maxim-uvarov:
As soon as I save the config file with trailing comma in the section of quick select patterns, I get this error.
image

It appeared several releases ago, maybe about 2 months. My config was unoperable, so I had to disable settings one by one. Finally I found that the only problematic section was `quick_select_patterns`. At that time I didn't find what was the problem, as I used my pattern exactly as it was in help (and this pattern I had been using for some time already).
So I disabled all the section. But later I was missing this killer feature a lot and decided to try to figure out what was the problem. And as far as I could understand it was with that trailing comma (even though in other parts of config they are fine).

maxim-uvarov:
here is my full config in case it matters:
```

-- Allow working with both the current release and the nightly
local config = {}
if wezterm.config_builder then
  config = wezterm.config_builder()
end


config.default_prog = { '/Users/user/.cargo/bin/nu', '-l'}
-- config.default_prog = { '/Users/user/.cargo/bin/nu077dfr', '-l', '-t', '30' }
-- config.debug_key_events = true

-- config.swap_backspace_and_delete = true
config.font = wezterm.font { family = 'JetBrains Mono' }
config.font_size = 16.0
config.initial_cols = 120
config.initial_rows = 120
config.skip_close_confirmation_for_processes_named = {
  'bash',
  'sh',
  'zsh',
  'fish',
  'tmux',
  'nu',
  'nu074',
  'nu076',
  'nu076df',
  'nu077',
  'nu077dfr'
}
config.scrollback_lines = 10000
config.keys = {
  -- Turn off the default CMD-m Hide action, allowing CMD-m to
  -- be potentially recognized and handled by the tab
  {
    key = 'Enter',
    mods = 'ALT',
    action = wezterm.action.DisableDefaultAssignment,
  },
  {
    key = 'Backspace',
    mods = 'CTRL',
    action = wezterm.action.DisableDefaultAssignment,
  }
}
config.mouse_wheel_scrolls_tabs = false
conifg.quick_select_patterns = {
  "[0-9A-F]{64}"
  -- '[0-9a-f]{7,40}',
}
config.debug_key_events = true     
config.launch_menu = {
  {
    args = { 'top' },
  },
  {
    -- Optional label to show in the launcher. If omitted, a label
    -- is derived from the `args`
    label = 'nu multithreads',
    -- The argument array to spawn.  If omitted the default program
    -- will be used as described in the documentation above
    args = { 'nu', '-l', '-t', '40' },

    -- You can specify an alternative current working directory;
    -- if you don't specify one then a default based on the OSC 7
    -- escape sequence will be used (see the Shell Integration
    -- docs), falling back to the home directory.
    -- cwd = "/some/path"

    -- You can override environment variables just for this command
    -- by setting this here.  It has the same semantics as the main
    -- set_environment_variables configuration option described above
    -- set_environment_variables = { FOO = "bar" },
  },
}

return config
```

phgz:
You wrote `conifg.quick_select_patterns` instead of `config.quick_select_patterns` -> Typo in `config`.

maxim-uvarov:
I'm so sorry to bother you in vain.
Thank you!!!

PR #3579, opened 2023-04-21
Add config option for mouse cursor when grabbing

### LINKED ISSUE #3560: Configure behaviour of mouse cursor when mouse is grabbed

From #859 (and the commit ae4bfae), the behaviour was changed so that if the app captures the mouse (or is mouse-reporting, I guess?) the arrow cursor is always used. 

I would rather this be a setting, something like `always_use_arrow_cursor_when_mouse_reporting`, or something like that.

zhiayang:
This should fix #3560 by adding a config option for it. Not sure if this is the best name, `use_arrow_mouse_cursor_when_mouse_reporting` might be a bit long.


zhiayang:
bump?

zhiayang:
Any chance that this can be looked at?

kouyk:
bumping this PR, it's very useful to have and I see no reason why it shouldn't be merged

PR #3600, opened 2023-04-23
chore: remove a never loop instead of get first

acehinnnqru:


wez:
Thanks, but I don't want this: it is a longer and more complicated way to express the same thing

PR #3613, opened 2023-04-25
Fixed double-skewing behavior

### LINKED ISSUE #3555: Fonts: Reduce default skew factor and/or permit user-specified skew factor
## Is your feature request related to a problem? Please describe.

When a font doesn't have built-in italics, wezterm (and most terminals) will approximate an italic variant by simply _skewing_ the letters. Currently, the skew factor is [fixed at 20%](https://github.com/wez/wezterm/blob/87cd9c5f6cc26809527344a9713aff19227dc243/wezterm-font/src/rasterizer/mod.rs#L25), and it doesn't look great. Here is a side by side. First image is wezterm, second is iTerm2, both using Fira Code Medium:
Screenshot 2023-04-18 at 01 17 59
Screenshot 2023-04-18 at 01 18 02

In fact, the letters are so skewed, that the cursor highlights portions of adjacent letters sometimes:
Screenshot 2023-04-18 at 01 29 22

### Describe the solution you'd like

Two fixes, both of which can happen independently:

1. I think the default skew factor should be reduced.  I haven't played with the values so I don't have a concrete recommendation for a number
2. In addition, make the skew factor configurable. E.g.,
```lua
config.font = wezterm.font("Fira Code", { weight="Medium", skew_factor = 0.10 })
```

### Describe alternatives you've considered

Looked at other font options. There's nothing rn that addresses this.

rozbb:
Previously, synthetiszed italic glyphs went through a face transform as well as a manual skewing operation. This resulted in a glyph being skewed twice. This PR removes the latter skewing operation. Here's a side by side with iTerm2 now:

Screenshot 2023-04-24 at 23 59 56
Screenshot 2023-04-25 at 00 10 06

This partially fixes #3555.

I played around with configurable skew factor but decided it was more effort than was worth it. iTerm2 doesn't let you adjust this factor, and I think that's fine.

# Remaining work

There's still a text highlighting bug when italics are synthesized. Notice that part of the `r` below is not highlighted, while part of the adjacent `a` is:
Screenshot 2023-04-25 at 00 04 54

For comparison, iTerm2 highlights _all_ of `r` and none of `a`, despite the fact that some of `r` is not inside the cursor's boundary. I'd imagine that the fix for this would be quite annoying, and I don't care too much.
Screenshot 2023-04-25 at 00 05 31

### DIFF
--- a/wezterm-font/src/rasterizer/freetype.rs
+++ b/wezterm-font/src/rasterizer/freetype.rs
@@ -1,8 +1,8 @@
 use crate::parser::ParsedFont;
-use crate::rasterizer::FontRasterizer;
+use crate::rasterizer::{FontRasterizer, FAKE_ITALIC_SKEW};
 use crate::units::*;
 use crate::{ftwrap, RasterizedGlyph};
-use ::freetype::{FT_GlyphSlotRec_, FT_Glyph_Format, FT_Matrix};
+use ::freetype::{FT_GlyphSlotRec_, FT_Matrix};
 use anyhow::bail;
 use config::{DisplayPixelGeometry, FreeTypeLoadFlags, FreeTypeLoadTarget};
 use std::cell::RefCell;
@@ -14,7 +14,6 @@ pub struct FreeTypeRasterizer {
     face: RefCell<ftwrap::Face>,
     _lib: ftwrap::Library,
     synthesize_bold: bool,
-    synthesize_italic: bool,
     freetype_load_target: Option<FreeTypeLoadTarget>,
     freetype_render_target: Option<FreeTypeLoadTarget>,
     freetype_load_flags: Option<FreeTypeLoadFlags>,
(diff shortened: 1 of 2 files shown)


wez:
Thanks!

rozbb:
Np! Thanks for the quick turnaround. Regarding the weird highlighting behavior, I think it might be easier than I thought. Might just be a simple change in the shader code. Though I've never really worked with shaders (outside of basic tutorials) so idk if that's the case.

wez:
re: highlighting, there are two different shaders, one for opengl and one for webgpu. However, I think if you wanted to fix the issue you'll need to look at this logic instead of the shader:

https://github.com/wez/wezterm/blob/154dfae9b9e81db6d046e15cd317bbd22b106e69/wezterm-gui/src/termwindow/render/screen_line.rs#L514-L520

that code explicitly constraints the the colors to terminal cells in order for things like ligatures and the cursor apply to the correct cells; care will be needed to avoid regressing:
* https://github.com/wez/wezterm/issues/478

rozbb:
What a can of worms 😅  An elegant approach would be to do highlighting _before_ skew is applied. But these things happen at drastically different levels, so I don't think that's much of an option. Another option is to de-skew in the shader. All a bit messy though :)

PR #3626, opened 2023-04-26
termwiz: make pest an optional dependency

quark-zju:
`pest` is only used for `termwiz::tmux_cc`. Not all termwiz users are interested in tmux integration. So let's make it optional.

----

Looking at the dependencies, I also wonder if other ones can be optional (looking for feedback):
- fancy-regex: only used by auto hyperlink recognition? Maybe it can be made optional? (`regex` is not exactly lightweight)
- wezterm-blob-leases: only used by image? Could `mod image` be gated by `use_image` feature?

### DIFF
--- a/mux/Cargo.toml
+++ b/mux/Cargo.toml
@@ -39,7 +39,7 @@ serial = "0.4"
 shell-words = "1.1"
 smol = "1.2"
 terminfo = "0.8"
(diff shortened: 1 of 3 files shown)


wez:
Obsoleted by your later 33b40fe62904cacfb3f37a99d8db53b9016ca4e1
Thanks!

PR #3666, opened 2023-05-02
Remove reserved native buttons space in full screen

junnplus:
When configuring `window_decorations= 'INTEGRATED_BUTTONS|RESIZE'`, there will be extra blank space in full screen mode.

image


wez:
Thanks!

PR #3674, opened 2023-05-03
Deduplicate entries in emoji menu

vimpostor:
In the CharSelect menu, the same emoji may appear multiple times, because multiple aliases might match the user query at the same time (this is very likely, as aliases tend to have similar names).
To fix this, instead of deduplicating the vector (which would have terrible performance), use a `HashMap` instead for free deduplication and pretty much identical performance (slightly worse for large n, slightly better for small n).

## Before
Screenshot_20230503_191433

Duplicate entries are shown.

## After
Screenshot_20230503_191504

Only entries with the best matching score are shown.

### DIFF
--- a/wezterm-gui/src/termwindow/charselect.rs
+++ b/wezterm-gui/src/termwindow/charselect.rs
@@ -18,6 +18,7 @@ use fuzzy_matcher::FuzzyMatcher;
 use serde::{Deserialize, Serialize};
 use std::borrow::Cow;
 use std::cell::{Ref, RefCell};
+use std::collections::HashMap;
 use std::path::PathBuf;
 use termwiz::input::Modifiers;
 use wezterm_term::{KeyCode, KeyModifiers, MouseEvent};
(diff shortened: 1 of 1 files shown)


wezterm-gui/src/termwindow/charselect.rs
```diff
@@ -268,45 +269,57 @@ fn compute_matches(selection: &str, aliases: &[Alias], group: CharSelectGroup) -
...
-                    .map(|score| MatchResult::new(row_idx, score, selection, aliases));
-                match &numeric_selection {
-                    Some(sel) => {
-                        let codepoints = entry.codepoints();
-                        if codepoints == *sel {
-                            Some(MatchResult {
+        for (row_idx, entry) in aliases.iter().enumerate() {
+            let character = match entry.character {
+                Character::Unicode { value: c, .. } => c,
+                Character::Emoji(d) => d.as_str().chars().next().unwrap(),
```
wez, line 279: The unwrap here caught my eye; I try to avoid unqualified unwraps and prefer returning errors or using `.expect` to annotate the reason why a panic is acceptable.

However, I think this logic is flawed: Emoji can be comprised of multiple codepoints so hashing based on the first codepoint will lead to falsely skipping certain results; this will be most prevalent for sequences that vary by skin tone, or that combine gender and vocation to produce alternative versions of certain emoji.

What I think is more appropriate here is taking the textual representation of the entry and hashing that.

In other words, use the result of calling `entry.glyph()` as the key to the hash table.

wezterm-gui/src/termwindow/charselect.rs
```diff
@@ -268,45 +269,57 @@ fn compute_matches(selection: &str, aliases: &[Alias], group: CharSelectGroup) -
...
+                None => alias_result,
+            } {
+                if let Some(other) = matches.get(&character) {
+                    // if the character was inserted already, replace it only if the score is improved
+                    if other.score < value.score {
+                        matches.insert(character, value);
+                    }
+                } else {
+                    matches.insert(character, value);
+                }
```
wez, line 319: Consider using something like this:

```rust
let existing = matches.entry(glyph).or_insert(value);
// Replace an existing entry only if the score improved
if value.score > existing.score {
   *existing = value;
}
```

wez (changes requested):
Thanks for this!
A couple of comments/suggestions on the code.

I feel somehow uneasy about this: as though there's a reason to show all of the aliases but I can't put my finger on what it is.  Trying to be rational about it: when no text is entered we show everything and allow browsing all of the results (good).  When filtering, I do think it makes sense to unclutter, but because of the way the fuzzy matching works, some matches are made in non-obvious ways and the results can be surprising--that is coming from fuzzy matching in the Launcher Menu and Command Palette though. I think the set of candidates here in charselect mode aren't so surprising, so this is probably fine. I don't think this blocks this PR, just sharing that something is gnawing at me about this and that it may take a bit of thinking through to satisfy myself that there aren't any weird gotchas for someone!

wezterm-gui/src/termwindow/charselect.rs
vimpostor: This isn't really possible, `or_insert()` will move `value` so you can't use it in the line below that.

The most elegant way to coalesce the two insert statements would be to use some negated `is_some_and()` `let` statement, but this isn't a stable Rust feature in wezterm's minimal Rust requirement.

vimpostor:
> I don't think this blocks this PR, just sharing that something is gnawing at me about this and that it may take a bit of thinking through to satisfy myself that there aren't any weird gotchas for someone!

I appreciate the review and if you feel like this is not a direction you want to go, it's totally fine to reject this PR. After all it's my own fault for not asking you beforehand if you would want to merge something like this.

That being said, I think the UX of not showing duplicate emoji is much better and aligns with the UX of popular messenger emoji pickers, such as Telegram or Whatsapp.

wezterm-gui/src/termwindow/charselect.rs
```diff
@@ -268,45 +269,54 @@ fn compute_matches(selection: &str, aliases: &[Alias], group: CharSelectGroup) -
...
-            })
-            .collect();
+                None => alias_result,
+            } {
+                if let Some(other) = matches.get(&glyph) {
+                    // if the character was inserted already, replace it only if the score is improved
+                    if other.score < value.score {
+                        matches.insert(glyph, value);
+                    }
+                } else {
```
wez, line 314: How about this as a potential alternative micro optimization then?
Should avoid repeating the hash lookup for the insertion case.

```suggestion
                if let Some(other) = matches.get_mut(&glyph) {
                    // if the character was inserted already, replace it only if the score is improved
                    if other.score < value.score {
                        *other = value;
                    }
                } else {
```

vimpostor: Done

wez:
Thanks!

PR #3703, opened 2023-05-09
add color config for char select

junnplus:
Now the char select window relies on the color configuration of pane select, making it impossible to distinguish between the two.

I want to configure them by adding two colors: `char_select_fg_color` and `char_select_bg_color`.


wez (approved):
Thanks!

PR #3728, opened 2023-05-17
Fix typo.

follower:
While I very much look forward to "future wezterms of wezterm", I think perhaps their existence should only be referred to by use of the codeword "version" for now. :)


wez:
Thanks!

PR #3737, opened 2023-05-18
packaging: Add Snap build

RAOF:


RAOF:
I am aware that previous attempts at a snap have been met with snapcraft bureaucracy; I am also offering to shepherd this through the “yes, this terminal emulator really needs classic” hoops on snapcraft.io.

Alternatively, this can either be closed or might still prove useful for people wanting to locally-build snaps.

snap/snapcraft.yaml
```diff
@@ -0,0 +1,73 @@
+name: wezterm # you probably want to 'snapcraft register <name>'
```
wez, line 1: I previously registered this name!

snap/snapcraft.yaml
```diff
@@ -0,0 +1,73 @@
...
+      - libxcb-shm0
+      - libxcb-util1
+      - libxcb-xkb1
+      - libxcb1
+      - libxkbcommon-x11-0
+      - libxkbcommon0
+    override-build: |
+      cargo build --release --features distro-defaults
+      mkdir -p $SNAPCRAFT_PART_INSTALL/bin
+      for component in wezterm wezterm-gui wezterm-mux-server strip-ansi-escapes ; do
```
wez, line 67: I've paged out pretty much everything I know about snaps... how do users explicitly launch one of these binaries, vs. just launching wezterm itself?

Sort of related; is there a reasonable way to export the shell integration script(s) (https://github.com/wez/wezterm/tree/main/assets/shell-integration) from the snap?  There isn't a way to do this with the AppImage and I doubt that is possible for snap, I'm just asking in case there is!

snap/snapcraft.yaml
```diff
@@ -0,0 +1,73 @@
...
+  rust-deps:
+    plugin: nil
+    build-packages:
+      - curl
+    override-pull: |
+      curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+  wezterm:
+    plugin: rust
+    source: .
+    parse-info: [assets/wezterm.appdata.xml]
```
wez, line 24: will this populate metadata such as the screenshots and other links that are shown from the flatpak/appimage? If not, do those need to be added here?

wez: And speaking of additional metadata, could you transcribe https://github.com/wez/wezterm/blob/main/.github/FUNDING.yml into appropriate `donation` entries?

wez (commented):
Thanks for this!

Yes, I got super frustrated by the intake process for the snap last time around. It looks like I previously registered wezterm; it shows up as a private snap when I login to https://snapcraft.io/snaps

CI wise, it looks like we'd want to add the build steps for this into the ubuntu 22 build?
I'm assuming that we can build and capture the artifact in PR and continuous commits, and then automate the publish when tagging.  That doesn't need to be in this PR; I'm just thinking ahead to how to keep this in good shape.

https://github.com/wez/wezterm/blob/main/ci/generate-workflows.py is the script that is used to generate the workflows for the various platforms in case you wanted to start thinking about that. I'd suggest looking at how the appimage is handled to get a sense of what might need to be changed to facilitate the snap build.

snap/snapcraft.yaml
RAOF: I believe there *is* some support for exporting shell completions from snaps.

Actually, now I think about it, because this will be a `classic` snap, we can just have an `install` hook copy the integration into the right place. Does the release machinery know where ”the right place” is, so I can copy it?

wez: this is where we build the .deb package and copy the shell integration and completion files; the shell integration lands in `/etc/profile.d`, but the destination for shell completion varies by shell:

https://github.com/wez/wezterm/blob/6b21e6ffde46d1345fdfc5a1437054fa8d964e7c/ci/deploy.sh#L248-L250

Note that on rpm based systems the shell completions live in completely different locations:
https://github.com/wez/wezterm/blob/6b21e6ffde46d1345fdfc5a1437054fa8d964e7c/ci/deploy.sh#L181-L182

snap/snapcraft.yaml
RAOF: > will this populate metadata such as the screenshots and other links that are shown from the flatpak/appimage?

I actually don't know about the screenshots; we might have to see once the snap gets into the store? It's quite possible that they are

Added the `donation:` stanza.

snap/snapcraft.yaml
RAOF: > I've paged out pretty much everything I know about snaps... how do users explicitly launch one of these binaries, vs. just launching wezterm itself?

Any binaries that want to end up in `$PATH` need to be explicitly added to the `apps:` section, above. Anything invoked by the `wezterm` binary itself is fine, though, as you've already got code to locate the relative path to these (as far as I can tell - everything *seems* to work :grin:).

I assumed that users didn't need to directly invoke the other binaries, but maybe actually we should expose `wezterm-gui` (with the `.desktop` file) *and* `wezterm` (without the `.desktop` file?), and maybe also `wezterm-mux-server`?

wez: Yeah, I think exposing/exporting all the binaries to the PATH makes sense, as there are some folks that want to use the mux server directly on remote hosts.

Only wezterm-gui needs to have a `.desktop` file.

wez:
it's been a while probably since either of us looked at this; what's the status?

RAOF:
Ah, sorry. Dropped off my radar a bit (partially because I've been building this snap locally!).

So, locally I've got everything but the shell-completion stuff done. I'm pretty sure it's *possible* to install the shell-completion stuff in an install hook, but it'd need logic for picking the right place to install them.

I guess that could be checking `lsb_release` and switching based on distro? Or we could just punt that for future work.

wez:
No worries; I too was swamped in the interim!

I think the shell integration scripts are more important to get installed than the completions, and thankfully those are unconditionally dropped into `/etc/profile.d` and don't care about the distro.

I'm happy to punt on the completions; those can actually be regenerated by the user if they need them, but for folks moving from the CI-provided deb/rpms to the snap, it would be a downgrade to their workflow if the integrations were no longer provided.

psadi:
Wanted to check if there are any plans to have this merged and published to the snap store.

PR #3738, opened 2023-05-18
env: Strip unwanted Snap variables from environment

RAOF:


env-bootstrap/src/lib.rs
```diff
@@ -11,6 +11,33 @@ pub fn set_wezterm_executable() {
     }
 }
 
+pub fn fixup_snap() {
+    if std::env::vars_os().any(|(key, _)| key == "SNAP") {
```
wez, line 15: This feels a bit clearer and easier to grok:

```suggestion
    if std::env::var_os("SNAP").is_some() {
```

env-bootstrap/src/lib.rs
```diff
@@ -11,6 +11,33 @@ pub fn set_wezterm_executable() {
...
+        // snapd sets a bunch of environment variables as a part of setup
+        // These are not useful to be passed through to things spawned by us.
+
+        // SNAP is the base path of the files in the snap
+        std::env::remove_var("SNAP");
+
+        // snapd also sets a bunch of SNAP_* environment variables
+        // This list may change over time, so for simplicity, assume
+        // anything in the SNAP_* namespace is set by snapd, an unset it.
```
wez, line 24:
```suggestion
        // anything in the SNAP_* namespace is set by snapd, and unset it.
```

wez (commented):
Thanks! A couple of minor comments!

env-bootstrap/src/lib.rs
RAOF: Oooh, yeah. That's better. I didn't notice that API! :grin:

wez:
Thanks!

PR #3741, opened 2023-05-18
fix encoding Tab with modifiers

cairijun:
On win32 we have some ["virtual modifiers" generated](https://github.com/wez/wezterm/blob/6bcb8eed56c2e8fa2d18687bff84f19587e0cb53/window/src/os/windows/window.rs#L2494), but they are causing problems when encoding `_ + Tab` to other platforms. `Tab` with modifiers are not working when using wezterm's builtin ssh client.

## Reproduce
1. `wezterm.exe ssh HOST:PORT -- showkey -a`
2. Press `Shift-Tab`
3. Only `0x09` (`\t`) is received, as if a single `Tab` was pressed
  image

## With this patch
image

## Logs with `debug_key_events = true`
[before.log](https://github.com/wez/wezterm/files/11507471/before.log)
[after.log](https://github.com/wez/wezterm/files/11507473/after.log)

### DIFF
--- a/termwiz/src/input.rs
+++ b/termwiz/src/input.rs
@@ -262,6 +262,9 @@ impl KeyCode {
             // We only want down events
             return Ok(String::new());
         }
+        // We are encoding the key as an xterm-compatible sequence, which does not support
+        // positional modifiers.
+        let mods = mods.remove_positional_mods();
 
         use KeyCode::*;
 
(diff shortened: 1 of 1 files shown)


termwiz/src/input.rs
```diff
@@ -365,6 +365,7 @@ impl KeyCode {
             }
 
             Tab => {
+                let mods = mods.remove_positional_mods();
```
wez, line 368: Hmm! Good find! However, I think this should probably be moved up to line 265 so that it applies to everything in this function, and not just Tab!

wez:
Thank you!

PR #3782, opened 2023-05-26
Fix sftp::FilePermissions::is_readonly to properly report readonly status

chipsenkbeil:
@wez I must have been asleep when I wrote this function because it's checking the wrong flags. We want it to see if any of the write permissions are set. If they are, then this is not readonly, otherwise it is. This is how the [std lib does it for Unix platforms](https://github.com/rust-lang/rust/blob/1a5f8bce74ee432f7cc3aa131bc3d6920e06de10/library/std/src/sys/unix/fs.rs#L570).


wez:
Thanks!

PR #3856, opened 2023-06-13
feat: handle horizontal mousewheel events

calops:
My apologies as this is a rough PR and I'll admit that I'm not sure I fully understand 100% of the code I touched. I don't *think* this is breaking anything though, and it enables horizontal scrolling as far as I've tested (it works in neovim).

There's a `TODO` left about the values in `mouse_report_button_number` as I really had no idea what to put there, so I simply incremented the other values. It seems to work but I don't know if that's correct, and I couldn't find any reference anywhere about this.

### DIFF
--- a/term/src/terminalstate/mouse.rs
+++ b/term/src/terminalstate/mouse.rs
@@ -56,6 +56,8 @@ impl TerminalState {
             MouseButton::Right => 2,
             MouseButton::WheelUp(_) => 64,
             MouseButton::WheelDown(_) => 65,
+            MouseButton::WheelLeft(_) => 66,
+            MouseButton::WheelRight(_) => 67,
         };
 
         if event.modifiers.contains(KeyModifiers::SHIFT) {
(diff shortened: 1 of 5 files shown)


term/src/terminalstate/mouse.rs
```diff
@@ -56,6 +56,8 @@ impl TerminalState {
             MouseButton::Right => 2,
             MouseButton::WheelUp(_) => 64,
             MouseButton::WheelDown(_) => 65,
+            MouseButton::WheelLeft(_) => 66, //TODO: check
```
wez, line 59: Confirmed; 66 is Button6 which is the left wheel, and 67 is Button7, the right wheel

wez (commented):
Thanks for this!
I think this looks good!

term/src/terminalstate/mouse.rs
calops: Awesome, thanks! Out of curiosity, where do you find this information?

wez: https://www.invisible-island.net/xterm/ctlseqs/ctlseqs.html#h3-Other-buttons 
plus other code here in wezterm's repo at:

https://github.com/wez/wezterm/blob/dbed5017f828e8719ad871a031acd4c09c0c98d0/termwiz/src/escape/csi.rs#L605-L606

calops: It should be gone now

wez:
Thanks!

Issue #581, opened 2021-03-29
Allow mouse bindings to bypass mouse-reporting


bew:
**Is your feature request related to a problem? Please describe.**
My open-link mouse-binding is Ctrl-click on wezterm, I'd like to have the exact same bindings to work whenever the program activated mouse-reporting or not.
Otherwise I must know if it has, and press the additional Shift key to bypass mouse reporting.

**Describe the solution you'd like**
A new field for mouse bindings like `bypass_mouse_reporting` defaulting to false.

**Describe alternatives you've considered**
Using the shift key, but to me it's not always consistent because I must know if a program has it enabled or not..

_Edit: Another mouse binding I'd like to use like this is Middle mouse click to paste, so I can paste in tmux (or others) with bracketedpaste always the same way._

wez:
This should be fixed now in `main`.

It typically takes about an hour before fixes are available as nightly builds for all platforms.  Linux builds are the fastest to build and are often available within about 20 minutes.  Windows and macOS builds take a bit longer.

Please take a few moments to try out the fix and let me know how that works out. You can find the nightly downloads for your system in the [wezterm installation docs](https://wezfurlong.org/wezterm/installation.html).

If you prefer to use packages provided by your distribution or package manager of choice and don't want to replace that with a nightly download, keep in mind that you can download portable packages (eg: a `.dmg` file on macOS, a `.zip` file on Windows and an `.AppImage` file on Linux) that can be run without permanently installing or replacing an existing package, and can then simply be deleted once you no longer need them. 

If you are eager and can [build from source](https://wezfurlong.org/wezterm/install/source.html) then you may be able to try this out more quickly.

PR #3861, opened 2023-06-16
docs(keyassignment): fix default OpenLinkAtMouseCursor keybind

owittek:
The default keybind to open a link is Shift + Left Click, not CTRL.

### DIFF
--- a/docs/config/lua/keyassignment/OpenLinkAtMouseCursor.md
+++ b/docs/config/lua/keyassignment/OpenLinkAtMouseCursor.md
@@ -5,10 +5,10 @@ a hyperlink, this action causes that link to be opened.
 
 ```lua
 config.mouse_bindings = {
-  -- Ctrl-click will open the link under the mouse cursor
+  -- Shift-click will open the link under the mouse cursor
   {
     event = { Up = { streak = 1, button = 'Left' } },
-    mods = 'CTRL',
+    mods = 'Shift',
     action = wezterm.action.OpenLinkAtMouseCursor,
   },
 }


owittek:
Actually it seems more like a bug since I've tried to modify this behavior to use `CTRL` instead of `SHIFT` by adding the snippet from the docs to my config but `SHIFT + Left Click` is still the way to open a link.

I'm running the latest macOS (13.4) in case that helps with testing.

bew:
This section is an example config, not the default.

By default links can be opened directly with a click, or might need shift if the program captures mouse events (shift allows to bypass this capture).

owittek:
> This section is an example config, not the default.
> 
> By default links can be opened directly with a click, or might need shift if the program captures mouse events (shift allows to bypass this capture).

Thank you for your answer. I guess I need to use `Shift` since I'm running everything in tmux so I didn't think about that. Is there some tmux integration to be either able to "just click the link" or to modify this behavior somehow?

Also I'm closing the PR, thank you!

bew:
> or to modify this behavior somehow

I actually asked about it some long time ago at #581, and we now have dedicated fields for making additional custom mouse bindings that apply when the program is capturing mouse events using `mouse_reporting=true`

See https://wezfurlong.org/wezterm/config/mouse.html#configuring-mouse-assignments (below the example)

owittek:
I also managed to get rid of the need to use `Shift` by removing `set -g mouse on` from my `.tmux.conf`. Might be an edge case worth documenting for other people running into this issue.

owittek:
Hey, since I kinda miss being able to scroll I'm once again looking into this. I went into your dotfiles for reference but I saw that you're still referencing the issue as open within them: https://github.com/bew/dotfiles/blob/60fa83e65e56e3550f42312b7d19d9a50be3306a/gui/wezterm/cfg_mouse.lua#L88

Would you mind helping me with a minimal config to enable opening Links with a simple left-click when `set -g mouse on` is set for tmux? @bew

bew:
Yeah I kinda got used to Ctrl-Shift-click as I was on tmux non-stop ^^'

I just updated my config to use `mouse_reporting=true` in my relevant mouse bindings if you want to take a 2nd look:
https://github.com/bew/dotfiles/commit/e37fcbbbe99798bbbacb3dccaa9de2fc79db79e5
(the 1st diff hunk is adding support for additional options to a function that helps me generate mouse bindings, 2nd diff hunk uses that for my Ctrl-click bindings, the 3rd hunk doesn't use my helper function and shows a simple binding to paste with middle-click both in normal output and in a program that enabled mouse reporting)

Basically you need to add new mouse bindings with `mouse_reporting = true` (this binding option is `false` by default, meaning that the binding only works when the program isn't grabbing mouse events)

Issue #3843, opened 2023-06-10
[bug] copy mode on wayland


justchokingaround:
### What Operating System(s) are you seeing this problem on?

Linux Wayland

### Which Wayland compositor or X11 Window manager(s) are you using?

Hyprland

### WezTerm version

wezterm 20230530-062729-95e44f21

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

Copying using `Copy Mode`, with the `CTRL-SHIFT-X` keybind on wayland does not work.

Note: it does work properly if an input was copied prior to using `Copy Mode`, using the mouse, in the same window.

### To Reproduce

1. Open a fresh wezterm window
2. Run any command
3. Try copying using `Copy Mode` with either `y` or `CTRL-SHIFT-C` (both don't work)

### Configuration

no config

### Expected Behavior

Copying should work without needing to use the mouse, not even once.

osandov:
I've been hitting this bug, too. I think I fixed it in #3929.

justchokingaround:
tested, and can confirm that it fixes my bug <3

PR #3929, opened 2023-07-03
Fix copy mode on Wayland

osandov:
On Wayland, copy mode often doesn't actually update the clipboard. Specifically, it only works one time after a pointer enter or pointer button event, then doesn't work again until the next event.

This is because the Wayland protocol serial number in CopyAndPaste::last_serial is only updated by pointer enter and pointer button events. So, subsequent copies using only the keyboard reuse the same serial number and get ignored. last_serial used to be updated for keyboard events, too, but that was (accidentally?) dropped in commit 0a00ffe98b11546c5fdecca8a77dd72cc609db9c.

Commit 0a00ffe98b11546c5fdecca8a77dd72cc609db9c also added another last_serial to WaylandConnection which is updated for keyboard events but isn't used anywhere as far as I can tell.

So, to fix this bug, let's get rid of CopyAndPaste::last_serial and replace it with WaylandConnection::last_serial, which is now updated for pointer and keyboard events.

closes: #3843

### DIFF
--- a/window/src/os/wayland/copy_and_paste.rs
+++ b/window/src/os/wayland/copy_and_paste.rs
@@ -16,13 +16,11 @@ use crate::Clipboard;
 #[derive(Default)]
 pub struct CopyAndPaste {
     data_offer: Option<WlDataOffer>,
-    pub(crate) last_serial: u32,
 }
 
 impl std::fmt::Debug for CopyAndPaste {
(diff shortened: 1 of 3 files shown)


justchokingaround:
i tested this, and it does indeed fix the bug in #3843

wez:
Looks good, thanks!

PR #3937, opened 2023-07-05
fix: issue 3935 handle F13-F24

ovidiu-ionescu:
Fixes: keys F13-F24 are not passed through on a PC122 keyboard [issue 3935](https://github.com/wez/wezterm/issues/3935)


wez:
Thanks for this!  For the sake of completeness, since I don't have a device where I can test this, could you also take a look at the kitty keyboard encoding for these keys?

There's some out-commented relevant code over here:

https://github.com/wez/wezterm/blob/69bb69b9ca30edb82e134a5835fcefadf8830fcc/wezterm-input-types/src/lib.rs#L1307-L1322

that may also require some plumbing for higher numbered F-keys over here:

https://github.com/wez/wezterm/blob/69bb69b9ca30edb82e134a5835fcefadf8830fcc/wezterm-input-types/src/lib.rs#L1009

and here:

https://github.com/wez/wezterm/blob/69bb69b9ca30edb82e134a5835fcefadf8830fcc/wezterm-input-types/src/lib.rs#L742

and here:

https://github.com/wez/wezterm/blob/69bb69b9ca30edb82e134a5835fcefadf8830fcc/wezterm-input-types/src/lib.rs#L1856-L1869

Testing methodology should be something like:

* Spawn a wezterm with kitty keyboard enabled: `wezterm -n --config enable_kitty_keyboard=true --config disable_default_key_bindings=true`
* In a shell in there, run `printf '\e[=31u' ; cat`
* Try pressing those extended F keys; you should see very large numbers like 57397 for F34
* https://github.com/kovidgoyal/kitty/blob/master/docs/keyboard-protocol.rst#functional-key-definitions has more details on the protocol itself

ovidiu-ionescu:
Sure, will do. My keyboard only has 24 Function keys labelled as such. Not sure if the extra 10 keys on the left are supposed to be the F25-F34 ones.

ovidiu-ionescu:
I can make wezterm send the kitty codes for F13-F24 but neovim does not seem to enter CSI mode and does not react to them. Is there a way to diagnose the dialog  establishing the mode between neovim and wezterm?

wez:
You can use https://wezfurlong.org/wezterm/config/lua/config/debug_key_events.html to have wezterm show what it is sending to the program in the terminal, which may be helpful

ovidiu-ionescu:
I think I found a problem in neovim, I found the place in the code where it ignores the F13 and higher kitty codes send by wezterm. 

Also collected a backtrace when wezterm crashed while doing this.

zeertzjq:
I don't think the escape sequences produced in this PR are correct.

https://sw.kovidgoyal.net/kitty/keyboard-protocol/ lists:
Name | CSI | Name | CSI
-- | -- | -- | --
F13 | 57376 u | F14 | 57377 u
F15 | 57378 u | F16 | 57379 u
F17 | 57380 u | F18 | 57381 u
F19 | 57382 u | F20 | 57383 u
F21 | 57384 u | F22 | 57385 u
F23 | 57386 u | F24 | 57387 

But the format used in this PR is `format!("{intro};{modifiers}{event_type}~")`, which ends with `~`, not `u`.

zeertzjq:
This is still incorrect, as F1-F12 should end with `~`, so that most terminal programs that don't interpret characters in Private User Area can still recognize them.

I think it should be something like this:
```rust
            Function(n) if *n < 25 => {
                // The spec says that kitty prefers an SS3 form for F1-F4,
                // but then has some variance in the encoding and cites a
                // compatibility issue with a cursor position report.
                // Since it allows reporting these all unambiguously with
                // the same general scheme, that is what we're using here.
                let intro = match *n {
                    1 => "\x1b[11",
                    2 => "\x1b[12",
                    3 => "\x1b[13",
                    4 => "\x1b[14",
                    5 => "\x1b[15",
                    6 => "\x1b[17",
                    7 => "\x1b[18",
                    8 => "\x1b[19",
                    9 => "\x1b[20",
                    10 => "\x1b[21",
                    11 => "\x1b[23",
                    12 => "\x1b[24",
                    13 => "\x1b[57376",
                    14 => "\x1b[57377",
                    15 => "\x1b[57378",
                    16 => "\x1b[57379",
                    17 => "\x1b[57380",
                    18 => "\x1b[57381",
                    19 => "\x1b[57382",
                    20 => "\x1b[57383",
                    21 => "\x1b[57384",
                    22 => "\x1b[57385",
                    23 => "\x1b[57386",
                    24 => "\x1b[57387",
                    _ => unreachable!(),
                };

                let c = match *n {
                    1..=12 => '~',
                    _ => 'u',
                };

                format!("{intro};{modifiers}{event_type}{c}")
            }
```

ovidiu-ionescu:
> This is still incorrect, as F1-F12 should end with `~`, so that most terminal programs that don't interpret characters in Private User Area can still recognize them.
> 
You are right; I did actually figure that out myself before reading your comment and my fix is quite similar.

ovidiu-ionescu:
@wez neovim seems to work Ok for me after the last commit.
Is there anything else you want done in this PR?

wez:
Thanks!

PR #4005, opened 2023-07-16
Doc update: Source install's cargo build option

seungjin:
This is documentation/blog update. 
Updated installation for source/UNIX update. 

Syncing with `README-DISTRO-MAINTAINER.md` .

### DIFF
--- a/docs/install/source.md
+++ b/docs/install/source.md
@@ -49,7 +49,7 @@ If your distribution has X11 but not Wayland, then you can build WezTerm without
 Wayland support by changing the `build` invocation:
 
 ```console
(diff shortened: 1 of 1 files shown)


wez:
Thanks for this, but I don't want the normal build-from-source to use distro-defaults

seungjin:
Then at least you need "--feature" parameter.
```console
$ cargo build --release --no-default-features vendored-fonts
$ cargo build --release --no-default-features --features vendored-fonts
```

Without `--features`, it spits out 
`error: unexpected argument 'vendored-fonts' found`.

PR #4017, opened 2023-07-17
add plugin_dir field to RepoSpec

MLFlexer:
Add plugin_dir to enable the listing of plugin directories with `wezterm.plugin.list()`, see [https://github.com/wez/wezterm/issues/4015](url)

### DIFF
--- a/lua-api-crates/plugin/src/lib.rs
+++ b/lua-api-crates/plugin/src/lib.rs
@@ -12,7 +12,7 @@ use wezterm_dynamic::{FromDynamic, ToDynamic};
 struct RepoSpec {
     url: String,
     component: String,
-    plugin_dir: String,
+    plugin_dir: PathBuf,
 }
 
 /// Given a URL, generate a string that can be used as a directory name.
@@ -56,11 +56,7 @@ impl RepoSpec {
             anyhow::bail!("invalid repo spec {url}");
         }
 
-        let plugin_dir = RepoSpec::plugins_dir()
-            .join(&component)
-            .to_str()
-            .ok_or(anyhow!("invalid plugin_dir string"))?
-            .to_string();
+        let plugin_dir = RepoSpec::plugins_dir().join(&component);
 
         Ok(Self {
             url,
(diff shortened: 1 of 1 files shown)


lua-api-crates/plugin/src/lib.rs
```diff
@@ -12,6 +12,7 @@ use wezterm_dynamic::{FromDynamic, ToDynamic};
 struct RepoSpec {
     url: String,
     component: String,
+    plugin_dir: String,
```
wez, line 15: Let's use PathBuf here so that we can centralize the path -> string conversion into one place (the ToDynamic trait) and keep the code in here simpler

```suggestion
    plugin_dir: PathBuf,
```

lua-api-crates/plugin/src/lib.rs
```diff
@@ -55,7 +56,17 @@ impl RepoSpec {
             anyhow::bail!("invalid repo spec {url}");
         }
 
-        Ok(Self { url, component })
+        let plugin_dir = RepoSpec::plugins_dir()
+            .join(&component)
+            .to_str()
+            .ok_or(anyhow!("invalid plugin_dir string"))?
+            .to_string();
```
wez, line 63:
```suggestion
            .join(&component);
```

lua-api-crates/plugin/src/lib.rs
```diff
@@ -66,12 +77,22 @@ impl RepoSpec {
             .ok_or_else(|| anyhow!("{path:?} isn't unicode"))?
             .to_string();
 
+        let plugin_dir = RepoSpec::plugins_dir()
+            .join(&component)
+            .to_str()
+            .ok_or(anyhow!("invalid plugin_dir string"))?
+            .to_string();
```
wez, line 84:
```suggestion
            .join(&component);
```

wez (changes requested):
Thanks for this!
I have a suggested change, but it otherwise looks fine

lua-api-crates/plugin/src/lib.rs
MLFlexer: good idea, that is a lot more clean

wez:
Thank you!

PR #4025, opened 2023-07-19
ssh: apply HostKeyAlgorithms in ssh config

cycleke:
Some ssh servers may only support certain key algorithms, so wezterm needs to use HostKeyAlgorithms in the ssh configuration file for the cryptographic handshake.


wez:
Thanks!
Note that libssh does read this value from the `~/.ssh/config` file. This PR just makes it possible for wezterm's lua config to pass the value down via the `ssh_option` portion of https://wezfurlong.org/wezterm/config/lua/SshDomain.html

PR #4050, opened 2023-07-23
Support [https://pop.system76.com/](POP!_OS) in get-deps

daniel-vainsencher:
A debian variant IIUC


daniel-vainsencher:
Thank you! I like wezterm, though I don't know all its features, recently learned to use the splitting of panes and it is very smooth with TUIs

PR #4072, opened 2023-07-30
docs: fix building without wayland command

shahamran:
The command provided to build from source without Wayland support is malformed:

```console
$ cargo build --release --no-default-features vendored-fonts
error: unexpected argument 'vendored-fonts' found
```

I think the intent was to disable default features and add the `vendored-fonts` feature, so I added the `--features` flag.


wez:
Thanks!

Issue #3440, opened 2023-04-03
force_reverse_video_selection


John-Gee:
Hello,

**Describe the solution you'd like**
Following the Discussion #3432 , I would like for wezterm to be able to reverse fg and bg colors on selection like other terminal emulators can do. Since I was told no setting currently exists for this, I am opening this FR.

**Additional context**
![reverse_color_selection](https://user-images.githubusercontent.com/29902477/229546181-54773b51-99a8-4355-a2f4-bbaead49c598.png)

Thank you!

wez:
FWIW, the logic for this stuff is already rather complex:

https://github.com/wez/wezterm/blob/45db43aa95f837c680da8dbe08f8f8a27b5cc680/wezterm-gui/src/termwindow/render/mod.rs#L490-L677

I'm not eager to try adding more options in here unless there is a really clear and pressing need.
There would need to be answers for questions like:

* How do you treat reverse video cursor coupled with reverse video selection?

and other combinations of factors/configs.

John-Gee:
Hmmm, I don't really know much about this so I'll take you word for it about complexity, but for reverse video cursor coupled with reverse video selection couldn't you disallow having both for a start? Since you already have the cursor one you could let that one take priority, and so if both are set the selection one would have no effect.

Tomiyou:
As far as I know, the logic for this should be fairly simple. If reverse video for selection is enabled in config, simply swap foreground and background color for selected text. I looked at the code, here is what that would look like IMO:
```
            // Selected text overrides colors
            (true, _, _, CursorVisibility::Hidden) => if self.config.force_reverse_video_selection {(
                params.bg_color,
                params.fg_color,
                params.cursor_bg,
            )} else {(
                params.selection_fg.when_fully_transparent(params.fg_color),
                params.selection_bg,
                params.cursor_bg,
            )},
```
While it is not a pressing need, this is the default behavior in gnome terminal, terminator, konsole, kitty and a couple more I think. It ensures that selected text is always very easy to read and contrasts the background nicely (no matter the colors), since the selection is simply the user theme color swapped. I am used to it to the point, that I actually can't live without it :sweat_smile:

I tried to implement this feature myself. I ran into an issue however, right now the selection in a line is rendered as a rectangle with a single color:
```
            let mut quad = self
                .filled_rectangle(
                    layers,
                    0,
                    euclid::rect(start, params.top_pixel_y, width, cell_height),
                    params.selection_bg,
                )
[...]

PR #4093, opened 2023-08-04
Add force_reverse_video_selection config

Tomiyou:
This config option allows selection to behave similarly as the cursor does when force_reverse_video_cursor config option is set to true. With this new config set to true, selected text's foreground and background colors are swapped. Addresses issue #3440


Tomiyou:
There is one thing missing from this PR. When the cursor itself is in reverse video mode, and the cursor lies inside the users selection, the cursors colors should be swapped again. Example: user has `force_reverse_video_cursor` option enabled and puts the cursor over the character `a` on the shell prompt. This character now has background and foreground colors swapped. Lets say the original BG color is black and original FG color is white. Reverse video swaps these colors, so BG color is now white and FG color is now black. But then the users selects text in his terminal, this character `a` and cursor included. Now the cursor's color should be swapped again, so the actual BG color is black again and FG color is white again.

I know it sounds complicated, but when using it, it intuitively makes sense. I would need some help writing that code, since `compute_cell_fg_bg()` function seems to use `params.selected` differently than what I assumed it means.

John-Gee:
@wez Is this too complicated for you to accept? I think some of users are really eager to have such a common functionality. Thank you!

LunarLambda:
Would love to have this

s4my:
Why isn't this merged yet?

wez:
Folks, I don't have a lot of spare time, and I'm out numbered by you all. Each time someone mentions me directly on an issue or a PR like this, without themselves making progress on the issue, all it achieves is using up what time I have to see if it was important.

If you want to see this PR move faster, I'm afraid that you'll just have to be patient and wait for me to have the right amount of time and inspiration to tackle the backlog.

PR #4130, opened 2023-08-10
Title argument

### LINKED ISSUE #4118: --title argument supprt
I'm missing the `--title` argument to override a window's title completely (as an argument to `wezterm start`), I think most terminals have this, so it would be nice.
e.g.:
```sh
wezterm start --title "Override Title" -- vim
```
(The title will still be "Override Title" even though vim is runing)

As far as I can tell there are no config options to do this (with `--config`), but correct me if I'm wrong.

diniamo:
Adds a `-T, --title` argument to `wezterm start`. E.G.:
```sh
wezterm start -T "I'm a terminal!"   # The title will be the string provided no matter what
wezterm start --title "Hello world!" # Same here
```

I probably missed a few things, and I'm not sure about a few others, please point out or ask anything that should be changed.

Closes #4118


rare-magma:
it would be great to have this feature in wezterm! @wez

ricardoseriani:
This feature will be very helpful, so I can have a fixed title in every terminal, making easy to search using rofi or any other solution.

Zeioth:
I'd like to bring attention to this PR. Because not having this is a deal breaker.

port19x:
I'm setting a 10 USD bounty on the merge of this PR.
Paid via github sponsorship

th3r00t:
Any update  on this?

axelkar:
C'mon @wez it's been (Edit: as rounded by GitHub) two years... No review even?

wezterm-gui/src/termwindow/mod.rs
```diff
@@ -85,12 +85,21 @@ use prevcursor::PrevCursorPos;
 const ATLAS_SIZE: usize = 128;
 
 lazy_static::lazy_static! {
+    static ref WINDOW_TITLE: Mutex<Option<String>> = Mutex::new(None);
```
axelkar, line 88: Implementing `--title` this way irreversibly (Lua can't control it anymore) overrides the window title for all windows. With single-instance being the default this is worsened even more.

axelkar (changes requested):

wezterm-gui/src/termwindow/mod.rs
diniamo: I would have done small nits, but I'm not going to do anything more, when I don't use Wezterm. You are free to use my code (if it helps at all), and/or open a new PR for this feature.

wez:
FWIW, I did take a peek at this when it first popped up, and didn't agree with the implementation approach.  I had planned to come back to this to comment to that effect, but I've had a lot happening in my life and it got away from me.

Since the OP is no longer using wezterm, let's close this one.

If someone wants to pick up the idea of setting an initial title, the implementation needs to set the title just for the first pane created by that wezterm invocation.

There are additional questions that should be asked, and answered, about how that might interact with an invocation that re-uses an existing gui session; I think there would need to be some additional work in the mux protocol to facilitate spawning with an initial title, which would then imply that spawning in general should have the option to set the title, and then there should also be fan out to all API surfaces that spawn to expose that argument.

There's a lot more to a fully implemented version of this PR that there appears.

PR #4151, opened 2023-08-15
x11: Use modifiers from X event if none were detected by XKB

bew:
This is a working tentative at fixing text expansion/injection done by Espanso by default, where XKB doesn't seem to detect the pressed modifiers `Ctrl+Shift` before pressing `v`.

ref: https://github.com/wez/wezterm/issues/3840#issuecomment-1677974127 (and following comments)

(( So with this patch, XKB's state is not updated and the key event' modifiers are kind of transient: only existing for this event. ))

Edit: waiting for your input on https://github.com/wez/wezterm/issues/3840#issuecomment-1680087150

### DIFF
--- a/window/src/os/x11/keyboard.rs
+++ b/window/src/os/x11/keyboard.rs
@@ -166,7 +166,7 @@ impl KeyboardWithFallback {
         events: &mut WindowEventSender,
     ) -> Option<WindowKeyEvent> {
         let want_repeat = self.selected.wayland_key_repeats(code);
-        self.process_key_event_impl(code + 8, pressed, events, want_repeat)
+        self.process_key_event_impl(code + 8, None, pressed, events, want_repeat)
     }
 
     pub fn process_key_press_event(
@@ -175,7 +175,8 @@ impl KeyboardWithFallback {
         events: &mut WindowEventSender,
     ) {
         let xcode = xkb::Keycode::from(xcb_ev.detail());
-        self.process_key_event_impl(xcode, true, events, false);
+        let xmods = Some(xcb_ev.state()); // modifiers from event
+        self.process_key_event_impl(xcode, xmods, true, events, false);
     }
 
     pub fn process_key_release_event(
@@ -184,18 +185,20 @@ impl KeyboardWithFallback {
         events: &mut WindowEventSender,
     ) {
         let xcode = xkb::Keycode::from(xcb_ev.detail());
-        self.process_key_event_impl(xcode, false, events, false);
+        let xmods = Some(xcb_ev.state()); // modifiers from event
+        self.process_key_event_impl(xcode, xmods, false, events, false);
     }
 
     fn process_key_event_impl(
         &self,
         xcode: xkb::Keycode,
+        xmods: Option<xcb::x::KeyButMask>,
         pressed: bool,
         events: &mut WindowEventSender,
         want_repeat: bool,
     ) -> Option<WindowKeyEvent> {
         let phys_code = self.selected.phys_code_map.borrow().get(&xcode).copied();
-        let raw_modifiers = self.get_key_modifiers();
+        let raw_modifiers = self.get_key_modifiers(xmods);
         let leds = self.get_led_status();
 
         let xsym = self.selected.state.borrow().key_get_one_sym(xcode);
@@ -380,7 +383,7 @@ impl KeyboardWithFallback {
         leds
     }
 
-    pub fn get_key_modifiers(&self) -> Modifiers {
+    pub fn get_key_modifiers(&self, fallback_xmods: Option<xcb::x::KeyButMask>) -> Modifiers {
         let mut res = Modifiers::default();
 
         if self.mod_is_active(xkb::MOD_NAME_SHIFT) {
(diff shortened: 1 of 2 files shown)


bew:
Hi @wez I'd be interested to know if you'd accept this change as a workaround for Espanso (and maybe other tools) that send mods in the XInput event?

wez:
Sorry it's taken a while to get to this.

Today I was screwed over by btrfs and had to reinstall my workstation. As part of that, I switched from Fedora to Ubuntu. My keyboard has a custom "paste" button that sends "shift-Insert" when pressed, and this key misbehaves in wezterm under ubuntu; the shift modifier goes missing from the actual Insert event. It worked fine in Fedora. I don't know what the difference is; both systems are Gnome XOrg systems.  There's probably something weird in the Ubuntu configuration that causes this, but I have no desire to investigate it.

Running with your PR seems to fix that, so yes, I am now suddenly motivated to see this PR get finished up! :)

window/src/os/x11/keyboard.rs
```diff
@@ -397,6 +400,27 @@ impl KeyboardWithFallback {
             // Mod4
             res |= Modifiers::SUPER;
         }
+
+        // When XKB doesn't have modifiers in its state but there are modifiers given in the X
+        // event, use these as fallback.
+        //
+        // This can happen (FIXME: WHY) for example when Espanso (a text expander/injector)
+        // simulate inputs.
+        if let (true, Some(event_xmods)) = (res.is_empty(), fallback_xmods) {
```
wez, line 409: one thing I'm a little unsure about here is how to distinguish between:

* xkb state knew the modifier state but legitimately produced an empty set
* xkb has no idea and produced an empty set of mods

I'm not sure if this is purely a theoretical distinction or whether there is something more tricky we need to do in order to be sure about this.

Is there a return code we can use to influence this choice?

wez: or maybe it doesn't matter and we should unconditionally "or" the event-provided modifiers in?

bew: I don't think we can do this distinction, but I'd argue it doesn't matter much IMO.

Currently the PR uses the event mods as a fallback only, however I'm thinking it could be different:
1. when the X event is synthetic (was crafted and sent by another app), use event mods _instead_ of xkb detected mods
2. when the X event has mods, use event mods _instead_ of xkb detected mods
3. merge xkb mods and event mods (as you suggested)

Not sure what's the best direction here..
I think the behavior 1 can be a good idea for synthetic events, because if an app sends an xevent to another app it might want a specific set of mods, but at the same time it could also have sent events to indicate mods are pressed followed by the keypress/release event..

Maybe we could do 1 & 2 or 1 & 3, WDYT ?

It would be interesting to have more data about your issue, can you give the output of `xev -event keyboard` and pressing your paste key?

wez: Just dropping by to share the logging, haven't had a chance to ponder the questions so far.

```
KeyPress event, serial 28, synthetic NO, window 0x2e00001,
    root 0x6d1, subw 0x0, time 64899989, (300,145), root:(344,1595),
    state 0x0, keycode 50 (keysym 0xffe1, Shift_L), same_screen YES,
    XLookupString gives 0 bytes:
    XmbLookupString gives 0 bytes:
    XFilterEvent returns: False

KeyPress event, serial 28, synthetic NO, window 0x2e00001,
    root 0x6d1, subw 0x0, time 64899989, (300,145), root:(344,1595),
    state 0x1, keycode 118 (keysym 0xff63, Insert), same_screen YES,
    XLookupString gives 0 bytes:
    XmbLookupString gives 0 bytes:
    XFilterEvent returns: False

KeyRelease event, serial 28, synthetic NO, window 0x2e00001,
    root 0x6d1, subw 0x0, time 64899990, (300,145), root:(344,1595),
    state 0x1, keycode 118 (keysym 0xff63, Insert), same_screen YES,
    XLookupString gives 0 bytes:
    XFilterEvent returns: False

KeyRelease event, serial 28, synthetic NO, window 0x2e00001,
    root 0x6d1, subw 0x0, time 64899991, (300,145), root:(344,1595),
    state 0x1, keycode 50 (keysym 0xffe1, Shift_L), same_screen YES,
    XLookupString gives 0 bytes:
    XFilterEvent returns: False
```

wez:
I took a peek at this this morning, and the situation I'm seeing frustrated me: in my case, what is happening is that there is a keyboard state notification from the X server that *sometimes* clears all the modifier masks prior to delivering the `Insert` key press.  I don't have a good explanation for what is happening or why.

I refactored your PR slightly to make the flow a bit easier to see; now instead of passing around an `Option<BtnMask>`, `Modifiers` is computed from the mask in the xcb event handlers and passed in and ORed in with the other modifier bits.

I've pushed this to main; it seems to resolve the issue that I'm facing and the I think the risk for misbehavior is relatively low.
We'll see if anyone surfaces with a complaint!

Thanks!

bew:
Much simpler indeed 👍, so I see you went with merging mods, we'll see if it's right :)

It's weird that this change makes your key work though, I don't see how it can be related `¯\_(ツ)_/¯`

PR #4160, opened 2023-08-18
Add zoom-pane subcommand

quantonganh:
Hi,

I'm currently working on integrating Helix with WezTerm to perform [interactive global search](https://github.com/helix-editor/helix/issues/196).

The mechanism I'm implementing involves sending a command to the bottom pane when a key is presseed:

```sh
case "$1" in
  "fzf")
    split_pane_down
    echo "hx-fzf.sh \$(rg --line-number --column --no-heading --smart-case . | fzf --delimiter : --preview 'bat --style=full --color=always --highlight-line {2} {1}' --preview-window '~3,+{2}+3/2' | awk '{ print \$1 }' | cut -d: -f1,2,3)" | $send_to_bottom_pane
    ;;
```

Once a selection is made, the file will be opened by Helix in the top pane. Now I want to hide the bottom pane but looks like WezTerm [does not support](https://github.com/wez/wezterm/discussions/3779) this functionality.

While I'm aware that it's possible to bind a key to [toggle the zoom state](https://wezfurlong.org/wezterm/config/lua/keyassignment/TogglePaneZoomState.html) of the top pane, this requires an additional key press. Furthermore, I'm looking to avoid the option of [killing](https://wezfurlong.org/wezterm/cli/cli/kill-pane.html) the bottom pane.

Thus, this PR has been initiated to introduce a CLI-based method for toggling the zoom state of the current pane.

### DIFF
--- a/wezterm-mux-server-impl/src/sessionhandler.rs
+++ b/wezterm-mux-server-impl/src/sessionhandler.rs
@@ -532,10 +532,23 @@ impl SessionHandler {
                             let tab = mux
                                 .get_tab(containing_tab_id)
                                 .ok_or_else(|| anyhow!("no such tab {}", containing_tab_id))?;
-                            tab.set_zoomed(false);
-                            if zoomed {
-                                tab.set_active_pane(&pane);
-                                tab.set_zoomed(zoomed);
+                            match tab.get_zoomed_pane() {
+                                Some(p) => {
+                                    let is_zoomed = p.pane_id() == pane_id;
+                                    if is_zoomed != zoomed {
+                                        tab.set_zoomed(false);
+                                        if zoomed {
+                                            tab.set_active_pane(&pane);
+                                            tab.set_zoomed(zoomed);
+                                        }
+                                    }
+                                }
+                                None => {
+                                    if zoomed {
+                                        tab.set_active_pane(&pane);
+                                        tab.set_zoomed(zoomed);
+                                    }
+                                }
                             }
                             Ok(Pdu::UnitResponse(UnitResponse {}))
                         },
--- a/wezterm-mux-server-impl/src/sessionhandler.rs
+++ b/wezterm-mux-server-impl/src/sessionhandler.rs
@@ -532,8 +532,11 @@ impl SessionHandler {
                             let tab = mux
                                 .get_tab(containing_tab_id)
                                 .ok_or_else(|| anyhow!("no such tab {}", containing_tab_id))?;
-                            tab.set_active_pane(&pane);
-                            tab.set_zoomed(zoomed);
+                            tab.set_zoomed(false);
+                            if zoomed {
+                                tab.set_active_pane(&pane);
+                                tab.set_zoomed(zoomed);
+                            }
                             Ok(Pdu::UnitResponse(UnitResponse {}))
                         },
                         send_response,
@@ -1021,7 +1024,7 @@ async fn split_pane(split: SplitPane, client_id: Option<Arc<ClientId>>) -> anyho
 
     Ok::<Pdu, anyhow::Error>(Pdu::SpawnResponse(SpawnResponse {
         pane_id: pane.pane_id(),
-        tab_id: tab_id,
+        tab_id,
         window_id,
         size,
     }))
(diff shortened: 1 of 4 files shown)


wezterm-mux-server-impl/src/sessionhandler.rs
```diff
@@ -945,6 +945,26 @@ impl SessionHandler {
...
+                    catch(
+                        move || {
+                            let mux = Mux::get();
+                            let (_domain_id, _window_id, tab_id) = mux
+                                .resolve_pane_id(pane_id)
+                                .ok_or_else(|| anyhow!("no such pane {}", pane_id))?;
+                            let tab = mux
+                                .get_tab(tab_id)
+                                .ok_or_else(|| anyhow!("no such tab {}", tab_id))?;
+                            tab.toggle_zoom();
```
wez, line 959: Thanks for diving in!

There's a nuanced difference between the behavior listed in the CLI and what is actually happening here.

The CLI logic takes a pane_id parameter (defaulting to the pane from the environment), but rather than manipulating the zoom status of that pane, it toggles whichever pane is zoomed in its containing tab, rather than the requested pane.

Note that there is already a `SetPaneZoomed` PDU in the mux protocol; it also happens to have this same scoping issue.

What I think the forward should be is:

1. Fix `SetPaneZoomed` to explicitly check the zoomed status of the pane.  To support this, add a method to `Tab` and `TabInner` called `get_zoomed_pane` that returns `self.zoomed.clone()`.  Then the code can use `tab.set_zoomed(false)` to unset the zoom state directly, and then, if zooming the target pane, `tab.set_active_pane()` to activate the target tab and `tab.set_zoomed(true)` to make it zoomed.

2. With that operation fixed, the rest of this PR doesn't need to modify the codec; all the logic can reside in the CLI.
 
For the CLI I think it would be better overall if we provided a choice of three actions:
  * Explicitly set the zoom state to either true or false.
  * To toggle the zoom state
 
eg: name the subcommand `zoom-pane` and give it parameters: `--zoom`, `--unzoom`, and `--toggle` to select the behavior.  I would lean towards `--zoom` being the default because that feels like the most likely action for `zoom-pane` to take by default.

The zoom and unzoom cases are straightforward and can call `client.set_zoomed` and pass the parameters through.

For the toggle case, it's a bit more involved:

* Use `client.list_panes()` to obtain tab and pane information (see `move_pane_to_new_tab.rs`)
* Resolve the tab_id of the target tab from that information
* Resolve the pane_id that is active + zoomed
* If the active + zoomed pane_id matches the target pane, then use `client.set_zoomed` to set zoomed to false for the target pane
* Otherwise, use `client.set_zoomed` to set zoomed to true for the target pane.

Why not add a `TogglePaneZoomState` PDU and handle this on the server side? Such a toggle action would aggregate multiple functions which would make the mux protocol more complex overall.  Having discrete, composable PDUs make things simpler and more flexible in the long run.

quantonganh: I've just made the changes based on your suggestion. Please review. Thank you.

wezterm/src/cli/zoom_pane.rs
```diff
@@ -0,0 +1,109 @@
...
+            } else {
+                client
+                    .set_zoomed(SetPaneZoomed {
+                        containing_tab_id,
+                        pane_id,
+                        zoomed: true,
+                    })
+                    .await?;
+            }
+        }
```
wez, line 106: I think this can simplify as well:

```suggestion
        if self.toggle {
            let is_zoomed = tab_id_to_active_zoomed_pane_id.get(&containing_tab_id).copied() == Some(pane_id);
            
                client
                    .set_zoomed(SetPaneZoomed {
                        containing_tab_id,
                        pane_id,
                        zoomed: !is_zoomed,
                    })
                    .await?;
        }
```

wezterm/src/cli/zoom_pane.rs
```diff
@@ -0,0 +1,109 @@
...
+
+        if self.unzoom {
+            client
+                .set_zoomed(SetPaneZoomed {
+                    containing_tab_id,
+                    pane_id,
+                    zoomed: false,
+                })
+                .await?;
+        }
```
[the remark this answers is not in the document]
quantonganh: You made me realized that we have to unset the default value for `zoom` if `unzoom`/`toggle` is present.

wez:
Thanks!

PR #4166, opened 2023-08-18
add `awesome` to default `tiling_desktop_environments`.

UserIsntAvailable:


docs/config/lua/config/tiling_desktop_environments.md
```diff
@@ -15,6 +15,7 @@ config.tiling_desktop_environments = {
   'X11 bspwm',
   'X11 i3',
   'X11 dwm',
+  'X11 awesome',
 }
```
wez, line 19: Adding this entry to this list is misleading when read with the version information at the top of the page.

Please amend this so that instead of adding it to this code block, you add a section below it:

```

{{since('dev')}}

The following additional entries are now part of the default value of `tiling_desktop_environments`:

 *  '"X11 awesome"'

```

wez:
Thanks! I have a suggested edit, but because it intersects with a fenced code block, I'm unable to propose and edit it directly in the PR.  Please amend the change in, and we can land this!

Thanks!

UserIsntAvailable:
Everything should be fine now :sweat_smile:

wez:
Thanks!

Issue #4173, opened 2023-08-20
Window padding background follows background color of adjacent cell


murtaza64:
It would be a nice-to-have aesthetic enhancement for the background of the padding around the terminal to follow the background color of adjacent cells. This would make statuslines in programs like vim look marginally better.

##Mockup screenshots

before:

![image](https://github.com/wez/wezterm/assets/13615693/862a81c8-c0f2-40f8-a06d-c71f0a4cc168)

after:
![image](https://github.com/wez/wezterm/assets/13615693/ac8964cf-79fd-417e-8292-11f8e6d502bb)

Not experienced with building terminals so no idea how difficult or performance-intensive this would be, but I feel like it would look nice :)

murtaza64:
I am bored so I will try my own hand at this before asking for the help of someone more experienced

wez:
Thanks for the suggestion.

However, I'm not interested in taking a PR to implement this enhancement because that would make it difficult to later make changes to how panes are composited and/or allowing changing the font size of individual panes.

murtaza64:
I spent the evening implementing the feature, and the way I implemented it it only considers the outer window borders, not the border between panes. If you have time. I would love if you could take a look (#4175). I'll attach some screenshots there

quot:
I would love to see this implemented. I really like using strong color schemes in TUI apps to know what is open at a glance. But having the padding color never change makes it really stand out.

This is with Helix's default theme:
![image](https://github.com/user-attachments/assets/45151dcd-2f4c-4df6-9428-a9956c8745ff)

This is 100% a nitpick on my side, and I completely understand skipping this if it would create too much work, but I just wanted to show my support for the feature request.

PR #4175, opened 2023-08-20
Implement padding background bleed

murtaza64:
Implements #4173, adding config options which make background colors bleed into the padding on the inside of the window border, which may for some people make things like statuslines in vim look a bit nicer.

This is my first PR here, and my first PR in Rust, so any help/criticism is appreciated, whether project-specific, rust style, etc.


murtaza64:
This is currently a proof of concept, remaining to be cleaned up, but here are some screenshots:

Horizontal bleed only
image

Horizontal and vertical bleed (note corners)
image

Multiple panes
image

murtaza64:
@wez after thinking about what you said in the issue some more, I do think this implementation will cause the problems you described, since it relies on the total row and column count of the window to figure out when to draw extra background rectangles. However, if you are planning to have each pane have its own row/col dimensions, I think it's possible to instead use the pane position alone to figure out if a particular line is touching the edge of the window. Is there any work on the updated panes I can take a look at to perhaps change my approach?

murtaza64:
Currently a bug in the bottom right corner implementation. Looking into it

murtaza64:
Maintainer doesn't want this feature (fair enough)

PR #4194, opened 2023-08-25
Update source.md

lengyijun:
Compile and install wezterm to /usr/local/bin


wez:
Thanks for this, however, I do not want to show how to "install" like this, because I foresee folks running this without understanding it, and then asking for help to uninstall it.

Issue #3803, opened 2023-05-30
default hyperlink_rules is wrong for Markdown link format


yutkat:
### What Operating System(s) are you seeing this problem on?

Linux Wayland

### Which Wayland compositor or X11 Window manager(s) are you using?

Sway

### WezTerm version

20230529-192946-62c40118

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

If the link is Markdown format, I cannot jump to the site by hyperlink. Why is a `)` included here?

https://github.com/wez/wezterm/blob/c6f4ff3626f66f5bc1c64978faf3dbf1e3b6106c/config/src/config.rs#L1619

Or the following rule should be applied first.

```rust
hyperlink::Rule::with_highlight(r"\((\w+://\S+)\)", "$1", 1).unwrap(),
```

### Configuration

no config

### Expected Behavior

This format will correctly link to `https://github.com`. 

```
[github](https://github.com)
```

Now it is linked to `https://github.com%29`.

nobeans:
I've just experiencing the same issue. It seems that the regular expression in [2] of the default settings is causing the problem:

```lua
  -- Matches: a URL in parens: (URL)
  {
    regex = '\\((\\w+://\\S+)\\)', -- [1]
    format = '$1',
    highlight = 1,
  },
  -- ...(snipped)...
  -- Then handle URLs not wrapped in brackets
  {
    regex = '\\b\\w+://\\S+[)/a-zA-Z0-9-]+', -- [2]
    format = '$0',
  },
```

The closing brace is included in the pattern, causing it to take precedence over the [1] pattern. I fixed it in my .wezterm.lua file as follows, and it seems to be working correctly:

```lua
  -- Then handle URLs not wrapped in brackets
  {
    regex = '[^(]\\b(\\w+://\\S+[)/a-zA-Z0-9-]+)',
    format = '$1',
    highlight = 1,
  },
```

I tested it with the following text to verify:

```
- (http://localhost/foo)
- [http://localhost/foo]
- {http://localhost/foo}
- <http://localhost/foo>
- http://localhost/foo
```

nobeans:
.wezterm.lua for workaround:

```lua
local wezterm = require 'wezterm'
local config = {}

config.hyperlink_rules = {
  -- Matches: a URL in parens: (URL)
  {
    regex = '\\((\\w+://\\S+)\\)',
[...]

PR #4212, opened 2023-08-28
Adjust hyperlink rules to handle links containing ")" appropriately (#3803)

nobeans:
The current hyperlink rules match eagerly for link that contain ")" like this:
```
- (http://localhost/foo)
   ^^^^^^^^^^^^^^^^^^^^^ --> "http://localhost/foo)"
- [FOO](http://localhost/foo)
        ^^^^^^^^^^^^^^^^^^^^^ --> "http://localhost/foo)"
```

This PR makes hyperlink rules match them appropriately like this:
```
- (http://localhost/foo)
   ^^^^^^^^^^^^^^^^^^^^ --> "http://localhost/foo"
- [FOO](http://localhost/foo)
        ^^^^^^^^^^^^^^^^^^^^ --> "http://localhost/foo"
```

See #3803 for more context.

### DIFF
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -1622,7 +1622,7 @@ pub fn default_hyperlink_rules() -> Vec<hyperlink::Rule> {
         hyperlink::Rule::with_highlight(r"<(\w+://\S+)>", "$1", 1).unwrap(),
         // Then handle URLs not wrapped in brackets
         // and include terminating ), / or - characters, if any
-        hyperlink::Rule::new(r"\b\w+://\S+[)/a-zA-Z0-9-]+", "$0").unwrap(),
+        hyperlink::Rule::with_highlight(r"[^(]\b(\w+://\S+[)/a-zA-Z0-9-]+)", "$1", 1).unwrap(),
         // implicit mailto link
         hyperlink::Rule::new(r"\b\w+@[\w-]+(\.[\w-]+)+\b", "mailto:$0").unwrap(),
     ]
--- a/docs/config/lua/config/hyperlink_rules.md
+++ b/docs/config/lua/config/hyperlink_rules.md
@@ -68,8 +68,9 @@ config.hyperlink_rules = {
   },
   -- Then handle URLs not wrapped in brackets
   {
-    regex = '\\b\\w+://\\S+[)/a-zA-Z0-9-]+',
-    format = '$0',
+    regex = '[^(]\\b(\\w+://\\S+[)/a-zA-Z0-9-]+)',
+    format = '$1',
+    highlight = 1,
   },
   -- implicit mailto link
   {


config/src/config.rs
```diff
@@ -1622,7 +1622,7 @@ pub fn default_hyperlink_rules() -> Vec<hyperlink::Rule> {
         hyperlink::Rule::with_highlight(r"<(\w+://\S+)>", "$1", 1).unwrap(),
         // Then handle URLs not wrapped in brackets
         // and include terminating ), / or - characters, if any
-        hyperlink::Rule::new(r"\b\w+://\S+[)/a-zA-Z0-9-]+", "$0").unwrap(),
+        hyperlink::Rule::with_highlight(r"[^(]\b(\w+://\S+[)/a-zA-Z0-9-]+)", "$1", 1).unwrap(),
```
wez, line 1625: Does this need to use a capturing group?  Seems like the diff could just be to prefix `[^(]` ?

nobeans: Yes, it needs. An extra character is included in a link if the capturing group is not used, like this:
```
        regex = '[^(]\\b\\w+://\\S+[)/a-zA-Z0-9-]+',
        format = '$0',
```
```
- [http://localhost/foo]
  ^^^^^^^^^^^^^^^^^^^^^ --> "[http://localhost/foo"
- http://localhost/foo
 ^^^^^^^^^^^^^^^^^^^^^ --> " http://localhost/foo"
```

rcloran:
I think this change breaks "normal" links (ie, surrounded by whitespace). I haven't tested the actual patch, but changed my config to include the regex:

```
config.hyperlink_rules = wezterm.default_hyperlink_rules()
if config.hyperlink_rules[4].regex == "\\b\\w+://\\S+[)/a-zA-Z0-9-]+" then
        config.hyperlink_rules[4].regex = '[^(]\\b(\\w+://\\S+[)/a-zA-Z0-9-]+)'
end
```

You can see what happens from the logs (patched config, then default config):

```
16:27:38.527  INFO   wezterm_gui::termwindow > clicking  https://accounts.myq-cloud.com/api/v6.0/accounts
16:27:49.942  INFO   wezterm_gui::termwindow > clicking https://accounts.myq-cloud.com/api/v6.0/accounts
```

That is, an extra leading space is captured.

I also noticed that any links already in the buffer will not be recalculated, so testing requires something like pasting the link in again, which left me confused for a while. Hopefully I haven't muddled anything up while I've been switching configs around.

rcloran:
> I think this change breaks "normal" links

Oh, I messed up  -- you even mentioned that problem in the review response to @wez... Adding `config.hyperlink_rules[4].format = '$1'` fixes it.

The one undesirable thing here is that because leading whitespace is now part of the pattern, the highlight extends to include it. But that's not a huge issue.

arcstur:
Hello, any progress here? :))

nobeans:
I'm not sure either. If there's an issue, I'm open to fixing it...?

wez:
I just haven't had time to review this

Tigerfyj:
I got a similar issue, but the actual link is appended another ".html".

What shows in the console:
```
  -  For more details, please see the link (https://foo.com/bar/baz.html)
```

What shows in the chrome address box: `https://foo.com/bar/baz.html).html`

Felixoid:
ended up here because of the same issue. How likely it would move forward?..

hankertrix:
This regex should work for links with only 1 pair of brackets:
```
\b\w+://[^\[({<>})\]\s]+[\[({][^\[({<>})\]\s]+?[\])}][^\[({<>})\]\s]*
```

This regex should work for links without brackets:

```
\b\w+://[^\[({\s]+\b
```

Test string used:
```
-  For more details, please see the link (https://foo.com/bar/baz.html)
16:27:38.527  INFO   wezterm_gui::termwindow > clicking  https://accounts.myq-cloud.com/api/v6.0/accounts
16:27:49.942  INFO   wezterm_gui::termwindow > clicking https://accounts.myq-cloud.com/api/v6.0/accounts

https://accounts.myq-cloud.com/api/v6.0/accounts

1. [![Sponsor WezTerm](https://img.shields.io/github/sponsors/wez?label=Sponsor%20WezTerm&logo=github&style=for-the-badge)](https://github.com/sponsors/wez)
2. [![C_(programming_language)](https://img.shields.io/github/sponsors/wez?label=Sponsor%20WezTerm&logo=github&style=for-the-badge)](https://en.wikipedia.org/wiki/C_(programming_language)_fantastic)
3. [C_(programming_language)](https://en.wikipedia.org/wiki/C_(programming_language))
4. (https://en.wikipedia.org/wiki/C_(programming_language))
5. [https://en.wikipedia.org/wiki/C_(programming_language)]
6. {https://en.wikipedia.org/wiki/C_(programming_language)}
7. <https://en.wikipedia.org/wiki/C_(programming_language)>
8. [Duck Duck Go](https://duckduckgo.com "search engine")
9. https://en.wikipedia.org/wiki/C_(programming_language)
```

wez:
Thanks for this, however, this was fixed differently in this PR that I merged recently:
* #6391

Issue #4226, opened 2023-08-30
Additions to the InputSelector


Danielkonge:
**Is your feature request related to a problem? Please describe.**
A clear and concise description of what the problem is.

I am trying to use the InputSelector more for choosing from longer lists and would like a few more options.

**Describe the solution you'd like**

My primary suggestion is to add a longer alphabet to choose from with one click when opening the InputSelector. 

Right now it only shows numbers [1-9] for the first 9 elements of the list, but it would be nice to show [1-9] for the first 9 elements and then [a-z] for the next 26 elements and possibly [A-Z] beyond that for anyone using small enough fonts to see that many elements. 

Alternatively, one could choose the exact labels themselves, but this would probably require slightly more work, and I assume my first suggestion would be easier to implement.

**Describe alternatives you've considered**

I have also thought about jumping straight to the filter part of the InputSelector to avoid picking out elements like above. It would be nice if there was a flag "start_with_filter" (or something like that) for the InputSelector that is "false" if you want to start the InputSelector as it does now and "true" if it essentially does what would be equivalent to starting the InputSelector and then pressing "/" now. 

Finally it would be nice if we were able to edit the text that currently says:
"Select an item and press Enter=accept Esc=cancel /=filter"

**Additional context**

Here is a picture where what I would like to add is "a. " in front of the line with .pylint.d, "b. " in front of the line with .ipython, and so on. 

<img width="481" alt="Screenshot 2023-08-30 at 17 22 09" src="https://github.com/wez/wezterm/assets/7075380/5c7dccee-15e2-4959-ba7d-7d0707c3e86c">

If I am pointed to the code that would need changing to implement this, I can try to take a look at it myself, but I don't really have any working experience with rust. Though I assume the above would hopefully be relatively simple?

wez:
These sound like reasonable requests.

If you feel like taking a crack at coding some of this, then I'd suggest starting with the alphabet, as that doesn't require changing any of the other config related structures to get it going.

It should be possible to re-use the logic from quick select, which is encapsulated in this function:
https://github.com/wez/wezterm/blob/99c9613966a73d5388344fd0df68d5977b493c7c/wezterm-gui/src/overlay/quickselect.rs#L59

You pass in the alphabet (eg: "1234567890abcdefghijklmnopqrstuvwxyz") and the number of entries in the list (or rather, the number of lines that can seen on the screen) and it will return a list of corresponding strings that represent each entry in the list.

Then the logic over here in the selector that matches a label by number:
[...]

PR #4227, opened 2023-08-31
Additions to the InputSelector (resolving #4226)

Danielkonge:
These changes should resolve my requests in #4226 . As mentioned in the issue this is my first time working in Rust, so please let me know if I am doing something weird in any of the submitted code. I have tested this by changing stuff around in my config, and everything seems to work as I would expect, but there might still be tiny bugs. 

I don't know if we might want to move some of the logic a bit? E.g., do we want global config options for the InputSelector? 

Also, I was thinking about adding colors in the selection when you press the first key in a two key key-sequence, but I haven't gotten around to that yet (and I am not entirely sure where to start).


wezterm-gui/src/overlay/selector.rs
```diff
@@ -368,6 +403,7 @@ pub fn selector(
         always_fuzzy: args.fuzzy,
         args,
         event_name,
+        selection: String::new(),
```
wez, line 406: see other comments for context on this!

```suggestion
        selection: String::new(),
        labels: vec![],
        max_items: 0,  -- and remove the unused size and max_items local vars above
```

wezterm-gui/src/overlay/selector.rs
```diff
@@ -178,25 +189,46 @@ impl SelectorState {
     }
 
     fn run_loop(&mut self, term: &mut TermWizTerminal) -> anyhow::Result<()> {
+        let max_items = self.max_items;
+        let alphabet = self.args.alphabet.to_lowercase();
+        let alphabet_has_j = alphabet.contains("j");
```
wez, line 194: This seems like a smart approach. Should `/` also have similar treatment? Are there other unmodified `Char` patterns below that might also need similar treatment?

FWIW, I think it would be ok to inline this check where the event is being matched; I'm less concerned about the runtime cost of `alphabet.contains` than I am about the readability of this function, and I don't think that having a separate local variable that is many lines apart from the logic helps with readability.

wezterm-gui/src/overlay/selector.rs
```diff
@@ -178,25 +189,46 @@ impl SelectorState {
...
                 }
                 InputEvent::Key(KeyEvent {
                     key: KeyCode::Char('k'),
                     ..
+                }) if !self.filtering && !alphabet_has_k => {
+                    self.move_up();
+                }
+                InputEvent::Key(KeyEvent {
+                    key: KeyCode::Char('J'),
+                    ..
```
wez, line 225: I'm not sure about this uppercase version of the movement keys.
In the launcher overlay we have CTRL-J/K as alternatives to CTRL-N/P; I'd be cool with having those here instead of this.

https://github.com/wez/wezterm/blob/99c9613966a73d5388344fd0df68d5977b493c7c/wezterm-gui/src/overlay/launcher.rs#L482-L493

wezterm-gui/src/overlay/selector.rs
```diff
@@ -222,6 +254,9 @@ impl SelectorState {
                     key: KeyCode::Backspace,
                     ..
                 }) => {
+                    if !self.filtering && !self.selection.is_empty() {
```
wez, line 257: pop is safe to call even when empty

```suggestion
                    if !self.filtering {
```

wezterm-gui/src/overlay/selector.rs
```diff
@@ -222,6 +254,9 @@ impl SelectorState {
...
+                        self.selection.pop();
+                    }
```
wez, line 259: there should probably be an `else` here, otherwise backspace will unconditionally mutate the filter term even it deleted an element from the selection

wez (commented):
Thanks for this!

re: global config, no, I think having the options be part of the key assignment is good.

re: colors, I'm unsure about that. We don't have a similar concept in quickselect and we haven't had anyone raise the desire or need to add it there, so I suspect that it probably isn't necessary.

I've made a few suggestions!

wezterm-gui/src/overlay/selector.rs
wez: the `max_items: 0` bit is here so that the logic to compute max_items and labels can be centralized to the same single location

docs/config/lua/keyassignment/InputSelector.md
```diff
@@ -22,6 +22,16 @@ upon the input.
...
   corresponding fields from the selected choice. Both will be `nil` if
   the overlay is cancelled without selecting anything.
+* `fuzzy` - a boolean that defaults to `false`. If `true`, InputSelector will start
+  in its fuzzy finding mode (this is equivalent to starting the InputSelector and
+  pressing / in the default mode).
+* `alphabet` - a string of unique characters. The characters in the string are used
+  to calculate one or two click shortcuts that can be used to quickly choose from
+  the InputSelector when not in fuzzy finding mode. Defaults to:
+  `"1234567890qwertyuiopasdfghjklzxcvbnm"`.
+* `fuzzy_description` - a string to display when in fuzzy finding mode. Defaults to:
```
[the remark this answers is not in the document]
Danielkonge: I think I already did this, but I forgot to push that change. It will be included when I update this.

wezterm-gui/src/overlay/selector.rs
```diff
@@ -69,25 +71,23 @@ impl SelectorState {
...
-            )),
+            Change::Text(format!("{}\r\n", truncate_right(desc, max_width))),
             Change::AllAttributes(CellAttributes::default()),
         ];
 
         let max_items = self.max_items;
+        let alphabet = &self.args.alphabet;
+        let mut labels = compute_labels_for_alphabet(alphabet, max_items + 1).into_iter();
+        let labels_len = labels.len();
+        let max_label_len = labels.clone().map(|s| s.len()).max().unwrap_or(0);
```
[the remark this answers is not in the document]
Danielkonge: Currently max_items doesn't label the last element on the screen (when the alphabet is large enough to do this). I might want to redefine max_items elsewhere so it has the expected meaning.

wezterm-gui/src/overlay/selector.rs
```diff
@@ -107,8 +107,19 @@ impl SelectorState {
                 attr.set_reverse(true);
             }
 
-            if row_num < 9 && !self.filtering {
-                changes.push(Change::Text(format!(" {}. ", row_num + 1)));
+            // from above we know that row_num <= max_items
+            // show labels as long as we have more labels left
+            // and we are not filtering
+            if !self.filtering && row_num < labels_len {
+                if let Some(s) = labels.next() {
```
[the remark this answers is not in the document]
Danielkonge: I changed to the suggested code now.

wezterm-gui/src/overlay/selector.rs
```diff
@@ -178,25 +189,46 @@ impl SelectorState {
     }
 
     fn run_loop(&mut self, term: &mut TermWizTerminal) -> anyhow::Result<()> {
+        let max_items = self.max_items;
+        let alphabet = self.args.alphabet.to_lowercase();
```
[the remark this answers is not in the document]
Danielkonge: Since compute_labels_for_alphabet starts like this:

```rust
pub fn compute_labels_for_alphabet(alphabet: &str, num_matches: usize) -> Vec<String> {
    let alphabet = alphabet
        .chars()
        .map(|c| c.to_lowercase().to_string())
        .collect::<Vec<String>>();
```

I think we need to make sure that we work with lower case for the alphabet (e.g. when checking if alphabet.contains("j")).  But I have moved it to the construction of SelectorState now.

wezterm-gui/src/overlay/selector.rs
Danielkonge: I have done it like this now.

wezterm-gui/src/overlay/selector.rs
Danielkonge: I have added an else.

wezterm-gui/src/overlay/selector.rs
Danielkonge: I have inlined it as `self.args.alphabet.contains("j")` now.

wezterm-gui/src/overlay/selector.rs
```diff
@@ -178,25 +189,46 @@ impl SelectorState {
     }
 
     fn run_loop(&mut self, term: &mut TermWizTerminal) -> anyhow::Result<()> {
+        let max_items = self.max_items;
```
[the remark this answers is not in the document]
Danielkonge: I have tried to do it like you suggested now.

docs/config/lua/keyassignment/InputSelector.md
```diff
@@ -44,6 +54,7 @@ config.keys = {
         end
       end),
       title = 'I am title',
+      fuzzy = true,
```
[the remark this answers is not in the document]
Danielkonge: Okay, is it fine that I edited the other example still?

wezterm-gui/src/overlay/selector.rs
Danielkonge: To do something similar for `/`, we would need to add another way to search, I think. I am not sure it makes sense to add that.

wezterm-gui/src/overlay/selector.rs
Danielkonge: I have done this now (though I kept the order).

wezterm-gui/src/overlay/selector.rs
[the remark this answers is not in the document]
Danielkonge: I have updated to use `self.filtered_entries.len().min(max_items+1)` now.

Danielkonge:
I changed a few things so that you can use upper case letters in the InputSelector "labels" for picking items and I added a fuzzy_description. I don't really plan to add more stuff now, but I am willing to edit anything you think should be changed. 

Also, as discussed in the Matrix room the error mentioned above seems to be a race condition problem, but it doesn't affect the use of the InputSelector since it "fixes" itself in about 15ms.

wezterm-gui/src/overlay/selector.rs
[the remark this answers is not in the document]
Danielkonge: I changed it now, so you can use uppercase letters. (Note: To do this I made a tiny edit to compute_labels_for_alphabet).

wezterm-gui/src/termwindow/paneselect.rs
```diff
@@ -72,7 +72,7 @@ impl PaneSelector {
 
         let panes = term_window.get_panes_to_render();
         let labels =
-            crate::overlay::quickselect::compute_labels_for_alphabet(alphabet, panes.len());
+            crate::overlay::quickselect::compute_labels_for_alphabet(alphabet, panes.len(), true);
```
wez, line 75:
```suggestion
            crate::overlay::quickselect::compute_labels_for_alphabet(alphabet, panes.len());
```

wez (commented):
Thanks!  Almost there: a couple more suggestions!

docs/config/lua/keyassignment/InputSelector.md
```diff
@@ -123,5 +164,68 @@ config.keys = {
...
 ```
 
+## Example of switching between a list of workspaces with the InputSelector
+
+```lua
+local act = wezterm.action
+local config = wezterm.config_builder()
+
+config.leader = { key = 'Space', mods = 'CTRL', timeout_milliseconds = 3000 }
```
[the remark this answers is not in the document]
Danielkonge: I can just make another key binding. I will change it to Ctrl-Shift-S for now if that is better. 

(I think the standard timeout_milliseconds are a bit too low for newcomers.)

wezterm-gui/src/overlay/quickselect.rs
```diff
@@ -108,29 +119,32 @@ mod alphabet_test {
 
     #[test]
     fn simple_alphabet() {
-        assert_eq!(compute_labels_for_alphabet("abcd", 3), vec!["a", "b", "c"]);
+        assert_eq!(
```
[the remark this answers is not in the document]
Danielkonge: I have taken the suggestions and added a few tests now.

docs/config/lua/keyassignment/InputSelector.md
```diff
@@ -22,6 +22,45 @@ upon the input.
   objects from the current pane and window, and `id` and `label` hold the
   corresponding fields from the selected choice. Both will be `nil` if
   the overlay is cancelled without selecting anything.
+* `fuzzy` - a boolean that defaults to `false`. If `true`, InputSelector will start
+  in its fuzzy finding mode (this is equivalent to starting the InputSelector and
+  pressing / in the default mode).
+* `alphabet` - a string of unique characters. The characters in the string are used
```
wez, line 28: Ah, one last thing to wrap this up, is to annotate these new fields with version information.  Something like this will work, but putting the field, description and version into a table might end up looking better.

```suggestion

{{since('nightly')}}

These additional fields are also available:

* `alphabet` - a string of unique characters. The characters in the string are used
```

The `since` macro accepts a version number or the special `nightly` string.  I use `nightly` to find and update the docs when I tag a release.

If you want to do this table style, you can use this for the version in which InputSelector was originally added, and to have it emit as an inline version number:

```
{{since('20230408-112425-69ae8472', inline=True)}}
```

Danielkonge: I have inlined some version information in the key table now.

wez:
Thank you!

config/src/keyassignment.rs
```diff
@@ -477,6 +477,27 @@ pub struct InputSelector {
...
     pub fuzzy: bool,
+
+    #[dynamic(default = "default_num_alphabet")]
+    pub alphabet: String,
+
+    #[dynamic(default = "default_description")]
+    pub description: String,
+
+    #[dynamic(default = "default_fuzzy_description")]
+    pub fuzzy_description: String,
```
bew, line 488: Hello, I think the word `description` isn't a great here, and `prompt` would better fit

Wdyt?

Danielkonge: I could go through the code and docs and change it, but I am not sure prompt would be a good idea. I would like to keep the naming similar to what you see in PromptInputLine (see https://wezfurlong.org/wezterm/config/lua/keyassignment/PromptInputLine.html ) and at the same time avoid confusion with this overlay. 

Of course, we could change `description` to `prompt` in PromptInputLine too. What do you think @wez?

wez: I think description is fine; prompt doesn't feel like quite the right word anyway, and this is consistent with promptinputline.

bew: > prompt doesn't feel like quite the right word anyway

:thinking: huh? My understanding of these words:

- `description` for me is a text that describes something, can be long, and doesn't ask for any action
- `prompt` for me is the text just before a text field input, it's a 'call-to-action' (e.g. ends with a `: `)

And from where I see `description` & `fuzzy_description` being used in the code, `prompt` matches more than `description`.

Or maybe I'm missing some meaning of the english language? (I'm not english native)

Danielkonge: I think description can be used to describe pretty much anything; also something asking for an action. In this case `description` can be used to describe either what your current InputSelector does, what different keys do or really anything else that you think helps you when using the InputSelector, which I think is better than prompt (it is better to be a bit more general, since it has different use cases). 

The way `description` is currently used in the code is not the only way to use it, it is just what I had in mind for myself.

bew: For the different use cases I'd use a separate options, one for the prompt and one for the description, if any.

This is what `fzf` does for example, where both the prompt (named `prompt`) and the description (named `header`) exist.
And since they are separate the layout can be changed so that for example the header is displayed before the prompt line or after (but before the items to select)

wez: `prompt` *is* a compatible word, it just feels a bit more "active" or stronger than I think is necessary here.  If I hadn't used description elsewhere I'd be more into it here, but as it stands: I don't have sufficiently strong feelings about it to actively want to change it

PR #4277, opened 2023-09-12
Added spawn_command_as_user to allow user impersonation when spawning processes into the Pty

sanitycheck:
Recently, I came across a scenario where I needed to be able to spawn processes into the Pty under different security contexts/user accounts using portable_pty.

I've added the method spawn_command_as_user to allow for user impersonation using the CreateProcessAsUser. Either a primary or impersonation token can be supplied parameter. CreateProcessAsUser requires both SeIncreaseQuotaPrivilege and SeAssignPrimaryTokenPrivileges user rights to be enabled on the calling process.

### DIFF
--- a/pty/src/serial.rs
+++ b/pty/src/serial.rs
@@ -116,6 +116,17 @@ impl SlavePty for Slave {
             port: Arc::clone(&self.port),
         }))
     }
+        /// Spawns a command in the security context of the primary token
+    #[cfg(windows)]
+    fn spawn_command_as_user(&self, cmd: CommandBuilder, h_token: &winapi::um::winnt::HANDLE) -> anyhow::Result<Box<dyn Child + Send + Sync>> {
+        ensure!(
+            cmd.is_default_prog(),
+            "can only use default prog commands with serial tty implementations"
+        );
+        Ok(Box::new(SerialChild {
+            port: Arc::clone(&self.port),
+        }))
+    }
 }
 
 /// There isn't really a child process on the end of the serial connection,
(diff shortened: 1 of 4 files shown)


pty/src/serial.rs
```diff
@@ -116,6 +116,17 @@ impl SlavePty for Slave {
             port: Arc::clone(&self.port),
         }))
     }
+        /// Spawns a command in the security context of the primary token
+    #[cfg(windows)]
+    fn spawn_command_as_user(&self, cmd: CommandBuilder, h_token: &winapi::um::winnt::HANDLE) -> anyhow::Result<Box<dyn Child + Send + Sync>> {
+        ensure!(
```
wez, line 122: FWIW, I think this should unconditionally error out with an explanation that it is unsupported for serial ports

pty/src/win/psuedocon.rs
```diff
@@ -107,6 +107,70 @@ impl PsuedoCon {
         Ok(())
     }
 
+    pub fn spawn_command_as_user(&self, cmd: CommandBuilder, h_token: HANDLE) -> anyhow::Result<WinChild>  {
```
wez, line 110: I'm not keen on the amount of code duplication between this and the existing `spawn_command`.

Can you refactor `spawn_command` to accept an `Option<HANDLE>` for the token?

I also noticed that the environment block and cwd aren't passed down in this case; please either enable that or add some comments explaining why that behavior is different.

wez (commented):
Thanks for this; I have some comments/questions!

pty/src/win/psuedocon.rs
sanitycheck: Hi @wez, thanks for taking a look at my PR! I've gone ahead and made the changes you've suggested. Let me know if there is anything else you need. Thanks!

PR #4298, opened 2023-09-16
Fix example comment about switching zero for Fira fonts

berquist:
I noticed that the comment about switching the form of the digit zero for Fira Mono and Fira Code is reversed from the actual behavior: the default is with the dot, not the line.


wez:
Thanks!

PR #4303, opened 2023-09-17
fix: clarify the docs for default `font_size`

swarnimarun:
Not quite sure but this has confused me a couple times.
So I think we should clean this up, I am not sure if just removing it is enough but let me know what else it should be refactored to.


wez:
Thanks, but the default font size is 12 points, and this PR would have removed the correct information.
I've updated this page to be clearer and correct.

PR #4322, opened 2023-09-20
wayland: update SCTK to fix startup on newer compositors

### LINKED ISSUE #3996: WezTerm doesn't start in Plasma 6

Linux Wayland

KWin 5.27.80 (upcoming Plasma 6)

### WezTerm version

20230713-174400-115e629e

### Describe the bug

WezTerm fails to start with the following error:

```
interface 'wl_surface' has no event 2
01:15:04.275  ERROR  wezterm_gui > running message loop: error during event_q.dispatch protocol_error=None: Resource temporarily unavailable (os error 11); terminating
```

### To Reproduce

_No response_

### Configuration

no config

### Expected Behavior

_No response_

### Logs

_No response_

### Anything else?

This is probably an issue in smithay-client-toolkit v0.16, because its examples fail with the same error in Plasma 6. Those examples in v0.17 work just fine.

hexchain:
Fixes #3996.

With this change, wezterm is now able to start in Wayland mode on Plasma 6.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4843,9 +4843,9 @@ checksum = "f67ad224767faa3c7d8b6d91985b78e70a1324408abcb1cfcc2be4c06bc06043"
 
 [[package]]
 name = "smithay-client-toolkit"
-version = "0.16.0"
+version = "0.16.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


colemickens:
This also fixes `wezterm` under sway-tip.

wez:
Thanks!

PR #4326, opened 2023-09-21
Contribute docs update

Danielkonge:
I think the previous pull request was approved without the last update. Here is what it was meant to say. (The approved one still had the mdbook stuff.)


wez:
Sorry, I didn't realize you'd submitted another PR; we conflicted in 4adbe89cb34885c24ed7663a6272dacfd6b6c415.  Can you rebase this on that? Thanks!

Danielkonge:
The fix you made was basically what I had too. (You can see it now.) The only real difference was that I moved the part about building it yourself (without serving it), since I think the part with serve and a link is better.

wez:
Thanks!

Issue #4328, opened 2023-09-22
Expose some Rust std functions in lua


eugenesvk:
**Is your feature request related to a problem? Please describe.**
Lua has a tiny standard library that doesn't support many basic niceties you'd expect when programming your config

**Describe the solution you'd like**
The proper solution would be to go beyond lua ;), but that's complicated

**Describe alternatives you've considered**
You've kindly offered to expose some of the nice things Rust has in its standard library in this https://github.com/wez/wezterm/discussions/2999#discussioncomment-7082218, so I'll paste a few functions that from a local "std" that eased some of the lua pain points
Thank you!

**Additional context**

[bigger std gist example](https://gist.github.com/eugenesvk/ed94e1137dd3a0cc297612640a35498a#file-std-lua)

### String functions

I think it's not possible to extend lua string class outside of lua? to achieve the ergonomics of `lua_string_var:lower():startswith("python")`, so will hide them

<details>
  <summary>String extensions</summary>

  - string.escmagic escapes lua special chars so you can add any string by adding escapes to those chars you find in a string
  - string.startswith/endswith/isin booleans to check whether a string starts/ends/contains another string
  - string.trim remove whitespace from sides
  - string.split split by a given separator
</details>

### Table functions

  - tValOrSelf(tab, key)  return value if key matches otherwise return self
  - tHasKey(tab, key) check whether a table has key
  - tHasVal(tab, value) check whether a table has value
  - tlen get table length
  - t2str(tbl, indent, out) convert table with a given indent to string

Table functions useful to concatenate configs from multiple semantically separated tables
[tables concat gist](https://gist.github.com/eugenesvk/ed94e1137dd3a0cc297612640a35498a#file-mytable-lua)

  - merge_all(...) Merge all the tables into one and return it
  - deepclone(original) Deep clone the given table
  - flatten_list(list) Flatten the given list of (item or (list of (item or ...)) to a list of item. (nested lists are supported)

### Math

  - a basic round to a given number of digits :), haven't found a non-buggy lua one not converting ±0.5 to 0

Maybe some timer extensions, but I've already mentioned that in a different issue https://github.com/wez/wezterm/issues/3186#issuecomment-1453776819
[timers gist](https://gist.github.com/eugenesvk/ed94e1137dd3a0cc297612640a35498a#file-timer-lua)

Danielkonge:
I have made a pull request with most of the table functions you want now. See #4336. Please check if this is similar to what you had in mind. Do you want to change anything?

[...]

PR #4330, opened 2023-09-23
basename, dirname and canonical_path addition.

Danielkonge:
This might solve #4248 and is a step towards adding more (rust) functionality as requested in #4328. 

I will add to the documentation later (this is just a first draft). 

Note that basename and dirname behave slightly different than the usual commands right now. For something like "/foo/bar/baz.rs" they behave the same, but if the path ends with "/." (e.g. "foo/bar/.") there is a difference. This doesn't seem important to me, but I can change it, if someone else thinks this difference is important. (This comes from the behavior of the rust functions file_name and parent.)

### DIFF
--- a/docs/config/lua/wezterm/basename.md
+++ b/docs/config/lua/wezterm/basename.md
@@ -24,7 +24,7 @@ from Rust's [`std::path::PathBuf`](https://doc.rust-lang.org/nightly/std/path/st
 local wezterm = require 'wezterm'
 local basename = wezterm.basename
 
-wezterm.log_info('baz.txt' == basename '/foo/bar/baz.txt')
+assert('baz.txt' == basename '/foo/bar/baz.txt')
 ```
 
 See also [dirname](dirname.md).
--- a/docs/config/lua/wezterm/basename.md
+++ b/docs/config/lua/wezterm/basename.md
@@ -24,7 +24,7 @@ from Rust's [`std::path::PathBuf`](https://doc.rust-lang.org/nightly/std/path/st
 local wezterm = require 'wezterm'
 local basename = wezterm.basename
 
-wezterm.log_info( 'baz.txt = ' .. basename '/foo/bar/baz.txt' )
+wezterm.log_info('baz.txt = ' .. basename '/foo/bar/baz.txt')
 ```
 
 See also [dirname](dirname.md).
--- a/docs/config/lua/wezterm/basename.md
+++ b/docs/config/lua/wezterm/basename.md
@@ -24,7 +24,7 @@ from Rust's [`std::path::PathBuf`](https://doc.rust-lang.org/nightly/std/path/st
 local wezterm = require 'wezterm'
 local basename = wezterm.basename
 
-wezterm.log_info('baz.txt = ' .. basename '/foo/bar/baz.txt')
+wezterm.log_info('baz.txt' == basename '/foo/bar/baz.txt')
 ```
 
 See also [dirname](dirname.md).
--- a/docs/config/lua/wezterm/basename.md
+++ b/docs/config/lua/wezterm/basename.md
@@ -1,0 +1,30 @@
+---
+title: wezterm.basename
+tags:
+ - utility
+ - filesystem
+---
+# `wezterm.basename(path)`
+
+{{since('nightly')}}
+
+This function returns a string containing the basename of the given path.
+The function does not check whether the given path actually exists.
+Due to limitations in the lua bindings, all of the paths
+must be able to be represented as UTF-8 or this function will generate an
+error.
+
+Note: This function is similar to the shell command basename, but it behaves
+slightly different in some edge case. E.g. `wezterm.basename 'foo.txt/.//'`
+returns `'foo.txt` since trailing `/`s are ignored and so is one `.`.
+But `wezterm.basename 'foo.txt/..'` returns `'..'`. This behaviour comes
+from Rust's [`std::path::PathBuf`](https://doc.rust-lang.org/nightly/std/path/struct.PathBuf.html#method.file_name).
+
+```lua
+local wezterm = require 'wezterm'
+local basename = wezterm.basename
+
+wezterm.log_info( 'baz.txt = ' .. basename '/foo/bar/baz.txt' )
+```
+
+See also [dirname](dirname.md).
--- a/docs/config/lua/wezterm/canonical_path.md
+++ b/docs/config/lua/wezterm/canonical_path.md
@@ -1,0 +1,37 @@
+---
+title: wezterm.canonical_path
+tags:
+ - utility
+ - filesystem
+---
+# `wezterm.canonical_path(path)`
+
+{{since('nightly')}}
+
+This function returns a string with the canonical form of a path if it exists.
+The returned path is in absolute form with all intermediate components normalized
+and symbolic links resolved.
+Due to limitations in the lua bindings, all of the paths
+must be able to be represented as UTF-8 or this function will generate an
+error.
+
+The function can for example be used get the correct absolute path for a path
+in a different format.
+```lua
+local wezterm = require 'wezterm'
+local canonical_path = wezterm.canonical_path
+
+wezterm.log_error( wezterm.home_dir .. ' = ' canonical_path( wezterm.home_dir .. "/.") )
+```
+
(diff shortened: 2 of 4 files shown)


Danielkonge:
Everything works as I would expect now, but I can change some edge case behavior if requested. Since I am basically just using Rust functions from `std::path` (or similar), the behavior is basically what you see there.

lua-api-crates/filesystem/src/lib.rs
```diff
@@ -29,6 +33,63 @@ async fn read_dir<'lua>(_: &'lua Lua, path: String) -> mlua::Result<Vec<String>>
...
+            Ok(utf8.to_string())
+        } else {
+            return Err(mlua::Error::external(anyhow!(
+                "path entry {} is not representable as utf8",
+                path.display()
+            )));
+        }
+    } else {
+        // parent returns None if the path terminates in a root or prefix
+        Ok("".to_string())
```
wez, line 73: I'm not sure about this one.
`dirname /` produces `/` in the shell.

Perhaps we should return the input path here?

I think the empty string is not a good return value here because it can be interpreted differently between unix and windows in different contexts.

lua-api-crates/filesystem/src/lib.rs
```diff
@@ -29,6 +33,63 @@ async fn read_dir<'lua>(_: &'lua Lua, path: String) -> mlua::Result<Vec<String>>
...
+    }
+}
+
+// if path exists return the canonical form of the path with all
+// intermediate components normalized and symbolic links resolved
+async fn canonical_path<'lua>(_: &'lua Lua, path: String) -> mlua::Result<String> {
+    let path = smol::fs::canonicalize(&path)
+        .await
+        .map_err(mlua::Error::external)?;
```
wez, line 82: Let's put the input path in here so that the error message is clearer, something like this:

```suggestion
        .map_err(|err| format!(mlua::Error::external("canonical_path('{path}'): {err:#}")))?;
```

docs/config/lua/wezterm/canonical_path.md
```diff
@@ -0,0 +1,45 @@
...
+
+Another common use case is to find the absolute path of a symlink. E.g., Dropbox is usually
+symlinked to `$HOME/Dropbox` on macOS, but is located at `$HOME/Library/CloudStorage/Dropbox`.
+```lua
+-- macOS only:
+local wezterm = require 'wezterm'
+local canonical_path = wezterm.canonical_path
+local home_dir = wezterm.home_dir
+
+wezterm.log_error(
```
wez, line 37: can we use assert for this as well?

docs/config/lua/wezterm/dirname.md
```diff
@@ -0,0 +1,37 @@
...
+```
+
+If you want only the directory name and not the full path, you can use
+`basename` and `dirname` together. E.g.:
+```lua
+local wezterm = require 'wezterm'
+local basename = wezterm.basename
+local dirname = wezterm.dirname
+
+wezterm.log_error('bar = ' .. basename(dirname '/foo/bar/baz.txt'))
```
wez, line 34: let's use assert here too

wez (commented):
Thanks!

lua-api-crates/filesystem/src/lib.rs
Danielkonge: I did consider returning `/`, but I am not sure if that would make sense in windows? (Basically I just know unix systems in general.)

Danielkonge: Or should we return whatever corresponds to root in windows? I can probably make that work too.

wez: What I suspect is that we only hit this case when the input is a root, so we can simply return the input string in this case?

Danielkonge: I have changed it so it returns the input for now.

Danielkonge:
I will stop working on this for today now. Please let me know if you think it needs any other changes.

lua-api-crates/filesystem/src/lib.rs
```diff
@@ -29,6 +33,60 @@ async fn read_dir<'lua>(_: &'lua Lua, path: String) -> mlua::Result<Vec<String>>
     Ok(entries)
 }
 
+// similar (but not equal) to the shell command basename
+async fn basename<'lua>(_: &'lua Lua, path: String) -> mlua::Result<String> {
```
wez, line 37: These path-name-only operations don't need to be async; it's a bit more efficient not to use async functions in the lua runtime, so it is desirable to avoid it if we don't need it!

```suggestion
fn basename<'lua>(_: &'lua Lua, path: String) -> mlua::Result<String> {
```

lua-api-crates/filesystem/src/lib.rs
```diff
@@ -29,6 +33,60 @@ async fn read_dir<'lua>(_: &'lua Lua, path: String) -> mlua::Result<Vec<String>>
...
+async fn dirname<'lua>(_: &'lua Lua, path: String) -> mlua::Result<String> {
+    let path_rs = Path::new(&path);
+    if let Some(parent_path) = path_rs.parent() {
+        if let Some(utf8) = parent_path.to_str() {
+            Ok(utf8.to_string())
+        } else {
+            return Err(mlua::Error::external(anyhow!(
+                "path entry {} is not representable as utf8",
+                &path
+            )));
```
wez, line 66:
```suggestion
            return Err(mlua::Error::external(anyhow!(
                "path entry {path} is not representable as utf8"
            )));
```

lua-api-crates/filesystem/src/lib.rs
```diff
@@ -29,6 +33,60 @@ async fn read_dir<'lua>(_: &'lua Lua, path: String) -> mlua::Result<Vec<String>>
...
+// similar (but not equal) to the shell command basename
+async fn basename<'lua>(_: &'lua Lua, path: String) -> mlua::Result<String> {
+    let path_rs = Path::new(&path);
+    if let Some(basename) = path_rs.file_name() {
+        if let Some(utf8) = basename.to_str() {
+            Ok(utf8.to_string())
+        } else {
+            return Err(mlua::Error::external(anyhow!(
+                "path entry {} is not representable as utf8",
+                &path
```
wez, line 45:
```suggestion
                "path entry {path} is not representable as utf8"
```

docs/config/lua/wezterm/basename.md
```diff
@@ -0,0 +1,30 @@
...
+slightly different in some edge case. E.g. `wezterm.basename 'foo.txt/.//'`
+returns `'foo.txt` since trailing `/`s are ignored and so is one `.`.
+But `wezterm.basename 'foo.txt/..'` returns `'..'`. This behaviour comes
+from Rust's [`std::path::PathBuf`](https://doc.rust-lang.org/nightly/std/path/struct.PathBuf.html#method.file_name).
+
+```lua
+local wezterm = require 'wezterm'
+local basename = wezterm.basename
+
+wezterm.log_info('baz.txt' == basename '/foo/bar/baz.txt')
```
wez, line 27: looks like the docs lost their asserts?

```suggestion
assert('baz.txt' == basename '/foo/bar/baz.txt')
```

docs/config/lua/wezterm/canonical_path.md
```diff
@@ -0,0 +1,43 @@
...
+
+{{since('nightly')}}
+
+This function returns a string with the canonical form of a path if it exists.
+The returned path is in absolute form with all intermediate components normalized
+and symbolic links resolved.
+Due to limitations in the lua bindings, all of the paths
+must be able to be represented as UTF-8 or this function will generate an
+error.
+
```
wez, line 17: For SEO purposes, I think it would be good to have a sentence here that mentions both `realpath` and `readlink`, which are similar operations that people may be used to in other languages/environments.

Danielkonge:
I made `basename` and `dirname` async mainly because all functions I saw in the Lua implementation at first were async, but I can see that other places have normal (non-async) functions, so doesn't it make the most sense to have these two not be async? (For `canonical_path` I think it makes sense that it is async.) Does it make any difference in this case?

Danielkonge:
I have changed this to a draft for now, while I work on implementing a Path object, which might be useful here.

Danielkonge:
I moved this to #4493.

PR #4334, opened 2023-09-24
Scroll viewport to bottom after erase scrollback

tradiff:
This PR addresses an issue identified while remotely controlling a Wezterm pane from Neovim, notably when the user manually scrolls up causing the cursor to go out of view, and then sends some remote commands. The sequence of actions leading to the problem is as follows:

This fixes a problem I noticed when remotely controlling a Wezterm pane from Neovim:
1. Be remotely controlling a wezterm pane
2. Manually scroll up in that pane so the cursor is no longer visible
3. Send the escape code to clear the scrollback buffer (`clear; printf "\e[3J"`) followed by something that will generate long output, such as running some unit testing.

Observed Behavior: After the scrollback buffer is cleared, the screen refreshes, making the cursor visible again. However, as the tests begin to run and generate lengthy output, the cursor moves out of view, and the lines remain scrolled to the top.

Expected Behavior: Given that clearing the scrollback buffer brings the cursor back into view, it would be logical for the viewport to continue scrolling along with any new output, keeping the cursor visible.

It is important to note that this issue only occurs when step 2 is included. Omitting this step will result in the original code behaving as expected.

Alternatively, here is a way to reproduce this from a single pane:
```lua
config.keys = {
  {
    mods = 'CMD',
    key = 'd',
    action = wezterm.action.Multiple {
      wezterm.action.SendString('clear; printf "\\e[3J"\n'),
      wezterm.action.SendString('ls\n'),
    },
  },
}
```

CMD+d will reproduce the problem (again, only if the user has manually scrolled up before hitting CMD+d)

This PR is the first Rust code I've written, and I'm also not very familiar with the Wezterm source. I welcome and encourage any feedback or meticulous reviews 😄

### DIFF
--- a/mux/src/lib.rs
+++ b/mux/src/lib.rs
@@ -23,7 +23,9 @@ use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
 use std::sync::{Arc, Weak};
 use std::thread;
 use std::time::{Duration, Instant};
-use termwiz::escape::csi::{DecPrivateMode, DecPrivateModeCode, Device, Mode};
+use termwiz::escape::csi::{
+    DecPrivateMode, DecPrivateModeCode, Device, Edit, EraseInDisplay, Mode,
+};
 use termwiz::escape::{Action, CSI};
 use thiserror::*;
 use wezterm_term::{Clipboard, ClipboardSelection, DownloadHandler, TerminalSize};
@@ -91,6 +93,7 @@ pub enum MuxNotification {
         old_workspace: String,
         new_workspace: String,
     },
+    EraseScrollback(PaneId),
 }
 
 static SUB_ID: AtomicUsize = AtomicUsize::new(0);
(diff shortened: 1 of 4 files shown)


wez:
Thanks for diving in!
There's a couple of things on my mind here that make this a bit tricky:

* The scrollback position is conceptually owned by the human user, and having it always (as in, literally always 100% of the time, not just in one specific scenario) get moved by automation is undesirable
* This PR is very focused on coupling the clearing of scrollback with the user's scroll position, which feels overly specific to that one action. What about others?
* This configuration option exists, and causes the scrollback to move to the bottom when interactively typing text: https://wezfurlong.org/wezterm/config/lua/config/scroll_to_bottom_on_input.html
* For the keybind repro scenario, it feels like there is a separate but related bug which is that `SendKey` and `SendString` should trigger the `scroll_to_bottom_on_input` logic
* For the remote control scenario, things are more complex because the scrollback position is a property of the GUI but in the context of the remote control, there is no guarantee that there is even a GUI present, and, conversely, to make things more complex, there can actually be multiple GUIs looking at the same pane, each with different scrollbar positions.
* Should automation-driven input also trigger `scroll_to_bottom_on_input`? We could add a `MuxNotification::InputReceived(PaneId)` that gets plumbed through, and have the gui handle it when it is received.  Given the recent work in https://github.com/wez/wezterm/issues/4788 I'm hesitant about adding unconditional/unbounded notification triggers, and it would suck to have every key stroke double-tap the scrolling logic in the common case, so care will need to be taken in the implementation.
* Maybe there should be a `scroll_to_bottom_on_output` option that is much more simply handled entirely in the GUI? I can see that being annoying in a lot of cases though, and the number of bug reports/questions about it will likely exceed its value.
* Maybe there's a much simpler way to handle this. Currently, the `get_viewport` method just returns whatever the GUI thinks that last position was, and there's some implicit adjustment to map into the scrollback history when retrieving the lines to paint at render time.  There could potentially be some fixup logic that happens in `get_viewport` to check whether the GUI's idea of the viewport has been completely invalidated and if so return `None` to indicate that it should use the bottom of the scrollback.

PR #4336, opened 2023-09-25
Adding some functions to work with Lua tables.

Danielkonge:
I will add documentation later, but here is a first draft of some functions for working with tables in lua. This gives some of what was asked for in #4328.


Danielkonge:
I think this is ready for feedback now.

docs/config/lua/wezterm.table/length.md
```diff
@@ -0,0 +1,26 @@
+# `wezterm.table.length(table)`
+
+{{since('nightly')}}
+
+This function returns the length of a Lua table (or array) passed to it.
+
+Note: The Lua function `#` also returns the length of an array, but
+`#` only works for arrays and not tables (with non-integer keys).
```
wez, line 8:
```suggestion
`#` only works for array-style tables with contiguous integer keys starting with index `1`, and not sparse arrays (with gaps in their integer keys), or object or other style of tables with non-integer keys.
```

docs/config/lua/wezterm.table/merge.md
```diff
@@ -0,0 +1,36 @@
...
+  a = 1,
+  b = '2',
+}
+
+local tbl2 = {
+  a = '1',
+  c = 3,
+}
+
+wezterm.log_error(merge { tbl1, tbl2 })
```
wez, line 30: I think it would be helpful to understand the result of this function if you could illustrate it with an assertion:

```lua
assert(merge{tbl1, tbl2}, {a='1', b='2', c=3})
```

docs/config/lua/wezterm.table/merge.md
```diff
@@ -0,0 +1,36 @@
+# `wezterm.table.merge(array_of_tables [, keep_first])`
+
+{{since('nightly')}}
+
+This function merges Lua tables passed to it in the form of an array.
```
wez, line 5: It wasn't immediately obvious to me how the merge might take place, so let's try to make it clear up-front:

```suggestion
This function merges a list of Lua object-style tables based on their keys.

The tables are passed to it in the form of an array.
```

docs/config/lua/wezterm.table/merge.md
```diff
@@ -0,0 +1,36 @@
...
+I.e., to merge the Lua tables `tbl1` and `tbl2`, we can pass them to
+the function as `{ tbl1, tbl2 }`. (See below.)
+
+By default this function merges tables with identical keys by taking
+the value from the last table in the array with each given key.
+
+The optional `keep_first` allows us to instead prefer values from the
```
wez, line 12: boolean arguments can be hard to intuit the effect, especially if the call looks something like `merge({a, b, c, d, e}, true)`.

Can you remove the boolean argument and split this into two separate lua functions: `merge` and `merge_first`?  (I'm also open to better naming of `merge_first`! Are there examples in other languages/runtimes that might inform naming for this?)
I think that will lead to much clearer code.

lua-api-crates/table-funcs/src/lib.rs
```diff
@@ -0,0 +1,205 @@
...
+    Ok(len)
+}
+
+fn has_key<'lua>(_: &'lua Lua, (table, key): (Table<'lua>, LuaValue)) -> mlua::Result<bool> {
+    Ok(table.contains_key(key).map_err(mlua::Error::external)?)
+}
+
+fn has_value<'lua>(_: &'lua Lua, (table, value): (Table<'lua>, LuaValue)) -> mlua::Result<bool> {
+    for pair in table.pairs::<LuaValue, LuaValue>() {
+        let (_, tbl_value) = pair.map_err(mlua::Error::external)?;
```
wez, line 107: Is the error here really an external error?  Does this work?
I think the same question applies to a number of other calls in this module; in general, errors returned from the mlua crate are already `mlua::Result` and shouldn't need to be mapped.

```suggestion
        let (_, tbl_value) = pair?;
```

wez (commented):
Thanks for this!
Sorry that it's taken a while to get around to reviewing it!

docs/config/lua/wezterm.table/merge.md
bew: Note: the sample assert suggested by @wez is not valid / doesn't do what you expect. Since there are no built-in table comparison functions in Lua we could just use a fake `assert_equal` function in docs like this, even if that function doesn't actually exist?

bew:
For `to_string` what about using this 'inspect' implementation?
(either rewrite it or embed it in `wezterm.inspect`)

https://github.com/kikito/inspect.lua
Iirc this the inspect impl used in some projects like neovim.

docs/config/lua/wezterm.table/merge.md
bew: Neovim has `vim.tbl_extend` & `vim.tbl_deep_extend`, where you pass the behavior you want in first param
https://neovim.io/doc/user/lua.html#vim.tbl_extend()
https://neovim.io/doc/user/lua.html#vim.tbl_deep_extend()

lua-api-crates/table-funcs/src/lib.rs
```diff
@@ -0,0 +1,205 @@
...
+    }
+    Ok(flat_vec)
+}
+
+fn length<'lua>(_: &'lua Lua, table: Table<'lua>) -> mlua::Result<Integer> {
+    // note that # only works correctly on arrays in Lua
+    let mut len: i64 = 0;
+    for _ in table.pairs::<LuaValue, LuaValue>() {
+        len += 1
+    }
```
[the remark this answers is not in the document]
Danielkonge: Adding `as i64` at the end, this works yes.

lua-api-crates/table-funcs/src/lib.rs
Danielkonge: Oh, I didn't think about that. I will try to cleanup all the errors after going through the other comments.

lua-api-crates/table-funcs/src/lib.rs
```diff
@@ -0,0 +1,205 @@
...
+                let old = ["\n", &extra_spaces].concat();
+                return Ok(res.replace(&old, "\n"));
+            }
+            Ok(res)
+        }
+        Err(err) => Err(mlua::Error::external(err)),
+    }
+}
+
+fn to_string_impl<'lua>(
```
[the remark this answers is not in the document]
Danielkonge: I also don't really like the current `to_string` implementation that much, but it is the way it is because I was trying to satisfy what I think #4328 is asking for. 

Maybe we should just remove this and use what I currently have for `to_string_fallback`? Or does the `ValuePrinter´ do anything extra compared to that?

docs/config/lua/wezterm.table/merge.md
wez: ah, then maybe:

```lua
assert(to_string(merge{tbl1, tbl2}), "{a='1', b='2', c=3}")
```

wez:
Since wezterm has custom embedded types that have their own string representation, it is important that what we provide for this string representation is:

* Aware of internal wezterm types (and thus cannot be implemented purely in lua, ruling out that inspect module)
* Ideally consistent with how wezterm prints strings in the debug overlay, and performs implicit string conversion in `wezterm.log_xxx` and its `print` implementation
* Not a parallel and slightly different implementation from ValuePrinter, which is already more code than I'd like to be maintaining; I'd hate to have to double the amount of that kind of code.

I think that means that we should stick to the `ValuePrinter` for this purpose.
That isn't actually a table-specific function, and could be exposed as `wezterm.to_string`.

docs/config/lua/wezterm.table/merge.md
Danielkonge: > Can you remove the boolean argument and split this into two separate lua functions: `merge` and `merge_first`? (I'm also open to better naming of `merge_first`! Are there examples in other languages/runtimes that might inform naming for this?) I think that will lead to much clearer code.

I can split it into two, but will wait until we agree on the details. Would you prefer two functions over string arguments like neovim?

For naming if we go with two functions, maybe `merge_keep` (more like the neovim naming) is better than `merge_first`? But this is not something I feel strongly about.

docs/config/lua/wezterm.table/merge.md
Danielkonge: > ah, then maybe:
> 
> ```lua
> assert(to_string(merge{tbl1, tbl2}), "{a='1', b='2', c=3}")
> ```

We can also do something like this, but it should wait until we decide what to do with `to_string`.

docs/config/lua/wezterm.table/merge.md
wez: Hmm, ok, you could add:

```rust
#[derive(Debug, FromDynamic, ToDynamic, Clone, PartialEq, Eq)]
enum ConflictMode {
   Keep,
   Force,
   Error,
}
impl_lua_conversion_dynamic!(ConflictMode);
```

then make the parameter an `Option<ConfictMode>`.  That should allow passing `"Keep"` as that parameter value. If that can be the first parameter without being unusable (eg: not requiring that the user do `merge(nil, {a, b})` that would be fine, but otherwise it could be the last parameter.

If we want to be influenced by nvim here, we could also rename this to `extend`.

Danielkonge:
For the last commit, it is only meant to be temporary for now, if you want to try out `to_string` using the `ValuePrinter` here. It should probably be moved to `wezterm.to_string` instead of `wezterm.table.to_string` later.

wez:
re: `utf8.charpattern`, 83fbba5a4fdee5e62fd9933304a4da9d0b9051cb handles that better in `ValuePrinter`.

wez:
6d58e5196bb3a433880643a05fe5d6f19e49572a improves handling of recursion in `ValuePrinter`.  With that fixed, I'm not sure that there is a need for `to_string_fallback`?

docs/config/lua/wezterm.table/merge.md
Danielkonge: > Hmm, ok, you could add:
> 
> ```rust
> #[derive(Debug, FromDynamic, ToDynamic, Clone, PartialEq, Eq)]
> enum ConflictMode {
>    Keep,
>    Force,
>    Error,
> }
> impl_lua_conversion_dynamic!(ConflictMode);
> ```

This code gives errors. Also, from what I can see isn't that macro for structs and not enums? (I am not used to reading rust macros)

I can just implement what is needed myself for this one case though.

docs/config/lua/wezterm.table/index.md
```diff
@@ -0,0 +1,17 @@
+# `wezterm.table` module
```
wez, line 1: This file is generated and shouldn't be added.

Please add `/docs/config/lua/wezterm.table/index.md` to `.gitignore` and remove this file from the PR!

Danielkonge: I updated the `.gitignore` and removed that file now.

docs/config/lua/wezterm.table/merge.md
Danielkonge: I added `wezterm.table.equal( tbl1, tbl2 )` so we can use that in the `assert`.

lua-api-crates/table-funcs/src/lib.rs
```diff
@@ -203,6 +232,34 @@ fn has_value<'lua>(_: &'lua Lua, (table, value): (Table<'lua>, LuaValue)) -> mlu
     Ok(false)
 }
 
+fn equal<'lua>(lua: &'lua Lua, (table1, table2): (Table<'lua>, Table<'lua>)) -> mlua::Result<bool> {
```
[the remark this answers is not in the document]
wez: I'd suggest something like:

```rust
fn lua_value_eq(value1: &LuaValue, value2: &LuaValue) -> bool {
   match (value1, value2) {
      (LuaValue::Table(a), LuaValue::Table(b)) => lua_table_eq(a, b),
      (a, b) => a.eq(b)
   }
}

fn lua_table_eq(table1: &Table, table2: &Table) -> bool {
   let mut table1_len = 0;
   for pair in table1.pairs::<LuaValue, LuaValue>() {
        match pair {
           Ok(key, value) => {
             table1_len += 1;
             match table2.get(key.clone()) {
                Ok(value2) => {
                    if !lua_value_eq(value, value2) {
                         return false;
                    }
                 }
                 Err(_) => return false,
           }
           Err(_) => return false,
        }
    }
    let table2_len = table2.pairs::<LuaValue, LuaValue>().count();
    table1_len == table2_len
}
```

lua-api-crates/table-funcs/src/lib.rs
```diff
@@ -22,27 +23,30 @@ pub fn register(lua: &Lua) -> anyhow::Result<()> {
...
+                _ => Err(mlua::Error::runtime(
+                    "Unknown string. Expected 'Keep', 'Force' or 'Error'".to_string(),
+                )),
+            },
             LuaValue::Error(err) => Err(err),
-            other => Err(mlua::Error::runtime(format!("Expected a Lua string. Got something of type: {}", other.type_name()))),
+            other => Err(mlua::Error::runtime(format!(
+                "Expected a Lua string. Got something of type: {}",
+                other.type_name()
+            ))),
```
wez: This should all be derivable via FromDynamic and ToDynamic. Please share the errors you mentioned when you tried my original suggestion

Danielkonge:
> [6d58e51](https://github.com/wez/wezterm/commit/6d58e5196bb3a433880643a05fe5d6f19e49572a) improves handling of recursion in `ValuePrinter`. With that fixed, I'm not sure that there is a need for `to_string_fallback`?

There is still one point where `to_string_fallback` is better than `to_string`.

With the following Lua code:
```lua
wezterm.log_info( "test1:", weztable.to_string({ a = {'a'}, 1, {2} }))
wezterm.log_info( "test2:", weztable.to_string_fallback({ a = {'a'}, 1, {2} }))
```

We get:
Screenshot 2023-12-02 at 01 01 49

`ValuePrinter` gets confused by the mix between an array-like table and non-integer entries. 

Otherwise I think `ValuePrinter` gives you all the information you need now, so the only question is if we want to change the formatting a bit.

lua-api-crates/table-funcs/src/lib.rs
Danielkonge:
```rust
error[E0277]: the trait bound `ConflictMode: wezterm_dynamic::todynamic::ToDynamic` is not satisfied
  --> lua-api-crates/table-funcs/src/lib.rs:30:1
   |
30 | impl_lua_conversion_dynamic!(ConflictMode);
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   | |
   | the trait `wezterm_dynamic::todynamic::ToDynamic` is not implemented for `ConflictMode`
   | required by a bound introduced by this call
   |
   = help: the following other types implement trait `wezterm_dynamic::todynamic::ToDynamic`:
             bool
             char
             isize
             i8
             i16
             i32
             i64
             usize
           and 152 others
note: required by a bound in `to_lua`
  --> /Users/daniel/personal/wezterm/luahelper/src/lib.rs:12:24
   |
12 | pub fn to_lua<'lua, T: ToDynamic>(
   |                        ^^^^^^^^^ required by this bound in `to_lua`
   = note: this error originates in the macro `impl_lua_conversion_dynamic` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0277]: the trait bound `ConflictMode: wezterm_dynamic::fromdynamic::FromDynamic` is not satisfied
  --> lua-api-crates/table-funcs/src/lib.rs:30:1
   |
30 | impl_lua_conversion_dynamic!(ConflictMode);
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `wezterm_dynamic::fromdynamic::FromDynamic` is not implemented for `ConflictMode`
   |
   = help: the following other types implement trait `wezterm_dynamic::fromdynamic::FromDynamic`:
             bool
             char
             isize
             i8
             i16
             i32
             i64
             usize
           and 152 others
note: required by a bound in `luahelper::from_lua`
  --> /Users/daniel/personal/wezterm/luahelper/src/lib.rs:20:26
   |
20 | pub fn from_lua<'lua, T: FromDynamic>(value: mlua::Value<'lua>) -> Result<T, mlua::Error> {
   |                          ^^^^^^^^^^^ required by this bound in `from_lua`
   = note: this error originates in the macro `impl_lua_conversion_dynamic` (in Nightly builds, run with -Z macro-backtrace for more info)

For more information about this error, try `rustc --explain E0277`.
error: could not compile `table-funcs` (lib) due to 2 previous errors
```

Danielkonge: Wait, I made a mistake. I will update the above.

Danielkonge: I forgot to add `FromDynamic` and `ToDynamic` to the `derive` in the above log. Here is the log with those added:

```rust
error: cannot find derive macro `FromDynamic` in this scope
  --> lua-api-crates/table-funcs/src/lib.rs:24:17
   |
24 | #[derive(Debug, FromDynamic, ToDynamic, Clone, PartialEq, Eq)]
   |                 ^^^^^^^^^^^

error: cannot find derive macro `ToDynamic` in this scope
  --> lua-api-crates/table-funcs/src/lib.rs:24:30
   |
24 | #[derive(Debug, FromDynamic, ToDynamic, Clone, PartialEq, Eq)]
   |                              ^^^^^^^^^

error[E0277]: the trait bound `ConflictMode: wezterm_dynamic::todynamic::ToDynamic` is not satisfied
  --> lua-api-crates/table-funcs/src/lib.rs:30:1
   |
30 | impl_lua_conversion_dynamic!(ConflictMode);
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   | |
   | the trait `wezterm_dynamic::todynamic::ToDynamic` is not implemented for `ConflictMode`
   | required by a bound introduced by this call
   |
   = help: the following other types implement trait `wezterm_dynamic::todynamic::ToDynamic`:
             bool
             char
             isize
             i8
             i16
             i32
             i64
             usize
           and 152 others
note: required by a bound in `to_lua`
  --> /Users/daniel/personal/wezterm/luahelper/src/lib.rs:12:24
   |
12 | pub fn to_lua<'lua, T: ToDynamic>(
   |                        ^^^^^^^^^ required by this bound in `to_lua`
   = note: this error originates in the macro `impl_lua_conversion_dynamic` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0277]: the trait bound `ConflictMode: wezterm_dynamic::fromdynamic::FromDynamic` is not satisfied
  --> lua-api-crates/table-funcs/src/lib.rs:30:1
   |
30 | impl_lua_conversion_dynamic!(ConflictMode);
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `wezterm_dynamic::fromdynamic::FromDynamic` is not implemented for `ConflictMode`
   |
   = help: the following other types implement trait `wezterm_dynamic::fromdynamic::FromDynamic`:
             bool
             char
             isize
             i8
             i16
             i32
             i64
             usize
           and 152 others
note: required by a bound in `luahelper::from_lua`
  --> /Users/daniel/personal/wezterm/luahelper/src/lib.rs:20:26
   |
20 | pub fn from_lua<'lua, T: FromDynamic>(value: mlua::Value<'lua>) -> Result<T, mlua::Error> {
   |                          ^^^^^^^^^^^ required by this bound in `from_lua`
   = note: this error originates in the macro `impl_lua_conversion_dynamic` (in Nightly builds, run with -Z macro-backtrace for more info)

For more information about this error, try `rustc --explain E0277`.
error: could not compile `table-funcs` (lib) due to 4 previous errors
```

lua-api-crates/table-funcs/src/lib.rs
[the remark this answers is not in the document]
Danielkonge: I have updated the code to something more like this, but I can't easily get around the `.clone()`s before `.pairs()`, since `equal` itself can't accept two references like `lua_table_eq` does.

If I change `lua_table_eq` to not use references, then I have to add something similar in `lua_value_eq` anyways. 

If you know a good way to get around this problem I can try to improve the code further.

lua-api-crates/table-funcs/src/lib.rs
```diff
@@ -232,32 +232,37 @@ fn has_value<'lua>(_: &'lua Lua, (table, value): (Table<'lua>, LuaValue)) -> mlu
...
+                    }
+                    Err(_) => return Ok(false),
+                }
             }
-        } else {
-            res = value1.eq(&value2);
+            Err(_) => return Ok(false),
         }
     }
+    let table2_len = table2.clone().pairs::<LuaValue, LuaValue>().count();
```
wez, line 260: removing this clone would be nice.  I think you can pass the Tables directly (eg: `Table` rather than `&Table`) to allow for that.

lua-api-crates/table-funcs/src/lib.rs
wez: I think you just need to `use wezterm_dynamic::{FromDynamic, ToDynamic};` at the top of the file?

Danielkonge: > I think you just need to `use wezterm_dynamic::{FromDynamic, ToDynamic};` at the top of the file?

Ah, I see what went wrong. I needed to update the `Cargo.toml` and add this. It works now (though it won't accept e.g. `keep` now, only `Keep`).

Danielkonge:
I plan to cleanup the docs and remove `to_string` and `to_string_fallback` in a few hours. I will instead make another very small pull request with only `wezterm.to_string`. 

Is there anything else in the `lib.rs` file that I still need to change/add/update now?

lua-api-crates/table-funcs/src/lib.rs
```diff
@@ -0,0 +1,254 @@
...
+                    tbl.set(key, inner_tbl)?;
+                } else {
+                    return Err(mlua::Error::runtime(format!(
+                        "The key {} is in more than one of the tables.",
+                        key.to_string()?
+                    )));
+                }
+            }
+        }
+    };
```
wez, line 149: Feels a lot simpler and more concise without the match:

```suggestion
    let behavior = behavior.unwrap_or_default();
            for (key, value) in tbl_vec {
                if !tbl.contains_key(key.clone())? {
                    tbl.set(key, value)?;
                } else if let LuaValue::Table(t) = value {
                    let inner_tbl = deep_extend(
                        lua,
                        (vec![tbl.get(key.clone())?, t], Some(behavior)),
                    )?;
                    tbl.set(key, inner_tbl)?;
                } else if behavior == ConflictMode::Error {
                    return Err(mlua::Error::runtime(format!(
                        "The key {} is in more than one of the tables.",
                        key.to_string()?
                    )));
                } else if behavior == ConflictMode::Force {
                    tbl.set(key, inner_tbl)?;
                }
            }
      
```

aarondill:
@Danielkonge 
i'm slightly confused about `clone`
it seems to be this lua equivalent: (sorry for the lack of indention, i'm writing this on a phone)
```lua
function clone(tbl)
local res ={}
for k, v in pairs(tbl) do
if type(v) == "table" then
res[k] = clone(v)
else
res[k] = v
end
return res
end
```

but the docs show usage like this:
```
local tbl1 = {a=1}
local tbl2 = clone(tbl1)
assert(tbl1 == tbl2)
```

since clone creates a new table, this assertion *should* fail every time this is run, as `==` checks for referencial equality


ps: I personally would prefer if `clone` took a second optional boolean `deep` (which perhaps defaults to true) that decides whether this is a recursive operation or act solely on the top level of the table (ie: is equal to the following code)

```lua
local tbl2 ={}
for k, v in pairs(tbl1) do
  tbl2[k] = v
end

```

lua-api-crates/table-funcs/src/lib.rs
```diff
@@ -0,0 +1,254 @@
...
+
+    Ok(tbl)
+}
+
+fn clone<'lua>(lua: &'lua Lua, table: Table<'lua>) -> mlua::Result<Table<'lua>> {
+    let table_len = table.clone().pairs::<LuaValue, LuaValue>().count();
+    let res: Table<'lua> = lua.create_table_with_capacity(0, table_len)?;
```
wez, line 156: The way that the full length is computed essentially clones the table and discards it, which feels more expensive than is perhaps necessary.  I would just create an empty one and fill it as we go; it feels like the lesser of two evils, and would be the same as using `{}` to create a new table in pure lua code anyway.


```suggestion
    let res: Table<'lua> = lua.create_table()?;
```

lua-api-crates/table-funcs/src/lib.rs
```diff
@@ -0,0 +1,254 @@
...
+
+    for pair in table.pairs::<LuaValue, LuaValue>() {
+        let (key, value) = pair?;
+        if let LuaValue::Table(tbl) = value {
+            let inner_res = clone(lua, tbl)?;
+            res.set(key, inner_res)?;
```
wez, line 162: I'd just skip the temporary and avoid the stress of coming up with a good name for it:

```suggestion
            res.set(key, clone(lua, tbl)?)?;
```

lua-api-crates/table-funcs/src/lib.rs
```diff
@@ -0,0 +1,254 @@
...
+fn flatten<'lua>(lua: &'lua Lua, arrays: Vec<LuaValue<'lua>>) -> mlua::Result<Vec<LuaValue<'lua>>> {
+    let mut flat_vec: Vec<LuaValue> = vec![];
+    for item in arrays {
+        match item {
+            LuaValue::Table(tbl) => {
+                let tbl_as_vec = tbl.sequence_values().filter_map(|x| x.ok()).collect();
+                let flat = flatten(lua, tbl_as_vec)?;
+                for j in flat {
+                    flat_vec.push(j);
+                }
```
wez, line 179:
```suggestion
                let mut flat = flatten(lua, tbl_as_vec)?;
                flat_vec.append(&mut flat);
```

aarondill:
as to `flatten`, it would be helpful if it took an (optional) second integer `max_depth` that would determine how deeply the arrays are flattened (ie values deeper are appended even if they are arrays)
```lua
flatten({ { 2, {3} }, { {4}, 5 } }, 1)
-- { 2, {3}, {4}, 5 }

flatten({ { 2, {3} }, { {4}, 5 } }, 2)
-- { 2, 3, 4, 5 }
```

lua-api-crates/table-funcs/src/lib.rs
```diff
@@ -0,0 +1,254 @@
...
+            }
+            LuaValue::Nil => (),
+            LuaValue::Thread(_) => (),
+            LuaValue::Error(err) => {
+                return Err(err);
+            }
+            other => {
+                flat_vec.push(other);
+            }
```
wez, line 188: I think skipping `nil` makes sense (although: given that lua treats `nil` as not assigned, I'd be surprised if we ever saw one as a value here), but all other types should just be flattened without caring about the value?
In particular, I wonder why threads are stripped out here; they are valid first class values.
```suggestion
            other => {
                flat_vec.push(other);
            }
```

lua-api-crates/table-funcs/src/lib.rs
```diff
@@ -0,0 +1,254 @@
...
+        }
+    }
+    Ok(flat_vec)
+}
+
+fn length<'lua>(_: &'lua Lua, table: Table<'lua>) -> mlua::Result<Integer> {
+    // note that # only works correctly on arrays in Lua
+    let len = table.pairs::<LuaValue, LuaValue>().count() as i64;
+    Ok(len)
+}
```
wez, line 198: I would just use usize here to keep this logic looking less arcane; any time `as <something>` is used
it suggests that something special is happening and requires some extra brain power on the part
of the reader.

```suggestion
/// note that the `#` operator only works correctly on arrays in Lua
fn length<'lua>(_: &'lua Lua, table: Table<'lua>) -> mlua::Result<usize> {
    Ok(table.pairs::<LuaValue, LuaValue>().count())
}
```

aarondill:
finally, I feel that has_key is unnecessary `has_key(tbl, k)` is equivalent to `tbl[k] ~= nil` (or simply the boolean `tbl[k]` if false isn't a possibility)

Although I can understand if you choose to keep it for symmetry with has_value (has_value could be renamed to has?)

wez (commented):
a couple more suggestions on the implementation in `lib.rs`.

The docs will need to be updated to match the conflict mode stuff.

It may be wise to provide some control over the depth of both the clone and flatten operations, as it is not always desirable/necessary to perform a fullly recursive clone.

Personally, I've never needed more nuance than either top-level or fully recursive, so I don't know if specifying an arbitrary depth limit is necessary.

aarondill (changes requested):

Danielkonge:
> @Danielkonge i'm slightly confused about `clone` it seems to be this lua equivalent: (sorry for the lack of indention, i'm writing this on a phone)
> 
> ```lua
> function clone(tbl)
> local res ={}
> for k, v in pairs(tbl) do
> if type(v) == "table" then
> res[k] = clone(v)
> else
> res[k] = v
> end
> return res
> end
> ```
> 
> but the docs show usage like this:
> 
> ```
> local tbl1 = {a=1}
> local tbl2 = clone(tbl1)
> assert(tbl1 == tbl2)
> ```
> 
> since clone creates a new table, this assertion _should_ fail every time this is run, as `==` checks for referencial equality

I changed `.clone()` since the first implementation. The first implementation cloned the Lua pointer to the table (se they were actually equal in that test), but I have changed it to instead clone the data of the table, since that seems more useful. But I haven't updated the docs yet (will start doing that soon). 

> ps: I personally would prefer if `clone` took a second optional boolean `deep` (which perhaps defaults to true) that decides whether this is a recursive operation or act solely on the top level of the table (ie: is equal to the following code)
> 
> ```lua
> local tbl2 ={}
> for k, v in pairs(tbl1) do
>   tbl2[k] = v
> end
> ```

I can add something like that, but it should probably be with a chosen behavior by a string similar to the new `extend`.

Danielkonge:
> as to `flatten`, it would be helpful if it took an (optional) second integer `max_depth` that would determine how deeply the arrays are flattened (ie values deeper are appended even if they are arrays)
> 
> ```lua
> flatten({ { 2, {3} }, { {4}, 5 } }, 1)
> -- { 2, {3}, {4}, 5 }
> 
> flatten({ { 2, {3} }, { {4}, 5 } }, 2)
> -- { 2, 3, 4, 5 }
> ```

I can add that too. Would you want to keep the table (i.e., a pointer to something that might change later) or add a new clone of the table when you don't flatten it?

Danielkonge:
> finally, I feel that has_key is unnecessary `has_key(tbl, k)` is equivalent to `tbl[k] ~= nil` (or simply the boolean `tbl[k]` if false isn't a possibility)
> 
> Although I can understand if you choose to keep it for symmetry with has_value (has_value could be renamed to has or contains?)

I don't think `has_key` is important, so I am fine with removing it. But I do think it makes sense to just have both `has_key` and `has_value` to make it clear what you are looking for, when using it. 

Maybe it would make sense to add a behavior option here too, so it can search for nested keys and/or values?

aarondill:
> I can add that too. Would you want to keep the table (i.e., a pointer to something that might change later) or add a new clone of the table when you don't flatten it?

Absolutely keep the original table (the pointer). If I need it to be cloned, its easy enough to iterate each item in the array and clone it individually
```lua
local tbl2 = flatten({ { 1, {2} }, {3} }, 1)
for k, v in pairs(tbl2) do
  if type(v) == 'table' then
    tbl2[k] = clone(v)
  end
end
```

wez:
FWIW, I'm fine with keeping `has_key`; it is easier for n00bs to search for, and having a function can be a lot more ergonomic when building up `filter` and `map` operations if/when we add those in the future; passing `table.has_key` is a lot more ergonomic than `function(table, key) table[value] ~= nil end`

lua-api-crates/table-funcs/src/lib.rs
```diff
@@ -0,0 +1,254 @@
...
+                if tbl.contains_key(key.clone())? {
+                    return Err(mlua::Error::runtime(format!(
+                        "The key {} is in more than one of the tables.",
+                        key.to_string()?
+                    )));
+                }
+                tbl.set(key, value)?;
+            }
+        }
+    };
```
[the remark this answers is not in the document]
Danielkonge: I can change it to something like this (do you want `Keep` to be the default case?).

I did actually have code with `if` statements inside the for loop at some point, but I changed it to the `match` syntax to minimize the number of `if` statements inside the `for` loop (especially the `Force` case [which I had as default] was nicer).

aarondill:
> I don't think `has_key` is important, so I am fine with removing it. But I do think it makes sense to just have both `has_key` and `has_value` to make it clear what you are looking for, when using it.
> 
> Maybe it would make sense to add a behavior option here too, so it can search for nested keys and/or values?

like I said, I understand the symmetry reason if you choose to go that route.

I do really like the idea of allowing searching for nested keys. Something like this would be awesome (are variadic arguments possible?):
```lua
local tbl1 = { a = { b = 2 } }
has_key(tbl1, 'a', 'b') -- true
has_key(tbl1, 'a', 'c') -- false
```

lua-api-crates/table-funcs/src/lib.rs
Danielkonge: I will update this with the above soon.

docs/config/lua/wezterm.table/length.md
```diff
@@ -0,0 +1,26 @@
+# `wezterm.table.length(table)`
+
+{{since('nightly')}}
+
+This function returns the length of a Lua table (or array) passed to it.
+
+Note: The Lua function `#` also returns the length of an array, but
+`#` only works for array-style tables with contiguous integer keys starting with index `1`, and not sparse arrays (with gaps in their integer keys), or object or other style of tables with non-integer keys.
```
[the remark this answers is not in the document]
Danielkonge: Since Lua unsets elements of a table by setting them equal to `nil`, I don't think we can get around this "unexpected" behavior (from my point of view I think this should be the expected behavior in Lua). I think I will just need to write better docs here and clarify what it counts (plus probably rename it to `count` as wez suggested).

lua-api-crates/table-funcs/src/lib.rs
Danielkonge: Oh, I didn't know that threads were valid values here (I think I misunderstood them). I will update it.

lua-api-crates/table-funcs/src/lib.rs
Danielkonge: I can do that. I assume that mlua will just translate `usize` to `i64` anyways?

lua-api-crates/table-funcs/src/lib.rs
wez: [coroutine.create](https://www.lua.org/manual/5.3/manual.html#pdf-coroutine.create) returns a thread, for example.  While I don't think anyone is directly using this on wezterm today, I think there are some libraries out there that use coroutines to implement some generator style patterns

aarondill:
> I do really like the idea of allowing searching for nested keys. Something like this would be awesome (are variadic arguments possible?):
> 
> ```lua
> local tbl1 = { a = { b = 2 } }
> has_key(tbl1, 'a', 'b') -- true
> has_key(tbl1, 'a', 'c') -- false
> ```
<details>
<summary>Here's an example lua implementation</summary>

```lua
---@param tbl table
---@param ... unknown
---@return boolean
local function has_key(tbl, ...)
  local keys = table.pack(...)

  assert(type(tbl) == "table", ("table expected, got %s"):format(type(tbl)))
  assert(keys.n > 0, "No keys provided")
  local count = 0
  local t = tbl
  for _, k in ipairs(keys) do
    if type(t) ~= "table" then return false end -- not a table
    local v = t[k]
    if v == nil then return false end
    t = t[k]
    count = count + 1
  end
  -- There's a nil key in the list. Nil keys are not allowed in lua
  if count < keys.n then return false end
  return true
end

local tbl = {
  a = 1,
  b = {
    c = 2,
    d = {
      e = 3,
      f = 4,
    },
  },
}

assert(has_key(tbl, "a"))
assert(not has_key(tbl, "a", "b"))
assert(has_key(tbl, "b"))
assert(has_key(tbl, "b", "c"))
assert(not has_key(tbl, "b", "a"))
assert(has_key(tbl, "b", "d"))
assert(has_key(tbl, "b", "d", "e"))
assert(has_key(tbl, "b", "d", "f"))
assert(not has_key(tbl, "b", "d", "g"))
assert(not has_key(tbl, "b", nil, "d", "f"))
```

</details>

Danielkonge:
> > I don't think `has_key` is important, so I am fine with removing it. But I do think it makes sense to just have both `has_key` and `has_value` to make it clear what you are looking for, when using it.
> > Maybe it would make sense to add a behavior option here too, so it can search for nested keys and/or values?
> 
> like I said, I understand the symmetry reason if you choose to go that route.
> 
> I do really like the idea of allowing searching for nested keys. Something like this would be awesome (are variadic arguments possible?):
> 
> ```lua
> local tbl1 = { a = { b = 2 } }
> has_key(tbl1, 'a', 'b') -- true
> has_key(tbl1, 'a', 'c') -- false
> ```

Something like that should be possible. But would you prefer this behavior or something like 
```lua
has_key(tbl1, 'b', 'Deep') -- true
has_key(tbl1, 'b', 'First') -- false
has_key(tbl1, 'c', 'Deep') -- false
```
where `'Deep'` and `'First'` is the behavior of the search? (Do you have a better name than `'First'`?)

Or do you think it is better to enter the "branch" of the table you are checking like what you wrote?

aarondill:
> Something like that should be possible. But would you prefer this behavior or something like
> 
> ```lua
> has_key(tbl1, 'b', 'Deep') -- true
> has_key(tbl1, 'b', 'First') -- false
> has_key(tbl1, 'c', 'Deep') -- false
> ```
> 
> where `'Deep'` and `'First'` is the behavior of the search? (Do you have a better name than `'First'`?)
> 
> Or do you think it is better to enter the "branch" of the table you are checking like what you wrote?

It seems to me that "Deep" would have to search every key in every table recursively to find the key.
I can't really see a use-case for a search like that, especially not a common one.

I'd prefer entering the 'branch' of the table, as it can be much more useful in common problems.
for a few examples from my own wezterm config:
"Is config.colors.foreground set?"
"is config.colors.copy_mode_active_highlight_bg.Color set?"
"is config.window_padding.left set?"

Danielkonge:
> It may be wise to provide some control over the depth of both the clone and flatten operations, as it is not always desirable/necessary to perform a fullly recursive clone.
> 
> Personally, I've never needed more nuance than either top-level or fully recursive, so I don't know if specifying an arbitrary depth limit is necessary.

If I do a clone of only the top level, what behavior would you want when I have something like `{ a=1, b=2, c={3} }`? Should the clone just copy the pointer to the `{3}` table for `c`? Or should it not include the nested tables at all? (If so, should `c` be `nil`?) If we just include the pointer, I think I will need to clarify the difference between a deep clone and a top-level clone very carefully in the docs, since they look similar when printing.

aarondill:
> I'd prefer entering the 'branch' of the table, as it can be much more useful in common problems. for a few examples from my own wezterm config: "Is config.colors.foreground set?" "is config.colors.copy_mode_active_highlight_bg.Color set?" "is config.window_padding.left set?"

Perhaps it would even be possible to return the value at the key path if it is set? I don't know how well rust interacts with multiple return types.
```lua
local has, color = has_key(config, 'copy_mode_active_highlight_bg', 'Color')
if has then
-- do something with color
else
-- color is nil, do something else
end
```

Danielkonge:
> > I'd prefer entering the 'branch' of the table, as it can be much more useful in common problems. for a few examples from my own wezterm config: "Is config.colors.foreground set?" "is config.colors.copy_mode_active_highlight_bg.Color set?" "is config.window_padding.left set?"
> 
> Perhaps it would even be possible to return the value at the key path if it is set? I don't know how well rust interacts with multiple return types (can you use Result with a tuple?).
> 
> ```lua
> local has, color = has_key(config, 'copy_mode_active_highlight_bg', 'Color')
> if has then
> -- do something with color
> else
> -- color is nil, do something else
> end
> ```
> 
> Since the rust code already has to traverse the entire table path, this would only add one `.get` call (since we now need to get the last value, instead of just checking presence)

We could add a functionality like that too, but wouldn't this make more sense to put into its own function called something like `get`? Though code like what you wrote would of course traverse the path to the key twice then, but that shouldn't be very expensive.

lua-api-crates/table-funcs/src/lib.rs
Danielkonge: I have updated this now. Looking at it again and looking at one of your other comments, do we actually want to just move this to the first `for` loop and just create a table without any capacity specified?

Danielkonge:
I added optional extra keys to `has_key` now, but it took longer than expected because of some annoyance with mlua's `.get()` and `.as_table()`, so I didn't get to the docs stuff yet. I will continue working on this tomorrow.

docs/config/lua/wezterm.table/has_value.md
```diff
@@ -0,0 +1,24 @@
+# `wezterm.table.has_value(table, value)`
+
+{{since('nightly')}}
+
+This function accepts a Lua table (or array) `table` and a value `value`.
+It returns `true` if `table` contains an entry with value equal to `value`
+and false otherwise.
```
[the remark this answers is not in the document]
Danielkonge: I added a small note similar to that now.

Danielkonge:
The docs are mostly fixed now, and I have cleaned up a few other things. @wez: In `extend` and `deep_extend` do you think it makes sense to just construct a `Table` without a specified size and move the second loop up to the first loop? Then we can simplify the code a bit.

lua-api-crates/table-funcs/src/lib.rs
```diff
@@ -0,0 +1,298 @@
...
+}
+
+/// note that the `#` operator only works correctly on arrays in Lua
+fn count<'lua>(_: &'lua Lua, table: Table<'lua>) -> mlua::Result<usize> {
+    Ok(table.pairs::<LuaValue, LuaValue>().count())
+}
+
+fn get<'lua>(
+    _: &'lua Lua,
+    (table, key, mut extra_keys): (Table<'lua>, LuaValue<'lua>, LuaMultiValue<'lua>),
```
wez, line 176: FWIW, I think this whole function can be simplified to avoid the special case of the initial key, something like this:

```rust
fn get<'lua>(
    _: &'lua Lua,
    (mut value, mut keys): (Table<'lua>, LuaMultiValue<'lua>)
) -> mlua::Result<LuaValue<'lua>> {
     for key in keys {
        match value {
             LuaValue::Table(tbl) => {
                  value = tbl.get(key)?;
             }
             _ => {
               // Cannot index non-table
               return Ok(LuaValue::Nil)
            }
         }
     }
     Ok(value)
}
```

docs/config/lua/wezterm.table/has_key.md
```diff
@@ -0,0 +1,38 @@
...
+{{since('nightly')}}
+
+This function accepts a Lua table `table` and a key `key`.
+It returns `true` if `table` contains a key equal to `key` (with non-nil value)
+and `false` otherwise.
+
+The function accepts an optional arbitrary number of extra arguments, that will
+all be intepreted as extra keys to check for recursively in the table. I.e., to
+check whether `table` has any non-nil value at `table.a.b.c`, we can use
+`wezterm.table.has_key(table, 'a', 'b', 'c')`.
```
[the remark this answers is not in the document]
Danielkonge: I can update it to `table[a][b][c]`, but what do you mean by not valid syntax? I am pretty sure that `table.a.b.c` works fine? (I might be misunderstanding something though...)

I will also update the wording.

lua-api-crates/table-funcs/src/lib.rs
```diff
@@ -0,0 +1,298 @@
...
+    let mut tbl_vec: Vec<(LuaValue, LuaValue)> = vec![];
+    for table in array_of_tables {
+        for pair in table.pairs::<LuaValue, LuaValue>() {
+            let (key, value) = pair?;
+            tbl_vec.push((key, value));
+        }
+    }
+    let tbl_len = tbl_vec.len();
+    // note we might allocate a bit too much here, but in many use cases we will be correct
+    let tbl: Table<'lua> = lua.create_table_with_capacity(0, tbl_len)?;
```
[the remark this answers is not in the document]
Danielkonge: I will update this soon.

lua-api-crates/table-funcs/src/lib.rs
Danielkonge: > FWIW, I think this whole function can be simplified to avoid the special case of the initial key, something like this:
> 
> ```rust
> fn get<'lua>(
>     _: &'lua Lua,
>     (mut value, mut keys): (Table<'lua>, LuaMultiValue<'lua>)
> ) -> mlua::Result<LuaValue<'lua>> {
>      for key in keys {
>         match value {
>              LuaValue::Table(tbl) => {
>                   value = tbl.get(key)?;
>              }
>              _ => {
>                // Cannot index non-table
>                return Ok(LuaValue::Nil)
>             }
>          }
>      }
>      Ok(value)
> }
> ```

I did consider that too, but choose the other way to hopefully have better error messages. If you prefer this way, I am fine with changing it.

lua-api-crates/table-funcs/src/lib.rs
```diff
@@ -0,0 +1,298 @@
...
+    lua: &'lua Lua,
+    (table, value, behavior): (Table<'lua>, LuaValue, Option<DepthMode>),
+) -> mlua::Result<bool> {
+    let behavior = behavior.unwrap_or_default();
+    match behavior {
+        DepthMode::Top => {
+            // we don't need a clone in this case
+            for pair in table.pairs::<LuaValue, LuaValue>() {
+                let (_, tbl_value) = pair?;
+                if tbl_value == value {
```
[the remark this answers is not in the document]
Danielkonge: I will just remove the behavior parameter for now then, and only support a shallow search in this PR. 

As to `==`, I think this should be equivalent to mlua's `.eq()`, but I can change it to that to make it clearer what is going on.

lua-api-crates/table-funcs/src/lib.rs
```diff
@@ -0,0 +1,298 @@
...
+                    return Ok(true);
+                }
+            }
+        }
+        DepthMode::Deep => {
```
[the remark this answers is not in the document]
Danielkonge: I can just remove it to simplify the code for now. I thought it might be useful to see if certain values were precent in one's config, but in that case you would usually know what key it should be set for anyways.

lua-api-crates/table-funcs/src/lib.rs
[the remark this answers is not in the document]
Danielkonge: Or do you mean `.equals()`? I think that is a custom method whereas `.eq()` is just derived.

lua-api-crates/table-funcs/src/lib.rs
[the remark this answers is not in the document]
Danielkonge: Actually looking at the updated code, I don't really think it adds much extra. Though I do agree that it will probably only be rarely used. Do you prefer to just remove it?

docs/config/lua/wezterm.table/clone.md
```diff
@@ -0,0 +1,57 @@
...
+assert(tbl == tbl_copy)
+assert(not (tbl == tbl_top_clone))
+assert(not (tbl == tbl_deep_clone))
+assert(equal(tbl, tbl_top_clone))
+assert(equal(tbl, tbl_deep_clone))
+
+tbl.a = 2
+assert(not equal(tbl, tbl_top_clone))
+assert(tbl_top_clone.a == 1)
+assert(not equal(tbl, tbl_deep_clone))
```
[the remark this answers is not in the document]
Danielkonge: I have added this now.

docs/config/lua/wezterm.table/extend.md
```diff
@@ -0,0 +1,78 @@
...
+}
+
+assert(equal(extend { tbl1, tbl2 }, { a = 2, b = { e = 5 }, c = 3, d = 4 }))
+assert(
+  equal(
+    extend({ tbl1, tbl2 }, 'Keep'),
+    { a = 1, b = { d = 4 }, c = 3, d = 4 }
+  )
+)
+-- This will return an error: extend({tbl1, tbl2}, 'Error')
```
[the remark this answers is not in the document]
Danielkonge: I have added something here too.

aarondill:
I can't quite read the source for "impl_lua_conversion_dynamic!" well enough to know if it already is, but is it possible to make the behavior enums case insensitive? if not then don't worry about it, but since we're using behavior strings and these won't be accessible as enums from lua, i'd prefer (personally) to use lowercase strings.

case insensitive would provide the best of both worlds, as well as ensuring that users don't have to remember whether each method needs capitalized or lowercase strings.

lua-api-crates/table-funcs/src/lib.rs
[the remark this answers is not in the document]
Danielkonge: I have updated this now and made a small fix of `deep_extend`too.

lua-api-crates/table-funcs/src/lib.rs
```diff
@@ -0,0 +1,298 @@
...
+        }
+    }
+    let tbl_len = tbl_vec.len();
+    // note we might allocate a bit too much here, but in many use cases we will be correct
+    let tbl: Table<'lua> = lua.create_table_with_capacity(0, tbl_len)?;
+
+    let behavior = behavior.unwrap_or_default();
+    for (key, value) in tbl_vec {
+        if !tbl.contains_key(key.clone())? {
+            tbl.set(key, value)?;
```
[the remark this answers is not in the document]
Danielkonge: > There's some inconsistency in behavior here:
> 
> ```lua
> local tbl1 = { a = 1 }
> local tbl2 = { a = {1,2,3} }
> 
> local res1 = extend_deep({tbl1, tbl2}, 'Force')
> local res2 = extend_deep({tbl2, tbl1}, 'Keep')
> ```
> 
> Even though both results superficially look the same:
> 
> ```
> {a = {1,2,3}}
> ```
> 
> `res1.a` is a clone of `tbl2.a` whereas `res2.a` is a reference to `tbl2.a`.
> 
> I think both should be clones.
> 
> What I think is that we should promote `wezterm.table.clone` to `wezterm.clone` and have it apply to an arbitrary lua value; that would be generally useful anyway and would simplify the logic in here.
> 
> Then we can call that here to make a deep copy of `value` before we assign it to the table.

I have made a few changes. I think both should be clones as you want now.

lua-api-crates/table-funcs/src/lib.rs
Danielkonge: I did something similar to what was mentioned here now, but I used `.get` from `luahelper` on the `UserData` instead of the above.

Danielkonge:
I think I have addressed most of the feedback now, but I still need to edit the following:

- `has_value` (code + docs): Do we want to just remove the behavior and only do the "Shallow" case?
- `clone` (code + docs): Do we want to move this to `wezterm.clone`? In that case, how should it clone `UserData`? 
- Should we rename `"Keep"` and `"Force"` to something like `"KeepFirst"` and `"TakeLast"`?

lua-api-crates/table-funcs/src/lib.rs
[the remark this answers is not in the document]
Danielkonge: I have cleanup up the `extend_deep` code more now. There were some edge cases, where it didn't behave as I would expect before. 

Now I do use clone, and I think the code itself is pretty clear now (as in, it is clear what the code is doing), but it did get a bit longer again because of different edge case behavior.

Danielkonge:
There are probably a few minor things left to edit in the code, but I think it is close enough that I can start to write some tests. @wez is there any other place in the code that does testing of similar Lua function implementations that I can base my tests on? Otherwise I will have to spend a little time and read the mlua docs more carefully.

Danielkonge:
@wez I kind of forgot about this. Could you give some feedback on the last two comments? Then I can go through and clean this up during this weekend (and maybe some of next week).

PR #4339, opened 2023-09-25
Fix mouse scroll on hovered but unfocused pane

### LINKED ISSUE #2373: Scroll unfocused pane

macOS

_No response_

### WezTerm version

20220807-193129-ac3a9c00

### Describe the bug

Seemingly similar to https://github.com/wez/wezterm/issues/1437. Cannot scroll on a pane which is unfocused, it will instead scroll the currently active pane.

### To Reproduce

Create two panes, attempt to scroll on the inactive one.

### Configuration

no config

### Expected Behavior

(It should be possible to enable the behavior that) scrolling over the inactive pane should scroll the hovered pane, not the currently focused one.

### Logs

_No response_

### Anything else?

I could imagine this behavior not being desirable to people, given the lack of mention of it that i can find in config/issues, so perhaps a there ought to be a config option?

tradiff:
Closes #2373

Scrolling a non-focused pane with the mouse wheel wasn't working. It turns out we are detecting the hovered pane, and `perform_key_assignment` does have the hovered pane, but it just wasn't passing this information in to `scroll_by_current_event_wheel_delta`. This resulted in `scroll_by_line` repeating the operation to get the current pane, and this doesn't have the hovering logic.

The fix is to pass the pane along in the call to `scroll_by_current_event_wheel_delta`, and it's subsequent call to `scroll_by_line`

### DIFF
--- a/wezterm-gui/src/termwindow/mod.rs
+++ b/wezterm-gui/src/termwindow/mod.rs
@@ -2360,22 +2360,18 @@ impl TermWindow {
         Ok(())
     }
 
-    fn scroll_by_current_event_wheel_delta(&mut self) -> anyhow::Result<()> {
+    fn scroll_by_current_event_wheel_delta(&mut self, pane: &Arc<dyn Pane>) -> anyhow::Result<()> {
         if let Some(event) = &self.current_mouse_event {
             let amount = match event.kind {
                 MouseEventKind::VertWheel(amount) => -amount,
                 _ => return Ok(()),
             };
-            self.scroll_by_line(amount.into())?;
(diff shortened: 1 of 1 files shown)


tradiff:
@wez, I just wanted to bring my PR to your attention and see if there's anything else I should update or improve. I totally understand that you're a busy person, so no rush at all 😄

Thanks for your time and all the work you do on wezterm. I appreciate any feedback you can provide when you have a moment.

wez:
Looks good, thanks!

Felixoid:
Many thanks! The scroll works now. Unfortunately, only partially, though.

PR #4343, opened 2023-09-26
read_dir add optional function first draft.

Danielkonge:
There is no documentation yet (I will do that later), but this has the functionality I hoped for.


Danielkonge:
I think this is ready for at least partial review now. Maybe don't focus too much on the exact implementation, but more on if the idea makes sense. 

I implemented a possibility of having multiple steps of sorting, but after using it, I don't think it is that useful and I have considered changing it so that you can add some formatting to the entry output instead.

aarondill:
I was trying to test this (fwiw, it seems to work for me) using the artifact from the ubuntu22.04 build action, but it seems like all my fonts are now bolded? Is this just an issue with the build artifacts or is it possibly an issue with the PR? If this is a known issue with PR build artifacts, do you know any way to fix it? I'd love to test this more thoroughly, but the font issues are distracting.

The left is the nightly release (installed from deb) and the right is this pr (running from the extracted tar, though i've also run it with the deb): The difference is slight, but noticeable. Especially on the prompt line.

Danielkonge:
> I was trying to test this (fwiw, it seems to work for me) using the artifact from the ubuntu22.04 build action, but it seems like all my fonts are now bolded? Is this just an issue with the build artifacts or is it possibly an issue with the PR? If this is a known issue with PR build artifacts, do you know any way to fix it? I'd love to test this more thoroughly, but the font issues are distracting.
> 
> The left is the nightly release (installed from deb) and the right is this pr (running from the extracted tar, though i've also run it with the deb): The difference is slight, but noticeable. Especially on the prompt line. 

This pull request should not change anything to do with fonts or rendering. Do you know when your nightly build is from? I can see that there were some font changes in late September. Maybe that is where something changed? (This PR is up to date with the latest commit.) 

If there is a difference between font rendering in this PR and the latest commit here in GitHub, then it sounds like a bug somewhere. In that case, we should try to figure out what is causing it.

aarondill:
@Danielkonge My current wezterm version `wezterm -V` is `wezterm 20230712-072601-f4abf8fd`

aarondill:
Okay, having updated, this seems to in fact be an issue with the nightly.

aarondill:
Since you seem to know when the change was, do you know how i could revert this in my own config? Thanks :smile:

Danielkonge:
> Since you seem to know when the change was, do you know how i could revert this in my own config? Thanks 😄

From what I can tell in the commits there have been a few different changes to the font code since July, so I don't know which exact one affects you here. 

Since you might want to update at some point anyway, I think it would be a better idea to update your code for font selection. Do you adjust the weight right now? 

The default weight should be `Regular`, so if that is too thick for you, you might prefer `Book` or `Demilight` (or even thinner). 

See: https://wezfurlong.org/wezterm/config/lua/wezterm/font.html

E.g., I like slightly thicker text than `Regular`, so my font setup is:
```
    config.font = wezterm.font({
        family = 'JetBrains Mono',
        weight = 'Medium',
        harfbuzz_features = { 'calt=0', 'clig=0', 'liga=0' },
    })
```
(I toggle ligatures with `LEADER-L`)

aarondill:
I seem to have fixed the font problem (by checking wezterm.version and setting "Thin" on the nightly, odd that this corresponds to what was previously Regular), but now I'm having selection issues.

On the nightly (left), the text color doesn't invert when selected (making it unreadable), while the stable (right) does invert the text color, which makes it so you can actually read it while selected. Is this a config option or a bug?



Thanks for the help btw. Sorry for invading your PR.

aarondill:
This actually seems to affect all highlights, including `less`'s messages: (look at the bottom)

Danielkonge:
> I seem to have fixed the font problem (by checking wezterm.version and setting "Thin" on the nightly, odd that this corresponds to what was previously Regular), but now I'm having selection issues.

I think the font used to be thinner than what people were used to from kitty or alacrity, so I assume it was just adjusted so the standard is the same as those now. (Though I am not sure about this.)

> On the nightly (left), the text color doesn't invert when selected (making it unreadable), while the stable (right) does invert the text color, which makes it so you can actually read it while selected. Is this a config option or a bug? 
> 
> I have this set, but it doesn't seem to work on nightly?
> 
> ```lua
> return {
> --- ...
>       -- the foreground color of selected text
>       selection_fg = "black",
>       -- the background color of selected text
>       selection_bg = "#fffacd",
> --- ...
> }
> ```

Hm, I guess you could have a theme overwriting it possibly? But I am not sure. Just to double check, what happens if you change "black" to "#000000"? (This should not be the problem.)

> Thanks for the help btw. Sorry for invading your PR.

That's fine. This PR is just for my work so far. I will soon close this PR and make a new one, where I better sort out everything.

aarondill:
Here's my full (theme relevant) config if it helps: https://github.com/aarondill/wezterm/blob/main/plugins/theme.lua

Danielkonge:
Looking at your config, I don't know which version of this pull request you are using, but I am playing around with making `read_dir` return `Path` objects instead of strings. (This is not final. I might keep it as strings to not mess with peoples config as what might be the case here?) So this https://github.com/aarondill/wezterm/blob/55a99df9d09d04bfa85fc3af86209c8fc836ea9c/plugin.lua#L111C4-L111C4 might cause you problems. Try to call `tostring` on `abs_name` just to make sure that you are working with strings and not getting some errors from that.

aarondill:
> Looking at your config, I don't know which version of this pull request you are using, but I am playing around with making `read_dir` return `Path` objects instead of strings. (This is not final. I might keep it as strings to not mess with peoples config as what might be the case here?) So this https://github.com/aarondill/wezterm/blob/55a99df9d09d04bfa85fc3af86209c8fc836ea9c/plugin.lua#L111C4-L111C4 might cause you problems. Try to call `tostring` on `abs_name` just to make sure that you are working with strings and not getting some errors from that.

actually, right now I'm just trying to get the nightly working. since I've already ruled this PR out from the issues. When I change the values, the color changes, but I think this is a possibly a consequence of the alpha value?

aarondill:
If I turn down the alpha though (more transparent) it just gets darker, which I suppose makes sense since my BG is black, but I can't seem to get the text to be readably without having a grey selection...

aarondill:
It seems to me like the alpha value might actually be broken right now, because when i decrease the alpha of the selection_bg, the text also decreases in brightness.

Danielkonge:
How does it look if you set `window_background_opacity = 1`? Also are you using Wayland? I remember hearing problems with Wayland and wezterm for a few things.

aarondill:
Here it is with no opacity:


I'm currently using AwesomeWM on Xorg.

Danielkonge:
Screenshot 2023-10-25 at 23 49 15

Just noticed this, so the alpha is probably a good guess for what is affecting you.

aarondill:
(top is stable, bottom is nightly)

aarondill:
> Screenshot 2023-10-25 at 23 49 15
> Just noticed this, so the alpha is probably a good guess for what is affecting you. Though this shouldn't be new...

It shouldn't be new, but it seems like maybe something changed between stable and nightly about this.
I can't seem to find any alpha value that can emulate the old behavior though

aarondill:
Do you think I should open a new issue about this? It seems like perhaps I've run into a bug in the rendering behavior.

Danielkonge:
> Do you think I should open a new issue about this? It seems like perhaps I've run into a bug in the rendering behavior.

It seems like a bug to me, yes. Have you tried to see what is the smallest config you can use to reproduce the bug. If you just have `selection_fg` and `selection_bg` set in your config and nothing else, does it work as expected then?

aarondill:
I seem to have found the issue. It's the font size. Even though it's rendering at the same size (Tiny in nightly vs Regular in stable), for some reason, it affects the color of the highlight.
```lua
return {
    "JetBrains Mono", -- <built-in>, BuiltIn
  }, { weight = "Thin" }), --- Try ExtraBlack to fix it. Though, the font size is now wrong.
  colors = {
    -- the foreground color of selected text
    selection_fg = "black",
    -- the background color of selected text
    selection_bg = "#fffacd",
  },
}
```
Thin:


ExtraBlack:

aarondill:
This is also an issue on the stable, but since I'm using Regular instead of Thin, the effect isn't as pronounced.

Danielkonge:
Yes, this seems to be the problem and an actual bug that you can make an issue for. I also have this effect of making the text thinner when selecting it when using `selection_fg = 'black`. But it is a weird bug, it doesn't seem to have this behavior with white?
 
Screenshot 2023-10-26 at 00 26 26


Screenshot 2023-10-26 at 00 26 49

aarondill:
I don't have the time to make that issue right now, because I've got some irl responsibilities to get to, but once I get some time later today, I'll go make an issue and hopefully it'll get figured out eventually. Considering how this was to _find_, I imagine it'll be complicated to fix, but here's hoping.

Danielkonge:
> I don't have the time to make that issue right now, because I've got some irl responsibilities to get to, but once I get some time later today, I'll go make an issue and hopefully it'll get figured out eventually. Considering how this was to _find_, I imagine it'll be complicated to fix, but here's hoping.

I don't think it will be too bad. We have narrowed it down a lot now. That hopefully makes it easier to find the bug and fix it.

lua-api-crates/filesystem/src/path.rs
```diff
@@ -286,7 +278,39 @@ impl UserData for Path {
...
+                .get::<_, mlua::Table>("string")?
+                .get::<_, mlua::Function>("format")?;
+            format.call::<_, LuaMultiValue>((lua_str, multi))
+        });
+        methods.add_method("byte", |lua: &'lua Lua, this, multi: LuaMultiValue| {
+            let lua_str = path_to_lua_str(lua, &this)?;
+            let format = lua
+                .globals()
+                .get::<_, mlua::Table>("string")?
+                .get::<_, mlua::Function>("format")?;
```
aarondill, line 296: Shouldn’t this be byte?

lua-api-crates/filesystem/src/path.rs
```diff
@@ -286,7 +278,39 @@ impl UserData for Path {
...
+            format.call::<_, LuaMultiValue>((lua_str, multi))
+        });
+        methods.add_method("byte", |lua: &'lua Lua, this, multi: LuaMultiValue| {
+            let lua_str = path_to_lua_str(lua, &this)?;
+            let format = lua
```
aarondill, line 293: You seem to have forgotten to change a few variable names while copy pasting.

lua-api-crates/filesystem/src/path.rs
Danielkonge: Yes, I had a few of these copying mistakes. I think it should be fixed now.

Danielkonge:
I moved this to #4493.

PR #4366, opened 2023-09-30
docs: update Arch install location

frankebel:
Arch Linux had a [git migration](https://archlinux.org/news/git-migration-announcement/) in May of this year. During that, the `[community]` repository was merged into `[extra]`. The old link was dead, so I updated it to the new location.


wez:
Thanks!

PR #4393, opened 2023-10-03
Start line selection at 0th cell

Tomiyou:
Starting line selection at 0th cell allows line selection to be extended using Shift+Mouse without breaking the selection start. This mostly happens when selecting a line then scrolling down and extending the selection.


bew:
Hello! Is this still useful to you?

PR #4397, opened 2023-10-04
Update shell-integration.md

Danielkonge:
I am not sure if another method than what is suggested here is recommended? In either case, I think some text explaining how to use the shell scripts should be added here.

### DIFF
--- a/docs/shell-integration.md
+++ b/docs/shell-integration.md
@@ -16,6 +16,11 @@ shell program to emit the escape sequences at the appropriate place.
 
 You can find some [examples for various shells in the wezterm
 repo](https://github.com/wez/wezterm/tree/main/assets/shell-integration).
+
+To use this file to setup shell integration in wezterm with Bash or Zsh, you can
+copy the file to your computer and source it via `. /path/to/file.sh` in your `.bashrc`
+or `.zshrc`, or you can install it at `/etc/profile.d` on most unix systems.
+
 Xonsh is supported via a [term-integrations](https://github.com/jnoortheen/xontrib-term-integrations) plugin.
 
 Starting with version 20210314-114017-04b7cedd, the Fedora and Debian packages


uncomfyhalomacro:
you dont need chmod +x here. install it at `/etc/profile.d` or just source it

Danielkonge:
> you dont need chmod +x here. install it at `/etc/profile.d` or just source it

I don't have a `/etc/profile.d` (I am using macOS), and I am pretty sure I did try just sourcing it, and it didn't work. I think that was why I wanted to include something like this in the docs in the first place. 

Maybe a better addition than what I wrote, would be to have a list of ways to use the script with a list of which operating system each method works for?

wez:
the file doesn't need to be executable in `/etc/profile.d`.  Those files are sourced via `. /path/to/file.sh` rather than `source /path/to/file.sh`.  I'd rather avoid the complication of executable mode here in the docs if you can confirm that the `.` approach works fine without it.

Danielkonge:
> the file doesn't need to be executable in `/etc/profile.d`. Those files are sourced via `. /path/to/file.sh` rather than `source /path/to/file.sh`. I'd rather avoid the complication of executable mode here in the docs if you can confirm that the `.` approach works fine without it.

Sourcing via `. /path/to/file.sh` seems to work, so I updated the PR. Is the new version better?

docs/shell-integration.md
```diff
@@ -16,6 +16,11 @@ shell program to emit the escape sequences at the appropriate place.
 
 You can find some [examples for various shells in the wezterm
 repo](https://github.com/wez/wezterm/tree/main/assets/shell-integration).
+
+To use this file to setup shell integration in wezterm with Bash or Zsh, you can
+copy the file to your computer and source it via `. /path/to/file.sh` in your `.bashrc`
+or `.zshrc`, or you can install it at `/etc/profile.d` if using Linux.
```
wez, line 22:
```suggestion
or `.zshrc`, or you can install it at `/etc/profile.d` on most unix systems.
```

wez:
Thanks!

PR #4413, opened 2023-10-09
Fix Gnome window buttons DPI scaling

Tomiyou:
Fixed Gnome window buttons not scaling properly on high DPI displays. Instead of a fixed size, Gnome window button size will now be calculated from top bar height instead. Since the height of the top is adjusted whenever the DPI changes, window buttons are now scaled properly.

### DIFF
--- a/wezterm-gui/src/termwindow/render/fancy_tab_bar.rs
+++ b/wezterm-gui/src/termwindow/render/fancy_tab_bar.rs
@@ -288,6 +288,7 @@ impl crate::TermWindow {
                     &font,
                     &metrics,
                     &self.config,
+                    tab_bar_height,
                 ),
             }
         };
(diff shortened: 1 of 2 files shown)


DanScharon:
This is the only reason why I keep the (ugly) default button style (Windows) for now. Without DPI scaling the Gnome buttons are just too small.
A pity this hasn't been reviewed yet

wezterm-gui/src/termwindow/render/fancy_tab_bar.rs
```diff
@@ -288,6 +288,7 @@ impl crate::TermWindow {
                     &font,
                     &metrics,
                     &self.config,
+                    tab_bar_height,
```
wez, line 291: this isn't the right approach. The font context and render metrics contain sufficient information about the size and dpi that the existing poly drawing stuff can use to scale appropriately.

Probably the right way to resolve this is to switch to using Points instead of Pixels in the gnome logic, similar to the way that the Windows flavor of these decorations works.

wez (commented):
Thanks for taking a stab at this! It's not the right way to go to resolve this though.

PR #4457, opened 2023-10-17
macos: fix leak of Window and View objects

0f-0b:
They were held strongly by [`Connection.windows`](https://github.com/wez/wezterm/blob/fec90ae04bf448d4b1475ba1d0ba1392846a70d6/window/src/os/macos/connection.rs#L24) and never released.


wez:
Thanks!

PR #4468, opened 2023-10-20
ci: Use sccache-action instead of rust-cache

jalil-salame:
It reduced ci times for fluke from 9min to 1min in MacOS (see [this issue](https://github.com/hapsoc/fluke/issues/118)) so I thought it would be worth a try here too.

`sccache` does not have a configurable `key` for the cache so it might not work at all for us.

### DIFF
--- a/ci/generate-workflows.py
+++ b/ci/generate-workflows.py
@@ -130,7 +130,7 @@ class CacheStep(ActionStep):
 
 class CacheRustStep(ActionStep):
     def __init__(self, name, key):
-        super().__init__(name, action="Swatinem/rust-cache@v2", params={"key": key})
+        super().__init__(name, action="mozilla-actions/sccache-action@v0.0.3")
 
 
 class CheckoutStep(ActionStep):
@@ -767,6 +767,9 @@ cargo build --all --release""",
         return steps
 
     def global_env(self):
+        self.env["CARGO_INCREMENTAL"] = "0"
+        self.env["SCCACHE_GHA_ENABLED"] = "true"
+        self.env["RUSTC_WRAPPER"] = "sccache"
         if "macos" in self.name:
             self.env["MACOSX_DEPLOYMENT_TARGET"] = "10.9"
         if "alpine" in self.name:
@@ -842,6 +845,15 @@ cargo build --all --release""",
                         "sed 's/root:!/root:*/g' -i /etc/shadow",
                     ),
                 ]
+            if "opensuse" in self.container:
+                steps += [
+                    # This holds the xcb bits
+                    RunStep(
+                        "Install tar",
+                        "zypper install -yl tar",
+                    ),
+                ]
+
         steps += self.install_newer_compiler()
         steps += self.install_git()
         steps += self.install_curl()
@@ -854,7 +866,8 @@ cargo build --all --release""",
 
         steps += self.install_openssh_server()
         steps += self.checkout()
-        steps += self.install_rust(cache="mac" not in self.name)
+        # We should be able to cache mac builds now?
+        steps += self.install_rust() # cache="mac" not in self.name)
         steps += self.install_system_deps()
         return steps
 
--- a/ci/generate-workflows.py
+++ b/ci/generate-workflows.py
@@ -73,15 +73,18 @@ class Step(object):
 
 
 class RunStep(Step):
-    def __init__(self, name, run, shell="bash", env=None):
+    def __init__(self, name, run, shell="bash", env=None, condition=None):
         self.name = name
         self.run = run
         self.shell = shell
         self.env = env
+        self.condition = condition
 
     def render(self, f, depth=0):
         indent = "  " * depth
         f.write(f"{indent}- name: {yv(self.name)}\n")
+        if self.condition:
+            f.write(f"{indent}  if: {self.condition}\n")
         if self.env:
             f.write(f"{indent}  env:\n")
             keys = list(self.env.keys())
@@ -98,17 +101,20 @@ class RunStep(Step):
 
 
 class ActionStep(Step):
-    def __init__(self, name, action, params=None, env=None, condition=None):
+    def __init__(self, name, action, params=None, env=None, condition=None, id=None):
         self.name = name
         self.action = action
         self.params = params
         self.env = env
         self.condition = condition
+        self.id = id
 
     def render(self, f, depth=0):
         indent = "  " * depth
         f.write(f"{indent}- name: {yv(self.name)}\n")
         f.write(f"{indent}  uses: {self.action}\n")
+        if self.id:
+            f.write(f"{indent}  id: {self.id}\n")
         if self.condition:
             f.write(f"{indent}  if: {self.condition}\n")
         if self.params:
@@ -122,14 +128,14 @@ class ActionStep(Step):
 
 
 class CacheStep(ActionStep):
-    def __init__(self, name, path, key):
+    def __init__(self, name, path, key, id=None):
         super().__init__(
-            name, action="actions/cache@v3", params={"path": path, "key": key}
+            name, action="actions/cache@v3", params={"path": path, "key": key}, id=id
         )
 
 
-class CacheRustStep(ActionStep):
-    def __init__(self, name, key):
+class SccacheStep(ActionStep):
+    def __init__(self, name):
         super().__init__(name, action="mozilla-actions/sccache-action@v0.0.3")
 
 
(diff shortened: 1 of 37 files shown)


jalil-salame:
You have to allow the new action in the repo configuration. It's `mozilla-actions/sccache-action@v0.0.3`

ci/generate-workflows.py
```diff
@@ -130,7 +130,7 @@ def __init__(self, name, path, key):
 
 class CacheRustStep(ActionStep):
     def __init__(self, name, key):
-        super().__init__(name, action="Swatinem/rust-cache@v2", params={"key": key})
+        super().__init__(name, action="mozilla-actions/sccache-action@v0.0.3")
```
jalil-salame, line 133: `sccache` has no way of adding a custom key yet(?). Might or might not trip us up

ci/generate-workflows.py
```diff
@@ -758,6 +758,9 @@ def update_homebrew_tap(self):
         return steps
 
     def global_env(self):
+        self.env["CARGO_INCREMENTAL"] = "0"
+        self.env["SCCACHE_GHA_ENABLED"] = "true"
+        self.env["RUSTC_WRAPPER"] = "sccache"
```
jalil-salame, line 763: There is probably a better way of doing this but the whole script is a big hack so...

ci/generate-workflows.py
```diff
@@ -845,7 +848,8 @@ def prep_environment(self, cache=True):
 
         steps += self.install_openssh_server()
         steps += self.checkout()
-        steps += self.install_rust(cache="mac" not in self.name)
+        # We should be able to cache mac builds now?
+        steps += self.install_rust() # cache="mac" not in self.name)
```
jalil-salame, line 852: `sccache` might fix macos caching or it might break because of this (it is necessary because otherwise `RUSTC_WRAPPER` is set but `sccache` is not installed).

I might be able to conditionally disable the env vars, but it is worth a try.

jalil-salame (commented):
Adding aditional explanations to the changes

jalil-salame:
I might restart the generate ci script rewrite effort...

Thanks for all the work! I've really enjoyed using wezterm and I hope I can do some more work on improving this c:

wez:
> You have to allow the new action in the repo configuration. It's `mozilla-actions/sccache-action@v0.0.3`

I've added this action to the allow list; I think you might need to push/bump again this PR for the actions to run.

jalil-salame:
OpenSuse Leap is failing because it cannot find `tar`. I am assuming it is not installed.

Should I add it as a dependency to `get-deps`, or as a CI step like `git` and `curl`?

jalil-salame:
`opensuse_tumbleweed` tests are failing. I don't think `sccache` could cause those failures, but I'm not optimistic.

jalil-salame:
Besides the `opensuse` errors, the results seem mixed:

<details><summary>Results: (click me)</summary>

## Clean Cache

| Type                | Build Time | Test Time | Cache Hits | Cache Misses |
|---------------------|-----------:|----------:|-----------:|-------------:|
| macos (Intel)       |      25:58 |     12:47 |         17 |         2780 |
| macos (ARM)         |      18:53 |       N/A |        N/A |          N/A |
| debian11            |      19:13 |     11:12 |        432 |         1095 |
| debian12            |      18:19 |     11:56 |         36 |         1491 |
| fedora38            |      17:06 |     11:39 |        380 |         1147 |
| ubuntu22.04         |      15:28 |     12:09 |        417 |         1110 |

## Warm? Cache

| Type                | Build Time | Test Time | Cache Hits | Cache Misses |
|---------------------|-----------:|----------:|-----------:|-------------:|
| centos7             |      21:33 |     13:52 |          0 |         1528 |
| centos8             |      14:13 |     10:10 |        530 |         1268 |
| centos9             |      18:09 |     11:26 |        630 |         1168 |
| windows             |      37:06 |     19:42 |          0 |          702 |
| fedora36            |      18:12 |     10:46 |        637 |         1158 |
| fedora37            |      14:34 |     07:58 |        376 |         1151 |
| debian10.3          |      16:31 |     09:22 |        543 |          984 |
| ubuntu20.04         |      13:11 |     10:06 |        524 |         1274 |

</details>

### Conclusion

The environment is not optimal to test the impact:

- `sccache` is competing for cache storage with other actions
- `cargo-nextest` was not cached for a few builds
  - `sccache` also speeds up the build of `cargo-nextest` which is both good (faster compiles) and bad (pollutes the cache)
- Some build have suspiciously few cache hits (centos7, macos, and windows stand out)

My guess is `sccache` is better, but it is hard to tell.

### Potential Improvements

`sccache` is not caching crate downloads (it is not supposed to). Adding a `cargo vendor` step might improve compile times further.

wez:
> OpenSuse Leap is failing because it cannot find `tar`. I am assuming it is not installed.

Thanks for looking at this!

I'm considering removing the opensuse builds from the GH CI because they are the most unreliable platforms: they often break in weird ways for days at a time.

I recently added COPR builds for several platforms (https://wezfurlong.org/wezterm/install/linux.html#installing-on-fedora-and-rpm-based-systems-via-copr) which makes it less critical to have GH run so many itself.  Provided that failures of the COPR build bubble up and are noticed by me; that's currently a missing piece.

> Should I add it as a dependency to `get-deps`, or as a CI step like `git` and `curl`?

My mental model is: it goes in `get-deps` unless we need it before we're in a position to run get-deps.

wez:
re: mixed results; I'm increasingly of the opinion that GH actions runners are under powered for the projects that I work on; the disk space in the runners isn't especially generous and the total amount of storage for caches is limiting when there are multiple platforms to support.

wezterm's CI hasn't quite reached the limits of cache space, but I did hit them over in kumomta and decided to move its CI to a locally hosted woodpecker ci, which makes a significant difference (40 minute unreliable builds on GH -> 3 mins cached or ~5 mins cold on my woodpecker box).  That works for kumomta because the number of platforms are small and I have just enough hardware for that.

Based on that experience, I suspect that adding more platforms to wezterm will introduce similar reliability issues with caching and my lead to disabling caching for some of them, or just punting them to other build systems (like COPR).

jalil-salame:
> re: mixed results; I'm increasingly of the opinion that GH actions runners are under powered for the projects that I work on; the disk space in the runners isn't especially generous and the total amount of storage for caches is limiting when there are multiple platforms to support.

I've also hit the limits of the CI machines in a very small project (which was installing ROCm and thus using 10+GiB of space).

`sccache` should help with cache utilization; the cache is now (mostly) shared for all Linux builds.

`cargo vendor` should be shared for all builds if I end up adding that.

jalil-salame:
Caching the dependencies with `cargo vendor` seems like a failed experiment; GH CI, with all its warts, does have a good network connection so we only save at most ~30s. It is better for crates.io though so we might want to keep it if it doesn't hurt anything.

jalil-salame:
Now that tar is installed in OpenSUSE, both are (expectedly) failing the tests!

Conclusion of the experiment:

**tl;dr** `sccache` is not a big improvement but it helps with cold caches and cache utilization. Vendoring is not useful for `wezterm` but maybe crates.io will like it.

1. `sccache` did not bring the massive improvement I had hoped, but it does do some nice things:
  - It uses the cache much better (I estimated ~5.5GiB of cached artifacts (GH rate limited me so I only had ~1/3 of the data))
  - It caches intermediate results (as seen in the cache hits, the cache is populated as the action runs, so a cold start is still very good (i.e. centos7 and centos8 both start at the same time, thus some of the artifacts are compiled by centos7 some by centos8 and are being shared by sccache reducing the load).
  - It works on macos :tada: 
2. Vendoring dependencies is negligible, the downloads take ~30s only
3. OpenSUSE is flaky
4. I dislike the Python script, next PR I do will replace it with xtasks or smth (not soon sadly)

wez:
Sorry it took a while to get back to this. I like the idea of using less cache space. I think the changes are reasonable also. Would you mind rebasing this PR? I think it is mergeable!

jalil-salame:
I'm gonna have to look at this again, after the rebase I apparently messed something up as the actions use both cache methods now :smiling_face_with_tear:

Hopefully I can do this tomorrow, otherwise it might take a week...

jalil-salame:
> I'm gonna have to look at this again, after the rebase I apparently messed something up as the actions use both cache methods now 🥲
> 
> Hopefully I can do this tomorrow, otherwise it might take a week...

I feel so dumb... That was all me, I did it... everything is working as intended...

For future confused me:

We use `sccache` to cache the compilation artifacts, and we use `cache` to cache the dependencies so that we are nice to `crates.io`.

I was planning on squashing everything into two commits; the python changes and the workflow changes. Are you okay with me doing that?

wez:
I will often just click the "squash and merge" button if I feel like the number of commits in a PR is a bit high/verbose.
I'm happy to do that here to save you some effort, or I can wait for you to amend your PR history!

wez:
Thank you again!

jalil-salame:
Looking at this a month after, the cache usage looks much better :tada:

```console
$ gh cache list --json sizeInBytes -q '([.[].sizeInBytes] | add) / (1024 * 1024)' --limit 16716
7539.66
$ gh cache list --json sizeInBytes -q '([.[].sizeInBytes] | add) / (1024 * 1024)' --limit 16716 --key sccache
6535.33
```

Numbers in MiB.

On other news; I am preparing myself mentally to rewrite the ci script so I can tackle #3360 . The basic idea I have is to use a `needs` dependency to chain the CI executions. I might also take the plunge and look at the diff of the schedule vs push actions, and see if I can merge them into a monstrosity that doesn't need two files (and is hopefully more maintainable). Does that sound any good?

PR #4472, opened 2023-10-22
docs: shell-integration: Document activation by default on Arch Linux

adamency:
Arch Linux also activates shell integation by default: https://gitlab.archlinux.org/archlinux/packaging/packages/wezterm/-/commit/fdc4c53519703466dcd0fd2baa41859448a01b05

This PR documents it in the appropriate place.


adamency:
Closed and reopened with ad hoc branch instead of main: https://github.com/wez/wezterm/pull/4473

PR #4476, opened 2023-10-22
updated updated cargoand remove compilation warnings

jesusalc:
updated updated cargo and removed compilation warnings


wez:
resolved independently

Issue #4495, opened 2023-10-27
`tab_bar_at_bottom` and `window_decorations` shift tab bar


0xdead10cd:
### What Operating System(s) are you seeing this problem on?

macOS

### WezTerm version

20230712-072601-f4abf8fd

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

No, and I'll explain why below

### Describe the bug

Placing the tab bar at the bottom of the window and removing the title bar but keeping the integrated buttons results in the tabs being shifted right instead of being fully left aligned:
<img width="574" alt="Screenshot 2023-10-26 at 10 18 41 PM" src="https://github.com/wez/wezterm/assets/93291476/6554fdd4-5fad-41c8-9328-23653d56c0b0">

Removing the integrated buttons results in the tabs actually being fully left aligned:
<img width="574" alt="Screenshot 2023-10-26 at 10 25 53 PM" src="https://github.com/wez/wezterm/assets/93291476/7e51a388-4cd9-4ad0-97ec-357817ed2765">

My guess is that the empty space is somehow being reserved for the integrated buttons even though the tab bar is at the bottom of the window.

### To Reproduce

Just open the terminal with the config shared. The tabs are shifted right.

### Configuration

```lua
local wezterm = require 'wezterm'

local config = {}

if wezterm.config_builder then
	config = wezterm.config_builder()
end

config.tab_bar_at_bottom = true
config.window_decorations = "INTEGRATED_BUTTONS|RESIZE"

return config
```

### Expected Behavior

The tabs are left aligned within the tab bar.

benbooth493:
I'm facing this exact same issue

wez:
Fixed in https://github.com/wez/wezterm/pull/4505 and in recent releases

PR #4505, opened 2023-10-30
Fixes tabbar identation when placed at the bottom of the window

0xdead10cd:
Fixes issue #4495 where title bar is indented instead of being fully left aligned.

**Before:**
Screenshot 2023-10-29 at 5 12 02 PM

**After:**
Screenshot 2023-10-29 at 5 10 42 PM

### DIFF
--- a/wezterm-gui/src/tabbar.rs
+++ b/wezterm-gui/src/tabbar.rs
@@ -372,6 +372,7 @@ impl TabBarState {
         if use_integrated_title_buttons
             && config.integrated_title_button_style == IntegratedTitleButtonStyle::MacOsNative
             && config.use_fancy_tab_bar == false
+            && config.tab_bar_at_bottom == false
         {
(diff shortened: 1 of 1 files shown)


0xdead10cd:
Setting `config.use_fancy_tab = true`, while `config.tab_bat_at_bottom = true` and `config.window_decorations = "INTEGRATED_BUTTONS|RESIZE"` with this code change still results in this behavior:

Screenshot 2023-10-29 at 5 31 40 PM

To fix this, I think that the requirement at [line 374](https://github.com/wez/wezterm/blob/fec90ae04bf448d4b1475ba1d0ba1392846a70d6/wezterm-gui/src/tabbar.rs#L374) that the padding be added if `config.use_fancy_tab_bar = false` be removed, but I've kept it so far in case it is there for a reason that's not obvious to me.

PR #4530, opened 2023-11-03
Fallback to SGR when rendering 16/256 colors when MaxColors is 16M

quark-zju:
When MaxColors is 16M, the "palette index" is treated by "setaf" as a true color, not the classic 256 color. Users of `PaletteIndex` expects 256 color and let's fallback to SGR.

For example, setaf on tmux-direct is:

    setaf=\E[%?%p1%{8}%<%t3%p1%d%e38:2::%p1%{65536}%/%d:%p1%{256}%/%{255}%&%d:%p1%{255}%&%d%;m,

It only special treats the first 8 colors (which means `rgb(0,0,0..8)` cannot be rendered with `setaf`).

Related: #4528

### DIFF
--- a/termwiz/src/render/terminfo.rs
+++ b/termwiz/src/render/terminfo.rs
@@ -136,8 +136,16 @@ impl TerminfoRenderer {
             }
 
             let has_true_color = self.caps.color_level() == ColorLevel::TrueColor;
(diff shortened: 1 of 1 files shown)


wez:
Thanks!

Issue #4523, opened 2023-11-02
Have `wezterm -e` start with `--always-new-process`


Ram-Z:
This is a follow up of https://github.com/wez/wezterm/issues/2782.

Most terminals will not attach to a running instance of a daemon and therefore are blocking when invoked with `-e`. Some tools rely on waiting for the spawned program to terminate.

In my case I had qutebrowsers external editor command set to `$TERMINAL -e vim {file}`. qutebrowser would create `file` and wait for `vim` to finish before reading and deleting it. With `TERMINAL=wezterm` this didn't work as the file got deleted immediately.

**Describe the solution you'd like**
`wezterm -e` should behave as if `wezterm start --always-new-process`.

**Describe alternatives you've considered**
I now handle `TERMINAL==wezterm` specially and start it as a new process, but I'd rather it just worked with `-e`.

vimpostor:
I agree, `-e` should start a blocking wezterm for compatibility reasons.

I think `wezterm start` should still **not** use `--always-new-process` by default, only the `-e` alias should star a new process by default. I can send in a patch later. I don't have a strong opinion though, so if @wez wants `start` and `-e` to be always consistent, then that's fine with me too.

Ram-Z:
> I think `wezterm start` should still **not** use `--always-new-process` by default.

I agree.

> I don't have a strong opinion though, so if @wez wants `start` and `-e` to be always consistent, then that's fine with me too.

Yeah, and if that means that `-e` continues to behave the way it does now, that's fine too.

wez:
I think this request makes sense, but I'm not sure whether the `clap` library can make it work that way. If you'd like to look into how to make that work, I'd be happy to review a PR for it!

vimpostor:
> but I'm not sure whether the `clap` library can make it work that way

There is no way to implement it while keeping it as a simple clap alias, because [clap uses a bijective mapping model](https://github.com/clap-rs/clap/issues/1820#issuecomment-615547678), so there is no way to have different behaviour between different aliases: All aliases are the same element to clap and you can't determine which alias was used.

However that's no proble, it is easy to fake this with only a few lines of code.

PR #4535, opened 2023-11-05
Always block when run with "wezterm -e"

vimpostor:
See #4523

There is no way to do this with a declarative clap macro, so we have to fake it instead.


wezterm-gui/src/main.rs
```diff
@@ -1161,7 +1164,17 @@ fn run() -> anyhow::Result<()> {
...
+            // Act as if the normal start subcommand was used,
+            // except that we always start a new instance.
+            // This is needed for compatibility, because many tools assume
+            // that "$TERMINAL -e $COMMAND" blocks until the command finished.
+            SubCommand::BlockingStart(s) => SubCommand::Start(StartCommand {
+                always_new_process: true,
+                ..s
+            }),
+            _ => sub,
+        },
```
wez, line 1177: I think you can simplify this to this:

```suggestion
        Some(SubCommand::BlockingStart(start)) => {
            // Act as if the normal start subcommand was used,
            // except that we always start a new instance.
            // This is needed for compatibility, because many tools assume
            // that "$TERMINAL -e $COMMAND" blocks until the command finished.
            SubCommand::Start(StartCommand {
                always_new_process: true,
                ..start
            })
        },
        Some(sub) => sub,
```

wez (commented):
An inventive approach!

I'm curious about how this impacts the `--help` output.  Since the `-e` alias has now been moved to a hidden command, they need updating anyway, so please also update the doc pages for the cli; the steps are:

```console
$ cargo build
$ ci/update-derived-files.sh
$ ci/build-docs.sh
```

What that does is run `wezterm CMD --help` for each subcommand and embed it into the docs.  Then the doc build runs to reformat the docs and minimize the diff.

then amend the updated files into this PR.

I think that it would be good to explicitly mention the `-e` flag somewhere reasonable and discoverable in the clap help.  To cover the bases, I think that means `wezterm --help`, `wezterm-gui --help` and `wezterm start --help`.

vimpostor:
> I think that it would be good to explicitly mention the `-e` flag somewhere reasonable and discoverable in the clap help. To cover the bases, I think that means `wezterm --help`, `wezterm-gui --help` and `wezterm start --help`.

Done, I mention `-e` explicitly in the help output now. Let me know if you prefer another format.

Issue #383, opened 2020-12-18
Font resolution regression


jsgf:
## Describe the bug

Current master wezterm fails to resolve my configured font, whereas it previously worked fine.

## Environment (please complete the following information):

 - OS: Linux X11
 - Version: wezterm 20201101-103216-403d002d-127-g4f9e36a4

## To Reproduce

`wezterm` prints
```
 2020-12-18T22:17:06.642Z ERROR mux::connui > Configuration Error: Unable to load a font specified by your font=wezterm.font('Hasklig', {bold=false, italic=false}) configuration. Fallback(s) are being used instead, and the terminal may not render as intended. See https://wezfurlong.org/wezterm/config/fonts.html for more information
```
and shows poor quality text in the window.

## Configuration

Hasklig is installed the normal Gnome desktop way, and appears in the "Fonts" tool as expected, and is used by other apps such as vscode.

```
local wezterm = require('wezterm')

return {
        font = wezterm.font("Hasklig"),
}
```

## Expected behavior
Wezterm find and use `Hasklig` font as it did before.
## Additional context

I built from source with one small local change to remove the openssl-vendor feature.

jsgf:
Also whatever fallback font it's using looks like it needs a new typewriter ribbon

![image](https://user-images.githubusercontent.com/147966/102667055-4e747e80-413d-11eb-9011-a652333e6fd9.png)

wez:
Looks like it's getting a mixture of noto color emoji for the numbers and deja vu sans for some others.

Can you run `RUST_LOG=trace wezterm 2> /tmp/trace` and get me the `grep font /tmp/trace` lines?  That'll help to understand what its looking up.

Do you know roughly the revision of the last version that worked for you?  No need to do an exhaustive bisect; there were some fontconfig related changes around Dec 14.  There were some more invasive changes a couple of weeks prior to that, so if you know roughly when you last built it might help narrow down which of those things messed it up for you.

wez:
(On the plus side: we're doing a much better job at scaling those color emoji number glyphs than we used to in this sort of situation!)

jsgf:
```
$ fc-match Hasklig
Hasklig-Regular.otf: "Hasklig" "Regular"
```

trace output: https://gist.github.com/jsgf/cfbd20c86d521dd659a7abc3cee9be17

wez:
[...]

PR #4563, opened 2023-11-12
Make RIS also clear the alternate screen

ninjalj:
xterm clears the alternate screen on RIS since #383. VTE and konsole already did that.

From esctest:
RISTests.test_RIS_ExitAltScreen


wez:
Thanks!

PR #4565, opened 2023-11-12
DECRQCRA: treat uninitialized cells as spaces

ninjalj:
The concept of uninitialized cells in wezterm is not the same as that on VT520 or that on xterm, so treat them as spaces, at least when asking for the checksum of a single cell, which is what esctest does.

### DIFF
--- a/term/src/terminalstate/mod.rs
+++ b/term/src/terminalstate/mod.rs
@@ -1979,7 +1979,17 @@ impl TerminalState {
                 checksum += u16::from(ch as u8);
             }
         }
-        checksum
+
+        // Treat uninitialized cells as spaces.
+        // The concept of uninitialized cells in wezterm is not the same as that on VT520 or that
+        // on xterm, so, to prevent a lot of noise in esctest, treat them as spaces, at least when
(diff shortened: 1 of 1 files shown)


ninjalj:
While from what I can tell, wezterm treats as null newly allocated (postfixes of) lines, the VT520 also treats as null anything "erased", "deleted" or newly "scrolled" into view (save for DECSEL, DECSED, DECERA and DECSERA), while xterm does some ad-hoc thing. More information at the discussion starting at  https://gitlab.gnome.org/GNOME/vte/-/issues/2526#note_1878466

Anyways, this distinction is useless for esctest purposes, so return space to remove a lot of noise in esctest output.

term/src/terminalstate/mod.rs
```diff
@@ -1979,7 +1979,11 @@ impl TerminalState {
                 checksum += u16::from(ch as u8);
             }
         }
-        checksum
+        if checksum == 0 {
```
wez, line 1982: Thanks for this!

Please add a comment here; basically the description from the PR plus a reference to this PR, so that it is easier for someone in the future to understand why this is here!

wez:
Thanks!

PR #4567, opened 2023-11-12
macos: floating window support

### LINKED ISSUE #4566: top level (floating) window on macos
I would love to have my wezterm window on top of other windows.

This can be done in Swift via:
```swift
window.level = .floating
```

After a small search on github and cocoa crate repo i figure out a possible solution, and tried it in the [`hello_world`](https://github.com/servo/core-foundation-rs/blob/master/cocoa/examples/hello_world.rs) example of `cocoa-foundation-rs/cocoa`

```rust
      // create Window
      let window = NSWindow::alloc(nil)
          .initWithContentRect_styleMask_backing_defer_(
              NSRect::new(NSPoint::new(0., 0.), NSSize::new(200., 200.)),
              NSWindowStyleMask::NSTitledWindowMask,
              NSBackingStoreBuffered,
              NO,
          )
          .autorelease();


      window.setLevel_(1); // this is what we need

      window.cascadeTopLeftFromPoint_(NSPoint::new(20., 20.));
      window.center();
```

It would be super nice to have this option in the `Window` menu, something such as "Keep on top" or "Floating Window". 

I wanted to open a PR for this but I'm not familiar with the codebase and I'm not sure where this should be added. If you can provide some indications I can try implementing this myself and open a PR :) 

not implementing

**Additional context**

- [`cocoa-rs`#`setLevel_` documentation](https://docs.rs/cocoa/latest/cocoa/appkit/trait.NSWindow.html#tymethod.setLevel_) 

- [Official Apple `CGWindowLevelKey` documentation](https://developer.apple.com/documentation/coregraphics/cgwindowlevelkey)
- [`NSWindow.Level` documentation](https://developer.apple.com/documentation/appkit/nswindow/level)

rawnly:
Added `ToggleFloatingWindow` on macos. 
Tested on macOS Sonoma (14.0)

## Demo
https://github.com/wez/wezterm/assets/16429579/10e0975b-6c09-4287-b5dc-e595688eed03

fixes #4566


rawnly:
any feedback on this? @wez

window/src/lib.rs
```diff
@@ -294,6 +294,14 @@ pub trait WindowOps {
     /// Set some text in the clipboard
     fn set_clipboard(&self, clipboard: Clipboard, text: String);
 
+    /// Set window level (macos only!)
+    #[cfg(target_os = "macos")]
+    fn set_level(&self, level: WindowLevel);
+
+    /// Get window level (macos only!)
+    #[cfg(target_os = "macos")]
+    fn level(&self) -> Future<WindowLevel>;
```
wez, line 303: The concept of an "always-on-top" window is not specific to macOS, even though this PR only implements it there.

The interface in this module should allow for things to work on all platforms, even if the implementation isn't present there yet.

My suggestion is to draw some inspiration from the `winit` crate which has this [WindowLevel](https://docs.rs/winit/latest/winit/window/enum.WindowLevel.html) enum:

```rust
pub enum WindowLevel {
    AlwaysOnBottom,
    Normal,
    AlwaysOnTop,
}
```

I think we could have essentially the same enum here in this module, and define a `set_window_level` method for setting it.  I think `set_level` is not specific or descriptive enough as a name.

I'm not convinced that we need a method to retrieve the window level; winit doesn't have one, and avoiding adding another `Future` thing here simplifies the implementation.

```suggestion
    /// Set window level. Depending on the environment, and user preference,
    /// this may have no effect.
    fn set_window_level(&self, level: WindowLevel) {
      // Default is NOP
    }
```

window/src/os/macos/window.rs
```diff
@@ -745,6 +745,19 @@ impl WindowOps for Window {
...
+            inner.set_level(level as i64);
+            Ok(())
+        });
+    }
+
+    #[cfg(target_os = "macos")]
+    fn level(&self) -> Future<WindowLevel> {
+        Connection::with_window_inner(self.id, |inner| Ok(inner.level().into()))
+    }
+
```
wez, line 760: This file is already macos specific, so there is no need for the `cfg` attribute here

```suggestion
    fn set_window_level(&self, level: WindowLevel) {
        Connection::with_window_inner(self.id, move |inner| {
            inner.set_window_level(level);
            Ok(())
        });
    }
```

window/src/os/macos/window.rs
```diff
@@ -1149,6 +1180,16 @@ impl WindowInner {
...
+    fn set_level(&mut self, level: i64) {
+        unsafe {
+            NSWindow::setLevel_(*self.window, level);
+        }
+    }
+
+    fn level(&self) -> i64 {
+        unsafe { NSWindow::level(*self.window) }
+    }
+
```
wez, line 1192: According to both:

* https://github.com/rust-windowing/winit/blob/b3c87caa7c040111d8dcae764fe903f54df77426/src/platform_impl/macos/appkit/window.rs#L313
* https://jameshfisher.com/2020/08/03/what-is-the-order-of-nswindow-levels/

The macOS floating concept has level 3 rather than the 22 you selected in here.

Winit chooses to use `-1` as one-below-normal for `AlwaysOnBottom`.

```suggestion
    /// <https://developer.apple.com/documentation/appkit/nswindow/level>
    /// <https://jameshfisher.com/2020/08/03/what-is-the-order-of-nswindow-levels/>
    fn set_level(&mut self, level: WindowLevel) {
        unsafe {
            NSWindow::setLevel_(*self.window, match level {
               WindowLevel::Normal => 0,
               WindowLevel::AlwaysOnTop => 3,
               WindowLevel::AlwaysOnBottom => -1,               
            });
        }
    }
```

window/src/os/macos/window.rs
```diff
@@ -1069,6 +1082,24 @@ impl WindowInner {
...
+impl From<i64> for WindowLevel {
+    fn from(value: i64) -> Self {
+        match value {
+            int_value if int_value == Self::Floating as i64 => Self::Floating,
+            int_value if int_value == Self::Normal as i64 => Self::Normal,
+            _ => Self::Normal,
+        }
+    }
+}
+
```
wez, line 1102:
```suggestion
```

wez (commented):
Thanks for diving in!

I think this PR needs to consider all platforms; both X11 and Windows have an AlwaysOnTop concept, so the configuration and actions need to expand and be ready for that, even if you're not in a position to implement and test for those platforms yourself.

I've made some suggestions with that in mind for your macOS implementation.

I think you could probably get close to making this work on windows and perhaps also X11 if you took a peek at the equivalent functionality in the `winit` crate. I'm not going to require that you do both of these in this PR, but it would be very helpful if you could take a crack at making it compile there via the CI as a kind of first-pass at that.

* https://github.com/rust-windowing/winit/blob/b3c87caa7c040111d8dcae764fe903f54df77426/src/platform_impl/linux/x11/window.rs#L1135
* https://github.com/rust-windowing/winit/blob/b3c87caa7c040111d8dcae764fe903f54df77426/src/platform_impl/windows/window.rs#L843

rawnly:
Thanks for the review I'll have a look at it asap!

wezterm-gui/src/termwindow/mod.rs
```diff
@@ -2507,6 +2507,28 @@ impl TermWindow {
             ToggleFullScreen => {
                 self.window.as_ref().unwrap().toggle_fullscreen();
             }
+            #[cfg(target_os = "macos")]
+            ToggleFloatingWindow => {
+                let window = self.window.clone().unwrap();
+
+                promise::spawn::spawn(async move {
+                    match window.level().await {
```
rawnly, line 2515: without a `level()` function to retrive the current window level how can we perform the toggle?

wez: I'm not opposed to adding a way to understand the level. Having a method that returns a Future makes it a bit awkward to do that in certain contexts.

Maybe adding `ALWAYS_ON_TOP` and `ALWAYS_ON_BOTTOM` to `WindowState` would satisfy this?

https://github.com/wez/wezterm/blob/f07cc837166dc5f29df56afb42af3d6887ec6a33/window/src/lib.rs#L102-L114

WindowState is exposed to lua via [window:get_dimensions](https://wezfurlong.org/wezterm/config/lua/window/get_dimensions.html)

https://github.com/wez/wezterm/blob/f07cc837166dc5f29df56afb42af3d6887ec6a33/wezterm-gui/src/scripting/guiwin.rs#L115-L131

That said, this is already another async method call and already awkward. I suppose the advantage is that we wouldn't need to make multiple different async calls to resolve more information about the window overall.

rawnly: looks nice, yeah async method is awkward, I agree with you. I'll try adding this to the state!

config/src/keyassignment.rs
```diff
@@ -501,13 +501,20 @@ fn default_fuzzy_description() -> String {
     "Fuzzy matching: ".to_string()
 }
 
+// FIXME: this is so ugly we should use `window::WindowLevel` instead
+#[derive(Debug, Clone, PartialEq, FromDynamic, ToDynamic)]
+pub enum SetWindowLevelLevel {
```
rawnly, line 506: I wasn't sure about adding the window crate here, just for an enum. Maybe we can move it somewhere else?

wezterm-gui/src/commands.rs
```diff
@@ -681,20 +681,28 @@ pub fn derive_command_from_key_assignment(action: &KeyAssignment) -> Option<Comm
...
+        SetWindowLevel(SetWindowLevelLevel::AlwaysOnTop) => CommandDef {
             brief: "Always on Top".into(),
             doc: "Toggles the window between floating and non-floating states to stay on top of other windows.".into(),
             keys: vec![],
             args: &[ArgType::ActiveWindow],
-            menubar: &["Window"],
+            menubar: &["Window", "Level"],
+            icon: None,
+        },
+        SetWindowLevel(SetWindowLevelLevel::Normal) => CommandDef {
```
rawnly, line 692: CleanShot 2023-11-30 at 03 08 16@2x

Since now there are 2 options (3 with the default level) I tought it was good to have a submenu with selectable levels instead of 2 toggles. Also it would be nice (in both cases) to be able to display a tick when selected something like the last item of window menu 

CleanShot 2023-11-30 at 03 07 13@2x

rawnly: Also naming i believe it can be improved

rawnly: actually after some reasoning maybe the 2 options "Toggle Always on Top"  and "Toggle Always on Bottom" have a better UX (thinking of assigning a shortcut, having a toggle is easier than selecting a level). btw having the tick on the left of the option would improve the ux a lot since you can understand the current state of the window.

config/src/keyassignment.rs
wez: Ah, currently `window` depends on `config`, so trying to reference a type from `window` here wouldn't be possible because it would introduce a dependency cycle.

It is OK to define `WindowLevel` here in config and use it in the `window` crate.

wezterm-gui/src/commands.rs
wez: Having ticks is out of scope of this PR. I'm not sure how feasible it is to accommodate that concept, because a number of the states potentially require some kind of async call to resolve, and it may not be possible to handle those in all contexts, or even to invalidate and update in all contexts.  So, I like the idea, but let's put it out of mind for the moment!

wez: re: Toggle, I think that having both a `Set` and a `Toggle` operation will be important.  Some users will want to be sure that they can set a specific state regardless of the current state in some actions, while others will want to assign a key press to the toggle behavior.

wezterm-gui/src/termwindow/mod.rs
```diff
@@ -2507,28 +2507,33 @@ impl TermWindow {
             ToggleFullScreen => {
                 self.window.as_ref().unwrap().toggle_fullscreen();
             }
-            #[cfg(target_os = "macos")]
-            ToggleFloatingWindow => {
+            SetWindowLevel(level) => {
                 let window = self.window.clone().unwrap();
+                let is_always_on_top = self.window_state.intersects(WindowState::ALWAYS_ON_TOP);
+                let is_always_on_bottom = self.window_state.intersects(WindowState::ALWAYS_ON_BOTTOM);
```
wez, line 2513: For this, what I'd envisioned is adding a method to `WindowState` that returns the effective `WindowLevel`:

```rust
impl WindowState {
   pub fn as_window_level(self) -> WindowLevel {
      if self.contains(Self::ALWAYS_ON_TOP) {
         WindowLevel::AlwaysOnTop
      } else if self.contains(Self::ALWAYS_ON_BOTTOM) {
         WindowLevel::AlwaysOnBotttom
      } else {
         WindowLevel::Normal
      }
   }
}
```

then code over here can much more simply reason about the current level

window/src/lib.rs
```diff
@@ -71,6 +71,13 @@ pub enum MouseCursor {
     SizeLeftRight,
 }
 
+#[derive(Debug)]
```
wez, line 74: as suggested in a comment above, I think you can and should move this into the `config` crate.

```suggestion
#[derive(Debug, Copy, Clone, PartialEq, Eq)]
```

rawnly:
Thanks for the feedback, I applied the changes as you suggested, now there are 3 ways to set levels, via `SetLevel` and via `ToggleAlwaysOn...`

I moved the `WindowLevel` inside the `config` crate on a separate `mod` (`window`) to be more clean

rawnly:
i left docs for the so i don't have to write them again every time 😅

window/src/os/macos/window.rs
```diff
@@ -1149,6 +1159,16 @@ impl WindowInner {
...
     }
 
+    fn set_window_level(&mut self, level: WindowLevel) {
+        unsafe {
+            NSWindow::setLevel_(*self.window, match level {
+                WindowLevel::AlwaysOnBottom => -1,
+                WindowLevel::Normal => 0,
+                WindowLevel::AlwaysOnTop => 3,
+            });
+        }
```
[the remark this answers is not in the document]
wez: the logic that reports WindowState needs to report the correct value, which for macos is really just in the `did_resize` function, so you need to update it:

https://github.com/wez/wezterm/blob/6d58e5196bb3a433880643a05fe5d6f19e49572a/window/src/os/macos/window.rs#L2771-L2777

window/src/os/macos/window.rs
```diff
@@ -2772,6 +2775,20 @@ impl WindowView {
...
 
+            let window_level = inner.window.as_ref().map_or(WindowLevel::Normal, |w| unsafe { w.load().level().into() });
+
+            let level_state = match window_level {
+                WindowLevel::AlwaysOnBottom => WindowState::ALWAYS_ON_BOTTOM,
+                WindowLevel::AlwaysOnTop => WindowState::ALWAYS_ON_TOP,
+                WindowLevel::Normal => WindowState::default(),
+            };
+
+            let screen_state = match (is_full_screen, is_zoomed) {
```
rawnly, line 2786: not sure about the readability of this

window/src/os/macos/window.rs
```diff
@@ -2772,6 +2775,20 @@ impl WindowView {
...
+                (true, _) => WindowState::FULL_SCREEN,
+                (_, true) => WindowState::MAXIMIZED,
+                _ => WindowState::default(),
```
rawnly, line 2789: Not sure about the readability of this

config/src/window.rs
```diff
@@ -0,0 +1,26 @@
...
+    AlwaysOnBottom = -1,
+    Normal = 0,
+    AlwaysOnTop = 3,
+}
+
+impl Default for WindowLevel {
+    fn default() -> Self {
+        WindowLevel::Normal
+    }
+}
```
wez, line 15: You can reduce the verbosity of this by changing how you declare the enum, and having rust derive it for you.

Also, I just noticed the integer values here; let's remove them as they are also a macos implementation detail that shouldn't bleed into the config crate.

```suggestion
#[derive(Default, Debug, Clone, ToDynamic, PartialEq, Eq, FromDynamic)]
pub enum WindowLevel {
    AlwaysOnBottom,
    #[default]
    Normal,
    AlwaysOnTop,
}
```

docs/config/lua/keyassignment/SetWindowLevel.md
```diff
@@ -0,0 +1,28 @@
+# `SetWindowLevel`
+
+Set window level sepcified by the argument value. eg: `AlwaysOnTop` keeps the current window on top of other windows.
+
+Accepted values: `AlwaysOnBottom` | `Normal` | `AlwaysOnTop`
```
wez, line 5:
```suggestion
Accepted values: 

 * `"AlwaysOnBottom"`
 * `"Normal"` (this is the default)
 * `"AlwaysOnTop"`
```

docs/config/lua/keyassignment/SetWindowLevel.md
```diff
@@ -0,0 +1,28 @@
...
+        key = '0',
+        mods = 'CMD|SHIFT',
+        action = act.SetWindowLevel("Normal")
+    }, 
+    {
+        key = ']',
+        mods = 'CMD',
+        action = act.SetWindowLevel("AlwaysOnTop")
+    }, 
+}
```
wez, line 27: this feels a bit more self-contained and focused on the assignment that is being discussed in this doc page

```suggestion
config.keys = {
    {
        key = '[',
        mods = 'CMD',
        action = wezterm.action.SetWindowLevel("AlwaysOnBottom")
    }, 
    {
        key = '0',
        mods = 'CMD|SHIFT',
        action = wezterm.action.SetWindowLevel("Normal")
    }, 
    {
        key = ']',
        mods = 'CMD',
        action = wezterm.action.SetWindowLevel("AlwaysOnTop")
    }, 
}
```

docs/config/lua/keyassignment/ToggleAlwaysOnBottom.md
```diff
@@ -0,0 +1,17 @@
+# `ToggleAlwaysOnBottom`
+
+Toggles the window to remain behind all other windows.
+
+```lua
+local wezterm = require 'wezterm'
+
```
wez, line 7:
```suggestion
```

docs/config/lua/keyassignment/ToggleAlwaysOnTop.md
```diff
@@ -0,0 +1,18 @@
+# `ToggleAlwaysOnTop`
+
+Toggles the window between floating and non-floating states to stay on top of other windows.
+
+```lua
+local wezterm = require 'wezterm'
+local config = {}
+
```
wez, line 8:
```suggestion
```

docs/config/lua/keyassignment/SetWindowLevel.md
```diff
@@ -0,0 +1,28 @@
+# `SetWindowLevel`
+
```
wez, line 2:
```suggestion

{{since('nightly')}}
```

docs/config/lua/keyassignment/ToggleAlwaysOnBottom.md
```diff
@@ -0,0 +1,17 @@
+# `ToggleAlwaysOnBottom`
+
```
wez, line 2:
```suggestion

{{since('nightly')}}
```

docs/config/lua/keyassignment/ToggleAlwaysOnTop.md
```diff
@@ -0,0 +1,18 @@
+# `ToggleAlwaysOnTop`
+
```
wez, line 2:
```suggestion

{{since('nightly')}}
```

window/src/os/macos/window.rs
```diff
@@ -1069,6 +1096,8 @@ impl WindowInner {
     }
 }
 
+/// @see https://developer.apple.com/documentation/appkit/nswindow/level
+
```
wez, line 1100: This comment should probably either be moved elsewhere or deleted

window/src/os/macos/window.rs
```diff
@@ -1149,6 +1178,13 @@ impl WindowInner {
         }
     }
 
+    fn set_window_level(&mut self, level: WindowLevel) {
+        unsafe {
+            NSWindow::setLevel_(*self.window, window_level_to_nswindow_level(level));
+            WindowView::did_resize(&mut **self.view, sel!(windowDidResize:), nil);
```
wez, line 1184:
```suggestion
            // Dispatch a resize event with the updated window state
            WindowView::did_resize(&mut **self.view, sel!(windowDidResize:), nil);
```

docs/config/lua/keyassignment/SetWindowLevel.md
```diff
@@ -0,0 +1,30 @@
...
+{{since('nightly')}}
+Set window level sepcified by the argument value. eg: `AlwaysOnTop` keeps the current window on top of other windows.
+
```
wez, line 5: Please update the docs to indicate that this is currently only implemented on macOS, and that these new assignments have no effect on other operating systems.

rawnly: same applies to the others right?

docs/config/lua/keyassignment/SetWindowLevel.md
```diff
@@ -0,0 +1,35 @@
...
+        key = ']',
+        mods = 'CMD',
+        action = wezterm.action.SetWindowLevel("AlwaysOnTop")
+    }, 
+}
+```
+
+> Note: 
+> This functionality is currently only implemented on macOS. 
+> The assigned values for window level will have no effect on other operating systems.
```
wez, line 35: Let's use the mkdocs admonition syntax for these, and move it up so that it is just below the `since` macro call at the top.

```suggestion
!!! note 
    This functionality is currently only implemented on macOS. 
    The assigned values for window level will have no effect on other operating systems.
```

rawnly:
any feedback so far?

rawnly:
@wez ?

wez:
Thanks!

PR #4578, opened 2023-11-14
wezterm.terminfo: add terminator to Sync capability

gpanders:
I'm not sure if this is strictly necessary, but according to the
terminfo specification [1], a conditional string should be terminated
with `%;`.

Quoting the relevant section:

    %? expr %t thenpart %e elsepart %;
         This forms an if-then-else.  The %e elsepart is optional.
         Usually the %? expr part pushes a value onto the stack, and
         %t pops it from the stack, testing if it is nonzero (true).
         If it is zero (false), control passes to the %e (else) part.

[1]: https://man7.org/linux/man-pages/man5/terminfo.5.html


wez:
Seems reasonable, thanks!

PR #4595, opened 2023-11-19
Adds config option for command palette rows

exastone:
Command palette defaulted to showing rows based on terminal rows, this allows configuration of the number of rows displayed by the command palette.

### DIFF
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -148,9 +148,7 @@ pub struct Config {
     #[dynamic(default = "default_command_palette_font_size")]
     pub command_palette_font_size: f64,
 
-    #[dynamic(default = "default_command_palette_rows")]
-    pub command_palette_rows: usize,
-
+    pub command_palette_rows: Option<usize>,
     #[dynamic(default = "default_command_palette_fg_color")]
     pub command_palette_fg_color: RgbaColor,
 
@@ -1571,10 +1569,6 @@ fn default_char_select_bg_color() -> RgbaColor {
 
 fn default_command_palette_font_size() -> f64 {
     14.0
-}
-
-fn default_command_palette_rows() -> usize {
-    0
 }
 
 fn default_command_palette_fg_color() -> RgbaColor {
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -148,6 +148,9 @@ pub struct Config {
     #[dynamic(default = "default_command_palette_font_size")]
     pub command_palette_font_size: f64,
 
+    #[dynamic(default = "default_command_palette_rows")]
+    pub command_palette_rows: usize,
+
     #[dynamic(default = "default_command_palette_fg_color")]
     pub command_palette_fg_color: RgbaColor,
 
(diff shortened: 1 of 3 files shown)


exastone:
https://github.com/wez/wezterm/assets/61954919/9d4a03ac-230d-4902-ae90-9bc2469f49a9

shows:
- command_palette_rows = 12
- command_palette_rows not set (defaults to 0, uses previous calculation to determine number of rows to display)
- command_palette_rows = 60 when terminal is less than 60 rows (defaults to previous calculation to determine number of rows to display)

config/src/config.rs
```diff
@@ -148,6 +148,9 @@ pub struct Config {
     #[dynamic(default = "default_command_palette_font_size")]
     pub command_palette_font_size: f64,
 
+    #[dynamic(default = "default_command_palette_rows")]
+    pub command_palette_rows: usize,
+
```
wez, line 153: Let's make this optional:

```suggestion
    pub command_palette_rows: Option<usize>,
```

config/src/config.rs
```diff
@@ -1570,6 +1573,10 @@ fn default_command_palette_font_size() -> f64 {
     14.0
 }
 
+fn default_command_palette_rows() -> usize {
+    0
+}
+
```
wez, line 1579: then we can simplify and remove this

```suggestion
```

docs/config/lua/config/command_palette_rows.md
```diff
@@ -0,0 +1,11 @@
+---
+tags:
+  - appearance
+  - command_palette
+---
+# `command_palette_rows = 14`
+
```
wez, line 7: Please add version information so that we can tell when this is released:

```suggestion
# `command_palette_rows = 14`

{{since('nightly')}}
```

docs/config/lua/config/command_palette_rows.md
```diff
@@ -0,0 +1,11 @@
...
+
+Specifies the number of rows displayed by the command palette.
+[ActivateCommandPalette](../keyassignment/ActivateCommandPalette.md).
+
```
wez, line 11:
```suggestion
If unset or `nil`, a default value based on the terminal display will be used.
```

wezterm-gui/src/termwindow/palette.rs
```diff
@@ -664,9 +664,14 @@ impl Modal for CommandPalette {
...
 
-        let max_rows_on_screen = ((term_window.dimensions.pixel_height * 8 / 10)
+        let mut max_rows_on_screen = ((term_window.dimensions.pixel_height * 8 / 10)
             / metrics.cell_size.height as usize)
             - 2;
+        if 0 < term_window.config.command_palette_rows
+            && term_window.config.command_palette_rows < max_rows_on_screen
+        {
+            max_rows_on_screen = term_window.config.command_palette_rows;
+        }
```
wez, line 674: Can be simplified to:

```suggestion
        if let Some(size) = term_window.config.command_palette_rows {
            max_rows_on_screen = max_rows_on_screen.min(size);
        }
```

wez (changes requested):
Thanks for this!

FWIW, I can't play the video you attached in firefox (it says that it is corrupt), or in chrome (it just doesn't play).

exastone:
> Thanks for this!
> 
> FWIW, I can't play the video you attached in firefox (it says that it is corrupt), or in chrome (it just doesn't play).

strange, plays on safari, chrome, and GitHub app. on firefox, if i refresh it shows either unsupported format type of corrupt file (just a mp4)

wez:
Thanks!

PR #4603, opened 2023-11-20
get-deps: support installing docs dependencies on debian

simpkins:
The python3-pip and mkdocs packages are required to run the ci/build-docs.sh script.


wez:
lgtm, thanks!

Issue #4247, opened 2023-09-04
Problem with meta keys in vim


mg979:
### What Operating System(s) are you seeing this problem on?

Linux X11

### Which Wayland compositor or X11 Window manager(s) are you using?

Fluxbox

### WezTerm version

wezterm 20230712-072601-f4abf8fd

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

I have a strange problem while using Vim. I'm posting here because it doesn't
happen with other terminal emulators (I tried kitty and xfce4-terminal-emulator).

It doesn't happen in Neovim, only in Vim.

I set up several mappings with the meta key, initially they work, then some key
combinations make them stop working.

OS: MX Linux 21

### To Reproduce

Minimal reproducible case in my PC:

testrc.vim:

```vim
set rtp=$VIMRUNTIME

set <m-`>=`
nnoremap <m-`> :terminal<cr>
tnoremap <m-`> exit<cr>
```

    $ vim -Nu testrc.vim

After vim starts, press <key>Alt-`</key> a couple of times, it should open the
terminal, then exit.

Then press `Ctrl-d`, and the <key>Alt-`</key> mapping doesn't work anymore.

### Configuration

It happens also with

    wezterm -n

### Expected Behavior

My mappings with meta keys should keep working.

### Logs

This is extracted from the debug output (date has been removed), it's the sequence for

    Alt-`

I made a diff of the output produced when the mapping stopped working, and they were identical.

```
INFO   wezterm_gui::termwindow::keyevent     > key_event RawKeyEvent { key: Physical(LeftAlt), modifiers: NONE, leds: (empty), phys_code: Some(LeftAlt), raw_code: 64, repeat_count: 1, key_is_down: true, handled: Handled(false) }
INFO   wezterm_gui::termwindow::keyevent     > key_event KeyEvent { key: LeftAlt, modifiers: NONE, leds: (empty), repeat_count: 1, key_is_down: true, raw: Some(RawKeyEvent { key: Physical(LeftAlt), modifiers: NONE, leds: (empty), phys_code: Some(LeftAlt), raw_code: 64, repeat_count: 1, key_is_down: true, handled: Handled(false) }) }
INFO   wezterm_gui::termwindow::keyevent     > send to pane DOWN key=LeftAlt mods=NONE
[...]

Issue #4589, opened 2023-11-16
enable_kitty_keyboard doesn't pass through CMD-? keys


jackielii:
### What Operating System(s) are you seeing this problem on?

macOS

### Which Wayland compositor or X11 Window manager(s) are you using?

Macos

### WezTerm version

wezterm 20230712-072601-f4abf8fd

Also tried nightly. Same behaviour

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

No, and I'll explain why below

### Describe the bug

It seems none of the `<CMD>+?` keys are working inside wezterm. I'm trying to map `<D-s>` in neovim to `:update`. `<D-` is the CMD/SUPER modifier: https://neovim.io/doc/user/intro.html#%3CD-

### To Reproduce

run cat and then press `<CMD>+s` 

on kitty it produce: `^[[115;9u`
on wezterm it produce `s`

### Configuration

wezterm's config:
```
return { enable_kitty_keyboard = true }
```

### Expected Behavior

`<CMD>+s` produce the right keycode

### Logs

The log seems empty in this case. 

The Debug Overlay:
wezterm version: 20231112-171650-721fbdf5 aarch64-apple-darwin
Window Environment: macOS 14.0 (23A344)
WebGPU: name=Apple M1 Pro, device_type=IntegratedGpu, backend=Metal, vendor=0, device=0
Enter lua statements or expressions and hit Enter.
Press ESC or CTRL-D to exit

jackielii:
I think this is fixed in my testing

Issue #4436, opened 2023-10-14
Super + Q keybinding problem


og900aero:
### What Operating System(s) are you seeing this problem on?

Linux X11

### Which Wayland compositor or X11 Window manager(s) are you using?

Dwm

### WezTerm version

20231004-113117-11dec45f

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

When I run wezterm, and run lf file manager, then press Super + Q (dwm binding -> kill client), then window is closed by dwm, but at the same time, in the background, wezterm also issues a ctrl + d "command" for some reason, which also covers another command for me in lf file manager, and this causes quite a problem.

### To Reproduce

1. Run wezterm
2. Run lf file manager
3. Then close window bye Super + Q (in dwm)
4. Then window is closed, and the background wezterm run ctrl + d keymap too

### Configuration

-- Pull in the wezterm API
local wezterm = require 'wezterm'

-- This table will hold the configuration.
local config = {}

-- In newer versions of wezterm, use the config_builder which will
-- help provide clearer error messages
if wezterm.config_builder then
  config = wezterm.config_builder()
end

-- This is where you actually apply your config choices
config.enable_wayland = false
config.default_cwd = "/home/Data/"
config.check_for_updates = true
config.show_update_window = true
config.check_for_updates_interval_seconds = 604800 -- weekly
config.scrollback_lines = 500
config.color_scheme = 'AdventureTime'
-- config.enable_tab_bar = false
config.hide_tab_bar_if_only_one_tab = true
-- config.font = wezterm.font('JetBrains Mono', { weight = 'Bold', italic = true })
config.font = wezterm.font('IosevkaTerm Nerd Font Mono', { weight = 'Bold', italic = true })
config.font_size = 15.0
config.line_height = 1.0
config.window_close_confirmation = 'NeverPrompt'
config.disable_default_key_bindings = true

wezterm.action.SpawnCommandInNewWindow {
  label = 'List all the files!',
  args = { 'ls', '-la' },
  cwd = '/home/Data',
}

config.keys = {
  {
    key = 'P',
    mods = 'CTRL',
    action = wezterm.action.ActivateCommandPalette,
  },
}

-- and finally, return the configuration to wezterm
return config

### Expected Behavior

If I use the parameter config.disable_default_key_bindings = true, wezterm should do nothing for Super + Q keybinding.

### Logs

[...]

PR #4605, opened 2023-11-21
fix: Correct modifier key handling in kitty protocol

gabyx:
**What does this PR do:**
Fix a bug in handling modifiers correctly for the kitty protocol.

Fixes: #4247, #4436, #4589

**How was it done?:**
There is a problem in `encode_kitty` function which I tried to debug and hopefully fixed it.
I guess the boolean `use_legacy` which determines to [encode legacy text keys](https://sw.kovidgoyal.net/kitty/keyboard-protocol/#legacy-text-keys) is wrong. 

- In the original any modifiers as `SUPER` (in nvim which sets `KittyKeyboardFlags` to `1`)
  went through this legacy `if` which is wrong. 
- The legacy text keys handling also forgot to add the `Shift` case which I added. [Point 3 here](https://sw.kovidgoyal.net/kitty/keyboard-protocol/#legacy-text-keys).

- I added `Meta` and `Hyper` which I try to add in a further PR)

@wez : As the whole protocol handling is quite difficult and I could not match quite all switched to the protocol.

### DIFF
--- a/wezterm-input-types/src/lib.rs
+++ b/wezterm-input-types/src/lib.rs
@@ -495,6 +495,8 @@ bitflags! {
         const LEFT_SHIFT = 1<<10;
         const RIGHT_SHIFT = 1<<11;
         const ENHANCED_KEY = 1<<12;
+        const HYPER = 1<<13;
+        const META = 1<<14;
     }
 }
 
@@ -516,6 +518,10 @@ impl TryFrom<String> for Modifiers {
                 mods |= Modifiers::CTRL;
             } else if ele == "SUPER" || ele == "CMD" || ele == "WIN" {
                 mods |= Modifiers::SUPER;
+            } else if ele == "META" {
+                mods |= Modifiers::META;
+            } else if ele == "HYPER" {
+                mods |= Modifiers::HYPER;
             } else if ele == "LEADER" {
                 mods |= Modifiers::LEADER;
             } else if ele == "NONE" || ele == "" {
@@ -1674,6 +1680,12 @@ impl KeyEvent {
         if raw_modifiers.contains(Modifiers::SUPER) {
             modifiers |= 8;
         }
+        if raw_modifiers.contains(Modifiers::HYPER) {
+            modifiers |= 16;
+        }
+        if raw_modifiers.contains(Modifiers::META) {
+            modifiers |= 32;
+        }
         if self.leds.contains(KeyboardLedStatus::CAPS_LOCK) {
             modifiers |= 64;
         }
@@ -1808,31 +1820,32 @@ impl KeyEvent {
                     *shifted_key
                 };
 
-                if !flags.contains(KittyKeyboardFlags::REPORT_ALTERNATE_KEYS)
+                let use_legacy = is_legacy_key
                     && event_type.is_empty()
-                    && is_legacy_key
-                    && !(flags.contains(KittyKeyboardFlags::DISAMBIGUATE_ESCAPE_CODES)
-                        && (self.modifiers.contains(Modifiers::CTRL)
-                            || self.modifiers.contains(Modifiers::ALT)))
-                {
-                    use_legacy = true;
-                }
+                    && !flags.contains(KittyKeyboardFlags::REPORT_ALTERNATE_KEYS)
+                    && !flags.contains(KittyKeyboardFlags::DISAMBIGUATE_ESCAPE_CODES);
 
                 if use_legacy {
                     // Legacy text key
+                    // https://sw.kovidgoyal.net/kitty/keyboard-protocol/#legacy-text-keys
+
                     let mut output = String::new();
                     if self.modifiers.contains(Modifiers::ALT) {
                         output.push('\x1b');
                     }
+
                     if self.modifiers.contains(Modifiers::CTRL) {
                         csi_u_encode(
                             &mut output,
-                            shifted_key.to_ascii_uppercase(),
+                            shifted_key.to_ascii_uppercase(), // Is this to uppercase intended?
                             self.modifiers,
                         );
+                    } else if self.modifiers.contains(Modifiers::SHIFT) {
+                        output.push(shifted_key.to_ascii_uppercase());
                     } else {
                         output.push(shifted_key);
                     }
+
                     return output;
                 }
 


wezterm-input-types/src/lib.rs
```diff
@@ -1808,31 +1820,32 @@ impl KeyEvent {
...
                     let mut output = String::new();
                     if self.modifiers.contains(Modifiers::ALT) {
                         output.push('\x1b');
                     }
+
                     if self.modifiers.contains(Modifiers::CTRL) {
                         csi_u_encode(
                             &mut output,
-                            shifted_key.to_ascii_uppercase(),
+                            shifted_key.to_ascii_uppercase(), // Is this to uppercase intended?
```
gabyx, line 1840: @wez : Is this intended to turn it into uppercase here? Or is this a typo/mistake where one wanted to tackle point 3 in the protocol : https://sw.kovidgoyal.net/kitty/keyboard-protocol/#legacy-text-keys

wezterm-input-types/src/lib.rs
```diff
@@ -516,6 +518,10 @@ impl TryFrom<String> for Modifiers {
                 mods |= Modifiers::CTRL;
             } else if ele == "SUPER" || ele == "CMD" || ele == "WIN" {
                 mods |= Modifiers::SUPER;
+            } else if ele == "META" {
```
wez, line 521: Defining `META` as its own thing needs some care and attention throughout the wezterm crates as well as the termwiz crate, as we've always treated it as an alias of `ALT`, which you can see happening on line 515 above.

This particular line in this PR has no effect because it is evaluated after that line.

Breaking it out into its own thing needs to consider:
* What happens for non-kitty encodings when this modifier is used? Is it simply ignored?
* How does it get set? This PR doesn't seem to add any OS-specific logic to map the incoming key press from the windowing layer to actually set either the META or HYPER flags
* What happens to folks that have defined key assignments using META as the modifier? Do those effectively stop working when they upgrade? Is that ok? (probably?)

wezterm-input-types/src/lib.rs
```diff
@@ -1800,30 +1812,31 @@ impl KeyEvent {
...
                     && is_legacy_key
                     && !(flags.contains(KittyKeyboardFlags::DISAMBIGUATE_ESCAPE_CODES)
                         && (self.modifiers.contains(Modifiers::CTRL)
                             || self.modifiers.contains(Modifiers::ALT)))
-                {
-                    use_legacy = true;
-                }
+                    && !self.modifiers.contains(Modifiers::SUPER)
+                    && !self.modifiers.contains(Modifiers::META)
+                    && !self.modifiers.contains(Modifiers::HYPER);
```
wez, line 1830: This is probably a bit more elegant if rephrased as something like:

```rust
!self.modifiers.intersects(Modifiers::SUPER|Modifiers::META|Modifiers::HYPER)
```

wezterm-input-types/src/lib.rs
```diff
@@ -3054,4 +3068,54 @@ mod test {
...
+                modifiers: Modifiers::META,
+                leds: KeyboardLedStatus::empty(),
+                repeat_count: 1,
+                key_is_down: true,
+                raw: None,
+                #[cfg(windows)]
+                win32_uni_char: None,
+            }
+            .encode_kitty(flags),
+            "q".to_string()
```
wez, line 3118: I think these test expectations are incorrect; these tests are failing

wez (commented):
Thanks for this! There's some more stuff to consider than is just in this PR.

Something that is absent is logic to populate these new key flags.
The only platform that supports those is unix.  IIRC, this is the function that is responsible:

https://github.com/wez/wezterm/blob/fde926722fb6ef05fb3be78624aff33095a283d1/window/src/os/x11/keyboard.rs#L383

wezterm-input-types/src/lib.rs
gabyx: Ah jeah, I was looking for that thing, something like `modifiers.any` but didnt exist and was too dump to construct this. Much better!

wezterm-input-types/src/lib.rs
gabyx: Jup still need to update these. I will reiterate till the test work, no worry.

gabyx:
@wez : Thanks for the reply, I  am not yet done fixing my tests, did not yet adapt the asserts (will do!).

- Thanks for the hint to populate the flags. So that means Meta and Hyper will not work on MacOS?.
- Is it possible to leave the modifiers in this PR and come up with another one to gracefullt add these under Unix at least?
  
 I am still not quite sure, if the kitty thing is all correct, but at least it seams so for the legacy_key part.
 I truely appreciate what you did here, did you look at the kitty's source code for all that `encode_kitty` stuff?
 Its such a convoluted complicated spec haha.

wez:
Only X11 (and Wayland) have META and HYPER modifiers as a concept. Neither Windows nor macOS have equivalents.

I think it is reasonable to define these new modifier flags. I don't think we need to go to extreme lengths to handle META/HYPER elsewhere or for backwards compatibility, but we do need to consider how to handle them for the non-kitty keyboard scenario, which may just be "pretend they don't exist".  Will need to compare and contrast behavior with other unix terminal emulators when using those keys.

I did not read the kitty source for this; only the spec docs, which I agree are not especially clear in a couple of places.

gabyx:
@wez: Ok, thanks for the information about Meta and Hyper. This is very useful. So to not clutter up this PR as a bugfix. lets go with the pragmatic approach and first merge this fix. Like it is presented now.
Is that good enough? Or anything else needed?

- So basically this PR fixes `SUPER` modifier key handling.

wez:
I think this PR is mergeable as-is.

Re: other modifiers and the future, are you proposing the addition of `MOD1..MOD5` to the `Modifiers` type in `wezterm-input-types`?  I think that those could be added (even though eg: `ALT` is already implicitly handled under a different name) in a similar way to the `LEFT_XXX` and `RIGHT_XXX` flags; those are populated if the system can make the distinction, but are otherwise non-essential and can then be inspected by eg: the kitty protocol encoding logic.

A potential wrinkle with that is: the flags type is 16 bits wide and there aren't enough available bits to represent 5 additional modifiers, so perhaps we should just add the mods that are not currently represented.  I'd like to avoid making `Modifiers` 32 bits wide if we can avoid it, but if it doesn't change the size of aggregate types (like the size of the entries in various key tables, where the cardinality * the size of the struct represents an increase in RAM usage) then there's no real argument against it.

gabyx:
@wez: Jeah it might be the only sane way to add Mod1-Mod5 to the table an check against where the other xkb types are handled. This addition will not hurt wezterm. But we still dont know which keys Mod1-Mod5 will be what “thing” for the kitty protocol. (to what Mod? did Hyper map?)

I think, Kitty is doing this by looking at the code: its checking the virtual modifier table with x11 calls and xkb and this is somewhat involved and as far as I could grasp was that this will correlate Super_L and other modifiers like Hyper_L/R to the 9 different modifiers xkb knows (Shift, Ctrl, Alt, Lock, Mod1-5) (`xmodmap -pm`). When we know this table (kitty does that at startup) 
we know for lets say Hyper_L maps to Mod3 (the user chose so) , if Mod3 was pressed, then we can encode a “Hyper” modifier in the kitty protocol.

So basically function: https://github.com/wez/wezterm/blob/fde926722fb6ef05fb3be78624aff33095a283d1/window/src/os/x11/keyboard.rs#L383

is somewhat only partially correct, it checks for Mod4 and relating this to Super. But the user might have changed this, it maybe has chosen that a certain key activates Super_L an mapped this to Mod3.

I hope my explanations were correct and understandable.
I think without implementing the exact same code as in kitty (will link to it, later)
wezterm preassumes some standard mapping from Alt, Shift, Ctrl, Super to Mod1-5.

gabyx:
The other thing is I am unaware of if Wayland has some otherstuff, it still uses xkb afaik. And the virtual modifier table is still a thing. Maybe we could really port the C code from kitty into rust: It basically this function: https://github.com/kovidgoyal/kitty/blob/72e14053f55b0165a47d6cd743a0c1de1944149a/glfw/xkb_glfw.c#L376

wez:
Thanks!

Issue #4626, opened 2023-11-30
fix: Implement modifier detection on wayland/x11


gabyx:
**Is your feature request related to a problem? 
The problem described in #4605 mentions the problem that the in Linux the modifiers in xkb are not correctly assigned to the 8 known modifiers known to xkb (e.g. `shift, ctrl, lock, mod1, mod2, .... , mod3`), instead wezterm uses a standard mapping which maps the standard setup in most keyboard layouts (e.g. US).

```
xmodmap -pm
```
reports

```shell
shift       Shift_L (0x32),  Shift_R (0x3e)
lock
control     Control_L (0x25),  Control_L (0x42),  Control_R (0x69)
mod1        Alt_L (0x40),  Alt_R (0x6c),  Alt_L (0xcc),  Meta_L (0xcd)
mod2        Num_Lock (0x4d)
mod3        ISO_Level5_Shift (0xcb)
mod4        Super_L (0x85),  Super_R (0x86),  Super_L (0xce),  Hyper_L (0xcf)
mod5        ISO_Level3_Shift (0x5c)
```

**Describe the solution you'd like**
To fully suppport the kitty protocol, its necessary to know which virtual modifier above is associated to which of the 8 modifiers, because xkb only reports these as a modifier mask. This then enables to correctly encode additional modifiers in the kitty protocol, such as `Meta`, ` Hyper`  etc.

Kitty takes the following approach for X11 and Wayland:
-  **X11**: ([a solutiuon over virtual modifier table](glfw_xkb_update_masks))
- **Wayland**: ([workaround with a loop over the keys](https://github.com/kovidgoyal/kitty/blob/master/glfw/xkb_glfw.c#L523), due to the inability to query the virtual modifier table in Wayland.

Its best advised to take the same road as Kitty above and implement the same behavior.
As you see above we can not really distinguish between Super left/right once we know that `mod4` has been pressed. This is also not so bad, since the [kitty protocol does not distinguis either](https://sw.kovidgoyal.net/kitty/keyboard-protocol).
The improvements then also enables to add these new modifiers without left/right (?) to the wezterm user command handling. 

**Describe alternatives you've considered**
[...]

PR #4629, opened 2023-11-30
feat: Identify modifiers in Wayland/X11

gabyx:
Fixes: #4626

### DIFF
--- a/window/src/os/x11/modifiers.rs
+++ b/window/src/os/x11/modifiers.rs
@@ -303,7 +303,9 @@ pub fn init_modifier_table_wayland(keymap: &xkb::Keymap) -> ModifierMap {
         let mut shifted: u32 = 1;
         let mut used_bits: u32 = 0;
 
-        macro_rules! assign {
+        // Algorithm can detect the same index for multiple modifiers
+        // but we only want one assigned to each of our modifiers.
+        macro_rules! try_assign_unique_index {
             ($mod:ident, $i:ident) => {
                 if mod_map.$mod.idx == xkb::MOD_INVALID
                     && (used_bits & shifted == 0)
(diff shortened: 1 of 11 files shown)


window/src/os/x11/modifiers.rs
```diff
@@ -0,0 +1,345 @@
...
+                    }
+                };
+            }
+
+            // We can handle exactly one keysym with exactly one bit set in the implementation
+            // below; with a lot more gymnastics, we could set up an 8x8 linear system and solve
+            // for each modifier in case there are some modifiers that are only present in
+            // combination with others, but it is not worth the effort.
+            if keysyms.len() == 1 && mods.is_power_of_two() {
+                S2!(xkb::keysyms::KEY_Shift_L, xkb::KEY_Shift_R, shift);
```
wez, line 180: why does just this one have `xkb::keysyms` as the namespace and the others are `xkb::`?

wez (commented):
At a high level, I think it is reasonable to do this kind of probe to reason about the keymap. We do something similar to detect dead keys in the Windows implementation.

I can't follow the code well enough to really understand it, and I don't understand why Wayland needs to be treated differently from X11 when the data model for both is the same keyboard map and state.

Most of my comments are suggestions for making it easier to follow and may not be the most constructive wrt. the higher level goal.

gabyx:
@wez: Thanks! for helping out on this: input really appreciated, also my compile switch is still wrong in the code (the code needs to switch in the ctor of the keyboard to select the appropriate impl for the modifier detection)
as far I could grasp, the wayland key press probing I understood, but the X11 analogous and no-workaround code is obviously not possible in Wayland (the reason I need to search again for, I guess its due to the calls which are not all in the xkb:: namespace and more in the x11 namespace. Once I implemented the X11 approach in the next days I probably understand more, why this is maybe not callable on wayland, I am on wayland and can test it.

I moved the code to modifiers.ra for the reason that its not that easy to understand, and that we can simply drop it or exchange it, or use the initi_modifiers_fallback which is the current thing wezterm does.
Also I try to not overdocument it but add some more key comments what we try to achieve. I thought it might be a good idea to keep as close to the kitty code as possible such that we can still look at it in the future and reason about it (esoecially the wayland probing)

gabyx:
@wez: the PR is far from finished, once X11 is implemented it needs documentation for the user what to consider about modifiers on wayland etc

gabyx:
@wez: The implementation for X11 is finished, last commit.. I think this code is better to understand and also gives some reason why it does not work on Wayland, xlib calls I think.

I added `scopeguard` as a new dependency as I was not sure what you use for that purpose. I needed a proper `defer` statement. Not sure maybe there is already the same in your library then I can remove it or maybe do it differently. I didnt fine something related to `defer` or `scope/guard`...

I have not answered the Windows and MacOs question about the Led stuff i removed. I still have to investigate this.

Implementation wise I think everything is now there similar to what kitty does, which is probably a good thing.
It would be nice if somebody could test the startup on X11. I might be able to pull it off by just switching on NixOS.

Thanks for the review.

I think I still need to add some documentation to highlight what is done on startup for X11 and Wayland such that users understand. Do you have a suggestion where I could add this?

wez:
It won't let me; I think you'll just have make a new PR from that branch

image

PR #4635, opened 2023-12-02
Windows rs

strottos:


window/src/os/windows/wgl.rs
```diff
@@ -277,10 +295,11 @@ impl GlState {
         }
 
         let res = unsafe { SetPixelFormat(hdc, format_id, &pfd) };
-        if res == 0 {
+        if res.is_err() {
```
wez, line 298: For something like this, where we are now getting an error as a result, I would prefer something like:

```rust
if let Err(err) = res {
   anyhow::bail!("SetPixelFormat function failed: {err:#}");
}
```

wez: or perhaps even just:

```rust
unsafe { SetPixelFormat(hdc, format_id, &pfd) }.context("SetPixelFormat")?;
```

(and can we remove the `unsafe` from here?)

window/src/os/windows/event.rs
```diff
@@ -11,29 +9,26 @@ unsafe impl Sync for EventHandle {}
...
-            return Err(IoError::last_os_error().into());
-        }
+        let handle = unsafe { CreateEventW(None, true, false, PCWSTR::null()) }?;
         Ok(Self(handle))
     }
 
     pub fn set_event(&self) {
         unsafe {
-            SetEvent(self.0);
+            SetEvent(self.0).unwrap();
```
wez, line 25: I try really hard to avoid bare `.unwrap()`.  Suggested alternatives:

* If the error is "totally impossible" and "can never happen", use `.expect("reason this can never happen")` so that the code documents why it is impossible
* If the error is possible at runtime and we just don't care, consider using `.ok()` instead of `.unwrap()` to discard the `must_use` attribute
* If the error is possible at runtime and something we might need to handle, change the signature of the calling function to some kind of `Result` and allow the caller to decide whether and how to handle it

window/src/os/windows/keycodes.rs
```diff
@@ -101,110 +100,110 @@ fn build_map() -> HashMap<WPARAM, PhysKeyCode> {
         (0x34, PhysKeyCode::K4),
         (0x36, PhysKeyCode::K6),
         (0x35, PhysKeyCode::K5),
-        (VK_OEM_PLUS, PhysKeyCode::Equal),
+        (VK_OEM_PLUS.0, PhysKeyCode::Equal),
```
wez, line 103: Not sure what the answer is, but the `.0` here is ugly and it would be great to eliminate it if possible.

window/src/os/windows/wgl.rs
```diff
@@ -173,7 +191,7 @@ impl GlState {
             }
 
             let extensions = if ext.GetExtensionsStringARB.is_loaded() {
-                unsafe { cstr(ext.GetExtensionsStringARB(hdc as *const _)) }
+                unsafe { cstr(ext.GetExtensionsStringARB(std::mem::transmute(hdc))) }
```
wez, line 194: in general, it is best to avoid `transmute` and take great care with it, because it will let you cast pretty much anything to anything without complaint, and is a great way to mask serious problems.  So please find a way to make this work without an inline `transmute` in here!

wez: (and same deal for other uses of `transmute` throughout!)

window/src/os/windows/window.rs
```diff
@@ -21,62 +21,87 @@ use std::any::Any;
...
+    GetFocus, GetKeyState, GetKeyboardLayout, GetKeyboardState, MapVirtualKeyW, ReleaseCapture,
+    SendInput, SetCapture, SetKeyboardState, ToUnicode, TrackMouseEvent, INPUT, INPUT_KEYBOARD,
+    KEYBDINPUT, KEYEVENTF_EXTENDEDKEY, KEYEVENTF_KEYUP, MAPVK_VK_TO_VSC, TME_LEAVE, TME_NONCLIENT,
+    TRACKMOUSEEVENT, VK_CAPITAL, VK_CONTROL, VK_DECIMAL, VK_LCONTROL, VK_LMENU, VK_LSHIFT, VK_LWIN,
+    VK_MENU, VK_NUMLOCK, VK_PACKET, VK_PROCESSKEY, VK_RCONTROL, VK_RMENU, VK_RSHIFT, VK_RWIN,
+    VK_SHIFT,
+};
+use windows::Win32::UI::Shell::{DragAcceptFiles, DragFinish, DragQueryFileW, HDROP};
+use windows::Win32::UI::TextServices::HKL;
+use windows::Win32::UI::WindowsAndMessaging::{
```
wez, line 70: This is big enough that it is starting to dominate the file, so I think that a wildcard import would be totally fine.

```suggestion
use windows::Win32::UI::WindowsAndMessaging::*;
```

bew:
I'll close this PR as it's very stale and has many conflicts with current `main`.

To follow progress on the migration from winapi to windows-rs, see #7797 where we'd like to incrementally swap out the libs instead of going with a 'oneshot' migration.

xtqqczze:
Switching to `windows-sys` for `filedescriptor` was a straightforward mechanical change, simpler than the changes suggested here. This makes me suspect that the `windows-sys` crate has changed since then.

bew:
> Switching to `windows-sys` for `filedescriptor` was a straightforward mechanical change, simpler than the changes suggested here. This makes me suspect that the `windows-sys` crate has changed since then.

yeah that's what I though as well when looking through this PR again after looking at yours 🤔

xtqqczze:
Oh, I see: these changes use the `windows` crate, not `windows-sys`. They aren't the same; for example, compare the signatures of `WSASocketW` in [`windows-sys`](https://docs.rs/windows-sys/latest/windows_sys/Win32/Networking/WinSock/fn.WSASocketW.html) and [`windows`](https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/Networking/WinSock/fn.WSASocketW.html).

With `windows-sys`, you essentially get the raw Windows ABI, whereas `windows` provides the same underlying API projected into a more strongly typed Rust interface.

For migrating from `winapi` (https://github.com/wezterm/wezterm/issues/7797), it seems like `windows-sys` seems like the better fit and might even make it possible to do in one shot. The higher-level `windows` crate could then be considered as part of a larger refactoring.

bew:
Let's keep discussion about the migration on #7797 🙏

PR #4638, opened 2023-12-03
Add `wezterm.to_string` function

Danielkonge:
Here is a draft of a `wezterm.to_string` function. I think it makes sense to keep it so that it only accepts on `Value`, but I can change it so it can accept multiple values if you prefer?

Also, for now I just use the `ValuePrinter` directly, so we don't need to be as careful with strings as in `printerhelper`, but I can use `printerhelper` if you prefer.

### DIFF
--- a/docs/config/lua/wezterm/to_string.md
+++ b/docs/config/lua/wezterm/to_string.md
@@ -1,0 +1,24 @@
+---
+title: wezterm.to_string
+tags:
+ - utility
+---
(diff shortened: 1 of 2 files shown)


Danielkonge:
I have committed the suggested change now. Did you want to add more clarification? E.g. show some uses on other kinds of values?

wez:
This is good, no need to add more, thanks!

PR #4653, opened 2023-12-08
feat: Identify modifiers on Wayland/X11

gabyx:
Fixes: #4626

- On Wayland the workaround goes by probing the keyboard for all keys and rebuilding the missing API which is not available as in X11.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6830,6 +6830,7 @@ dependencies = [
  "promise",
  "raw-window-handle",
  "resize",
+ "scopeguard",
  "serde",
  "shared_library",
  "shlex",
--- a/termwiz/src/escape/csi.rs
+++ b/termwiz/src/escape/csi.rs
@@ -46,7 +46,7 @@ fn csi_size() {
     assert_eq!(std::mem::size_of::<Cursor>(), 12);
     assert_eq!(std::mem::size_of::<Edit>(), 8);
     assert_eq!(std::mem::size_of::<Mode>(), 24);
-    assert_eq!(std::mem::size_of::<MouseReport>(), 8);
+    assert_eq!(std::mem::size_of::<MouseReport>(), 12);
     assert_eq!(std::mem::size_of::<Window>(), 40);
     assert_eq!(std::mem::size_of::<Keyboard>(), 8);
     assert_eq!(std::mem::size_of::<CSI>(), 32);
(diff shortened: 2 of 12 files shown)


gabyx:
@wez: The implementation for X11 is finished, last commit.. I think this code is better to understand and also gives some reason why it does not work on Wayland, xlib calls I think.

I added `scopeguard` as a new dependency as I was not sure what you use for that purpose. I needed a proper `defer` statement. Not sure maybe there is already the same in your library then I can remove it or maybe do it differently. I didnt find something related to `defer` or `scope/guard`...

I have not answered the Windows and MacOs question about the Led stuff i removed. I still have to investigate this.

Implementation wise I think everything is now there similar to what kitty does, which is probably a good thing.
It would be nice if somebody could test the startup on X11. I might be able to pull it off by just switching on NixOS.

Thanks for the review.

I think I still need to add some documentation to highlight what is done on startup for X11 and Wayland such that users understand. Do you have a suggestion where I could add this?

gabyx:
@wez First have to check the tests again.

wez:
I got cc'd on this thread, which seems relevant to your efforts! I haven't had time to do much more than a quick skim:
* https://github.com/kovidgoyal/kitty/issues/6913

gabyx:
> I got cc'd on this thread, which seems relevant to your efforts! I haven't had time to do much more than a quick skim:
> 
> * [Modifier key events have their modifier bits set differently on Linux vs OSX kovidgoyal/kitty#6913](https://github.com/kovidgoyal/kitty/issues/6913)

Thanks I have seen the discussion, I am aware what it is about, and dont think its too important for now for wezterm.
It has nothing to do so far with the solution in this PR. However one should check again on how kitty implements the solution for providing modifiers on key-presses of modifiers keys. I dont like too much the solution proposed. As it seems its quite a debate about provifing mdofiers of modifier-keys on press or release and for wayland only release works well because the modifier event comes after the keypress... But to not get confused this PR only procedurally detects modifiers on startup and saves their index to query them on actual keypresses during run.

wez:
FWIW, I recently made some changes around modifier handling for X11 that likely collide with this PR

gabyx:
@wez: I will add a new PR with the rebased stuff. 
Closing this. It will contain the past (rebased: cleaned) stuff for introspection with the new PR.

PR #4712, opened 2023-12-22
Correct missing parens type-o in parse.md

andresrinivasan:


wez:
Thanks for diving in, however, this is not a typo; the parentheses are not required for a simple single argument being passed to a function call.  All of the examples in the docs are processed by https://github.com/azzamsa/gelatyx which runs https://github.com/JohnnyMorganz/StyLua over the lua, so that the examples are both syntactically valid and consistent throughout the docs.

PR #4727, opened 2023-12-28
Improve integrated titlebar buttons on macos

yuraiz:
https://github.com/wez/wezterm/assets/63008755/5caf97e6-640f-4e10-bac0-e7dc4d1128f5

I found a way to move traffic lights, it seems to work as I want to, but needs some improvements

TODO:
- [x] Correctly refresh buttons every time it's needed
- [ ] Handle not only the fancy variant
- [ ] Check how it works with other preferences
- [x] Handle RTL
- [X] Correctly handle the margin for the window buttons in the tab bar code

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1276,10 +1276,11 @@ checksum = "5729f5117e208430e437df2f4843f5e5952997175992d1414f94c57d61e270b4"
 
 [[package]]
 name = "deranged"
-version = "0.3.8"
+version = "0.3.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f2696e8a945f658fd14dc3b87242e6b80cd0f36ff04ea560fa39082368847946"
+checksum = "b42b6fa04a440b495c8b04d0e71b707c585f83cb9cb28cf8cd0d976c315e31b4"
 dependencies = [
+ "powerfmt",
  "serde",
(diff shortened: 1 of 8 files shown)


micsama:
its good!

micsama:
will it be merge？ its looks cool

yuraiz:
> will it be merge?

It isn't finished yet and not even ready for review

yuraiz:
Now I don't see any bugs, but probably not every case is handled or some of them handled incorrectly.
I think PR is mostly ready. I'll check how I want to use it with fancy_title_bar = false, maybe it's better to just hide the native buttons in that case.

RTL is handled:

Screenshot 2024-10-01 at 23 55 25

For testing I enabled rtl using the following code inside of the app:
```rust
// somewhere in window/src/os/macos/window.rs
 unsafe {
    let user_defaults = <id as cocoa::foundation::NSUserDefaults>::standardUserDefaults();

    cocoa::foundation::NSUserDefaults::setBool_forKey_(
        user_defaults,
        YES,
        *nsstring("AppleTextDirection"),
    );
    cocoa::foundation::NSUserDefaults::setBool_forKey_(
        user_defaults,
        YES,
        *nsstring("NSForceRightToLeftWritingDirection"),
    );
}
```
(applies after restart, call with NO to disable).

PR #4731, opened 2023-12-29
Feat: Send pane content to editor

desdic:
I wanted to create a feature where I could send the pane content to a file and then edit it with my favorite editor (I couldn't find a way to get the entire content btw.)

Its a bit crude since its my first rust modification and I don't have access to windows or MacOS to test on.


wez:
Thanks for diving in!

Have you seen the example here in the docs?

https://wezfurlong.org/wezterm/config/lua/wezterm/on.html#custom-events

You can get all of the lines from the scrollback rather than just the viewport, and make it open in a tab instead of a window, by modifying it a little bit. Something like this should do it, but I haven't tested this:

```lua
local io = require 'io'
local os = require 'os'
local act = wezterm.action

wezterm.on('trigger-vim-with-scrollback', function(window, pane)
  local scrollback_text = pane:get_lines_as_text(pane:get_dimensions().scrollback_rows)

  -- Create a temporary file to pass to vim
  local name = os.tmpname()
  local f = io.open(name, 'w+')
  f:write(scrollback_text)
  f:flush()
  f:close()

  -- Open a new tab running vim and tell it to open the file
  window:perform_action(
    act.SpawnCommandInNewTab {
      args = { 'vim', name },
    },
    pane
  )

  -- Wait "enough" time for vim to read the file before we remove it.
  -- The window creation and process spawn are asynchronous wrt. running
  -- this script and are not awaitable, so we just pick a number.
  --
  -- Note: We don't strictly need to remove this file, but it is nice
  -- to avoid cluttering up the temporary directory.
  wezterm.sleep_ms(1000)
  os.remove(name)
end)

return {
  keys = {
    {
      key = 'E',
      mods = 'CTRL',
      -- See also https://wezfurlong.org/wezterm/config/lua/wezterm/action_callback.html
      action = act.EmitEvent 'trigger-vim-with-scrollback',
    },
  },
}
```

I'm not enthusiastic about putting the equivalent of this directly in wezterm for a couple of reasons:

* The choice of editor is subjective and to make this feature work for everyone, a lot of configuration around the editor selection and formatting the command will be needed
* Some folks will want panes, some tabs and some a window. There's a lot of configuration possibilities that are difficult to express without a lot of additional overhead in pre-defining various types to support those choices
* The editor invocation doesn't actually do anything that wezterm cares about; there's no actionable-to-wezterm outcome from spawning the process, so there's no state to capture to justify adding built-in integration

Because of the above, and because it is already possible to configure and customize the behavior very deeply, I don't want to take this PR.

I appreciate your efforts!

desdic:
Hi and thank you .. I totally missed this custom event so yes its make totally sense not to add my feature since its already there with more choices .. so I'm closing this PR

Issue #4693, opened 2023-12-18
Switching panes too quickly on SSH domain causes rapid SetFocusedPane and then OOM


osandov:
### What Operating System(s) are you seeing this problem on?

Linux Wayland

### Which Wayland compositor or X11 Window manager(s) are you using?

GNOME 45

### WezTerm version

wezterm 20231203-124028-e3cd2e93

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

When I'm connected to an SSH domain and I move too quickly between panes (e.g., go from a pane in the top left to a pane in the bottom right without a pause in between the motions), WezTerm goes crazy. The pane focus starts jumping around when I press any keys, and then eventually WezTerm uses too much memory and gets OOM killed. Running with debug logs shows a flood of `encode_async SetFocusedPane size=6`.

### To Reproduce

* Connect to SSH domain
* Create a grid of 4 panes
* Hold Ctrl+Shift and mash arrow keys for a few seconds

### Configuration

I reproduced this with my config stripped down to only the `ssh_domains`:
```lua
return {
  ssh_domains = {
    {
      name = 'dl4',
      remote_address = 'dl4',
    },
  }
}
```

### Logs

```
...
01:07:58.721  DEBUG  codec                                  > encode_async SetFocusedPane size=6
01:07:58.721  DEBUG  codec                                  > encode_async SetFocusedPane size=6
01:07:58.721  DEBUG  codec                                  > encode_async SetFocusedPane size=6
01:07:58.721  DEBUG  codec                                  > encode_async SetFocusedPane size=6
01:07:58.721  DEBUG  codec                                  > encode_async SetFocusedPane size=6
01:07:58.721  DEBUG  codec                                  > encode_async SetFocusedPane size=6
01:07:58.722  DEBUG  codec                                  > encode_async SetFocusedPane size=6
01:07:58.722  DEBUG  codec                                  > encode_async SetFocusedPane size=6
01:07:58.722  DEBUG  codec                                  > encode_async SetFocusedPane size=6
01:07:58.722  DEBUG  codec                                  > encode_async SetFocusedPane size=6
01:07:58.722  DEBUG  codec                                  > encode_async SetFocusedPane size=6
01:07:58.722  DEBUG  codec                                  > encode_async SetFocusedPane size=6
01:07:58.722  DEBUG  codec                                  > encode_async SetFocusedPane size=6
[...]

Issue #4390, opened 2023-10-02
Switching panes in multiple directions at once causes the focused pane to rapidly switch when any key is pressed


illode:
### What Operating System(s) are you seeing this problem on?

Linux Wayland

### Which Wayland compositor or X11 Window manager(s) are you using?

KDE Plasma 5.27.8

### WezTerm version

20231002-075348-a61cc8eb

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

If I try to change to two different panes e.g. `Alt+Left+Right` (using my keybinds), the pane switching completely breaks. It doesn't always happen, but it happens often enough to be a huge pain.

It's easiest to explain with a video:

https://github.com/wez/wezterm/assets/114271894/99963be2-2517-4779-a976-938c12504d46

First I spam switch panes with `Alt+Random Arrow Key` (custom keybind), then once it breaks I spam a random key (z). You can see the focus randomly flicker between panes.

### To Reproduce

1. Use my configuration
2. Create multiple panes (`Ctrl+p -> s -> <Arrow Key>` with my config, but I don't think it really matters how they're made)
3. Switch to multiple panes at once using `Alt+<Up>+<Right>` or some other combination of arrow keys. There must a pane in the directions pressed for it to break.

### Configuration

local wezterm = require 'wezterm'
local act = wezterm.action

return {

  -- Keyboard Shortcuts
  leader = { key = 'p', mods = 'CTRL', timeout_milliseconds = 5000 },

  keys = {
    -- Key Tables
    { key = 'r', mods = 'LEADER', action = act.ActivateKeyTable { name = 'resize_pane', one_shot = false } },
    { key = 's', mods = 'LEADER', action = act.ActivateKeyTable { name = 'split_pane', one_shot = true } },
    { key = 'm', mods = 'LEADER', action = act.ActivateKeyTable { name = 'move_pane', one_shot = false } },

    -- Close Active Pane
    { key = 'w', mods = 'LEADER', action = act.CloseCurrentPane { confirm = false } },    
    -- Change Active Pane
    { key = 'LeftArrow', mods = 'ALT', action = act.ActivatePaneDirection 'Left' },
    { key = 'RightArrow', mods = 'ALT', action = act.ActivatePaneDirection 'Right' },
    { key = 'UpArrow', mods = 'ALT', action = act.ActivatePaneDirection 'Up' },
[...]

Issue #4517, opened 2023-10-31
Focus shifted randomly around panes


crides:
### What Operating System(s) are you seeing this problem on?

macOS

### WezTerm version

20230712-072601-f4abf8fd

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

No, and I'll explain why below

Edit: I didn't because I have a big session, and since it's not really possible to save the pane contents now, I don't want to lose them and restore the processes.

### Describe the bug

A weird state where the focus would jump around all of the panes in the current tab randomly. It seems like each key/mouse event will trigger a random pane focus. For example, when holding `e` for a while, each pane will get some amount of `e`, but the amounts are different.

It's also not possible to switch to other tabs, as any attempts to do so will just focus back to the current pane.

Edit: It's not possible to get rid of the issue until all but one of the panes are closed within one tab 

### To Reproduce

 I'm not sure how this was caused in the first place, but maybe I'll run into it again and I'll see

Edit: it happened again. one thing that is common between the 2 occurrences is that wezterm hangs briefly before things went bad. One of the unfocused panes was producing a lot of output, and wezterm stopped updating that pane for 5~10 seconds, and then when it seem to have recovered, the focusing shifting happens.

### Configuration

https://github.com/crides/dotfiles/blob/d03c95f3c6ae79fa555b8aaf9b332f68613c12e8/config/wezterm/wezterm.lua

### Anything else?

No logs in the debug window

wez:
I can't do much with this without a small and reproducible set of instructions.

PR #4737, opened 2024-01-01
Fix pane focus rapidly switching back and forth

osandov:
Switching panes multiple times too quickly sometimes causes the focus to get stuck rapidly switching back and forth between two panes by itself. This happens both locally and on an SSH domain. The root cause is similar in both cases.

In the local case, GuiFrontend handles a MuxNotification::PaneFocused event by calling Mux::focus_pane_and_containing_tab, which calls through Tab::set_active_pane -> TabInner::set_active_pane -> TabInner::advise_focus_change. TabInner::advise_focus_change then calls Mux::notify with a new MuxNotification::PaneFocused event, which is a redundant notification focusing the same pane again.

This is normally harmless other than a small amount of wasted work; TabInner::set_active_pane notices that the focused pane didn't change and doesn't generate yet another event.

However, if another real focus change is queued between the original focus change and the redundant one, we get into trouble. For example, if the user switches to pane 0 and then immediately to pane 1, the event queue contains [PaneFocused(0), PaneFocused(1)]. When the first event is handled, pane 0 is focused, and a redundant notification is generated, so the event queue contains [PaneFocused(1), PaneFocused(0)]. Then after the next event is handled, pane 1 is focused, and we add another redundant notification, so the event queue contains [PaneFocused(0), PaneFocused(1)]. Repeat ad nauseam.

Fix this by not generating the notification from
TabInner::advise_focus_change when it would be redundant, which we indicate with a new boolean parameter.

The mux server case is very similar, except that Pdu::SetFocusedPane in SessionHandler is the vector for the bug.

This bug appears to have been introduced by commit f71bce1727bbf1c384ca63a650a2aadec5afe841. I verified that `wezterm cli activate-pane-direction` still works over SSH after this fix.

closes: #4390
closes: #4693

### DIFF
--- a/mux/src/tab.rs
+++ b/mux/src/tab.rs
@@ -20,6 +20,12 @@ pub type Cursor = bintree::Cursor<Arc<dyn Pane>, SplitDirectionAndSize>;
 static TAB_ID: ::std::sync::atomic::AtomicUsize = ::std::sync::atomic::AtomicUsize::new(0);
 pub type TabId = usize;
 
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+pub enum NotifyMux {
+    Yes,
+    No,
+}
+
 #[derive(Default)]
 struct Recency {
     count: usize,
@@ -698,8 +704,8 @@ impl Tab {
         self.inner.lock().get_active_idx()
     }
 
-    pub fn set_active_pane(&self, pane: &Arc<dyn Pane>) {
-        self.inner.lock().set_active_pane(pane)
+    pub fn set_active_pane(&self, pane: &Arc<dyn Pane>, notify: NotifyMux) {
+        self.inner.lock().set_active_pane(pane, notify)
     }
 
     pub fn set_active_idx(&self, pane_index: usize) {
@@ -1747,7 +1753,7 @@ impl TabInner {
         self.active
     }
 
-    fn set_active_pane(&mut self, pane: &Arc<dyn Pane>) {
+    fn set_active_pane(&mut self, pane: &Arc<dyn Pane>, notify: NotifyMux) {
         if let Some(item) = self
             .iter_panes_ignoring_zoom()
             .iter()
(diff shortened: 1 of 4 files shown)


mux/src/tab.rs
```diff
@@ -698,8 +698,8 @@ impl Tab {
         self.inner.lock().get_active_idx()
     }
 
-    pub fn set_active_pane(&self, pane: &Arc<dyn Pane>) {
-        self.inner.lock().set_active_pane(pane)
+    pub fn set_active_pane(&self, pane: &Arc<dyn Pane>, notify: bool) {
```
wez, line 701: Please switch these `notify: bool` parameters to an enum, as booleans are hard to intuit the meaning of in the code, and are problematic if refactoring later adds additional boolean parameters; it can be tricky to get the ordering right.  Using enums solves both of these problems.

I'm not fantastic at naming, so if you come up better names for both the enum and its variants, I'm all ears!

```rust
#[derive(Debug, Clone, Copy, Eq, PartialEq)]
pub enum FocusNotify {
   Yes,
   No
}
```

so that the call sites look like:

```
tab.set_active_pane(&pane, FocusNotify::No);
```

```suggestion
    pub fn set_active_pane(&self, pane: &Arc<dyn Pane>, notify: FocusNotify) {
```

wez (commented):
Thanks for this! I've only quickly skimmed this without really following the call stacks through at the moment, and have one quick and easy bit of feedback!

osandov:
Done. I went with `NotifyMux` as the name, although I'm happy to change it to anything.

wez:
Thanks! Now, do `wezterm cli activate-tab`, `wezterm cli activate-pane-direction` and `wezterm cli activate-pane` still work after this change?

osandov:
Ugh, I tested `wezterm cli activate-pane-direction` and it worked, but the other two no longer work over SSH :( they all still work locally. I also just tested having multiple clients attached to the same SSH session, and this also broke syncing the pane focus updates between those.

My guess is that we need to avoid sending a redundant update back to the source it originally came from, but we need to send updates everywhere else. Is that possible?

P.S. There are some vaguely related issues I've run into with SSH sessions that I hadn't written up as issues yet: rotating panes through the command palette gets undone on the next keypress, and resizing panes by dragging the borders flickers. Ignoring `Pdu::TabResized` events in the client "fixes" those for me, but it breaks syncing clients, too. So I wonder if there's a more general solution here.

wez:
Thanks for checking!

I'm not in the right frame of mind to seriously design this right now, but I'll throw out a couple of possible approaches and let you noodle on them.

1. We know who the client is and could pass that through as part of the NotifyMux enum you added. Then it could get filtered out from being sent back to the originating client.  Not sure if I like this idea because it might break some subtle wakeup somewhere.
2. We have an `InputSerial` concept that likely could be applied here. I honestly can't remember the details of how it works without looking through the code, but look for `InputSerial` and `input_serial` to get the gist of it.  There could be some filtering applied by the client that will discard focus notifications if the input serial is older than some checkpoint that it has established already.  I don't have a good suggestion for how to frame that checkpoint off the topic, but here's where the serial is used today to avoid other unpleasantness:

https://github.com/wez/wezterm/blob/4921f139d35590ab35415021221a2a6f5cf10ab3/wezterm-client/src/pane/renderable.rs#L333-L346

crides:
Tested. This does fix the problem when it's caused by the terminal hanging on much output (i.e. #4517).

osandov:
I'm not going to get around to resolving the other muxing issues here, so I'm going to close this. Anyone else is welcome to pick this up.

PR #4752, opened 2024-01-06
fix: Proper block cursor shape

hyperpuncher:
Fixes https://github.com/wez/wezterm/issues/4231

Before:
before
After:
after

### DIFF
--- a/wezterm-gui/src/customglyph.rs
+++ b/wezterm-gui/src/customglyph.rs
@@ -262,7 +262,7 @@ impl PolyStyle {
                 let mut stroke = Stroke::default();
                 stroke.width = width;
                 if self == PolyStyle::OutlineHeavy {
-                    stroke.width *= 3.6; // NOTE: Changing this makes block cursor unproportinal at different font sizes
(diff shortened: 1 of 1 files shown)


wez:
Thanks for this! How do the other block glyphs look?
There are a number of scripts to run and text files you can cat in the `test-data` directory to eyeball for anything weird.

hyperpuncher:
Oh, didn't know about this tests, will test it and report.

hyperpuncher:
So, after testing many different combinations I think `stroke.width` needs to be multiplied by `3.1`.
With `3.6` couple of glyphs were fuzzy on all font sizes and hinting settings.

3.6
fuzzy_1
3.1
sharp_1

3.6
fuzzy_2
3.1
sharp_2

3.6
fuzzy_3
3.1
sharp_3

But on `3.1` there is one defect, if font size is higher than 22 cursor becomes slightly unproportional.

3.6
proportional
3.1
unproportional

Also found out that that  `legacy-box` is always thin on the right egde no matter what I change.

legacy-box

wez:
Thanks; I think this is good enough for the time being!

PR #4764, opened 2024-01-08
Ensure order of omitted escape code parameters is retained

### LINKED ISSUE #4730: Sixel parser ignores `P2` parameter if `P1` is blank

Linux Wayland

kwin

### WezTerm version

20231228-084719-ff274374

### Describe the bug

A sixel sequence follows the pattern:

```
DCS  P1;  P2;  P3;  q  s..s  ST
```

where `P1` is the macro parameter and `P2` selects how the terminal draws the background color.

If `P1` is blank but `P2` is set to `1` as follows:

```
\x1bP;1q ...
```

then unspecified pixel are set to the current background color instead of remaining at their current color as expected.

### To Reproduce

Run the following twice in the terminal:

```bash
echo -e '\eP0;1q;#1;2;100;0;0#1??}}GG}}??}}??--#0;2;0;100;100\e\\'
```

image

The second sixel graphic has a cyan background, which is not expected: it should be transparent, because `P2` was set to `1`. 

The third sixel graphic shows the expected output (`P1` is not blank and `P2` is `1`, resulting in transparent background).

### Configuration

default

### Expected Behavior

I would expect sixel device control string parameters to be parsed even if preceding parameters are blank.

### Logs

_No response_

### Anything else?

_No response_

joouha:
This fixes #4730.


wez:
Thanks!

PR #4767, opened 2024-01-09
Expose as_command API on CommandBuilder

mehulkar:


wez:
What's your goal here? In case it isn't clear, the object this returns is not connected to a PTY. This method is an implementation detail that is used by internal logic that is used to setup a child process.

What do you gain from exposing it outside the crate compared to using `std::process::Command` directly?

mehulkar:
I'm still at experimental phase, but I'm essentially wanting to use our existing implementation to set a process group. Currently that looks something like this:

```rust
let command = tokio::process::Command::new();
let std_cmd = command.as_std();
let child = std_cmd.group.spawn()
```

The `.group()` implementation is just `CommandGroupBuilder::new(std_cmd)`, so I was hoping to do this:

```rs
let pty_system = NativePtySystem::default();
let pair = pty_system.openpty(PtySize::default()).unwrap();
let cmd = CommandBuilder::new();
let group = CommandGroupBuilder::new(cmd.as_command()); // ← new API from this PR
group.spawn();
```

After typing this up and understanding your reply I see how this won't work, since:

> the object this returns is not connected to a PTY

Maybe what needs to happen instead is some kind of option to `spawn_command()` that can set the group instead? (and then I can remove `command-group`?

wez:
Can you expand on the group thing a bit more?
What are you trying to do?
The default behavior for a new pty is tied up with the use of the `setsid(2)` system call that happens here:
https://github.com/wez/wezterm/blob/1c0901957d3eb3deaa8d9406040f234ee81733e5/pty/src/unix.rs#L255-L258

mehulkar:
In turborepo, we spawn child processes for a bunch of scripts (provided by the user) in parallel. We use a process group so we can kill all these process (i.e. forward signals). 

I want to keep this behavior. Is establishing as a session leader necessary for this pty? Is it possible to spawn a child process with a connected pty that can be part of a process group set from the outside?

wez:
Perhaps what you want is to add a unix specific `set_process_group` method to `CommandBuilder` that stores an optional process group id in the builder, then, in the code I pointed to, do something like:

```rust
// process_group is pre-extracted from the builder in the same was as controlling_tty
if let Some(pgrp) = process_group {
   if libc::setpgrp(pgrp) == -1 {
      return Err(io::Error::last_os_error());
   }
} else {
  if libc::setsid() == -1 {
      return Err(io::Error::last_os_error());
   }
}
```

PR #4775, opened 2024-01-11
Fix freetype_load_target

### LINKED ISSUE #4770: freetype_load_target doesn't work

Linux X11

Compiz

### WezTerm version

20240103-143005-4921f139

### Describe the bug

freetype_load_target is ignored.

### To Reproduce

_No response_

### Configuration

  config.freetype_load_target = "Light"
  config.freetype_render_target = "HorizontalLcd"
  config.freetype_interpreter_version = 40
  config.font = wezterm.font({ family = 'Roboto Mono', weight = 600 })
  config.font_size = 9

### Expected Behavior

freetype_load_target = "Light":

zsh_003

freetype_load_target = "Normal":

zsh_004

### Logs

_No response_

### Anything else?

I think there should be extra parentheses in the render_mode_to_load_target function, otherwise the load_flags gets calculated incorrectly:

((render_mode as u32) & 15) << 16

dkoukola:
Fixes #4770

### DIFF
--- a/wezterm-font/src/ftwrap.rs
+++ b/wezterm-font/src/ftwrap.rs
@@ -51,7 +51,7 @@ fn render_mode_to_load_target(render_mode: FT_Render_Mode) -> u32 {
     // for these in the bindings so we do some bit magic for
     // ourselves.  This is how the FT_LOAD_TARGET_() macro
     // assembles these bits.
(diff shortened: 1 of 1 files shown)


wez:
Oh, that's a good find, thank you!

Issue #2297, opened 2022-07-24
Crash on resume when monitor setup changed


titilambert:
### What Operating System(s) are you seeing this problem on?

Linux Wayland

### Which Wayland compositor or X11 Window manager(s) are you using?

sway

### WezTerm version

20220720-074534-41ee60c9

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

Wezterm crashes on resume (after suspend) when an external monitor is plugged or unplugged during the suspend
```
not a valid new object id (4278190080), message data_offer(n)
15:53:49.353  ERROR  wezterm_gui > running message loop: error during event_q.read_events protocol_error=None: Invalid argument (os error 22); terminating
```

### To Reproduce

1. Launch wezterm
2. suspend the laptop
3. plug a new monitor
4. resume
5. it crashes

### Configuration

```
local wezterm = require 'wezterm';
local act = wezterm.action

return {
  tab_bar_at_bottom = true,
  window_close_confirmation = "AlwaysPrompt",
  skip_close_confirmation_for_processes_named = {},
  harfbuzz_features = {"calt=0", "clig=0", "liga=0"},
   window_padding = {
    left = 1,
    right = 1,
    top = 1,
    bottom = 1,
  },
  -- font = wezterm.font("DejaVu Sans Mono"),
  font = wezterm.font("Noto Mono"),
  keys = {
    {key="LeftArrow", mods="SHIFT", action=act.ActivateTabRelative(-1)},
    {key="RightArrow", mods="SHIFT", action=act.ActivateTabRelative(1)},
    {key="LeftArrow", mods="SHIFT|CTRL", action=act.MoveTabRelative(-1)},
    {key="RightArrow", mods="SHIFT|CTRL", action=act.MoveTabRelative(1)},
    -- {key="w", mods="SHIFT|CTRL", action=wezterm.action.CloseCurrentTab{confirm=true}},
    -- {key="w", mods="SHIFT|CTRL", action=wezterm.action.CloseCurrentPane{confirm=true}},
  }
}
```

### Logs

```
not a valid new object id (4278190080), message data_offer(n)
15:53:49.353  ERROR  wezterm_gui > running message loop: error during event_q.read_events protocol_error=None: Invalid argument (os error 22); terminating
```

wez:
Possible a dupe of https://github.com/wez/wezterm/issues/2293

[...]

PR #4777, opened 2024-01-11
Upgrade smithay-client-toolkit 0.17

tzx:
Saw https://github.com/wez/wezterm/issues/4504, when I was looking into https://github.com/wez/wezterm/issues/4483.

Understand that this requires a large refactor, so I am attempting it here. Currently, typing works, but there are a lot of missing features that I am noting down, possibly more...:

- [x] Mouse input doesn't work at all
- [x] rendering cursor icons
- [ ] window decorations (It currently uses the Fallback one from SCTK, but not wezterm's own custom one)
- [x] clipboard support
- [x] primary clipboard support (~~seems like I rather update to 0.18 to make this easier~~, I did it in 0.17 but 0.18 has an abstraction)
- [x] drag and drop support
- [x] ime support
- [x] wlr-output-management-unstable
- [x] all rendering frontends supported (egl seems weird because I hardcoded stuff so I need to fix that)


wez:
Thank you for starting on this! It's not a small undertaking!

tmccombs:
I'd be willing to help with this some. But I'm not familiar with the wezterm codebase at all. I am familiar with rust. 

What could I do that would be most helpful?

tzx:
Hey @wez,

I was looking at old commits and I was wondering why you kept [output.rs](https://github.com/wez/wezterm/blob/main/window/src/os/wayland/output.rs) which uses wlr-output-management protocol, when it is disabled due to #2297 from https://github.com/wez/wezterm/commit/e2bf468393f77e227e518ac4acb8f1c74edc003e. The option is not documented and the discussion did mention that a client like wezterm shouldn't be using this. Care to explain why it remained or should I just remove it?

wez:
> Hey @wez,
> 
> I was looking at old commits and I was wondering why you kept [output.rs](https://github.com/wez/wezterm/blob/main/window/src/os/wayland/output.rs) which uses wlr-output-management protocol, when it is disabled due to #2297 from [e2bf468](https://github.com/wez/wezterm/commit/e2bf468393f77e227e518ac4acb8f1c74edc003e). The option is not documented and the discussion did mention that a client like wezterm shouldn't be using this. Care to explain why it remained or should I just remove it?

I wanted to keep the option of being able to reason about the display to support functions like https://wezfurlong.org/wezterm/config/lua/wezterm.gui/screens.html

I'm not overly attached to it though, so if it makes things easier, you can certainly remove that particular implementation.

salimp2009:
Not related to this but in case you have not seen
there is also wgpu issue related to vulkan backend with Nvidia GPU which has been resolved but the version you are using needs to include that specific merge or patch the dependency as needed. 

it was also effecting the performance . After the patch is applied then it works  (I tried this using ice crate )
https://github.com/gfx-rs/wgpu/pull/4967

I used this to patch the ice crate (but had to use the latesT version from github)

```
[patch.crates-io]
wgpu = { git = "https://github.com/gfx-rs/wgpu", rev = "ac8756c" }
```
image

wez:
Note that newer versions of wgpu have fatal resource leaks at the moment, so we are blocked from upgrading until they are resolved upstream.
* https://github.com/wez/wezterm/issues/4826
* https://github.com/wez/wezterm/issues/4825

tzx:
Alright @wez ,

I (believe I) have most of the stuff for 0.17. The one thing missing is the themed frame (and the themed pointer to go along with it), but I currently now have it using the fallback frame that SCTK uses. Some thorough testing would be also nice, as I was not testing every capability that's possible from wezterm. 

Unfortunately, I am now too busy and don't have enough energy to work more on this. This task consumed a decent chunk of my time, and I now understand why you didn't want to start on this. I didn't expect the API to changed that much. I do want to mention that I tried to keep some of the object/types similar to what was there previously, but I think SCTK 0.17 does expect something else (Right now, a lot of object data associated with Wayland is empty, but the inner is coupled with the manager, see `TextInputState` and `TextInputData` as an example with their two "inners"). I wanted to handle that in a separate refactoring as this current task is already a lot.

Hopefully, you or someone else can pick this up. I apologize for not being able to finish this, but I hope that I helped on getting this task started.

wez:
@tzx many thanks for doing this work! I appreciate it a great deal! I'll see if I can persuade one or two people to run with this and we can figure out where to go from there!

jokeyrhyme:
@tzx amazing, Herculean effort :D

I'll check the `todo!()` macros and compare some window and settings behaviours across the compositors I use (GNOME, sway, COSMIC) and report back

In terms of continuing this work, if I were up to actually contributing code, should we fork the fork and raise PRs against it, fork the fork and submit a new PR here, or add more people as collaborators to the fork to preserve this PR?

wez:
Since we just cut a release, which gives folks an up to date stable build to target, I'm conceptually ok with merging this and making nightly a bit wonky for wayland users while we quickly try to close any gaps that might cause work to be lost, and then we can take a bit longer to polish up the rest.

jokeyrhyme:
I'll update this comment with results from other configurations, but to start with...

Testing what we have so far on COSMIC (pretty sure this offers server-side decorations) with tiling disabled

- `cargo run --bin wezterm-gui --release -- --config 'window_decorations="NONE"'` -> SCTK panic "trying to resize hidden frame"
- `cargo run --bin wezterm-gui --release -- --config 'window_decorations="RESIZE"'` -> `todo!()` panic
- `cargo run --bin wezterm-gui --release -- --config 'window_decorations="TITLE"'` -> `todo!()` panic
- `cargo run --bin wezterm-gui --release -- --config 'window_decorations="INTEGRATED_BUTTONS|RESIZE"'` -> `todo!()` panic
- `cargo run --bin wezterm-gui --release -- --config 'window_decorations="TITLE|RESIZE"'` -> SCTK panic "trying to resize hidden frame"

COSMIC, with tiling enabled:

- `cargo run --bin wezterm-gui --release -- --config 'window_decorations="NONE"'` -> SCTK panic "trying to resize hidden frame"
- `cargo run --bin wezterm-gui --release -- --config 'window_decorations="RESIZE"'` -> works
- `cargo run --bin wezterm-gui --release -- --config 'window_decorations="TITLE"'` -> works
- `cargo run --bin wezterm-gui --release -- --config 'window_decorations="INTEGRATED_BUTTONS|RESIZE"'` -> works
- `cargo run --bin wezterm-gui --release -- --config 'window_decorations="TITLE|RESIZE"'` -> SCTK panic "trying to resize hidden frame"

tzx:
Ahh, I forgot about turning off window_decorations in my config when working on frame decorations when I was trying wezterm on weston... Hope it's an easy fix.

jokeyrhyme:
@tzx don't worry about it
If there are other settings you think I ought to be interested in, let me know
But, I _will_ eventually dive in and figure out what to look at based on your comment ( https://github.com/wez/wezterm/pull/4777#issuecomment-1913394104 ) and the `todo!()` s
So don't stress about it :)

jokeyrhyme:
Okay, I've had a chance to play with this

I'm no expert in Wayland or SCTK, although learning a lot just reading through the code

What I'd like to do is prepare a PR that can be merged soon after this PR, and my approach is to hard code CSD via SCTK Fallback Frame, ignoring user and compositor preference, with the user able to move/resize/maximise the window without hitting any `todo!()`

Then separate future PRs can look at updating to SCTK 0.18, and restoring user control over decorations

jokeyrhyme:
I've created a branch here: https://github.com/jokeyrhyme/wezterm/tree/smithay-0-17-conflicts-after-4824-and-4866

I figured it was worth tackling the `git` conflicts separately before trying to do much else

This incorporates the changes to resize_increment behaviour from https://github.com/wez/wezterm/pull/4824 and https://github.com/wez/wezterm/pull/4866

tmccombs:
I tested this branch on Hyprland and found the following issues:

1.  For every key press, I see an error that looks like:  "ERROR  window::os::wayland::window > set_cursor: Cursor not found" on stderr
2. With `window_decorations` of NONE, or "TITLE | RESIZE" I get the "trying to resize hidden frame" panic
3.  With `window_decorations` of "TITLE" or "RESIZE" It opens, but there are strange gray lines to the bottom and left:
image
4.  The text of the terminal seems offset, possibly due to decorations, but the decorations aren't actually rendered. This is most obvious if I open a TUI with a background color, like vim:
image
5. Probably related to 4, If I go down to the bottom of the window, then the last line oveflows past the end of  the window:
image

tmccombs:
Actually... If I make the window floating, it makes a little more sense: 
image

It looks like it is trying to draw the whole window offset from the frame Hyprland is using for the window. Maybe there is a problem between the reference frame of  the window including the client decorations and without.

Although, ideally I'd use `window_decorations="NONE"` on hyprland.  But that currently panics.

tmccombs:
A few thoughts on this:

1. We should be passing in the appropraite decorations mode to `xdg_shell::new_window` (second argument) instead of always using `Decorations::FollowServer`. 
2. If we are suing server-side decorations (which I don't think we know until after getting a configure back from the server, I'm not entirely sure where the code that handles that is), then we shouldn't create the `window_frame` for the decorations at all. I think that panic is happening because we are trying to resize the frame after hiding it.

tmccombs:
if someone else doesn't get to it, I think I know what needs to be done to fix this, and a general idea of how to fix this at least using the fallback frame. 

Although using https://github.com/PolyMeilex/sctk-adwaita might look nicer.

jokeyrhyme:
@tmccombs thanks!

I believe this PR is effectively frozen, so I've created some follow-up PRs that address some of the panics by using the fallback frame in all cases (for now) as a temporary stop-gap

- https://github.com/wez/wezterm/pull/4905
- https://github.com/wez/wezterm/pull/4906

wez:
I've rebased this and pushed it to `main`, for (mostly) better and (ideally temporarily) slightly worse!
Thanks again @tzx!

catdevnull:
Hi, just wondering, would this somewhat improve the fractional scaling situation in Wayland?

tmccombs:
Here's my draft at attempting to improve this: 

https://github.com/wez/wezterm/pull/4970

dedguy21:
start if i enter:
```env WAYLAND_DISPLAY=1 wezterm```

added that to the .desktop app script, and opens everytime now on hyprland.

PR #4781, opened 2024-01-12
x11: Fix comment around modifiers merging, tweak changelog

bew:
refs: #4151

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -141,8 +141,8 @@ As features stabilize some brief notes about them will accumulate here.
   that is pasted into the terminal. This happens even if bracketed paste
   mode is disabled. #4765
 * Sixel parser ignores P2 parameter if P1 is blank. Thanks to @joouha! #4730
-* X11: synthesize raw modifiers to resolve some issues with synthetic/injected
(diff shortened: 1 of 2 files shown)


bew:
Friendly ping @wez, this is a short comment PR

wez:
re: build failure, I think I'll just remove centos 7. It's EOL and not worth the effort to try to keep it going any more.

I was holding off on this PR for the moment, because I'm wondering if unconditionally merging of the modifiers is the cause of https://github.com/wez/wezterm/issues/4776 but we haven't had feedback on that yet.

Also, something weird I've noted once or twice in the past week is that if I resize wezterm, sometimes vim mysteriously gets suspend as though I typed ctrl-z.  I don't know if this is just some crazy unrelated thing or if there is some weird stuff happening with input on this system--it doesn't seem like resizing with the mouse would be connected to input, but also: it seems weird that the modifiers spontaneously get reset to 0 here. I haven't had time/motivation to sit down and debug it.

bew:
Your modifiers reset is definitely weird 😬
But this PR doesn't change any code, only makes it clearer imo (by moving & reworking the comment).

I'm thinking your weird behavior could be tracked in a separate issue?
(which might end up reverting / reworking the code for mods merge)

PR #4810, opened 2024-01-19
fix: changed the criteria to calculate the sunset / sunrise progression

mikyk10:
Fixes https://github.com/wez/wezterm/issues/4809


wez:
Thank you!

PR #4822, opened 2024-01-20
x11: fix resize increments

jknockel:
Set atom with property (type) `WM_NORMAL_HINTS (WM_SIZE_HINTS)` instead of `WM_SIZE_HINTS (CARDINAL)`.

The `use_resize_increments` feature didn't work at all for me prior.

### DIFF
--- a/window/src/os/x11/window.rs
+++ b/window/src/os/x11/window.rs
@@ -1558,8 +1558,8 @@ impl XWindowInner {
         self.conn().send_request_no_reply(&xcb::x::ChangeProperty {
             mode: PropMode::Replace,
             window: self.window_id,
-            property: xcb::x::ATOM_WM_SIZE_HINTS,
-            r#type: xcb::x::ATOM_CARDINAL,
+            property: xcb::x::ATOM_WM_NORMAL_HINTS,
+            r#type: xcb::x::ATOM_WM_SIZE_HINTS,
             data,
         })?;
 


wez:
Thanks for this; what do you think about setting both sets of properties? It's been a while since I used this, but it did work in whatever WM I was running when I built it at the time.

jknockel:
> Thanks for this; what do you think about setting both sets of properties? It's been a while since I used this, but it did work in whatever WM I was running when I built it at the time.

How curious -- it's surprising to me that the previous version worked with a window manager, since the previous version seems to contain a simple kind of mixup (namely, using `WM_SIZE_HINTS` as a property instead of as a type).

I did find some [behavior](https://gitlab.freedesktop.org/xorg/app/xprop/-/blob/0d2837162ef0e54647485eee2ba4b88012fbb1b4/xprop.c#L1538) in xprop that might explain why the previous version might still work with a window manager, as xprop uses both property and type to try to parse an atom, so it's possible that the window manager in question was doing something similar. However, if the window manager were doing something similar, then it would be only accidentally working with `WM_SIZE_HINTS (CARDINAL)`, and it would still be sufficient to only set `WM_NORMAL_HINTS (WM_SIZE_HINTS)`. (Note too that a more high level tool, xwininfo, only correctly interprets `WM_NORMAL_HINTS (WM_SIZE_HINTS)` and ignores `WM_SIZE_HINTS (CARDINAL)`.)

Also, in looking at how other terminals do this, both gnome-terminal and xfce4-terminal use `gtk_window_set_geometry_hints()` which only sets `WM_NORMAL_HINTS (WM_SIZE_HINTS)`. So if there exists a window manager that only works with `WM_SIZE_HINTS (CARDINAL)` and not `WM_NORMAL_HINTS (WM_SIZE_HINTS)`, then it is probably broken with a lot of other software that uses resize increments as well.

So, while it's impossible to say with 100% certainty without knowing the window manager in question, I believe for the above reasons that it would be safe to not set `WM_SIZE_HINTS (CARDINAL)`.

wez:
OK, sounds good to me!

PR #4824, opened 2024-01-21
x11: improve use_resize_increment behavior

jknockel:
These changes fix an issue where the number of terminal cells can change when (1) using X11, (2) during zooming in or out (e.g., when pressing `ctrl+-` or `ctrl+=`), (3) with `use_resize_increment` enabled.

As a desirable side effect of these changes, when resizing the terminal, many X11 window managers will now report the current cell dimensions of the terminal (e.g., 80x24) as an overlay when resizing the window.  Previously this would be a different number (e.g., 81x25) that was near to but not exactly the cell dimensions.

Depending on the amount of padding, we could previously resize by dragging the edge of the window to get a negative size displayed by the window manager's overlay (e.g., 80x-1 for a very short window displaying zero terminal lines and only partially showing the top of the tabs).  It is also desirable to avoid 80x0 or 0x24 dimensions as these can render strangely.  For example, when there is a lot of padding, an 80x0 size can still show part of the top of the first line despite it being described as a cell height of zero. Therefore, we now hint to the window manager that the terminal should be a minimum of 1x1 cells.

Note that I have only been able to test these changes on X11.

This is my first time using rust, so I apologize for any unidiomatic code or for any unfamiliarity with this code base. :)

Note that these changes don't make much sense without [!4822](https://github.com/wez/wezterm/pull/4822), but I wanted to split that one off because I am hoping that one will be more straightforward to review.

### DIFF
--- a/wezterm-gui/src/main.rs
+++ b/wezterm-gui/src/main.rs
@@ -43,6 +43,7 @@ mod markdown;
 mod overlay;
 mod quad;
 mod renderstate;
+mod resize_increment_calculator;
 mod scripting;
 mod scrollbar;
 mod selection;
--- a/wezterm-gui/src/termwindow/resize.rs
+++ b/wezterm-gui/src/termwindow/resize.rs
@@ -1,5 +1,5 @@
 use crate::utilsprites::RenderMetrics;
-use ::window::{Dimensions, Window, WindowOps, WindowState};
+use ::window::{Dimensions, ResizeIncrement, Window, WindowOps, WindowState};
 use config::{ConfigHandle, DimensionContext};
 use mux::Mux;
 use std::rc::Rc;
@@ -101,18 +101,14 @@ impl super::TermWindow {
             }
         }
 
-        window.set_resize_increments(
-            if self.config.use_resize_increments {
-                self.render_metrics.cell_size.width as u16
-            } else {
-                1
-            },
-            if self.config.use_resize_increments {
-                self.render_metrics.cell_size.height as u16
-            } else {
-                1
-            },
-        );
+        window.set_resize_increments(if self.config.use_resize_increments {
+            ResizeIncrement {
+                x: self.render_metrics.cell_size.width as u16,
+                y: self.render_metrics.cell_size.height as u16,
+            }
+        } else {
+            ResizeIncrement::disabled()
+        });
 
         if let Err(err) = self.recreate_texture_atlas(None) {
             log::error!("recreate_texture_atlas: {:#}", err);
--- a/wezterm-gui/src/termwindow/resize.rs
+++ b/wezterm-gui/src/termwindow/resize.rs
@@ -1,3 +1,4 @@
+use crate::resize_increment_calculator::ResizeIncrementCalculator;
 use crate::utilsprites::RenderMetrics;
 use ::window::{Dimensions, ResizeIncrement, Window, WindowOps, WindowState};
 use config::{ConfigHandle, DimensionContext};
@@ -146,7 +147,7 @@ impl super::TermWindow {
 
         let border = self.get_os_border();
 
-        let (size, dims) = if let Some(cell_dims) = scale_changed_cells {
+        let (size, dims, ri_calc) = if let Some(cell_dims) = scale_changed_cells {
             // Scaling preserves existing terminal dimensions, yielding a new
             // overall set of window dimensions
             let size = TerminalSize {
@@ -191,7 +192,18 @@ impl super::TermWindow {
                 dpi: dimensions.dpi,
             };
 
-            (size, dims)
+            let ri_calc = ResizeIncrementCalculator {
+                x: self.render_metrics.cell_size.width as u16,
+                y: self.render_metrics.cell_size.height as u16,
+                padding_left: padding_left,
+                padding_top: padding_top,
+                padding_right: padding_right,
+                padding_bottom: padding_bottom,
+                border: border,
+                tab_bar_height: tab_bar_height as usize,
+            };
+
+            (size, dims, ri_calc)
         } else {
             // Resize of the window dimensions may result in changed terminal dimensions
 
@@ -238,7 +250,18 @@ impl super::TermWindow {
                 dpi: dimensions.dpi as u32,
             };
 
-            (size, *dimensions)
+            let ri_calc = ResizeIncrementCalculator {
+                x: self.render_metrics.cell_size.width as u16,
+                y: self.render_metrics.cell_size.height as u16,
+                padding_left: padding_left,
+                padding_top: padding_top,
+                padding_right: padding_right,
+                padding_bottom: padding_bottom,
+                border: border,
+                tab_bar_height: tab_bar_height as usize,
+            };
+
+            (size, *dimensions, ri_calc)
         };
 
         log::trace!("apply_dimensions computed size {:?}, dims {:?}", size, dims);
@@ -256,10 +279,7 @@ impl super::TermWindow {
         self.update_title();
 
         window.set_resize_increments(if self.config.use_resize_increments {
-            ResizeIncrement {
-                x: self.render_metrics.cell_size.width as u16,
-                y: self.render_metrics.cell_size.height as u16,
-            }
+            ri_calc.into()
         } else {
             ResizeIncrement::disabled()
         });
--- a/wezterm-gui/src/termwindow/resize.rs
+++ b/wezterm-gui/src/termwindow/resize.rs
@@ -63,12 +63,7 @@ impl super::TermWindow {
         self.emit_window_event("window-resized", None);
     }
 
-    pub fn apply_scale_change(
-        &mut self,
-        dimensions: &Dimensions,
-        font_scale: f64,
-        window: &Window,
-    ) {
+    pub fn apply_scale_change(&mut self, dimensions: &Dimensions, font_scale: f64) {
         let config = &self.config;
         let font_size = config.font_size * font_scale;
         let theoretical_height = font_size * dimensions.dpi as f64 / 72.0;
@@ -100,15 +95,6 @@ impl super::TermWindow {
                 self.fonts.change_scaling(prior_font, prior_dpi);
             }
         }
-
-        window.set_resize_increments(if self.config.use_resize_increments {
-            ResizeIncrement {
-                x: self.render_metrics.cell_size.width as u16,
-                y: self.render_metrics.cell_size.height as u16,
-            }
-        } else {
-            ResizeIncrement::disabled()
-        });
 
         if let Err(err) = self.recreate_texture_atlas(None) {
             log::error!("recreate_texture_atlas: {:#}", err);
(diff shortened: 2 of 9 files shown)


wezterm-gui/src/termwindow/resize.rs
```diff
@@ -302,9 +294,37 @@ impl super::TermWindow {
                 // pixel geometry which is considered to be a user-driven resize.
                 // Stashing the dimensions here avoids that misconception.
                 self.dimensions = dims;
+                // Our new size might not be divisible by the current
+                // increment, and our current size might not be divisible by
+                // the new increment, so we disable the increment while
+                // resizing and then set the new increment afterward.
+                window.set_resize_increments(1, 1, 0, 0);
```
wez, line 301: one thing I'm not 100% sure about here is the ordering of this wrt. to the final (re)application of the configured values below.  What I mean by that is that this conditional block ("Queue up...") has effects that happen partially asynchronously before they fully apply and settle, triggering this resize logic again in a later pass.  For example, tiling window managers can and do re-adjust the size, and on macOS, if the window would exceed the screen size, it will change it.

I don't know if that matters here: I think resize increments are not really compatible with tiling window managers, so I don't know if we need to sweat the details of what happens there and just tell folks to pick which paradigm they want to use.

window/src/os/x11/window.rs
```diff
@@ -1525,16 +1525,16 @@ impl XWindowInner {
...
         let hints = xcb_size_hints_t {
-            flags: XCB_ICCCM_SIZE_HINT_P_RESIZE_INC,
+            flags: XCB_ICCCM_SIZE_HINT_P_MIN_SIZE | XCB_ICCCM_SIZE_HINT_P_RESIZE_INC | XCB_ICCCM_SIZE_HINT_BASE_SIZE,
             x: 0,
             y: 0,
             width: 0,
             height: 0,
-            min_width: 0,
-            min_height: 0,
+            min_width: (base_width + x).into(),
```
wez, line 1536: FWIW, the macOS equivalent of this appears to be: https://developer.apple.com/documentation/appkit/nswindow/1419670-contentminsize?language=objc

window/src/os/x11/window.rs
```diff
@@ -1543,8 +1543,8 @@ impl XWindowInner {
             min_aspect_den: 0,
             max_aspect_num: 0,
             max_aspect_den: 0,
-            base_width: 0,
-            base_height: 0,
+            base_width: base_width.into(),
```
wez, line 1546: I don't see a macOS equivalent of this, but perhaps it can be emulated by calculating https://developer.apple.com/documentation/appkit/nswindow/1419124-contentlayoutrect?language=objc appropriately

wez (commented):
Thanks for this!

re: macOS and Windows; I don't expect you to run these on those systems, because I think this change is largely platform agnostic (beyond passing more data to the platform-specific parts), and because Windows doesn't have any logic at all for this, and macOS has a very simple subset.

For Wayland, the logic for increments is handled entirely in the window crate.  I'd appreciate it if you could run eg: Weston or Wayfire in your X11 session to get a little internal wayland compositor window and test your changes there; I think having the base width and height work there would be nice.  That said, wayland support in wezterm is currently not great, and there are efforts in #4777 to effectively rewrite it, so communication with the folks looking at that would be good to avoid losing or breaking stuff around that work.

I think the logic in here looks sound, and my comments are largely stylistic.

wezterm-gui/src/termwindow/resize.rs
jknockel: I think that there are at least two invariants we want to be maintained.  First is we want the window manager to always see a consistent version of (1) our requested resize increments and (2) our requested window size.  By consistent here, I mean that our resize increments should divide the window size.  I *think* we are OK with this invariant, at least if we make the basic assumption that the window manager is processing events in the order delivered (if it's not doing that then I don't know how we can ever guarantee anything).

But I think that there is a second one that involves consistency between the client's (wezterm's) internal state and ConfigureNotify events, especially if those ConfigureNotify events come in "late".  For instance, if I press ctrl+-, normally something like this would happen (take scale to be a scaling factor, termsize to be terminal dimensions as measured in cells, and windowsize to be the window dimensions as measured in pixels):

* Initial (scale = scale0, termsize = termsize0, windowsize = windowsize0)
* Ctrl+- (scale = scale1, send ConfigureRequest(scale1 * termsize))
* ConfigureNotify (windowsize = scale1 * termsize, termsize = windowsize1 / scale1)

So far so good!

What if I press ctrl+- twice very quickly?

* Initial (scale = scale0, termsize = termsize0, windowsize = windowsize0)
* Ctrl+- (scale = scale1, send ConfigureRequest(scale1 * termsize))
* Ctrl+- (scale = scale2, send ConfigureRequest(scale2 * termsize))
* ConfigureNotify (windowsize = scale1 * termsize, termsize = windowsize1 / scale2) // WRONG termsize
* ConfigureNotify (windowsize = scale2 * termsize, termsize = windowsize2 / scale2) // correct termsize is restored

OK, we have the wrong terminal size for a moment, but at least it is eventually corrected.  But what if I'm spamming the key and this happens:

* Initial (scale = scale0, termsize = termsize0, windowsize = windowsize0)
* Ctrl+- (scale = scale1, send ConfigureRequest(scale1 * termsize))
* Ctrl+- (scale = scale2, send ConfigureRequest(scale2 * termsize))
* ConfigureNotify (windowsize = scale1 * termsize, termsize = windowsize1 / scale2) // WRONG termsize
* Ctrl+- (scale = scale3, send ConfigureRequest(scale3 * termsize_wrong)) // ConfigureRequest sent with WRONG size
* ConfigureNotify (windowsize = scale3 * termsize, termsize = windowsize2 / scale3) // a different wrong termsize
* ConfigureNotify (windowsize = scale3 * termsize_wrong, termsize = windowsize3 / scale3 = termsize_wrong) // original wrong termsize :(

So I believe that we also want some invariant which is something like that we have consistent scale and window sizes whenever we send ConfigureRequests, which means that we will need to somehow be sure that we have received all pending ConfigureNotify requests from the window manager.

And actually, this race has nothing to do with this merge request.  I can reproduce this race condition, or at least one like it, on the main branch without even using a tiling window manager, just by spamming the ctrl+- and ctrl+= keys.  Specifically, spamming ctrl+- or ctrl+= will eventually change my termsize.

If this race condition exists, then this might be tricky to fix, as I'm not sure what synchronization primitive to use to fix this.  If `XSync()` is under the hood sending an `GetInputFocus` request and waiting for the response, then what would `XSyncWindowManager()` look like?  Create an unmapped dummy window, reconfigure it, and then wait for the `ConfigureNotify` event?  If we make the same basic assumption as above that the window manager is processing events in the order delivered, then it might work...

wez: FWIW, we're using xcb rather than libx11 under the hood, so there is no XSync.  Also: this synchronization problem potentially applies to other platforms where the model is different. For example, on Windows, we may receive multiple messages one after the other, or we may get recursively called by windows' internal message dispatching when we trigger a size change while processing an event.

So... not easy to resolve in isolation.

In the context of this PR: if your changes work about as well as the existing state of things, then I'm not going to require you to resolve this synchronization issue here.

jknockel: > FWIW, we're using xcb rather than libx11 under the hood, so there is no XSync. 

Of course -- I was just speaking in terms of pseudocode there, although if I recall correctly, XCB also uses GetInputFocus to implement synchronization in the *_checked variants of its functions.

> Also: this synchronization problem potentially applies to other platforms where the model is different. For example, on Windows, we may receive multiple messages one after the other, or we may get recursively called by windows' internal message dispatching when we trigger a size change while processing an event.
> 
> So... not easy to resolve in isolation.

True. But this is really bugging me now.  I'd like to get back to it at some point, if even just for the X11 case.

> In the context of this PR: if your changes work about as well as the existing state of things, then I'm not going to require you to resolve this synchronization issue here.

Acknowledged, and thanks for your other feedback too.  It was very helpful.  I'll implement the recommendations as soon as I get a chance.

wez:
sorry, rebasing on 516add5a4b162839e781555fa342682fca539ef1 should fixup the test failures that are my fault

jknockel:
I've attempted to address your code comments!

There is one other change that I made, which is that I no longer set the disabled (1, 1, 0, 0) `ResizableIncrement` before calling `set_inner_size()` as I found at least on the window managers that I tested it wasn't actually necessary and that we can just call `set_resize_increment()` as usual before calling `set_inner_size()` instead.  I can't remember what compelled me to originally think that setting the disabled `ResizableIncrement` before calling `set_inner_size()` was necessary, but I suspect that I was already working on refactoring the code to implement `base_width`/`base_height` and that I had mistakenly refactored the `set_resize_increment()` call to be after the `set_inner_size()` call without having realized the importance of the order of these calls. I think that not setting the disabled `ResizableIncrement` should be safe in that the code is now closer to the way it was before this merge request when we were effectively just calling `set_resize_increment()` before `set_resize_increment()` anyways, but since I don't believe there is any harm either (at least on X11) in setting the disabled `ResizableIncrement` before `set_inner_size()`, that way is still an option, especially if it helps another platform.

I haven't implemented Wayland or OS X yet, but I have tested on Weston, and there don't appear to be regressions at least.  Are implementing Wayland and OS X required for this merge request?  If so, I can take a stab at Wayland next.

wezterm-gui/src/main.rs
```diff
@@ -43,6 +43,7 @@ mod markdown;
 mod overlay;
 mod quad;
 mod renderstate;
+mod ricalc;
```
wez, line 46: Let's use a more verbose name for this, as otherwise I'm sure that I won't remember what this is in a month or two :)

```suggestion
mod resize_increment_calculator;
```

wez:
This is looking great, thanks!

> I haven't implemented base_width/base_height on Wayland or OS X yet, but I have tested on Weston, and there don't appear to be regressions at least. Are implementing Wayland and OS X required for this merge request? If so, I can take a stab at Wayland next.

Not required to be in this PR.

If Wayland is easy enough to do then I'd appreciate you updating it. I do want to note that Wayland support is currently being rewritten in https://github.com/wez/wezterm/pull/4777 so if rounding out this feature somehow ends up to be a lot of work, I'd be fine with deferring it.

I also won't require that you make this work on macOS, but if you do have access to it, again, I would appreciate your attention there!

jknockel:
I've added wayland and macos support.  I can't of course test whether macos works, but do you know if there is a way I can at least type check it from linux? Right now my strategy is to push changes and hope the automatic build doesn't fail... :sweat_smile:

wez:
FWIW, it's OK to keep throwing things at this PR if you want to iterate.

```
 error[E0599]: no method named `contentMinSize_` found for struct `objc::rc::StrongPtr` in the current scope
    --> window/src/os/macos/window.rs:1266:18
     |
1265 | /             self.window
1266 | |                 .contentMinSize_(NSSize::new(min_width.into(), min_height.into()));
     | |                 -^^^^^^^^^^^^^^^ method not found in `StrongPtr`
     | |_________________|
     | 
```

for this sort of error: those methods are sometimes added through traits in some of the objc ecosystem.
The ones you want to use are not always present in the ecosystem, but you can use a lower-level objc call like this to call a method by name:

https://github.com/wez/wezterm/blob/46732d049a162c81e173229006a9b47c262e6114/window/src/os/macos/connection.rs#L40

so for the code in this PR, something like this may work:

```rust
     let () = msg_send![*self.window, setContentMinSize: NSSize::new(min_width.into(), min_height.into())]; 
```

jknockel:
> so for the code in this PR, something like this may work:
> 
> ```rust
>      let () = msg_send![*self.window, setContentMinSize: NSSize::new(min_width.into(), min_height.into())]; 
> ```

Let's give it a shot! :crossed_fingers:

wez:
Testing it on X11 and macOS; it works great, thank you!

PR #4839, opened 2024-01-22
fix: fixed palette selected row position on move down

metiftikci:
Reproduce:

- Open command palette (ctrl + shift + p)
- Move down 2 pages
- Move up half page
- Move down
- See selected row goes to bottom


wez:
Thank you!

PR #4843, opened 2024-01-23
Add additional outputs for nix's flake package

davidsierradz:
Based on https://github.com/wez/wezterm/pull/3547#issuecomment-1890599432 and https://github.com/wez/wezterm/pull/3547#issuecomment-1890790535, I just added the additional outputs to the default package (terminfo, icon, desktop file, shell scripts), copied from https://github.com/NixOS/nixpkgs/blob/3dc440faeee9e889fe2d1b4d25ad0f430d449356/pkgs/applications/terminal-emulators/wezterm/default.nix

### DIFF
--- a/nix/flake.nix
+++ b/nix/flake.nix
@@ -125,6 +125,34 @@
             --add-needed "${pkgs.vulkan-loader}/lib/libvulkan.so.1" \
             $out/bin/wezterm-gui
         '';
+
+        postInstall = ''
+          mkdir -p $out/nix-support
+          echo "${passthru.terminfo}" >> $out/nix-support/propagated-user-env-packages
+
+          install -Dm644 assets/icon/terminal.png $out/share/icons/hicolor/128x128/apps/org.wezfurlong.wezterm.png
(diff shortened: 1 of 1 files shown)


happenslol:
Looks fine to me, since it's copied directly from the official nix package there shouldn't be any issues. I'd still like to test this however, which I'll get to by tuesday I hope.

happenslol (approved):
I've tested this and everything works fine, thanks!

nix/flake.nix
```diff
@@ -125,6 +125,34 @@
...
+        passthru = {
+          terminfo =
+            pkgs.runCommand "wezterm-terminfo"
+            {
+              nativeBuildInputs = [pkgs.ncurses];
+            } ''
+              mkdir -p $out/share/terminfo $out/nix-support
+              tic -x -o $out/share/terminfo ${src}/termwiz/data/wezterm.terminfo
+            '';
+        };
```
SuperSandro2000, line 155: I definitely wrote that 😂 hit me up if you need help with the nix part

PR #4846, opened 2024-01-24
macos/clipboard: use try_quote instead of quote

moni-dz:
[`shlex::quote`](https://github.com/comex/rust-shlex/security/advisories/GHSA-r7qv-8r2h-pg27) is deprecated, use `try_quote` to remove the warning in macOS


wez:
Thanks!

PR #4847, opened 2024-01-24
fix kitty image protocol display parameters

jonboh:
Hi! I've made some changes into the way that images get displayed to properly handle the [kitty image protocol](https://sw.kovidgoyal.net/kitty/graphics-protocol). The changes should not affect the iterm2 and sixel implementations.

I've also added in `test-data/kitty-png.py` some functions with test cases (intended to be checked manually) that should cover all the changes. To run each of the test just uncomment the one to be run and comment the rest.

Here is a list of changes:
- Fix x,y,w,h,c,r parameter behaviour
- Fix cell padding (X, Y). It must be lower than the cell width/height (previously it could pad without limit). Now for bigger input the padding is limited.
- images with id=0 handling 

Here are some screenshots comparing kitty, the current wezterm `master` branch, and this pr branch.

### Kitty
image

### Wezterm 05eadb7cb
image

### This PR
image

### Kitty
image

### Wezterm 05eadb7cb
image

### This PR
image

For these I've used this picture, although any image should generate similar results.
ferris

This PR should solve most of the issues described in https://github.com/wez/wezterm/issues/986#issuecomment-1120622597

I'll be happy to make any modification if needed :)


wez:
Thanks for looking into this!
Could you also run a couple of tests to confirm that sixel and iterm2 image protocol handling isn't impacted by this change?
https://wezfurlong.org/wezterm/cli/imgcat.html is the easiest way to play around with the latter.
The former is a bit more ad-hoc.

The thing that caught my eye, although I haven't read your diff in detail yet, is that there were some changes around how the cursor position is handled, so it would be good to confirm that `wezterm imgcat --no-move-cursor` still functions correctly.

jonboh:
Here's what I get running `img2sixel` and `wezterm imgcat`
I had checked `img2sixel`, and `wezterm imgcat`, but not with `--no-move-cursor` thanks for pointing it out, it seems the behaviour on all three cases is the same 👍.

## Wezterm 05eadb7cb
image

## kitty-image branch (this pr)
image

term/src/terminalstate/image.rs
```diff
@@ -115,6 +131,15 @@ impl TerminalState {
             .context("computing xpos")?;
 
         let cursor_x = self.cursor.x;
+
+        let width_in_cells = fullcells_width + (remainder_width_cell > 0) as usize;
+        let height_in_cells = fullcells_height + (remainder_height_cell > 0) as usize;
```
wez, line 136: stylistically, I think these conversions from bool to integer are a bit hard to follow; they are certainly succinct, but it's not super clear what is happening, or perhaps more important, whether that what happens is intentional.

I propose adding a little helper function, something like:

```rust
use num_traits::{One, Zero};

/// Returns `1` if `b` is true, else `0`,
fn one_or_zero<T: Zero+One>(b: bool) -> T{
    if b {
        T::one()
    } else {
        T::zero()
    }
}
```

then the logic here could be:

```suggestion
        let width_in_cells = fullcells_width + one_or_zero(remainder_width_cell > 0);
        let height_in_cells = fullcells_height + one_or_zero(remainder_height_cell > 0);
```

term/src/terminalstate/image.rs
```diff
@@ -202,6 +221,11 @@ impl TerminalState {
             }
         }
 
+        // adjust cursor position if the drawn cells move beyond current cell
+        let x_padding_shift = (draw_width as usize + cell_padding_left as usize
+            > cell_pixel_width * width_in_cells) as i64;
+        let y_padding_shift = (draw_height as usize + cell_padding_top as usize
+            > cell_pixel_height * height_in_cells) as i64;
```
wez, line 228: I think these are candidates for `one_or_zero` as well:

```suggestion
        let x_padding_shift = one_or_zero(draw_width as usize + cell_padding_left as usize
            > cell_pixel_width * width_in_cells);
        let y_padding_shift = one_or_zero(draw_height as usize + cell_padding_top as usize
            > cell_pixel_height * height_in_cells);
```

wez:
If you wouldn't mind, could you run through these issues and confirm whether this PR addresses them? Thanks!

* https://github.com/wez/wezterm/issues/1663
* https://github.com/wez/wezterm/issues/1156 (I think this is the id 0 thing you're fixing here?)
* https://github.com/wez/wezterm/issues/2084 this one is definitely related to id 0
* https://github.com/wez/wezterm/issues/4233
* https://github.com/wez/wezterm/issues/3918
* https://github.com/wez/wezterm/issues/2422

wez:
So this PR will close all of those issues? Excellent work, thank you!

jonboh:
Yeah, #2761  can also be closed, as its changes are already integrated (the part about the id=0)
thank you for reviewing it and wezterm as a whole! :D

PR #4848, opened 2024-01-24
chore: add signal getter on exit status

chris-olszewski:
Hi Wez, thanks for the great pty library.

Adding a getter for the signal on exit status.

I have a need to differentiate between processes that exit with a code of 1 and those that are killed by a signal. I'm currently using `exit_status.to_string().contains("Terminated by")` to determine if an exit status has a signal associated with it.


wez:
Seems reasonable; thank you!

PR #4850, opened 2024-01-25
fix: render bar cursors on topmost layer

### LINKED ISSUE #3047: Text overlapping bar cursor
When using the bar cursor option the text for that cell overlaps with the rendering of the cursor

A config option to render the bar cursor as solid, not showing the color of the text underneath 

Thought the cursor_border color option might have helped but it didn't

**Additional context**
Screenshot from 2023-02-02 16-59-43
You'll see the leftmost part of the "w" under the cursor is overlapping with the cursor itself. Ideally I would like an option to always render the bar cursor over the top for consistency. Apologies if there is already a way to do this that I couldn't figure out.

ErrorNoInternet:
Closes #3047.

### DIFF
--- a/wezterm-gui/src/termwindow/render/screen_line.rs
+++ b/wezterm-gui/src/termwindow/render/screen_line.rs
@@ -357,7 +357,7 @@ impl crate::TermWindow {
                 };
                 let mut quad = layers
                     .allocate(cursor_layer)
-                    .context(format!("layers.allocate({cursor_layer})"))?;
+                    .with_context(||format!("layers.allocate({cursor_layer})"))?;
                 quad.set_hsv(hsv);
                 quad.set_has_color(false);
 
--- a/wezterm-gui/src/termwindow/render/screen_line.rs
+++ b/wezterm-gui/src/termwindow/render/screen_line.rs
@@ -13,6 +13,7 @@ use std::rc::Rc;
 use std::time::Instant;
 use termwiz::cell::{unicode_column_width, Blink};
 use termwiz::color::LinearRgba;
+use termwiz::surface::CursorShape;
 use wezterm_bidi::Direction;
 use wezterm_term::color::ColorAttribute;
 use wezterm_term::CellAttributes;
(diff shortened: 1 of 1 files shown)


wezterm-gui/src/termwindow/render/screen_line.rs
```diff
@@ -350,7 +351,12 @@ impl crate::TermWindow {
                 + (phys(params.cursor.x, num_cols, direction) as f32 * cell_width);
 
             if cursor_shape.is_some() {
-                let mut quad = layers.allocate(0).context("layers.allocate(0)")?;
+                let mut quad = match cursor_shape.unwrap() {
```
wez, line 354: in general, the bare `.unwrap()` is a bad code smell in Rust, especially right after a check like `.is_some()`.

A better way to phrase this block is:

```rust
if let Some(shape) = cursor_shape {
    let cursor_layer = match shape {
         // allow bar cursors to draw over the top of text
         CursorShape::BlinkingBar | CursorShape::SteadyBar => 2,
         // but all other cursors sit below it for legibility
         _ => 0,
    };
    let mut quad = layers.allocate(cursor_layer).with_context(||format!("layers.allocate({cursor_layer})"))?;
```

wez (commented):
Thanks for this.

I'm not totally sure about taking this change as-is though. It is intentional that the text sits above the cursor as it stands a better chance of being visible/legible in various situations.
I realize that that is less critical for bar cursors.
I want to noodle on this a bit before deciding.

wezterm-gui/src/termwindow/render/screen_line.rs
ErrorNoInternet: Yes, but at the bottom (line ~412) we also pass `Option<CursorShape>` to `.cursor_spite()`. I'll do `Some(shape)`.

ErrorNoInternet:
It was bothering me because I have fish autosuggestions that cover my cursor

wezterm-gui/src/termwindow/render/screen_line.rs
```diff
@@ -349,8 +350,14 @@ impl crate::TermWindow {
...
-            if cursor_shape.is_some() {
-                let mut quad = layers.allocate(0).context("layers.allocate(0)")?;
+            if let Some(shape) = cursor_shape {
+                let cursor_layer = match shape {
+                    CursorShape::BlinkingBar | CursorShape::SteadyBar => 2,
+                    _ => 0,
+                };
+                let mut quad = layers
+                    .allocate(cursor_layer)
+                    .context(format!("layers.allocate({cursor_layer})"))?;
```
wez, line 360: don't pay the cost of making the string unless this fails

```suggestion
                    .with_context(||format!("layers.allocate({cursor_layer})"))?;
```

wez:
Thanks!

PR #4862, opened 2024-01-26
fix: disproportionate cursor

hyperpuncher:
Fixes disproportionate cursor on 4K resolution at any font size and on 1440p with font sizes bigger than 22.

Before:
3 1-linux
After:
3 01-linux


jankatins:
Are you sure you added different images? At least I cannot see a difference between them just by looking at them.

hyperpuncher:
Yep, on the first image lines are thicker on the top and bottom.

jankatins:
Yes, you are right, now I also see it, thanks! It's amazing what kind of attention to details goes into this!

wez:
Thanks!

PR #4866, opened 2024-01-26
wayland: consider base in resize increment calculation

jknockel:
This corrects the calculation so that the # of terminal cells will not change on a scale change.


wez:
Thanks!

PR #4875, opened 2024-01-27
WIP fix for i3 terminals sending mouse press before focus event

tpoliaw:
I don't expect this is the correct place for this fix but the thinking
is that the window only gets the mouse press event if the mouse is over
the terminal window. In this case it makes sense (to me?) that if the
window doesn't have focus, the click should still be passed to the
application.

### DIFF
--- a/wezterm-gui/src/termwindow/mouseevent.rs
+++ b/wezterm-gui/src/termwindow/mouseevent.rs
@@ -755,7 +755,7 @@ impl super::TermWindow {
         }
 
         let allow_action = if self.is_click_to_focus_window || !is_focused {
-            matches!(&event.kind, WMEK::VertWheel(_) | WMEK::HorzWheel(_))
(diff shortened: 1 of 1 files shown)


wez:
The original bug report that started this was:
* https://github.com/wez/wezterm/issues/1140
the crux of the issue was that neovim would get confused if it received focus and click events out of order.
The resolution was to ignore them when they didn't make sense.
I think this PR will effectively regress the fix for that original issue.

I think that if you want to send them anyway, which is probably reasonable, that it may make sense to implicitly generate a focus event when we get a click and we are not already focused, so that neovim users remain happy.

tpoliaw:
I can't reproduce the neovim issue with this in place (maybe neovim fixed it on their end?) but I take your point. Is receiving a focusIn event when the window already has focus going to cause problems and/or, would it be better to emit a Press(Left) event if a release is received before the equivalent press?

PR #4888, opened 2024-01-28
Set default cargo target to wezterm-gui

mlwarner:
I was following the steps in the CONTRIBUTING file and I was prompted to specify a bin when I used `cargo run`. This change makes `cargo run` choose the `wezterm-gui` package if nothing is specified.

https://doc.rust-lang.org/cargo/reference/workspaces.html#the-default-members-field

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,7 @@ members = [
   "wezterm-open-url",
   "wezterm-ssh"
 ]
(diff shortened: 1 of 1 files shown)


wez:
Thanks, but I don't really want the gui to be the default. I'd rather that we fixup the contributing docs to match the instructions shown on https://wezfurlong.org/wezterm/install/source.html

PR #4891, opened 2024-01-28
fix typo in installation example

vo1dwoken:
The installation example mentions the Ubuntu 20 package, while the provided example is for the Ubuntu 22 package


wez:
Thanks!

PR #4898, opened 2024-01-29
Upgrade rcgen to 0.12

hack3ric:
This brings ring version from 0.16 to 0.17, which enables build on platforms like RISC-V and LoongArch.


wez:
Thanks!

PR #4906, opened 2024-01-30
fix(wayland)!: hardcode client-side decorations; hardcode scale=1.0

jokeyrhyme:
- this PR addresses some panics in https://github.com/wez/wezterm/pull/4777#issuecomment-1913452992 by hardcoding client-side decorations (instead of trying to negotiate for server-side decorations) and hardcoding scale=1.0 (instead of using detected monitor/surface values)
- this does mean user preferences are not honoured, however, the terminal is in a mostly usable state

future PRs:
- restore/improve visual experience of CSD title-bar per https://github.com/wez/wezterm/pull/4777#issuecomment-1915849032
- use detected output/surface scale instead of hardcoding to 1.0
- other ideas from: https://github.com/wez/wezterm/pull/4777#issuecomment-1913394104
- hidden/absent CSD when SSD is offered by compositor
- update to smithay-client-toolkit 0.18 (let's do this after we've got everything back to normal again)

### DIFF
--- a/window/src/os/wayland/window.rs
+++ b/window/src/os/wayland/window.rs
@@ -1,5 +1,6 @@
 use std::any::Any;
 use std::cell::{RefCell, RefMut};
+use std::cmp::max;
 use std::convert::TryInto;
 use std::io::Read;
 use std::num::NonZeroU32;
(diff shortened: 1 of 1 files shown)


wez:
FWIW, I had this cherry-picked and rebased on top of my rebase of #4777 (which is now in `main`) and it would exit immediately after launch, so there is something not quite right here.  My test is simple:

* Run `wayfire` in a window of its own
* `WAYLAND_DISPLAY=wayland-1 ./target/debug/wezterm`

that works in `main` but exits with this PR.

My display is hidpi, in case that is relevant

jokeyrhyme:
Huh, my display is scale=1.25 here, let me see...

window/src/os/wayland/window.rs
```diff
@@ -782,14 +761,22 @@ impl WaylandWindowInner {
...
-                        let desired_pixel_height = pixel_height - (pixel_height % y as i32);
+                    self.window_frame.set_resizable(true);
+                    if let Some(incr) = self.resize_increments {
+                        let min_width = incr.base_width + incr.x;
+                        let min_height = incr.base_height + incr.y;
+                        let extra_width = (pixel_width - incr.base_width as i32) % incr.x as i32;
+                        let extra_height = (pixel_height - incr.base_height as i32) % incr.y as i32;
+                        let desired_pixel_width = max(pixel_width - extra_width, min_width as i32);
+                        let desired_pixel_height =
+                            max(pixel_height - extra_height, min_height as i32);
```
jokeyrhyme, line 772: It seems these changes from https://github.com/wez/wezterm/pull/4866 were lost

jokeyrhyme:
Bah, I pressed "Resolve conflicts" in GitHub and it did a `git merge`

And there's still conflicts :S

wez:
I'm going to close this because wayland got basically rewritten since this PR was created.
Thank you!

PR #4924, opened 2024-02-01
Add a CloseWithoutClear copy mode key assignment

LeszekSwirski:
Closing the copy overlay currently unconditionally clears the viewport,
in particular resetting scroll. For the search overlay, we don't
necessarily want to scroll back to the prompt after finding a match --
indeed, the old search overlay (which didn't use copy mode) had this
behaviour.

Add a CloseWithoutClear key assignment which has this desired behaviour,
and make it the default for ESC in search mode.

### DIFF
--- a/config/src/keyassignment.rs
+++ b/config/src/keyassignment.rs
@@ -673,7 +673,6 @@ pub enum CopyModeAssignment {
     PageUp,
     PageDown,
     Close,
-    CloseWithoutClear,
     PriorMatch,
     NextMatch,
     PriorMatchPage,
@@ -690,6 +689,7 @@ pub enum CopyModeAssignment {
     JumpBackward { prev_char: bool },
     JumpAgain,
     JumpReverse,
+    ResetViewport,
 }
 
 pub type KeyTable = HashMap<(KeyCode, Modifiers), KeyTableEntry>;
--- a/config/src/keyassignment.rs
+++ b/config/src/keyassignment.rs
@@ -673,6 +673,7 @@ pub enum CopyModeAssignment {
     PageUp,
     PageDown,
     Close,
+    CloseWithoutClear,
     PriorMatch,
     NextMatch,
     PriorMatchPage,
--- a/config/src/keyassignment.rs
+++ b/config/src/keyassignment.rs
@@ -689,7 +689,6 @@ pub enum CopyModeAssignment {
     JumpBackward { prev_char: bool },
     JumpAgain,
     JumpReverse,
-    ResetViewport,
 }
 
 pub type KeyTable = HashMap<(KeyCode, Modifiers), KeyTableEntry>;
--- a/wezterm-gui/src/overlay/copy.rs
+++ b/wezterm-gui/src/overlay/copy.rs
@@ -1206,7 +1206,6 @@ impl Pane for CopyOverlay {
                     JumpBackward { prev_char } => render.jump(false, *prev_char),
                     JumpAgain => render.jump_again(false),
                     JumpReverse => render.jump_again(true),
-                    ResetViewport => render.set_viewport(None),
                 }
                 PerformAssignmentResult::Handled
             }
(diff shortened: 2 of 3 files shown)


config/src/keyassignment.rs
```diff
@@ -673,6 +673,7 @@ pub enum CopyModeAssignment {
     PageUp,
     PageDown,
     Close,
+    CloseWithoutClear,
```
wez, line 676: I don't think `Clear` is the correct way to describe what happens.

How about we tweak this PR:

* `Close` - closes without changing the viewport. This is a break in backwards compatibility for folks that are using this assignment in their configs, but I think there are relatively few people that are doing that so the impact should be small
* Change the `copy_key_table` function (https://github.com/wez/wezterm/blob/6e85ba49fdd0a550d1806289f8410b8b7047bf48/wezterm-gui/src/overlay/copy.rs#L1577)  to use something like the following instead of `Close`:

```rust
fn close_and_scroll_to_bottom() -> KeyAssignment {
  KeyAssignment::Multiple(vec![
     KeyAssignment::CopyMode(CopyModeAssignment::Close),
     KeyAssignment::ScrollToBottom
  ]) 
}

...

(
    WKeyCode::Char('\x1b'),
    Modifiers::NONE,
    close_and_scroll_to_bottom(),
),
```

this way it should be easier to reason about and compose key assignment actions, without having to pre-define actions for every combination.

wez (commented):
Thanks for diving in!

config/src/keyassignment.rs
LeszekSwirski: I used "Clear" since the operation being skipped is "ClearViewport" -- maybe "ResetViewport" is more accurate vs. "ScrollToBottom"

Splitting up into two composed key assignment actions sounds reasonable; the reason I went with keeping "Close" unchanged is to not break people's existing configs in case they remapped it, but I have no strong opinion here.

LeszekSwirski: Done (as ResetViewport, and keybindings without a helper function for consistency with surrounding code).

wezterm-gui/src/overlay/copy.rs
```diff
@@ -1580,22 +1580,34 @@ pub fn copy_key_table() -> KeyTable {
         (
             WKeyCode::Char('c'),
             Modifiers::CTRL,
-            KeyAssignment::CopyMode(CopyModeAssignment::Close),
+            KeyAssignment::Multiple(vec![
+                KeyAssignment::CopyMode(CopyModeAssignment::Close),
+                KeyAssignment::CopyMode(CopyModeAssignment::ResetViewport),
```
LeszekSwirski, line 1585: Maybe these should actually be the other way around, to preserve the old behaviour? Not sure.

wezterm-gui/src/overlay/copy.rs
```diff
@@ -1580,22 +1580,34 @@ pub fn copy_key_table() -> KeyTable {
         (
             WKeyCode::Char('c'),
             Modifiers::CTRL,
-            KeyAssignment::CopyMode(CopyModeAssignment::Close),
+            KeyAssignment::Multiple(vec![
```
wez, line 1583: please factor this using a function as I suggested previously

LeszekSwirski: I kept this intentionally outside of a helper because a) the documented default config [has inlined Multiples](https://github.com/wez/wezterm/pull/4924/files#diff-aabffc061ef2a12a200c36153cd0551622f2bf61aab2967cdcc6bf11ebc546bbR20), so it felt nice to be consistent, and b) there's one case where we already call `Close` in a `Multiple` ([along with a copy to clipboard](https://github.com/wez/wezterm/pull/4924/files#diff-cefae38b87fd0f76c51a4bc47e630e5bd3a3e4a84821e15c2b9c919f337169edR1861)), so using a helper there would end up with nested `Multiple`s and that felt icky.

I can still add a helper if you prefer, but maybe then I should also add a helper for lua configs? Are nested `Multiple`s ok?

wez: I want the rust code to be concise so that it is easier to maintain: I have to live with this code forever. The documented defaults are informational and derived by automation and do not need to have a helper function.

Nested Multiples should be fine.

wezterm-gui/src/overlay/copy.rs
LeszekSwirski: Swapped them around.

config/src/keyassignment.rs
```diff
@@ -689,6 +689,7 @@ pub enum CopyModeAssignment {
     JumpBackward { prev_char: bool },
     JumpAgain,
     JumpReverse,
+    ResetViewport,
```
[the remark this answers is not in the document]
LeszekSwirski: Good point, thanks -- I missed that.

LeszekSwirski:
Added a helper and dropped ResetViewport, PTAL (and let me know if you prefer me to squash the individual commits)

LeszekSwirski (commented):
Submitting pending comments

wezterm-gui/src/overlay/copy.rs
LeszekSwirski: Fair enough, changed.

LeszekSwirski:
Ping

LeszekSwirski:
Ping again (@LeszekSwirski)

LeszekSwirski:
@wez anything more to do here?

LeszekSwirski:
Regular ping for review, @wez

wez:
Thank you!

Issue #4928, opened 2024-02-01
Termwiz: support Kitty keyboard enhancements


yyogo:
Currently termwiz supports parsing [Kitty protocol keyboard events](https://docs.rs/termwiz/0.22.0/termwiz/input/enum.KeyboardEncoding.html#variant.Kitty), but it has no way to tell the terminal to enable/disable them, or to query if they are supported.

**Describe the solution you'd like**
The ideal solution would be to automatically and transparently enable Kitty enhancements whenever they are supported. The user should also be able to check which enhancements are enabled, and to enable/disable them manually as needed.

**Describe alternatives you've considered**
1. Allow users to extend Terminal's capabilities, by allowing raw read/write
2. Implement `DeviceAttributes` querying but leave the feature-specific logic to the user, and expose a mechanism for writing CSs directly to the terminal (i.e something like `Terminal::write_control_sequence(csi: CSI)`). 

**Additional context**
- https://sw.kovidgoyal.net/kitty/keyboard-protocol/

wez:
I think expanding the `ProbeHints` and `Capabilities` to make them aware of kitty keyboard is reasonable.
The implementation could be sort of similar to the existing mouse reporting stuff.

I don't have time to work on this, but I'd be willing to review a PR for it.

yaymukund:
Hello, I notice that #4951 is over a year old. Is this still being worked on? From that PR:

>  The question of where to put the actual probe is made complicated here because this is the first feature to require active probing.

Is there more clarity on how this should be designed? I'd be happy to give it a shot.

PR #4951, opened 2024-02-04
termwiz:  Kitty keyboard enhancements support

yyogo:
`Terminal` now enables [enhanced keyboard reporting](https://sw.kovidgoyal.net/kitty/keyboard-protocol/) (on `set_raw_mode()`) if it is supported and enabled in capabilities. It also pops the state on `drop`.

See #4928 .

* Add `keyboard_enhancement` field to `ProbeHints` and `Capabilities`.
* Add `enhanced_keyboard_state` method to `ProbeCapabilities`.

Some questions / considerations before merging: 
- Should we keep it enabled by default? Can it break existing apps?
- Should `Capabilities` be responsible for probing instead of `Terminal`?

### DIFF
--- a/termwiz/examples/key_tester.rs
+++ b/termwiz/examples/key_tester.rs
@@ -1,4 +1,4 @@
-use termwiz::caps::Capabilities;
+use termwiz::caps::{Capabilities, ProbeHints};
 use termwiz::input::{InputEvent, KeyCode, KeyEvent, Modifiers};
 use termwiz::terminal::{new_terminal, Terminal};
 use termwiz::Error;
@@ -9,7 +9,8 @@ const CTRL_C: KeyEvent = KeyEvent {
 };
 
 fn main() -> Result<(), Error> {
-    let caps = Capabilities::new_from_env()?;
+    let caps =
+        Capabilities::new_with_hints(ProbeHints::new_from_env().probe_for_enhanced_keyboard(true))?;
     let mut terminal = new_terminal(caps)?;
     terminal.set_raw_mode()?;
 
--- a/termwiz/src/caps/mod.rs
+++ b/termwiz/src/caps/mod.rs
@@ -116,6 +116,11 @@ builder! {
         /// invisible and reset, and directly emit those sequences.
         /// This can improve rendered text compatibility with pagers.
         force_terminfo_render_to_use_ansi_sgr: Option<bool>,
+
+        /// Whether Kitty enhanced keyboard protocol should be used
+        /// Note: termcap does not report this capability so it is assumed
+        /// to be supported by `Capabilities` and enabled if possible by `Terminal`.
+        keyboard_enhancement: Option<bool>,
     }
 }
 
@@ -164,6 +169,7 @@ pub struct Capabilities {
     bracketed_paste: bool,
     mouse_reporting: bool,
     force_terminfo_render_to_use_ansi_sgr: bool,
+    keyboard_enhancement: bool,
 }
 
 impl Capabilities {
@@ -293,6 +299,7 @@ impl Capabilities {
 
         let bracketed_paste = hints.bracketed_paste.unwrap_or(true);
         let mouse_reporting = hints.mouse_reporting.unwrap_or(true);
+        let keyboard_enhancement = hints.keyboard_enhancement.unwrap_or(true);
 
         let force_terminfo_render_to_use_ansi_sgr =
             hints.force_terminfo_render_to_use_ansi_sgr.unwrap_or(false);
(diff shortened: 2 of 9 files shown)


wez (commented):
Thanks for this!
The question of where to put the actual probe is made complicated here because this is the first feature to require active probing.

The intent of `Capabilities` is that it holds the pre-probed information about the terminal capabilities. So far all of that information is either provided explicitly or derived from environment variables.

To add active probing is a little awkward because we need to instantiate a Terminal in order to use `ProbeCapabilities`, and to do that we need to have already created `Capabilities`.

Here's what I suggest:

Add a new method to the `Terminal` trait: `set_keyboard_encoding(&self, encoding: KeyboardEncoding)`. (see: https://docs.rs/termwiz/latest/termwiz/input/enum.KeyboardEncoding.html ).  That method will be responsible for applying the specified keyboard encoding mode.  The implementation should remember which mode it enabled, so that it can revert back to basics when Drop'd.  Keyboard encoding is independent of cooked/raw mode.

Then `Capabilities` could be expanded with an `probe_for_enhanced_keyboard: Option<bool>` that can be opted into.  When it is enabled, the `new_terminal` implementation can include that probing logic:

```rust
pub fn new_terminal(caps: Capabilities) -> Result<impl Terminal> {
    let probe_keyboard = caps.probe_for_enhanced_keyboard();
    let mut term = SystemTerminal::new(caps)?;

   if probe_keyboard {   
      let mut probe = term
         .probe_capabilities()
         .ok_or_else(|| anyhow!("Terminal has no prober?"))?;

      let keyboard = match probe.kitty_keyboard_protocol() {
         Ok(Some(flags)) => KeyboardEncoding::Kitty(flags),
         _ => KeyboardEncoding::Xterm
     };
     term.set_keyboard_encoding(keyboard)?;
   }

   Ok(term)
}
```

I don't want to make the probing the default behavior at this time, as there are situations where probing is undesirable; for example, if you are running in a terminal that is broken in the face of the probing escape sequences, or if you are on a very slow link, or if you want to keep things intentionally very basic.

That means that the embedding application will likely do something like this to use kitty keyboard:

```rust
let hints = ProbeHints::new_from_env().probe_for_enhanced_keyboard(true);
let caps = Capabilities::new_with_hints(hints)?;
let mut term = termwiz::terminal::new_terminal(caps)?;
```

yyogo:
Thanks for the review. The reason I put the probing in `set_raw_mode` instead of `new_terminal` is that I figured it needs to be in raw mode to work correctly. How about setting raw in  `ProbeCapabilities::new` and restoring to whatever it was before when it's dropped? Maybe via a helper object? (I kind of like the idea of refactoring Terminal so that all mode settings use RAII / drop semantics, but I think that is beyond the scope of this PR)

Fair enough as for the default. In my limited testing, it seems like some keys aren't being parsed correctly under some terminals (even in Kitty strangely enough). But that's a story for another PR.

wez:
the keyboard encoding mode is not at all connected to raw/cooked modes; those are aspects of termios handled by the kernel and have nothing to do with how the terminal emulator encodes its input.

yyogo:
@wez I understand that, however for the probing to work the terminal has to be in raw mode, so that the CSI queries aren't echoed and / or processed incorrectly.

wez:
hmm, I didn't think it is strictly necessary, but sure, I'm ok with enabling it and restoring the original mode when dropped. I think the nuance there is: the probe may be constructed with the terminal in either mode, so we shouldn't unconditionally revert to cooked mode on drop, but rather restore the mode that was in place before we set it to raw.

FWIW, here's where the probe stuff is used currently in `wezterm imgcat`:
https://github.com/wez/wezterm/blob/22150bfba69031ae6e09de430aa2370d06290448/wezterm/src/main.rs#L502-L528

(and looking at that, I can see that it puts the terminal into raw mode, so it looks like I'd forgotten about that!)

yyogo:
@wez I'd appreciate your thoughts on the current implementation:

- Added a `TermiosGuard` wrapper for the TTY handler that restores termios state when dropped, and changed unix terminal to use it
  - The guard is passed to `ProbeCapabilities` to set and restores raw mode (this introduces a breaking change since `ProbeCapabilities` now has a nontrivial drop so it must be dropped before using the terminal again)
- Add the `probe_for_enhanced_keyboard` probe hint and the `enhanced_keyboard_state` probe cap. method
- Add `set_keyboard_encoding()` to `Terminal`
- Add `get_keyboard_encoding()` and `keyboard_enhancement_supported()` to unix terminal
- Set keyboard encoding in `set_raw_mode` if the hint is enabled and the terminal supports it
- Restore keyboard encoding on drop / cooked mode

One minor pain point is the `modifyOtherKeys` thing, I'm having trouble including it in the current `keyboard_encoding` model, I don't think it's critical though.

PR #4955, opened 2024-02-04
windows: account for dpi in window size

jknockel:
On Windows, to set a window's size, in addition to the size of the client area, the width and height of the window's "frame" must be included in the dimensions passed to SetWindowPos() which burdens us in needing to know the exact size of the window frame so that we can properly account for its dimensions.  Previously, we used AdjustWindowRect() to account for the frame's dimensions, but the size of a window's frame can change depending on the DPI of the monitor on which it is placed, and it appears that neither AdjustWindowRect() nor AdjustWindowRectEx() account for this automatically.  Instead, we use AdjustWindowRectExForDpi(), passing in the window's DPI, so that we properly calculate the window's size.

### DIFF
--- a/window/src/os/windows/window.rs
+++ b/window/src/os/windows/window.rs
@@ -38,6 +38,7 @@ use winapi::shared::winerror::S_OK;
 use winapi::um::imm::*;
 use winapi::um::libloaderapi::GetModuleHandleW;
 use winapi::um::shellapi::{DragAcceptFiles, DragFinish, DragQueryFileW, HDROP};
+use winapi::um::shellscalingapi::{GetDpiForMonitor, MDT_EFFECTIVE_DPI};
 use winapi::um::sysinfoapi::{GetTickCount, GetVersionExW};
 use winapi::um::uxtheme::{
     CloseThemeData, GetThemeFont, GetThemeSysFont, OpenThemeData, SetWindowTheme,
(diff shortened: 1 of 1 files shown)


window/src/os/windows/window.rs
```diff
@@ -390,6 +396,15 @@ fn decorations_to_style(decorations: WindowDecorations) -> u32 {
     }
 }
 
+unsafe fn get_primary_monitor_dpi() -> u32 {
+    let zero_point = POINT { x: 0, y: 0 };
+    let primary = MonitorFromPoint(zero_point, MONITOR_DEFAULTTOPRIMARY);
```
wez, line 401: I think switching this to `MonitorFromWindow(std::ptr::null_mut(), MONITOR_DEFAULTTOPRIMARY)` would be more consistent with the logic in `create_window` and remove issues with `0,0` not being on the primary monitor in some unusual user-configurations

window/src/os/windows/window.rs
```diff
@@ -390,6 +396,15 @@ fn decorations_to_style(decorations: WindowDecorations) -> u32 {
     }
 }
 
+unsafe fn get_primary_monitor_dpi() -> u32 {
```
wez, line 399: I'd just remove the `unsafe` from here; you can make it safe by asserting that the resolved `primary` monitor is not null.

wez (commented):
Thanks! A couple of minor comments, but this looks good!

jknockel:
I believe I've addressed your feedback.

wez:
Looks good, thank you!

PR #4970, opened 2024-02-06
Better decorations in wayland

tmccombs:
This is an attempt to improve window decorations on wayland. It waits to create the window decorations  until we get a configure event from the compositor, so that we can know whether the compositor wants us to use CSD or not.


wez:
How do you feel about also rebasing/revising this one? I think it will help with a couple of outstanding issues!

window/src/os/wayland/window.rs
```diff
@@ -1312,6 +1357,14 @@ impl CompositorHandler for WaylandState {
...
         _new_factor: i32,
     ) {
+        // TODO: uncomment after upgrading SCTK to 0.18
+        // let window_id = SurfaceUserData::from_wl(surface).window_id;
+        // WaylandConnection::with_window_inner(window_id, |inner| {
+        //     if let Some(frame) = inner.window_frame.as_mut() {
+        //         frame.set_scaling_factor(new_factor as f64);
+        //     }
+        //     Ok(())
+        // });
```
wez, line 1367: We're on SCTK 0.19 now, so should we go ahead and do this?

```suggestion
        let window_id = SurfaceUserData::from_wl(surface).window_id;
        WaylandConnection::with_window_inner(window_id, |inner| {
            if let Some(frame) = inner.window_frame.as_mut() {
                frame.set_scaling_factor(new_factor as f64);
            }
            Ok(())
        });
```

window/src/os/wayland/frame.rs
```diff
@@ -2,6 +2,8 @@
 //! in smithay_client_toolkit 0.11 which is Copyright (c) 2018 Victor Berger
 //! and provided under the terms of the MIT license.
 
+// TODO: update this for SCTK 0.18 and use this instead of the FallbackFrame
```
wez, line 5: What do we need to do here now that we're on SCTK 0.19? Does it need to be part of this PR?

wez (commented):
Thank you!

FWIW, under KDE Plasma I experience this with this pr; the window briefly appears and then we immediately terminate:

```console
$ ./target/debug/wezterm -n --config 'window_decorations="RESIZE"'
Io error: Broken pipe (os error 32)
07:24:04.792  ERROR  wezterm_gui > running message loop: Io error: Broken pipe (os error 32); terminating
```

window/src/os/wayland/frame.rs
tmccombs: No, it doesn't need to be, and probably shouldn't be part of this PR. 

Basically it would need to be re-written to implement the DecorationsFrame trait.

Alternatively, just delete this and use the [adwaita implementation](https://github.com/PolyMeilex/sctk-adwaita) instead.

window/src/os/wayland/window.rs
tmccombs: yep, done

window/src/os/wayland/window.rs
```diff
@@ -838,25 +872,33 @@ impl WaylandWindowInner {
...
+                let window = self.window.as_ref().unwrap();
+
+                match self.decorations.as_mut() {
+                    Some(frame) if !frame.is_hidden() => {
+                        log::trace!("Resizing frame");
+                        frame.resize(w.try_into().unwrap(), h.try_into().unwrap());
+                        let outer_size = frame.add_borders(w, h);
+                        let (x, y) = frame.location();
+                        window.set_window_geometry(
+                            x.try_into().unwrap(),
```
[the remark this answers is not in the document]
tmccombs: hmm, so the x and y values are actually expected to be negative. 

 As far as I understand, the way this works is the origin (0,0) is the top left of the _inner_ surface. But you set the window geometry to have a negative x and y, to tell the compositor that the outer window should have space for the decorations. 
 
 Looking at examples of using the wayland_csd_decorations crate, it looks like it can be worked around by using `window.xdg_surface().set_window_geometry` instead of the `set_window_geometry` on the sctk XdgSurface. 
 
 I just pushed a fix for that panic. I might get a chance to test it more this weekend.

tmccombs:
Well. On Sway, I don't get the pipe error, but it is extremely buggy.  Resizing doesn't seem to work at all. 

Possibly I messed something when resolving conflicts. 

I'm not sure when I'll have the time to dig deeper into this. I wouldn't object if someone else wanted to pick this up, otherwise, I'll put this on my todo list to fix.

valpackett:
On niri (smithay) I actually get the following with this PR:

```
wl_shm_pool#26: error 1: invalid width or height (0x24)
Protocol error 1 on object wl_shm_pool@26:
16:19:11.347  ERROR  wezterm_gui > running message loop: Protocol error 1 on object wl_shm_pool@26: ; terminating
```

I'll look into sctk-adwaita.

HearseDev:
@tmccombs check out #7095

tmccombs:
I'll close in favor of that

PR #4977, opened 2024-02-06
nix: init overlay

notohh:
adds a nix overlay.

tested against my machine (NixOS 24.05 x86_64-linux), compiles and runs as expected

cc: @happenslol


happenslol:
LGTM, this is not that deep honestly :stuck_out_tongue:

happenslol (approved):
LGTM!

PR #4982, opened 2024-02-07
windows: set  PSEUDOCONSOLE_INHERIT_CURSOR

### LINKED ISSUE #4784: [Windows] using portable_pty causes terminal to be cleared while trying to stream to stdout

Windows

_No response_

### WezTerm version

"0.8.1"

### Describe the bug

While trying to stream output from the pty, the current terminal's output will be cleared. I attached a video recording:

https://github.com/wez/wezterm/assets/4332460/327789b8-763d-4266-902d-3e1fb41f30a0

### To Reproduce

repro repo:
https://github.com/Cammisuli/portable-pty-windows

1. On Windows, run `cargo run`

### Configuration

na

### Expected Behavior

Being able to configure this behaviour with the underlying ConPTY instance. 

### Logs

_No response_

### Anything else?

Diving deep into Windows ConPTY (and stumbling across issues in other libraries, like this https://github.com/microsoft/node-pty/issues/423) there are mentions of using `PSEUDOCONSOLE_INHERIT_CURSOR` to configure conpty. And based on other users reactions, it appears to work for them. 

Microsoft docs https://learn.microsoft.com/en-us/windows/console/createpseudoconsole

Cammisuli:
This adds `PSEUDOCONSOLE_INHERIT_CURSOR` as a flag for `CreatePseudoConsole`

There's some remarks in the Microsoft docs([learn.microsoft.com/en-us/windows/console/createpseudoconsole](https://learn.microsoft.com/en-us/windows/console/createpseudoconsole)) about handling the cursor, and I'm not sure if we need to handle that in the actual Wezterm application somewhere. 

If we do, then I can possibly just make setting this flag optional?

Fixes #4784

### DIFF
--- a/pty/src/win/psuedocon.rs
+++ b/pty/src/win/psuedocon.rs
@@ -83,7 +83,7 @@ impl PsuedoCon {
                 size,
                 input.as_raw_handle() as _,
                 output.as_raw_handle() as _,
-                PSEUDOCONSOLE_RESIZE_QUIRK | PSEUDOCONSOLE_WIN32_INPUT_MODE,
+                1,
                 &mut con,
             )
         };
--- a/pty/src/win/psuedocon.rs
+++ b/pty/src/win/psuedocon.rs
@@ -24,6 +24,7 @@ use winapi::um::winnt::HANDLE;
 
 pub type HPCON = HANDLE;
 
+pub const PSUEDOCONSOLE_INHERIT_CURSOR: DWORD = 0x1;
 pub const PSEUDOCONSOLE_RESIZE_QUIRK: DWORD = 0x2;
 pub const PSEUDOCONSOLE_WIN32_INPUT_MODE: DWORD = 0x4;
 #[allow(dead_code)]
@@ -83,7 +84,9 @@ impl PsuedoCon {
                 size,
                 input.as_raw_handle() as _,
                 output.as_raw_handle() as _,
-                1,
+                PSUEDOCONSOLE_INHERIT_CURSOR
+                    | PSEUDOCONSOLE_RESIZE_QUIRK
+                    | PSEUDOCONSOLE_WIN32_INPUT_MODE,
                 &mut con,
             )
         };


wez:
Does this change prevent win32 input mode from being used?

Cammisuli:
From the Windows documentation, `CreatePseudoConsole` only accepted `PSEUDOCONSOLE_INHERIT_CURSOR`. 

We're using using the Conpty with the writer, and we can send messages to the process. Or is there something else that I should check?

Cammisuli:
For context, this is how we're using it: https://github.com/nrwl/nx/blob/202f14177833e021f86d646969fd2e2fee8dbb97/packages/nx/src/native/command.rs#L129-L217

Cammisuli:
Ah I see, a classic case of outdated docs 😅 

For my reference here's the flag https://github.com/microsoft/terminal/pull/6309

wez:
Could you try building wezterm (or waiting for the windows build to complete in CI and downloading wezterm from the artifacts in the CI run), and give it a quick sniff test to make sure that it seems to behave reasonably?

I think there are a couple of cases to try:

* Launch wezterm-gui from explorer
* From a command prompt window where the cursor is not in the home position, launch:
   * wezterm.exe
   * wezterm-gui.exe

and sanity check that the cursor is always consistently in the home position and doesn't actually inherit it from the spawning command prompt.

Cammisuli:
Running wezterm-gui.exe from the explorer shows up blank until I resize the window. I thought it might've been related to my changes, but I downloaded another artifact from the github runs, and it happens there as well. I'm not sure if this is expected with these dev builds because it happens with multiple artifacts. 

Here's a recording about launching wezterm from a terminal with different cursor positions:

https://github.com/wez/wezterm/assets/4332460/4658062e-e8b6-4f6a-86cd-a3af87e629e3

Cammisuli:
@wez do you have any more thoughts about this?

Cammisuli:
Please let me know if there's anything else that we need to do 🙂

wez:
OK, I think this is probably fine. Let's merge it and see if anyone finds an edge case!
Thank you!

PR #4986, opened 2024-02-09
nix: fix version string

kouyk:
Derive the version string using available information. This will work well for clones of this repository, but not pure sources (e.g. GitHub source downloads).

Additionally, don't override the version found in existing .tag to allow actual version to be specified manually. The side effect of this is that the actual version might differ from that of the derivation, but there are no trivial workarounds to set the derivation version based on existing .tag file given that the flakes are stored in a subdirectory instead of the root of the repository.

### DIFF
--- a/nix/flake.nix
+++ b/nix/flake.nix
@@ -87,13 +87,19 @@
         cargo = pkgs.rust-bin.stable.latest.minimal;
         rustc = pkgs.rust-bin.stable.latest.minimal;
       };
+
     in {
       packages.default = rustPlatform.buildRustPackage rec {
         inherit buildInputs nativeBuildInputs;
 
         name = "wezterm";
         src = ./..;
-        version = self.shortRev or "dev";
+        version = if self ? submodules
+          then builtins.concatStringsSep "-"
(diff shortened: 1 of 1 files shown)


kouyk:
cc: @happenslol, @gabyx, @edmundmiller

nix/flake.nix
```diff
@@ -113,7 +119,9 @@
         '';
 
         postPatch = ''
-          echo ${version} > .tag
+          if [ ! -f .tag ]; then
+            echo ${version} > .tag
```
wez, line 123: it's important (for my sanity) that the version number correlates with the version number that wezterm's build would have generated for the same git revision. Please share examples of what is being injected here!

kouyk: this pr aligns the version number with that of normal releases, instead of using just the first 8 characters of commit hash without date and time information. for this particular change, it continues to use whatever version is defined to (in this PR `wezterm-00000000-000000-xxxxxxxx`), unless there is an existing tag file which is useful for cases when the source tar is downloaded and the version cannot be reliably retrieved.

these few lines are simply bash, not nix.

bew:
Replaced with #7854

PR #4990, opened 2024-02-10
Add support for pane swapping in Key Assignment and Lua API.

bogdan2412:
Adds a couple of variants:
- SwapActivePaneDirection / tab:swap_active_pane_direction analogous to ActivatePaneDirection.
- SwapActivePaneByIndex / tab:swap_active_pane_by_index analogous to ActivatePaneByIndex.

### DIFF
--- a/config/src/keyassignment.rs
+++ b/config/src/keyassignment.rs
@@ -292,6 +292,20 @@ impl PaneDirection {
     }
 }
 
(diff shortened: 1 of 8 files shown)


bogdan2412:
This PR doesn't add a CLI command, though happy to push that through separately assuming folks are happy with this PR.

bogdan2412:
I ended up looking at an issue with swapping not working with ClientDomains in #5009 and the work there will affect what this patch should look like, so for now you can skip over the code, though happy to hear thoughts on direction.

PR #4991, opened 2024-02-10
Fix Emits of Additional ANSI Characters

### LINKED ISSUE #4975: Non-standard modifier keys emit original ANSI characters for that key on keydown

Linux Wayland

Mutter

### WezTerm version

20240205-070437-39d2b6ca

### Describe the bug

[My keyboard layout](https://neo-layout.org/) uses non-standard keys as modifiers to add additional layers. For example, if I press `Mod3 + m`, a `%` will be inserted. Mod3 is a modifier, comparable to the shift keys. Mod3 is mapped to the keys that are normally Caps Lock (Mod3 Left) or the `| \` key (Mod3 Right) on an ANSI QWERTY layout (`' #` on my German keyboard), Mod4 is mapped to `< > |` (Mod4 left; no ANSI QWERTY equivalent for key) and right alt (Mod4 right).

If I press a modifier that is on a key that would normally result in a character being inserted (such as `| \` for Mod3, but not Caps Lock), wezterm will insert a single character corresponding to the “original” character that would be on that key for ANSI QWERTY. For example, if I press `Mod3 Right + m` to insert a `%`, wezterm will insert a `\%`. So the modifiers work, but an additional character is inserted once upon keydown of the modifier key. If I simultaneously press the shift key, `|` will be inserted instead of `\`.

This problem first occurred for me for `20240203-110809-5046fc22`, it is not present for `20240128-202157-1e552d76`. It happens for the native package (Arch Linux/pacman), AppImage and flatpak.

### To Reproduce

Activate keyboard layout with modifier keys on keys that normally print characters (such as Neo or NeoQwertz), start wezterm, press key combination such as `Mod3 + m`, with `Mod3` corresponding to the `\ |` key. On keydown of the modifier key, the original character on that key will be inserted.

### Configuration

no config

### Expected Behavior

No characters should be inserted when pressing the modifier keys on their own.

### Logs

_No response_

### Anything else?

_No response_

schrieveslaach:
This commit ensures that ANSI chars won't be emitted when pressing non-standard modifier keys.

Fixes #4975

### DIFF
--- a/window/src/os/x11/keyboard.rs
+++ b/window/src/os/x11/keyboard.rs
@@ -396,7 +396,13 @@ impl KeyboardWithFallback {
                             }
                             _ => sym,
                         }
-                    } else if kc.is_none() && key_code_from_sym.is_none() {
+                    } else if kc.is_none()
+                        && key_code_from_sym.is_none()
+                        // Make sure that non-standard modifier keys are not mapped
+                        // to a fallback because that would result in extra emits of
+                        // the original ANSI characters
+                        && !xsym.is_modifier_key()
+                    {
                         // Not sure if this is a good idea, see
                         // <https://github.com/wezterm/wezterm/issues/4910> for context.
                         match fallback_feed {


schrieveslaach:
@op3, thanks for testing. Do you see more missing keys? If not, I would update my PR with the aforementioned missing keys.

op3:
So, `ISO_Level3_Lock` and `ISO_Level5_Lock` would be missing, obviously.

Another key combination that should not result in any characters is `ISO Level5 Shift + 4`. This key combination is not assigned and thus is not supposed to do anything, but the current code tries to find a fallback (which would be `4`).

Obviously, this is much less of a problem than the modifier keys, because one would usually not press this key combination. Still, I would consider this to be incorrect behavior. And adding “4” to the list of exceptions also does not feel right (too specific …).

There are a few more unassigned key combinations. The W, A, and Z keys on layer 5 come to mind.

schrieveslaach:
@wez, does this fix provide a solution that you would consider to be merged?

wez:
Thanks for this; only just getting around to looking at it.
I think the question on my mind is, are there any cases where this change has a negative impact? eg: for non-neo layout users?

schrieveslaach:
@wez, sorry for the late response.

>  I think the question on my mind is, are there any cases where this change has a negative impact? eg: for non-neo layout users?

I couldn't find any issue when using the regular German layout and the behavior with and without my fix is the same.

@tsacha, I also included ISO_Level3/5_Latch

apraga:
This PR works with ergol so merging it would be great !

stefanfrede:
What can be done to speed up the merging of this request? As far as I can tell, this will help solve the problem I and a handful of others are having. So if there is anything I can do, please let me know.

willghatch:
I'd like to chime in that while this patch is an improvement (and I would love to see it landed), I think more generally it should handle all modifiers this way.  Eg. in my keyboard layout as mentioned in https://github.com/wez/wezterm/issues/6518#issuecomment-2617029954 the problem also happens with Hyper_L and Hyper_R.

schrieveslaach:
@willghatch, thanks for the information. I updated the PR and also use now an existing method of xkbcommon.

@wez, any chance to review this PR because it is very minimal now.

willghatch:
I've been using the old version of this patch, plus a couple lines for hyper keys, for over a month now.  This new version looks even better.  I'll switch to using it, and I hope it can land.

Darkproduct:
I'm an idiot. I forgot to change the branch to `fix-additional-emit`.

It works perfectly. I'll use this branch until it is hopefully merged. Thank you so much.

willghatch:
As an update to my previous comment, I've been using wezterm with the current version of this patch since March.  With my keyboard layout, Wezterm is basically unusable without this patch, but this patch fixes the problem.  This patch makes a huge impact for people who use the popular NEO keyboard layout or a custom keyboard layout on Linux.  This patch is solid gold.

Note that this patch fixes https://github.com/wezterm/wezterm/issues/6518 as well as https://github.com/wezterm/wezterm/issues/4975 since they are duplicates.

swy7ch:
Hi all,

Any news on this? I'm using the `fr(ergol)` layout, and it is bugged in wezterm since it uses a dead key for accentuation :(

schrieveslaach:
The issue still persists on main. I rebased this branch and tested again that this change fixes the issue. @wez, do you mind to have a quick look and maybe merging it?

bnjbvr:
This also seems to fix #5866 for me, fwiw!

schrieveslaach:
@bnjbvr, thanks for the note. Please note, I rebased and tested this PR again and still works for me.

@wez, @bew, what can I add to this PR to push it across the finish line?

schrieveslaach:
@bew, I noticed that you labeled the PR with X11 and that made me realize that the code change is related to the X11 subsystem (totally forgot about it). However, I'm running wezterm under Wayland (GNOME has no x11 session anymore). Is the label wrong or should I investigate why the code is used on Wayland too?

bnjbvr:
I think my earlier scan of the code showed that the same code was used on Wayland, resulting in fixing the issue for me on Wayland. I seem to recall I came to this conclusion by going up the chains of callers via LSP, fwiw

honnorat:
This PR fixes #5866, which is a blocker for me.
There is no conflict with `wezterm:main`, @wez is it possible to merge it soon ?

PR #4994, opened 2024-02-11
Fix inconsistent spawn behaviour when using default_mux_server_domain.

bogdan2412:
When using an exec domain as a default_mux_server_domain, you currently witness the following behavior:
- Newly started mux-servers create a pane in the exec domain.
- Splitting an existing pane will create the new pane in the exec domain.
- Calling [wezterm cli spawn] will create a new tab or window in the exec domain.
- Creating a new tab from the Gui does *not* run in the exec domain.

This appears to be due to the fact that ClientDomain hardcodes an assumption that the default domain_id of the remote mux-server is 0.

Instead of having ClientDomain issue a SpawnV2 RPC explicitly asking for domain_id 0, simply use DefaultDomain.

### DIFF
--- a/wezterm-client/src/domain.rs
+++ b/wezterm-client/src/domain.rs
@@ -20,7 +20,6 @@ use wezterm_term::TerminalSize;
 pub struct ClientInner {
     pub client: Client,
     pub local_domain_id: DomainId,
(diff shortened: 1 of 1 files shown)


wez:
Good catch! Thank you!

PR #4995, opened 2024-02-11
Fix kitty image leaving behind images after removing image placements

### LINKED ISSUE #2422: Kitty image protocol fails to delete images after resize by placement id

Linux Wayland

Reproducible on X and Wayland with any window manager, tested on:
- Wayland Sway, Gnome
- X11 Gnome

### WezTerm version

20220814-222723-f2ee48bd

### Describe the bug

Deleting image by placement id via kitty protocol `\x1b_Ga=d,d=i,i={image_id},p={placement_id}\x1b\\` right after resize (this is done in response to `SIGWINCH` to redraw the screen) leaves them behind sometimes. Looking at `kitty_remove_placement` it looks like `placement_id` is resolved to coordinates which might change after resize?

### To Reproduce

I do not have small reproducible example, I have a program that renders scalable graphic icons and displays them as glyphs when I resize it leaves some glyphs behind
```bash
$ git clone https://github.com/aslpavel/sweep-rs.git
$ cd sweep-rs/
$ ./scripts/demo.py
# resize wezterm multiple times
```
## Broken output (images left behind)
wezterm-broken

## How it looks on kitty
wezterm-expected

### Configuration

```lua

return {
   -- environment
   set_environment_variables = {
      COLORTERM="truecolor",
   },
   term = "xterm-24bit",

   -- font
   font_size = 11.0,
         -- "PragmataPro",
         "Iosevka",
         -- "Victor Mono",
         "Material Design Icons",
         "Symbols Nerd Font",
   }),
   bold_brightens_ansi_colors = false,

   -- colors
   colors = {
      foreground = "#ebdbb2",
      background = "#282828",
      cursor_bg = "#ebdbb2",
      cursor_fg = "#ebdbb2",
      cursor_border = "#ebdbb2",
      scrollbar_thumb = "#504945",
      ansi = {"#282828", "#cc241d", "#98971a", "#d79921", "#458588", "#b16286", "#689d6a", "#ebdbb2"},
      brights = {"#7c6f64", "#fb4934", "#b8bb26", "#fabd2f", "#83a598", "#d3869b", "#8ec07c","#fbf1c7"},
      tab_bar = {
         background = "#32302f",
         active_tab = {
            bg_color = "#076678",
            fg_color = "#ebdbb2",
         },
         inactive_tab = {
            bg_color = "#458588",
            fg_color = "#3c3836",
         },
         inactive_tab_hover = {
            bg_color = "#83a598",
            fg_color = "#3c3836",
         },
         new_tab = {
            bg_color = "#458588",
            fg_color = "#ebdbb2",
         },
         new_tab_hover = {
            bg_color = "#076678",
            fg_color = "#ebdbb2",
         }
      }
   },

   -- tab bar
   use_fancy_tab_bar = false,
   tab_max_width = 20,

   default_cursor_style = "SteadyBar",
   enable_scroll_bar = true,
   enable_wayland = true,
   enable_kitty_graphics = true,
}
```

### Expected Behavior

Image should be deleted by its `(id, placement_id)` on delete command even after resize

### Logs

_No response_

### Anything else?

_No response_

jonboh:
[DRAFT] This PR is not intended to be merged right now.

This should fix https://github.com/wez/wezterm/issues/2422, but the fix in commit 319a5376e3d09587520f2550a9ca6e7e8655472c will have serious performance issues for cases with big scrollbacks and lots of image updates. The "fix" is to ignore `PlacementInfo` (that might be stale) and go through all the scrollback removing the image.

The main issues seems to be that `StableRowIndex` can become stale whenever a window resizing changes the amount of columns enough so as to add more lines by wrapping other lines. After wrapping, the image will get displaced (this is in kitty's implementation does not happen), and so the `PlacementInfo` that is stored for that `image_id` and `placement_id` is no longer valid.

It seems to me that it would be a better approach to keep the original way of removing the images by their `PlacementInfo` (instead of going through all the scrollback), and prevent the images from being moved by lines being wrapped. The problem I see here is that this handling of the cell properties of those lines will need to be done wherever the lines are being wrapped and I'm not sure you'll be too happy about adding logic about images there (for being so unrelated).

Let me know if you have any ideas or thoughts about how we could address this.

### Kitty(reference):
kitty_resize

### Wezterm(main):
wezterm_resize

### Wezterm(this PR):
wezterm_fix_resize

Another issue tangentially related to this is that images are not occluded when a new pane appears:
image
In this case there isn't a way to update displayed images when the geometry of the pane changes. A way to handle these geometry changes in general would also solve the resizing.

### DIFF
--- a/term/src/terminalstate/kitty.rs
+++ b/term/src/terminalstate/kitty.rs
@@ -101,7 +101,11 @@ impl TerminalState {
             verbosity
         );
         if image_id != 0 {
-            self.kitty_remove_placement(image_id, placement.placement_id);
+            if let Some(place_id) = placement.placement_id {
+                if place_id != 0 {
+                    self.kitty_remove_placement(image_id, placement.placement_id);
+                }
+            }
         }
         let img = Arc::clone(self.kitty_img.id_to_data.get(&image_id).ok_or_else(|| {
             anyhow::anyhow!(
--- a/term/src/terminalstate/kitty.rs
+++ b/term/src/terminalstate/kitty.rs
@@ -301,9 +301,7 @@ impl TerminalState {
     ) {
         let seqno = self.seqno;
         let screen = self.screen_mut();
-        let range =
-            screen.stable_range(&(info.first_row..info.first_row + info.rows as StableRowIndex));
-        for idx in range {
+        for idx in 0..screen.scrollback_rows() {
             let line = screen.line_mut(idx);
             for c in line.cells_mut() {
                 c.attrs_mut()


wez (commented):
Thanks for looking into this!

The rewrap/resize case is clearly something that I didn't think all the way through for this.

I think that we should add some logic to fixup `PlacementInfo` on resize/rewrap; while this PR makes the removal more correct it seems like the fundamental problem is that our data is stale, and that that will impact other areas.

I think what is needed is to amend the `StableRowIndex` and other fields in the placement in `Screen::rewrap_lines` and/or `Screen::resize`.  There is some logic to compute an adjusted cursor position that you might be able to adapt, but there are a couple of open issues around the cursor position moving in surprising ways in some edge cases when sizing smaller and then larger again, so don't trust the existing logic blindly!

bew:
Hello @jonboh, this is an old one ! 😅
I'm slowly learning about the Kitty protocol and its impl & shortcomings in Wezterm, and am starting to get ready to review related PRs to try to improve the Kitty situation for next release.

👉 Would you be willing merge/rebase on `main` and continue working on this topic?
Or would you prefer if someone else took the lead on this? 🤔

jonboh:
you can take it up
Note that the changes in this branch were not intended to be merged (thats why it was left as draft). As I remember this was a bit hacky and could be done better with more work.

PR #5001, opened 2024-02-12
fix linux `divine_process_list`

### LINKED ISSUE #4993: `divine_process_list().foreground` always gives the shell

Linux X11

awesomewm

### WezTerm version

not a upstream version; based on `a7d9cfd25`

### Describe the bug

`LocalPane::divine_process_list().foreground` always seems to return the shell. 

I'm doing this: 
```rust
        let proc_list = self.divine_process_list(CachePolicy::FetchImmediate);
        dbg!((self.pane_id(), proc_list.as_ref().map(|p| &p.foreground.argv), self.get_foreground_process_name(CachePolicy::FetchImmediate), self.get_foreground_process_info(CachePolicy::FetchImmediate).as_ref().map(|p| &p.argv)));
```

with neovim open, what I get is:
```
[mux/src/localpane.rs:821] (self.pane_id(), proc_list.as_ref().map(|p| &p.foreground.argv),
    self.get_foreground_process_name(CachePolicy::FetchImmediate),
    self.get_foreground_process_info(CachePolicy::FetchImmediate).as_ref().map(|p|
            &p.argv)) = (
    0,
    Some(
        [
            "/home/steven/.local/bin/nu",
        ],
    ),
    Some(
        "/usr/bin/nvim",
    ),
    Some(
        [
            "nvim",
        ],
    ),
)
```

I don't think nushell is the problem here, because bash and zsh both give this:
```
[mux/src/localpane.rs:821] (self.pane_id(), proc_list.as_ref().map(|p| &p.foreground.argv),
    self.get_foreground_process_name(CachePolicy::FetchImmediate),
    self.get_foreground_process_info(CachePolicy::FetchImmediate).as_ref().map(|p|
            &p.argv)) = (
    1,
    Some(
        [
            "/usr/bin/python3",
            "-c",
            "import sys; sys.path = [p for p in sys.path if p != \"\"]; import neovim; neovim.start_host()",
            "/home/steven/dotfiles/config/nvim/plugged/semshi/rplugin/python3/semshi",
        ],
    ),
    Some(
        "/usr/bin/nvim",
    ),
    Some(
        [
            "nvim",
        ],
    ),
)
```

Okay maybe `semshi` (a vim plugin for python stuff) is doing weird stuff, so I tried `ed`:
```
[mux/src/localpane.rs:821] (self.pane_id(), proc_list.as_ref().map(|p| &p.foreground.argv),
    self.get_foreground_process_name(CachePolicy::FetchImmediate),
    self.get_foreground_process_info(CachePolicy::FetchImmediate).as_ref().map(|p|
            &p.argv)) = (
    0,
    Some(
        [
            "/home/steven/.local/bin/nu",
        ],
    ),
    Some(
        "/usr/bin/ed",
    ),
    Some(
        [
            "ed",
        ],
    ),
)
[mux/src/localpane.rs:821] (self.pane_id(), proc_list.as_ref().map(|p| &p.foreground.argv),
    self.get_foreground_process_name(CachePolicy::FetchImmediate),
    self.get_foreground_process_info(CachePolicy::FetchImmediate).as_ref().map(|p|
            &p.argv)) = (
    1,
    Some(
        [
            "/usr/bin/bash",
        ],
    ),
    Some(
        "/usr/bin/ed",
    ),
    Some(
        [
            "ed",
        ],
    ),
)
```

### To Reproduce

1. add the above code somewhere and a keybinding to trigger it
2. run some persistent stuff and press the binding to trigger the debug

### Configuration

no config other than the keybind

### Expected Behavior

_No response_

### Logs

_No response_

### Anything else?

_No response_

crides:
Fixes #4993

The position of the `starttime` field in `/proc/pid/stat` is incorrect.

`man procfs` under `/proc/pid/stat` says:
```
              (4) ppid  %d
                     The PID of the parent of this process.
... snip ...
              (22) starttime  %llu
                     The  time the process started after system boot.  Before Linux 2.6, this
                     value was expressed in jiffies.  Since Linux 2.6, the value is expressed
                     in clock ticks (divide by sysconf(_SC_CLK_TCK)).

                     The format for this field was %lu before Linux 2.6.
```
which has a difference of 18. previously, the difference in the code is 19, so we were indexing `vsize` (the VM memory usage)

### DIFF
--- a/procinfo/src/linux.rs
+++ b/procinfo/src/linux.rs
@@ -73,7 +73,7 @@ impl LocalProcessInfo {
                 name: name.to_string(),
                 status: fields.first()?.to_string(),
                 ppid: fields.get(1)?.parse().ok()?,
-                starttime: fields.get(20)?.parse().ok()?,
+                starttime: fields.get(19)?.parse().ok()?,
             })
         }
 


wez:
Ah, a great find, thank you! I'm curious how this matters on posix, but I'll follow up on your original issue

PR #5002, opened 2024-02-12
fix(flatpak): Remove cwd from desktop entry

briandipalma:
`cwd` provided from desktop entry overrides Lua configured cwd so you end up with different behaviour when you launch from DE launcher and from running `flatpak run ...`.

I'm guessing this is the file that's used to create the flatpak desktop entry.


briandipalma:
I installed the flatpak today and started configuring wezterm: https://github.com/briandipalma/iac/blob/main/dotfiles/wezterm/wezterm.lua#L9

But this line was not having any effect

```lua
config.default_cwd = wezterm.home_dir .. "/dev/iac"
```

when I ran the flatpak from the desktop environment launcher (Pop_OS/GNOME). Yet it did work when I ran the flatpak from the command line with `flatpak run org.wezfurlong.wezterm`. So obviously I went to dig into the desktop file which is what the launcher uses. I noticed the `--cwd` and from what I can tell in the documentation that flag overrides the config file. So I guessed that was the problem. Deleting it fixed the problem.

Do we need to specify `--cwd` in the desktop file? I would have thought that, based on the documentation, wezterm would open up in the users home directory as fallback anyway.

bew:
There is unfortunately no easy solution that works for everyone here...
See more on recent PR #7338 and how it conflicts with #2103..

I think a proper solution would be to indeed remove that but also add additional desktop entry fields to tell consumers of the desktop file how to call wezterm to explicitly set CWD or execute a program.
👉 Follow tracking issue on this for updates: https://github.com/wezterm/wezterm/issues/6008.

Let's close this for now and we'll revisit this with #6008 🤔

Issue #3237, opened 2023-03-13
Provide a way to save the current layout


jason0x43:
**Is your feature request related to a problem? Please describe.**

One of the features I miss most from tmux when using Wezterm's native tabs and panes is the ability to save and later restore the active layout. This is useful when updating the terminal client, when restarting my laptop, or just when I want to close whatever I was doing for work on Friday and restore it on Monday.

This is related to #1949.

**Describe the solution you'd like**

Wezterm currently provides APIs to initialize a layout at startup. As far as I'm aware, though, there isn't an API to save the current layout. That would be ideal -- a function that could be called, such as from a keybinding or some other event, that would save a snapshot of the domain to a file, and a complementary function that could restore such a snapshot.

Ideally, the snapshot would save at least the layout within each window (tabs and panes), tab and pane titles, and CWD in each pane. The ability to optionally save and restore the scroll buffers of all the panes would also be worthwhile.

I'm less concerned with saving window positions and sizes. Other tools make snapping windows into position easy enough; the worthwhile part would be restoring the general context of whatever I was working on when I saved the layout.

**Describe alternatives you've considered**

Use a terminal multiplexer than can save layouts, such as tmux with the resurrect plugin.

**Additional context**

RazaGR:
I want to add that it would be also great to have the ability to also save workspaces and what it was running 

@wez any idea if this will be implemented? Thanks

wez:
No one is working on this at this time. I have no plans to work on this any time soon. If you search for other issues around saving/restoring state you will discover the set of tricky concerns that need to be considered before attempting to implement this feature.

vvzen:
+1, since this would be lovely and quite useful!

I have recently switched from iTerm2+tmux to just using wezterm on macOS and it has been a beautiful journey so far. The only missing piece right now is the fact that now I have to shut down my laptop to do an OS update and I'm gonna lose all of my carefully crafted tabs and workspaces, unless I find I way to serialize everything down and restore it later.

When in tmux-land, I was using https://github.com/tmux-plugins/tmux-resurrect .

Even if this is not being planned right now, do you think there could be a way to achieve something similar (hacking it away) it by creating the required workspaces and tabs when the `gui-startup` event fires (https://wezfurlong.org/wezterm/config/lua/gui-events/gui-startup.html) ? I'm asking since I'm still kinda new and if you think it's not worth the effort I'll avoid even going in this rabbit hole :D 

Thanks!

wez:
[...]

Issue #3091, opened 2023-02-10
Preserve tab history on exit/restore on launch


eugenesvk:
**Is your feature request related to a problem? Please describe.**
iTerm has this cool feature where it __automatically__ __saves__ all the tabs' data when you quit the app and __restores__ it when you reopen the app.
It can't restore the state of the shell, but still having the same tabs with the same content is rather helpful in not breaking the workflow on restarts

**Describe the solution you'd like**
Have an option to automatically save all tab data on close and restore it on next startup when I quit the app

**Describe alternatives you've considered**
Not sure what thse would be. As far as I understand it could be possible to save the position by hooking into quit/startup events, but not the history itself

**Additional context**
Searched through the issues/discussions and there are some that are tangentially related, but not the same, they mostly cover only the composition/layout, but not the content, sorry if I missed something
  - https://github.com/wez/wezterm/issues/256
  - https://github.com/wez/wezterm/issues/1949
  - https://github.com/wez/wezterm/discussions/1665 and https://github.com/wez/wezterm/issues/1531

wez:
I consider this to be a duplicate of #256

Issue #1949, opened 2022-05-03
Save and load layouts


JakeHL:
**Is your feature request related to a problem? Please describe.**
I find it frustrating when I have to open up a terminal, configure my pane how I like them and run the relevant tools in each pane.

**Describe the solution you'd like**
This suggestion is heavily inspired by [layouts](https://zellij.dev/documentation/layouts.html) in [zellij](https://github.com/zellij-org/zellij)

I think it would be great to have some form of config file that can be loaded that will automatically configure pane/tabs and the commands that are run inside them. These could be saved inside of source control alongside projects.

For example if I'm frequently opening a text editor, file browser, and terminal in wezterm for editing code maybe it could be loaded by typing `wz text-editor.yml` or something.

configs could be made by hand or exported from an existing layout.

I would vote to use yaml for the config, however I'm aware toml is more idiomatic in rust applications. Maybe lua would be a good idea to be in keeping with the config file.

an example of what the config could look like if it were yaml. I'm absolutely positive this could be better thought out.
```yaml
split:
  direction: vertical
  sections: 
    - split:
        direction: horizontal
      sections:
      - pane:
          cmd: 'br .'
          cwd: '.'
          width: '20%'
      - pane:
          cmd: 'nvim .'
    - pane:
        cmd: 'zsh'
        height: '10%'
```

**Describe alternatives you've considered**
Remembering the last used layout by directory was an option I considered, but having editable configs would make it really nice to store specific layouts alongside projects.

**Additional context**
I currently use zellij for this functionality however I really like the multiplexing in wezterm. Thanks for the great work!

wez:
FWIW, when I think of layout, I think of window positioning rather than application composition. A layout feels like something that should be able to be applied or re-applied to make the arrangement conform to a pre-defined layout, rather than (in zellij's current implementation) an initial composition.

The "start up these apps in this arrangement" concept would fit wezterm better as an application startup event with a context parameter that allows setting up multiple windows, tabs and panes.

The "apply this layout to my existing set of windows/tabs/panes" concept (eg: like switching between tiling window modes in some window managers) is a bit different and requires the ability to influence the pane arrangement in a tab. 

There's some plumbing needed for these concepts. I've linked some relevant issues below just so that it is easier for me to thread things together and close out the appropriate issues when things start to happen.

Related issues:
* https://github.com/wez/wezterm/issues/674
* https://github.com/wez/wezterm/issues/256
* https://github.com/wez/wezterm/issues/1253
[...]

Issue #1326, opened 2021-11-17
Save scrollback history to file


yyogo:
It's often desirable to save the entire scrollback history to a file, e.g after a long command has finished but it's output wasn't captured, or when running in a remote shell.

Ideally, wezterm would have this implemented as a mux server command accessible by the CLI, or as an action (that would include an overlay for filename input or some other configuration). 

# Prior art
* `tmux` has the `capture-pane` command

# Alternatives
* Copy mode -> `g0vG$` -> copy & paste to a file

bew:
There is an example of something close in https://wezfurlong.org/wezterm/config/lua/wezterm/on.html#custom-events

yyogo:
> There is an example of something similar in https://wezfurlong.org/wezterm/config/lua/wezterm/on.html#custom-events You can use a custom lua action to do this.

Alright, thanks! 

I suppose this could also be done from the CLI, right? I haven't been able to use `cli proxy`, maybe it's a bug

PR #5013, opened 2024-02-15
[Feature] Resurrection: saving and restoring terminal layout/contents/commands

crides:
Drafting as this is WIP, but want to seek feedback/get discussion going on this.

Currently:
- content/scrollback saving and restoring works. This is done by having a reverse attribute to CSI converter, so that the saved contents are in "plain text" (other than attributes; just like `tmux`)
- basic layout saving/restoring works. The layout is saved to a slightly minimized tree in JSON form.
- basic shell/cwd/program works. shell, cwd and foreground* program just gets obtained thru existing APIs, and shell and cwd are restored thru pane spawning/splitting; program is just converted back to a shell string and send to the pane. Note: foreground is technically not the right term, as we grab the newest of the first level children of the shell and spawn it back.
- the whole state is wrapped in a `.tar.zstd` for compression. each pane's content gets its own file (just like `tmux`)
- the save/restore functionalities are exported as lua APIs so users can do whatever they want
  locally I'm using this for testing (manually saving/restoring):
  ```lua
        { mods = 'CTRL|SHIFT', key = 'S', action = wezterm.action_callback(function(_, _, _)
            wezterm.mux.save_state_to("state_test")
        end) },
        { mods = 'CTRL|SHIFT', key = 'R', action = wezterm.action_callback(function(_, _, _)
            wezterm.mux.restore_state_from("state_test")
        end) },

  ```
- there's basically 0 error handling. that definitely needs to be improved

This has not been thoroughly tested, and I'll integrate it into my workflow and see how it goes


crides:
possibly related:
- #1949 
- #3237 
- #1326 
- #3091 

I remember there should be one more related issue (more about content than layout) but I can't find it (though it's pretty short/less important)

mux/src/pane.rs
```diff
@@ -192,6 +373,37 @@ pub trait Pane: Downcast + Send + Sync {
 
     fn get_logical_lines(&self, lines: Range<StableRowIndex>) -> Vec<LogicalLine>;
 
+    fn get_logical_lines_raw(&self, lines: Range<StableRowIndex>) -> String {
```
wez, line 376: I don't believe that we need this function or the complicated `diff_attrs` function above at all. Look at:

https://github.com/wez/wezterm/blob/22f9f8d288521508bd4cdc924d9f209d78b36483/lua-api-crates/termwiz-funcs/src/lib.rs#L268

which is how we render a set of `Line` into ansi escape sequences to support that various lua functions like https://wezfurlong.org/wezterm/config/lua/pane/get_lines_as_escapes.html

mux/src/pane.rs
```diff
@@ -316,6 +528,10 @@ pub trait Pane: Downcast + Send + Sync {
     fn exit_behavior(&self) -> Option<ExitBehavior> {
         None
     }
+
+    fn save(&self) -> Option<SavedPaneState> {
```
wez, line 532: naming: match whatever naming we come up with for the mux level save function as discussed above

But, conceptually speaking the command resolution and replay needs some additional consideration that I'll break out into a separate non-code comment.

mux/src/tab.rs
```diff
@@ -751,6 +751,35 @@ impl Tab {
     pub fn get_zoomed_pane(&self) -> Option<Arc<dyn Pane>> {
         self.inner.lock().get_zoomed_pane()
     }
+
+    pub fn save(&self) -> Option<SavedTab> {
```
wez, line 755: naming here also

mux/src/tab.rs
```diff
@@ -751,6 +751,35 @@ impl Tab {
...
+        fn convert_node(t: &Tree) -> Option<SavedTree> {
```
wez, line 756: can we leverage/reuse some or all of the `pane_tree` and `build_from_pane_tree` logic in this file that is used by the mux protocol here?

mux/src/window.rs
```diff
@@ -263,4 +263,11 @@ impl Window {
             self.invalidate();
         }
     }
+
+    pub fn save(&self) -> Option<SavedWindow> {
```
wez, line 267: naming here also

termwiz/src/color.rs
```diff
@@ -257,6 +257,17 @@ impl From<SrgbaTuple> for ColorSpec {
     }
 }
 
+impl From<ColorAttribute> for ColorSpec {
```
wez, line 260: hopefully not needed if you switch to `lines_to_escapes`

termwiz/src/surface/line/clusterline.rs
```diff
@@ -9,17 +9,17 @@ use std::num::NonZeroU8;
 
 #[cfg_attr(feature = "use_serde", derive(Serialize, Deserialize))]
 #[derive(Debug, Clone, PartialEq)]
-struct Cluster {
-    cell_width: u16,
-    attrs: CellAttributes,
+pub struct Cluster {
```
wez, line 12: hopefully the set of changes in this file are not needed with `lines_to_escapes`

termwiz/src/surface/line/line.rs
```diff
@@ -2,10 +2,11 @@ use crate::cell::{Cell, CellAttributes, SemanticType, UnicodeVersion};
 use crate::cellcluster::CellCluster;
 use crate::hyperlink::Rule;
 use crate::surface::line::cellref::CellRef;
-use crate::surface::line::clusterline::ClusteredLine;
+pub use crate::surface::line::clusterline::ClusteredLine;
```
wez, line 5: also hope that this file can be reverted

termwiz/src/surface/line/storage.rs
```diff
@@ -6,7 +6,7 @@ use serde::{Deserialize, Serialize};
 
 #[cfg_attr(feature = "use_serde", derive(Serialize, Deserialize))]
 #[derive(Debug, Clone, PartialEq)]
-pub(crate) enum CellStorage {
+pub enum CellStorage {
```
wez, line 9: and this; these are all intentionally opaque implementation details

mux/src/lib.rs
```diff
@@ -1439,3 +1551,106 @@ impl wezterm_term::DownloadHandler for MuxDownloader {
...
+            split_panes(domain, *s.right, tab, new.pane_id(), id_map).await;
+        }
+    })
+}
+
+#[derive(Clone, Serialize, Deserialize)]
+pub struct SavedPaneState {
+    id: usize,
+    cwd: Option<SerdeUrl>,
+    root_argv: Vec<String>,
```
wez, line 1622: the domain and other important info are missing from this, but I think the approach here needs to be different; I'll cover this more in the main overall review comment

wez (commented):
Thanks for tackling this!  There are a couple of existing things that I think you can build on top of that will make this simpler (see inline comments on the code), but there are also a couple of things that I think might make you frown a little bit, because there's more to reconcile before we can merge in this functionality.

From a security and privacy perspective: storing the terminal output on disk is a hazard, because it may contain private/privileged/secret information. At the very least we need to ensure that the files that we write for that are private.  In theory, the `UmaskSaver` that we create at process start time should mask off the group+other bits from the files that we save so you may not strictly need to take any extra steps around that most basic safeguard.

However: I know that there was a CVE reported against VTE when it stored scrollback unencrypted in plain text as part of its infinite scrollback handling, and there are parallels between that and doing it here. I'd feel more comfortable if we had a reasonable solution around this, although the logistics are a bit different: for scrollback that is discarded when the process exits, the key material can also be discarded. But when we explicitly want to load things again later, then we need a persistent key that we can use to decrypt it later so there is additional configuration and infrastructure required to deal with that. 

There are some headaches here. If we don't have a solution for this then we need to ensure that the save/restore functionality is opt-in and that is it very clear to the user that it is not encrypted and that they need to satisfy themselves that it is suitable for their environment.

In terms of deciding the program and re-launching it, there are a couple of things to consider; here are a couple of issues that may not seem immediately relevant:

* https://github.com/wez/wezterm/issues/580
* https://github.com/wez/wezterm/issues/2080

why are they relevant here?

* The command that we spawned into the pane is not necessarily the foreground or youngest process currently running into the pane.  I'm not sure that it is wise to grab the foreground/youngest process TBH.
* Not all panes are local panes with processes (eg: could be a remote ssh session, or a serial port) so querying the process list will never work for those
* A pane may be a "local" pane but may be in a different domain (such as an ExecDomain or WslDomain) that has a slightly different way to get launched
* The command that was spawned could be dangerous! Consider someone launching `rm -rf` into a pane.  Zellij's session resumption prompts the user to confirm that it is safe to re-launch. We could potentially shortcut an unconditional prompt with some basic heuristics. For example, if the process passes the `mux-is-stateful` checks (which include looking at https://wezfurlong.org/wezterm/config/lua/config/skip_close_confirmation_for_processes_named.html) then we could assume that a basic shell invocation need not be prompted about

What I had in mind around this was a very rough idea that we might save the `CommandBuilder` that was used to spawn a given pane so that we could retrieve and reason about it later via something like `Pane:get_spawning_command(&self) -> Option<CommandBuilder>`

Then there could be a general mechanism to handle the "pane isn't running, show a prompt to see if it should be (re)started" state.

There's complications around this:

* The existing logic that cares about the process having exited and showing the hold message is only present for local panes. There's a decent chunk of code to reconcile the exit status of the process and the configured exit behavior and it is all specific to a local process.
* There is currently no way to represent a not-yet-spawned process/pane, and I don't think we should intoduce that concept to the Pane interface itself because it would make it more awkward to reason about the state of a Pane.

Instead, I think we can draw from the technique used in `mux/src/ssh.rs` and/or things like the copy and quickselect overlays to shim in a wrapping Pane instance.

So, when mapping the saved version of a pane in, a separate `ResumedPane` would be built instead. We'd pass in the original `CommandBuilder` and the saved scrollback.  The pane would render those, then it would handle the safe-to-restart logic and show a prompt if required.  When it is safe/confirmed to spawn the actual command it would then do so in the appropriate domain and switch its mode of operation to either delegate to that newly spawned pane, or to replace itself in its containing tab with that newly spawned pane.

I'm also open to other ideas on how to deal with this. I do think it is important to handle the various different kinds of panes holistically, or otherwise find a composable/reusable way to deal with the (re)start behavior.

mux/src/pane.rs
crides: Ah okay, that'll make life easier. When I started on this the lua function didn't exist yet.

crides:
Thanks for the very detailed reply!

> but there are also a couple of things that I think might make you frown a little bit, because there's more to reconcile before we can merge in this functionality.

No problem. I would like to do things in a correct way for other users, but first I have some things I would like clarifying in your comment.

---

> In theory, the `UmaskSaver` that we create at process start time should mask off the group+other bits from the files that we save so you may not strictly need to take any extra steps around that most basic safeguard.

Yep, the state saves as `0600` right now

> But when we explicitly want to load things again later, then we need a persistent key that we can use to decrypt it later so there is additional configuration and infrastructure required to deal with that.

Encryption would be nice, but it'd also be nice if the user can easily dump the contents without having wezterm load it first. I'm not sure of a good way to do both yet.

> If we don't have a solution for this then we need to ensure that the save/restore functionality is opt-in and that is it very clear to the user that it is not encrypted and that they need to satisfy themselves that it is suitable for their environment.

I'm not sure about the security parts, but right now the feature is totally opt in. Nothing extra will be done unless the user triggers `save_state_to`/`restore_state_from` in their config. Given we use lua for configs, I think exposing the saving/restoring API + some timer/cron utilities would be easy and enough to achieve what `tmux-resurrect` has, instead of wrapping a bunch of stuff behind config options

---

As for spawning commands, I don't quite understand the problems yet:

> - The command that we spawned into the pane is not necessarily the foreground or youngest process currently running into the pane.

No. But the foreground command running on top of the shell should be. Also, we can't grab the second or deeper children of the shell, because that'll break the spawning relationships

> - Not all panes are local panes with processes (eg: could be a remote ssh session, or a serial port) so querying the process list will never work for those

I'll need to look into those, but in those cases I think we shouldn't restore recursively all the layers? As in, we should just attempt to restore the pane history and connection, but not care about what's running inside of those panes.

> - A pane may be a "local" pane but may be in a different domain (such as an ExecDomain or WslDomain) that has a slightly different way to get launched

I'll need to look into those

> - The command that was spawned could be dangerous! Consider someone launching `rm -rf` into a pane. Zellij's session resumption prompts the user to confirm that it is safe to re-launch. We could potentially shortcut an unconditional prompt with some basic heuristics. 

`tmux-resurrect` handles this by [only allowing default and config whitelisted processes to be restored](https://github.com/tmux-plugins/tmux-resurrect/blob/master/docs/restoring_programs.md). I think this is better than just ask for confirmation*. Maybe we can use a combination of whitelisting, and asking for confirmation for non-whitelisted things.

*: because, if I have 50 panes with editors and repls inside some of them, then I'd need to go to every pane to restore them

wez:
> > * Not all panes are local panes with processes (eg: could be a remote ssh session, or a serial port) so querying the process list will never work for those
> 
> I'll need to look into those, but in those cases I think we shouldn't restore recursively all the layers? As in, we should just attempt to restore the pane history and connection, but not care about what's running inside of those panes.

If I have a 3-pane layout, where I have a regular local shell on the left, a shell in a docker container (implemented as an ExecDomain that essentially ran `docker exec ...`) in the middle, and an ssh session (implemented as an SshDomain that used the internal ssh client to connect to a remote host) in the right pane, there's no reason that the save/restore shouldn't be able to handle those three things in the same.

* For the local pane case: what was run by the user was the shell, and I think the principle of least surprise is to resume their shell (and show their scrollback up to that point).  I think running whatever the most recent command was in the shell (assuming that we figure it out correctly) might be interesting, but would be confusing when that program terminates and it doesn't leave them in their shell
* For the docker container case, I don't think we can guarantee to see inside the container to understand what it was running purely from process tree introspection, and again, I think just dropping back into the command that was used to spawn the pane is the least surprising and probably most anticipated outcome
* For the ssh case we definitely cannot see into what it was running, but it is similar to the other two cases where I think the reasonable expectation is to restore the prior scrollback and drop you into a freshly logged in shell there

> > * The command that was spawned could be dangerous! Consider someone launching `rm -rf` into a pane. Zellij's session resumption prompts the user to confirm that it is safe to re-launch. We could potentially shortcut an unconditional prompt with some basic heuristics.
> 
> `tmux-resurrect` handles this by [only allowing default and config whitelisted processes to be restored](https://github.com/tmux-plugins/tmux-resurrect/blob/master/docs/restoring_programs.md). I think this is better than just ask for confirmation*. Maybe we can use a combination of whitelisting, and asking for confirmation for non-whitelisted things.
> 
> *: because, if I have 50 panes with editors and repls inside some of them, then I'd need to go to every pane to restore them

Yeah, I think a combination of an allow list and some reasonable heuristics would make sense for this.  That said, I don't think the prompt would be onerous: what I had in mind is that the pane is restored and you will see a simple text prompt in that pane to initiate a restart, so it's not like it would be a series of 50 modals that you had to consider immediately on restore, you would only have to confront it if/when you wanted to.

Destroy666x:
I don't think security is that important if it's something that can be willingl enabled in Lua.

Mikilio:
Maybe it could be made secure by storing the data in /tmp and introducing a Lua setting to specify your own command to store/retrieve it to disk. The defaults would be a naive copy, and we could add a warning that this configuration is insecure.

For people who want more security, they can then encrypt their restoration data, filter it as they see fit or use any third party program to handle backups from tempfs in general.

PR #5026, opened 2024-02-17
docs(lua/config/front_end): default value of OpenGL

cultab:
Reflects changes made as of `20240128-202157-1e552d76`.

Low hanging fruit I noticed at random. From other pages seems like we should be keeping the change history of default values and behavior, As such I've only appended a "sinse block".


wez:
Thanks!

PR #5031, opened 2024-02-18
gui: pass window to set_inner_size()

jknockel:
This fixes a regression from a0974a25 where `TermWindow::set_inner_size()` can be called before the `TermWindow`'s `window` has been set. The consequence of this can be a new window that is not of the correct size.

I can only reproduce this issue on Windows, and although I seem to be able to consistently reproduce it now I would think that I would have noticed this issue earlier had it always reproduced in the past.


wez:
Thanks!

PR #5032, opened 2024-02-18
gui: don't repaint during resize

jknockel:
Wait until `set_inner_size()` has finished before repainting.  If we repaint while our request to resize is in progress, our understanding of the current size of the window may be inconsistent with the actual size, causing the rendered contents of the window to bounce around.

The easiest way for me to reproduce the original issue is to ctrl+- or ctrl+= on either X11 or Weston. The proposed changes seem to fix the issue on both of those platforms. I couldn't reproduce the original issue on Windows, but the proposed changes didn't seem to make anything any worse on Windows either!

### DIFF
--- a/wezterm-gui/src/termwindow/mod.rs
+++ b/wezterm-gui/src/termwindow/mod.rs
@@ -370,6 +370,7 @@ pub struct TermWindow {
     pub dimensions: Dimensions,
     pub window_state: WindowState,
     pub resizes_pending: usize,
+    is_repaint_pending: bool,
     pending_scale_changes: LinkedList<resize::ScaleChange>,
     /// Terminal dimensions
     terminal_size: TerminalSize,
@@ -696,6 +697,7 @@ impl TermWindow {
             dimensions,
             window_state: WindowState::default(),
             resizes_pending: 0,
+            is_repaint_pending: false,
             pending_scale_changes: LinkedList::new(),
             terminal_size,
             render_state,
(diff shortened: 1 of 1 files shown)


wez:
Hmm, how does this look in practice? If we don't repaint during a live resize then I think there are two likely outcomes:
* When resizing larger: either transparent or random data in the gaps for the new larger size
* the window contents are distorted until the resize completes

the behavior may also depend on the selected front end and gpu drivers, and may also be OS-dependent.

If you keep a hold on the window border and keep resizing quickly larger and smaller for an extended period, how does it look?

wez:
if this change is scoped solely to wezterm-initiated resizes then I suppose my questions are irrelevant

jknockel:
> if this change is scoped solely to wezterm-initiated resizes then I suppose my questions are irrelevant

Yes these changes only pertain to program-generated calls to `set_inner_size()`, not to user resizing.  With the proposed changes, we defer any repaint triggered while a `set_inner_size()` is in progress until immediately afterwards, so if there are issues on a platform with window contents being undefined immediately after a resize before we repaint then this change shouldn't make that situation any worse.

As an aside, I am working on a separate merge request to handle the problem of the window contents jumping around for user-generated resizes on X11.  Part of the approach currently involves setting the X11 background color of the window as well as its bit gravity so that the window's contents are always defined and don't jump around, but there is unfortunately I believe more required.

wez:
Thanks!

PR #5033, opened 2024-02-18
x11: fix window contents jumping on async resize

jknockel:
We can't hope to keep the renderer's idea of the window size consistent with the actual window size when the user can asynchronously change its size at any possible moment. Therefore, we create a child window which we render to instead and which we synchronously resize in response to user-generated, asynchronous resize events. Since by default the contents of X11 windows are undefined on resize, we take the additional steps of setting our windows' X11 background color and bit gravity as appropriate as well. These changes are intended to address the contents of the window jumping around in response to user-generated, asynchronous resizes.

I'm currently marking this as a draft because I am aware of one outstanding issue where we should probably be calling `eglWaitClient()` before resizes to the child window, and I am unsure how to idiomatically do that in this code!  In practice, not doing so doesn't seem to cause any issue in my testing, but I believe that there may be some race condition where the GPU is still rendering when we synchronously resize the child window where we could still end up seeing the window contents jumping around.  I believe that calling `eglWaitClient()` since the last OpenGL operations were performed would address this race condition unless there is something I'm failing to account for (like, I'm hoping that nothing is ever rendering to the window concurrently in a separate thread).  In the meantime, I've marked the two places where I believe `eglWaitClient()` calls should go in code comments so that it is hopefully clearer what I think is necessary to get this completely right.

### DIFF
--- a/window/src/os/x11/window.rs
+++ b/window/src/os/x11/window.rs
@@ -59,6 +59,7 @@ impl CopyAndPaste {
 
 pub(crate) struct XWindowInner {
     pub window_id: xcb::x::Window,
+    pub child_id: xcb::x::Window,
     conn: Weak<XConnection>,
     pub events: WindowEventSender,
     width: u16,
@@ -98,6 +99,9 @@ impl Drop for XWindowInner {
                     .context("flush pending requests prior to issuing DestroyWindow")
                     .ok();
                 conn.send_request_no_reply_log(&xcb::x::DestroyWindow {
+                    window: self.child_id,
+                });
+                conn.send_request_no_reply_log(&xcb::x::DestroyWindow {
                     window: self.window_id,
                 });
             }
@@ -120,7 +124,7 @@ unsafe impl HasRawDisplayHandle for XWindowInner {
 unsafe impl HasRawWindowHandle for XWindowInner {
     fn raw_window_handle(&self) -> RawWindowHandle {
         let mut handle = XcbWindowHandle::empty();
-        handle.window = self.window_id.resource_id();
+        handle.window = self.child_id.resource_id();
         handle.visual_id = self.conn.upgrade().unwrap().visual.visual_id();
         RawWindowHandle::Xcb(handle)
     }
@@ -133,11 +137,11 @@ impl XWindowInner {
         let gl_state = match conn.gl_connection.borrow().as_ref() {
             None => crate::egl::GlState::create(
                 Some(conn.conn.get_raw_dpy() as *const _),
-                self.window_id.resource_id() as *mut _,
+                self.child_id.resource_id() as *mut _,
             ),
             Some(glconn) => crate::egl::GlState::create_with_existing_connection(
                 glconn,
-                self.window_id.resource_id() as *mut _,
+                self.child_id.resource_id() as *mut _,
             ),
         };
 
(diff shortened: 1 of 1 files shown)


window/src/os/x11/window.rs
```diff
@@ -1754,15 +1801,23 @@ impl WindowOps for XWindow {
 
     fn set_inner_size(&self, width: usize, height: usize) {
         XConnection::with_window_inner(self.0, move |inner| {
-            inner
-                .conn()
-                .send_request_no_reply_log(&xcb::x::ConfigureWindow {
+            let conn = inner.conn();
+            // XXX: Can I call eglWaitClient() or ensure that it has already been called by now?
```
wez, line 1805: I'm not sure about this specific question, but if this window is in GL mode, rather than webgpu mode, then https://docs.rs/glium/latest/glium/struct.SyncFence.html#method.wait is the primitive that should be used, so presumably you'd want to do something with that. I'm unclear exactly what set of operations are being awaited and whether it is actually valid to wait on the fence here

jknockel: Oof, I've never used sync fences before, but they look like they could work.  All I would really like to do is wait until OpenGL is done drawing and swapping buffers so that I can safely resize the window.  `eglWaitClient()` has the nice semantics of guaranteeing exactly that, although if I can call `glFinish()` then that should also work.

And webgpu... I don't know what the equivalents of any of these things would be in that, but I imagine that there must be an equivalent of at least one of them... :thinking:

jknockel:
I've managed to convince myself that `eglWaitClient()` or some analog is not required before resizing. For one thing, I can't reproduce any race without that synchronization. Furthermore, there are many authoritative sources which only synchronize after the resize but not before. The original GLUT3 library [appears](https://github.com/markkilgard/glut/blob/8cd96cb440f1f2fac3a154227937be39d06efa53/lib/glut/glut_event.c#L373) to only call `glXWaitX()` after resizing but not `glXWaitGL()` before resizing, `glXWaitX()` being at worst an `XSync()` and `glXWaitGL()` being the GLX equivalent of `eglWaitClient()`. Moreover, the "OpenGL® on Silicon Graphics Systems" guide [states](https://irix7.com/techpubs/007-2392-003.pdf):

> Note: The X and OpenGL command streams are asynchronous, meaning that the order
> in which OpenGL and X commands complete is not strictly defined. In a few cases, it is
> important to explicitly synchronize X and OpenGL command completion. For example,
> if an X call is used to resize a window within a widget program, call glXWaitX() before
> calling glViewport() to ensure that the window resize operation is complete.

The guide makes no mention of `glXWaitGL()`, and a later code sample in the guide only calls `glXWaitX()`.

As such, I think I am happy with the current level of synchronization, and so I will mark this merge request ready for review. :)

Of course, I haven't mentioned webgpu at all, but I cannot reproduce any race using webgpu either, and I suspect (hope?) that a similar reasoning would apply for why this merge request's current level of synchronization is sufficient for webgpu as well.

jknockel:
I just realized that this approach breaks transparent backgrounds.  But perhaps if I set the parent window to be fully transparent...

jknockel:
> I just realized that this approach breaks transparent backgrounds. But perhaps if I set the parent window to be fully transparent...

Yes this appears to work.

wez:
Thanks!

PR #5034, opened 2024-02-19
Add notes about cask and quarantine issues with MacOS

joshuataylor:
I forget about this every time. Should also help with search engines, as the error is pretty common.

Thanks MacOS 🤦

### DIFF
--- a/docs/install/macos.md
+++ b/docs/install/macos.md
@@ -34,8 +34,20 @@ If you'd like to use a nightly build:
 
 ```console
 $ brew tap homebrew/cask-versions
-$ brew install --cask wezterm-nightly
+$ brew install --cask wez/wezterm/wezterm-nightly --no-quarantine
 ```
+
+> 1. `--cask` needs to be set, as the cask name exists in homebrew already and will result in the following error:
+> ```
+> Error: Cask wezterm-nightly exists in multiple taps:
+  homebrew/cask-versions/wezterm-nightly
+  wez/wezterm/wezterm-nightly
+  ```
+> 2. `--no-quarantine` is set, otherwise you will receive the warning that Wezterm cannot be opened because the developer cannot be verified. This message is from [Gatekeeper](https://support.apple.com/en-mide/102445).
+> 3. To remove the above, you can remove the quarantine flag by using (-r is recursive, -d is delete, .app is a directory on MacOS):
+> ```sh
+xattr -rd com.apple.FinderInfo /Applications/WezTerm.app
+````
 
 to upgrade to a newer nightly (normal `brew upgrade` will not upgrade it!):
 


docs/install/macos.md
```diff
@@ -34,9 +34,21 @@ If you'd like to use a nightly build:
...
+$ brew install --cask wez/wezterm/wezterm-nightly --no-quarantine
 ```
 
+> 1. `--cask` needs to be set, as the cask name exists in homebrew already and will result in the following error:
+> ```
+> Error: Cask wezterm-nightly exists in multiple taps:
+  homebrew/cask-versions/wezterm-nightly
+  wez/wezterm/wezterm-nightly
+  ```
+> 2. `--no-quarantine` is set, otherwise you will receive the warning that Wezterm cannot be opened because the developer cannot be verified. This message is from [Gatekeeper](https://support.apple.com/en-mide/102445).
```
wez, line 46: What I would expect here is just a warning that you downloaded something from the internet.
wezterm is signed; if are you seeing a warning that the package you are installing is not signed or not verifiable then something has gone wrong somewhere and you should not use the package.

joshuataylor: I received this when installing nightly, is nightly not signed or is that a generic warning of when you install something via the internet?

Happy to test further on a clean machine if needed. :+1:

wez: all builds of wezterm produced by my CI on macOS are signed using my developer identity. Please share the wording from the message that you see when you don't use the quarantine option.

joshuataylor: Without:
```
“WezTerm” is an app downloaded from the Internet. Are you sure you want to open it?

Homebrew Cask downloaded this file today at 2:24PM. Apple checked it for malicious software and none was detected.
```

`brew uninstall --cask wezterm-nightly` and then install with just that flag fixes it.

If I then do another uninstall and try without the flag, the same message appears.

I've tested this in a virtualised MacOS using [tart](https://tart.run) and observe the same results.

Let me know if this makes sense..

SCR-20240219-tjnn

matthewberryman: > otherwise you will receive the warning that Wezterm cannot be opened because the developer cannot be verified

This wording is wrong, the developer can be verified, it is because you don't have the Security setting to allow applications downloaded from "App Store *and* identified developers set (per screenshot), that you are seeing the warning, and the `--no-quarantine` flag, which also bypasses this particular GateKeeper check along with the others. 

I would argue, especially if you are installing brew casks, that you would want to change the setting, rather than use `--no-quarantine`, which would bypass other checks including unsigned binaries.

Screenshot 2024-02-20 at 08 42 34

joshuataylor:
Agreed, let me reword it.

PR #5035, opened 2024-02-19
[feature] tab min width

matthewberryman:
I find that after my stroke I am a bit lacking in fine motor coordination, so in both the fancy and retro tab modes the minimum tab width as is a bit small, and so am often closing tabs by accident (when no process is running, anyway) and have difficulty navigating between tabs.

This change adds a minimum tab width to both tab modes (which defaults to the previous hardcoded minimum for retro mode).

Being a bit new to rust, I haven't written any tests, but this seems to be a thing which is possibly hard to test, and `tab_max_width` seems to be untested (at least I did a bit of a search for tests using `tab_max_width` and couldn't find any, though happy to work on tests if someone can give me a bit of a pointer to how to go about it.


kenchou:
I also really need this feature. plz
@wez

docs/config/lua/config/tab_min_width.md
```diff
@@ -0,0 +1,14 @@
...
+---
+# `tab_min_width`
+
+Specifies the minimum width that a tab can have in the
+tab bar when using either the of the tab modes.  
+
+Defaults to 5 glyphs in width.
+
+```lua
+config.tab_max_width = 5
```
mjs, line 13: This should be `tab_min_width`

IceSentry:
Is this blocked by something?

rrojan (approved):
LGTM :trophy: 
Why is this PR blocked though... @matthewberryman

matthewberryman:
> LGTM 🏆 Why is this PR blocked though... @matthewberryman

A question for @wez

dkulchinsky:
Would love to see this getting merged & released

@wez 🙏🏼

xpufx:
I believe this can be kind of emulated with pad_right and pad_left, which might be why this wasn't merged.

krasoffski:
I also would like to have `tab_min_width`.

HaaaaE:
@wez plz

mmcc007:
Another data point for why a minimum width matters, from a 1360px window with 24 tabs
(fancy tab bar, `show_close_tab_button_in_tabs = false`, `Liberation Sans Narrow`):

| tab bar `font_size` | characters visible per tab |
|---|---|
| 7pt  | all names fit, none legible |
| 8pt  | ~6-7 |
| 10pt | ~4 |
| 11pt | ~2-3 |

Because the width per tab is just `window_width / tab_count`, every increase in
legibility costs characters one-for-one, and there is no size that is both readable
and shows a useful part of the name.

Two things I tried that I think are worth recording, since both look like they should
work and neither does:

- **Shrinking distant tabs in `format-tab-title`** so tabs near the active one could
  take the freed width. When tabs do not fit, every tab is given the *same* width — a
  2-character label is padded out to the full tab width and the space is not handed
  back. Content-proportional sizing seems to apply only when everything already fits.
- **Expanding the label on hover.** `format-tab-title` does receive a `hover` flag, but
  a hovered tab is only highlighted, never widened (compared screenshots with the
  pointer driven onto a tab). There is no tooltip primitive either.

So without `tab_min_width` there is currently no configuration that keeps labels
legible at high tab counts. I have opened #8052 for the follow-on question of what the
bar should do once tabs stop shrinking (overflow + horizontal scrolling) — that only
becomes meaningful once this lands.

bew (commented):
Hello! For this feature I think we'll need to cater for the case where tabs don't fit in the window width, so we'll need to way to scroll through the items in the bar 🤔

(note: I'm not sure how to impl this right now)

docs/config/lua/config/tab_min_width.md
```diff
@@ -0,0 +1,14 @@
+---
+tags:
+  - tab_bar
+---
+# `tab_min_width`
+
+Specifies the minimum width that a tab can have in the
+tab bar when using either the of the tab modes.
```
bew: Missing word? What modes are you talking about?

docs/config/lua/config/tab_min_width.md
```diff
@@ -0,0 +1,14 @@
+---
+tags:
+  - tab_bar
+---
+# `tab_min_width`
```
bew: Put the default in the title
```suggestion
# `tab_min_width = 5`
```

docs/config/lua/config/tab_min_width.md
```diff
@@ -0,0 +1,14 @@
...
+
+Specifies the minimum width that a tab can have in the
+tab bar when using either the of the tab modes.  
+
+Defaults to 5 glyphs in width.
+
+```lua
+config.tab_min_width = 5
+```
```
bew: Need 'see also' link to `tab_max_width`

PR #5050, opened 2024-02-21
ci: include PopOS on deploy file

josemiguelo:
When running the `ci/deploy.sh` script on a PopOS-powered  machine,  such script don't finish the deb package assembly since Pop it's not included on the distro options case:

image

Just adding the case to finish the build


wez:
Thanks!

PR #5057, opened 2024-02-22
fix: Swap horizontally/vertically split pane nomenclature

wokoman:
I checked multiple times that I'm not crazy, but it really seems that `horizontally` and `vertically` are swapped for pane splitting.

Splitting something **horizontally** means dividing it by a line going from left to right, resulting in top and bottom halves.
Splitting something **vertically** means dividing it by a line going from top to bottom, resulting in left and right halves.
Right? [Sauce.](https://english.stackexchange.com/questions/293520/split-horizontally-or-vertically-which-one-is-which) 🥸

Hopefully I found all the right places to change this.

Ran `rustup component add rustfmt-preview && cargo test --all && cargo fmt --all` with no issues. Thanks Wez!


wez:
This is a tricky one because depending on whether you think that a horizontal split operation happens *around* the horizon or *along* the horizon, whatever we pick for this is going to feel wrong for someone.

The nomenclature and effect used in wezterm was borrowed from iterm, and I agree that it can feel weird. That is why the associated text talks about where the resulting split panes end up in the more recently added operations, and why I would suggest the use of `--left`, `--right`, `--top`, `--bottom` in preference to `--horizontal` or `--vertical`.

I think changing the label for a horizontal split to talk about vertically splitting (for the folks that think it is opposite than the current situation) is even more confusing for everyone, so I'm not in favor of this change, and the obvious "let's just swap the vertical and horizontal labels" is 100% off the table because that would break the configurations of a lot of users, as well as be confusing for the other group of people who think the orientation is correct.

RioNight:
For what it's worth, in vim, a vertical split splits the window into a right window and left window. This is sort of opposite to what Wezterm does.

wokoman:
@wez Got it, thanks for the explanation. I'll just live with it, that I'm in the other camp. Closing 🫡

00sapo:
Sounds weird to me as well: in the whole field of linux, a vertical split is what wezterm names "horizontal", and viceversa. No reason to keep it like this in the long term, imho

PR #5073, opened 2024-02-25
feat(nix): package macos application bundle when building flake

smsegal:
Previously, the nix flake was missing the macOS packaging step that actually produces the `.app` bundle. 

I copied this snippet from the nixpkgs wezterm source here: https://github.com/NixOS/nixpkgs/blob/73de017ef2d18a04ac4bfd0c02650007ccb31c2a/pkgs/applications/terminal-emulators/wezterm/default.nix#L114-L121

I can verify that building on macOS with this added produces a runnable application bundle.


smsegal:
Bumping. Anything I can do to get this merged?

PR #5087, opened 2024-02-27
No need --cask flags anymore

Goooler:
No need to add it for Brew installation/uninstallation now.

### DIFF
--- a/ci/wezterm-homebrew-macos.rb.template
+++ b/ci/wezterm-homebrew-macos.rb.template
@@ -46,7 +46,7 @@ cask "wezterm" do
       /usr/local/bin/    for x86 Mac,
       /opt/homebrew/bin/ for M1 Mac.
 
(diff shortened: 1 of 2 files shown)


wez:
are you sure about this? ISTR that there was some ambiguity in the past and that it was in fact necessary to use `--cask` to guarantee that wezterm's authentic CI-produced and signed binaries were installed

Goooler:
I installed wezterm without this option from official cask repo.

rawnly:
According to homebrew docs this is needed

https://formulae.brew.sh/cask/1password
https://formulae.brew.sh/cask/wezterm

Goooler:
This is necessary if there is both a cask and a formula present.

wez:
We're going to keep the flag because I feel like there have been at least two different occasions where we've been bitten by not having it.

PR #5092, opened 2024-02-28
fix: fix feature = "cargo-clippy" deprecation

rex4539:
https://blog.rust-lang.org/2024/02/28/Clippy-deprecating-feature-cargo-clippy.html

### DIFF
--- a/codec/src/lib.rs
+++ b/codec/src/lib.rs
@@ -9,7 +9,7 @@
 //! of this code; in this way the client and server can more gracefully
 //! manage unknown enum variants.
 #![allow(dead_code)]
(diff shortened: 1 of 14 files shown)


codec/src/lib.rs
```diff
@@ -9,7 +9,7 @@
 //! of this code; in this way the client and server can more gracefully
 //! manage unknown enum variants.
 #![allow(dead_code)]
-#![cfg_attr(feature = "cargo-clippy", allow(clippy::range_plus_one))]
+#![cfg_attr(clippy, allow(clippy::range_plus_one))]
```
wez, line 12: From the docs, it sounds like all the cfg_attr stuff can be removed?

```suggestion
#![allow(clippy::range_plus_one)]
```

Issue #4455, opened 2023-10-17
Mouse scroll speed insanly fast


exoess:
### What Operating System(s) are you seeing this problem on?

Linux Wayland

### Which Wayland compositor or X11 Window manager(s) are you using?

Hyprland

### WezTerm version

wezterm 20231004-113117-11dec45f

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

I barely touch the mouse and it'll jump to the bottom or top of the buffer in an instant, I already adjusted the `alternate_buffer_wheel_scroll_speed` and it changed nothing

### Configuration

`config.alternate_buffer_wheel_scroll_speed = 1`

### Expected Behavior

normal scrolling like any other terminal

bew:
Are you using a separate mouse or a touchpad?

exoess:
no, I'm using my laptops touchpad

bew:
This might be a duplicate of #3142 then

exoess:
oh, okay

exoess:
is there a fix yet?

bew:
Doesn't seem like it, I don't use Wayland personally.

I'd recommend you to close this issue as a duplicate.

exoess:
okay

Issue #3142, opened 2023-02-21
Scrolling too fast in Wayland


hgaiser:
### What Operating System(s) are you seeing this problem on?

Linux Wayland

### Which Wayland compositor or X11 Window manager(s) are you using?

Gnome

### WezTerm version

20230217-120342-3b39aa55

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

Scrolling under Wayland is significantly faster than scrolling on X11. Presumably this is because of differences between `libinput` and `xf86-input-libinput`. A full scroll from top to bottom on my laptops touchpad results in roughly 300 lines scrolling on Wayland, while it scrolls 70 lines on X11.

### To Reproduce

Scroll in Wayland and compare it to X11.

### Configuration

no config

snaggen:
Just did a comparison with VTE based Tilix and my mouse scroll wheel which have discrete scroll ticks. 
Under tilix one tick scroll 3 lines. Under Wezterm one tick scroll 10 lines. This is under Fedora Gnome using Wayland. I have not tested this under X11. 

I actually like the more aggressive scroll. However, it might be a trackpad under Wayland thing, so my comparison is irrelevant. Just thought I point out this for information.

hgaiser:
10 lines per tick seems like quite a lot, I'll test today what I'm getting with a scroll wheel. I've used tilix before as well, I thought it didn't scroll discrete lines?

I should've mentioned that the lines scrolled per tick on my touchpad seem the same in Wayland vs X11, but the distance traveled on my touchpad between ticks feels different. On Wayland I need to move my fingers significantly less before a scroll event is triggered.

Maybe I'll see if I can measure how many ticks are generated when moving from top to bottom in both Wayland and X11.

hgaiser:
I added a log statement [here](https://github.com/wez/wezterm/blob/e6ffeaf28573c26299e3de9f0b007d1a8f6d9466/wezterm-gui/src/termwindow/mod.rs#L2253) to test a few things, these are my results:

1. On X11, every scroll event scrolls for exactly 5 lines, whether it's a scroll wheel or touchpad event doesn't matter. Scrolling over the entire vertical space of my touchpad generated 25 ticks, so 125 lines were scrolled.
2. On Wayland:
   1. Scroll wheel ticks scroll for exactly 10 lines.
[...]

PR #5104, opened 2024-03-03
Add Wayland Scroll Factor

mike-lloyd03:
Wayland users are reporting that mouse and trackpad scrolling is much to fast under Wayland.

#3142
#4455

Using @SuperSandro2000's recommendation [here](https://github.com/wez/wezterm/issues/3142#issuecomment-1821787347), I've added a config option to modify the scroll factor on Wayland (`config.wayland_scroll_factor`).

I've found that setting this value to `0.25` is ideal on my environment (Hyprland, Zellij).

I haven't modified the documentation yet because I wanted to see if this was a desirable approach first. I suppose it's possible to derive a suitable scroll factor from the window's scale factor, but I do not have enough familiarity with the Wayland protocol to implement this. If the config option route is acceptable, I will update the docs.

This also fixes a typo in `CONTRIBUTING.md` as proof that I actually read it.

### DIFF
--- a/window/src/os/wayland/window.rs
+++ b/window/src/os/wayland/window.rs
@@ -685,7 +685,7 @@ impl WaylandWindowInner {
         }
 
         if let Some((value_x, value_y)) = PendingMouse::scroll(&pending_mouse) {
-            let factor = self.get_dpi_factor() as f64;
+            let factor = self.get_dpi_factor() as f64 * self.config.wayland_scroll_factor;
 
             if value_x.signum() != self.hscroll_remainder.signum() {
                 // reset accumulator when changing scroll direction
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -111,7 +111,7 @@ to reviewing the code.
 
 ## Submitting a Pull Request
 
-After considering all of the above, and once you're prepared your contribution
+After considering all of the above, and once you've prepared your contribution
 and are ready to submit it, you'll need to create a pull request.
 
 If you're new to GitHub pull requests, read through
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -839,6 +839,9 @@ pub struct Config {
 
     #[dynamic(default = "default_ulimit_nproc")]
     pub ulimit_nproc: u64,
+
+    #[dynamic(default = "default_wayland_scroll_factor")]
+    pub wayland_scroll_factor: f64,
 }
 impl_lua_conversion_dynamic!(Config);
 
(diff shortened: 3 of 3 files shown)


wez:
Thanks for diving in!

I'm a bit reticent to merge this because what I suspect is that there is something else going on in how the mouse scroll events are interpreted.  This commit is certainly an easy fix but I think that it could eg: produce different results with different compositors, or when different devices are attached.

I think a little bit more research is needed. For example, are there other Rust wayland apps that scroll more reasonably? If so, what are they doing to process the scroll events?

SuperSandro2000:
Not sure if that helps, but the only other program where I encountered similar behavior is VSCode but there you can manually specific the scroll factor.

valpackett:
The setting shouldn't be specific to any display backend, but rather universally available. A terminal is quite a particular special case, so it's to be expected that users might want faster/slower scrolling there!

> what are they doing to process the scroll events?

Firefox for example can be configured to interpret touchpad scroll deltas either as lines (in which case it basically uses the raw value, there's a configurable scale factor but it defaults to 1.0; it is generally way-too-fast out of the box, and still the default on non-nightly (ugh)), and as pixels (in which case it multiplies the value by both a configurable factor that defaults to 40.0, [and by the window scale](https://searchfox.org/mozilla-central/rev/f9139f56bbcc0d587966c007178910ea31df7d58/widget/gtk/nsWindow.cpp#5134)).

gptlang:
Would just a rename of the config work in this case? Being able to configure scroll sensitivity is quite useful beyond just wayland & without weird speed bugs

gptlang:
I can't believe this still hasn't been merged. I've been using `foot` because of this. Might just run from the fork

gptlang:
This doesn't work in neovim.

FSund:
> I think a little bit more research is needed. For example, are there other Rust wayland apps that scroll more reasonably? If so, what are they doing to process the scroll events?

A natural comparison might be Alacritty
- [input.rs/mouse_wheel_input](https://github.com/alacritty/alacritty/blob/03c2907b44b4189aac5fdeaea331f5aab5c7072e/alacritty/src/input/mod.rs#L706)
- [event.rs/scroll](https://github.com/alacritty/alacritty/blob/03c2907b44b4189aac5fdeaea331f5aab5c7072e/alacritty/src/event.rs#L667)

window/src/os/wayland/window.rs
```diff
@@ -685,7 +685,7 @@ impl WaylandWindowInner {
         }
 
         if let Some((value_x, value_y)) = PendingMouse::scroll(&pending_mouse) {
-            let factor = self.get_dpi_factor() as f64;
+            let factor = self.get_dpi_factor() as f64 * self.config.wayland_scroll_factor;
```
gitworkflows, line 688:
```suggestion
            let dpi_factor = self.get_dpi_factor() as f64;
            let scroll_factor = dpi_factor * self.config.wayland_scroll_factor;
```

gitworkflows (changes requested):
Improve scroll factor calculation

PR #5135, opened 2024-03-09
fix #4985 - reimplement get_appearance for wayland

### LINKED ISSUE #4985: get_appearance toggling skips first change when starting in dark mode

Linux Wayland

Sway

### WezTerm version

20240203-110809-5046fc22

Yes, the toggling didn't work at all.

### Describe the bug

I'm using `get_appearance` as shown [here](https://wezfurlong.org/wezterm/config/lua/window/get_appearance.html?h=get_app) to toggle between a like and dark theme.  When I start a new process (i.e., when there aren't any wezterm windows open or if I run  `wezterm start --always-new-process`) and I am already in dark mode, switching my desktop to light mode does not change wezterm's appearance. If I then switch back to dark mode, and switch again to light mode, wezterm's appearance is updated. Subsequent changes work correctly.

This does not happen when I start wezterm and am already in light mode. In this instance, changing to dark mode immediately switches to dark mode, and subsequent changes work correctly.

I am using `darkman` as my colors-scheme switcher.

### To Reproduce

This is easily reproduced  under sway using darkman.

- `darkmanctl set dark` to get into dark mode
- With no wezterm instances running, run wezterm
- toggle to light mode using `darkmanctl toggle` (nothing will happen to wezterm)
- toggle back to dark mode with `darkmanctl toggle`
- toggle once more to light mode with `darkmanctl toggle` (wezterm is now using the light theme)

### Configuration

```lua
wezterm.add_to_config_reload_watch_list(wezterm.config_dir)

function get_appearance()
  if wezterm.gui then
    return wezterm.gui.get_appearance()
  end
  return 'Dark'
end

function scheme_for_appearance(appearance)
  if appearance:find 'Dark' then
    return 'Modus-Vivendi'
  else
    return 'Modus-Operandi-Tinted'
  end
end

function weight_for_appearance(appearance)
   if appearance:find 'Dark' then
      return "Medium"
   else
      return "Light"
   end
end

return {
  window_decorations = 'RESIZE',
  term = 'wezterm',
  enable_wayland = true,
  color_scheme = scheme_for_appearance(get_appearance()),
  font = wezterm.font {
     family = "Iosevka Comfy Motion",
     weight = weight_for_appearance(get_appearance()),
     harfbuzz_features = { "calt" },
  },
  font_size = 11,
  use_fancy_tab_bar = false,
}
```

### Expected Behavior

The theme is changed to light theme immediately

### Logs

Nothing shows up in the logs

### Anything else?

I tried this using nightly, but nightly didn't seem to work at all with `get_appearance`. The theme was always the light theme no matter what I did. I tried this on the nightly appimage and by building it from source.

MartinNowak:
- was dropped with wayland reimplementation (#4777, 3eaba4e3d6)
- get appearance from xdg desktop portal
- advise all windows of appearance changes to reload config

fixes #4985

### DIFF
--- a/window/src/os/wayland/connection.rs
+++ b/window/src/os/wayland/connection.rs
@@ -13,7 +13,7 @@ use wayland_client::{Connection as WConnection, EventQueue};
 
 use crate::screen::{ScreenInfo, Screens};
 use crate::spawn::SPAWN_QUEUE;
-use crate::{Connection, ConnectionOps, ScreenRect};
+use crate::{Appearance, Connection, ConnectionOps, ScreenRect};
 
 use super::state::WaylandState;
 use super::WaylandWindowInner;
@@ -46,7 +46,11 @@ impl WaylandConnection {
         Ok(wayland_connection)
     }
 
-    pub(crate) fn advise_of_appearance_change(&self, _appearance: crate::Appearance) {}
+    pub(crate) fn advise_of_appearance_change(&self, appearance: crate::Appearance) {
+        for win in self.wayland_state.borrow().windows.borrow().values() {
+            win.borrow_mut().appearance_changed(appearance);
+        }
+    }
 
     fn run_message_loop_impl(&self) -> anyhow::Result<()> {
         const TOK_WL: usize = 0xffff_fffc;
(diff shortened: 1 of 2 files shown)


wez:
Thank you!

PR #5150, opened 2024-03-11
Add missing docs, fix typo, bring to attention error.

gwww:
For `appearance.md` the URL no longer exists. I put link to the synced color schemes, but don't think that's the "right" approach. At least you know about it now and can decide.

For `window_frame.md` added `font` and `font_size`. The docs now match how I understand them. Hope that I've got it right.

Lastly, fixed a typo in `swallow_mouse_click_on_focus.md`.


wez:
Looks good, thank you!

Issue #4503, opened 2023-10-29
Parsing of ProxyCommand doesn't use Hostname with %h


oceanusxiv:
### What Operating System(s) are you seeing this problem on?

macOS

### WezTerm version

20230712-072601-f4abf8fd

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

No, and I'll explain why below

### Describe the bug

When wezterm reads the SSH config file's ProxyCommand, it substitutes `%h` with the host, instead of the host name specified by the `HostName` field of the config. For particularly exotic proxy commands this can cause issues.

I believe the relevant line is 
https://github.com/wez/wezterm/blob/fec90ae04bf448d4b1475ba1d0ba1392846a70d6/wezterm-ssh/src/config.rs#L535

### To Reproduce

Run wezterm in verbose `wezterm ssh -v username@host` with some config file setup in `~/.ssh/config` like such

```
Host host
    ProxyCommand sh -c "some_command %h"
    HostName different_name
    User ubuntu
```

and observe the resulting proxy command being issued, which would most likely be
`sh -c "some_command host"` instead of the more correct `sh -c "some_command different_name"`.

### Configuration

no config

### Expected Behavior

The substitution of `%h` should use the `HostName` specified in the config file for that target (if it was specified).

wheatdog:
References:

https://github.com/openssh/openssh-portable/blob/88351eca17dcc55189991ba60e50819b6d4193c1/ssh.c#L1215-L1221

https://github.com/openssh/openssh-portable/blob/88351eca17dcc55189991ba60e50819b6d4193c1/ssh.c#L1432

https://github.com/openssh/openssh-portable/blob/master/sshconnect.h#L63

PR #5163, opened 2024-03-14
Fix SSH config %h using Host

emc2314:
This commit fix issue #4503

### DIFF
--- a/wezterm-ssh/src/config.rs
+++ b/wezterm-ssh/src/config.rs
@@ -532,7 +532,13 @@ impl Config {
         }
 
         let mut token_map = self.tokens.clone();
-        token_map.insert("%h".to_string(), host.to_string());
+        token_map.insert(
+            "%h".to_string(),
+            result
+                .get("hostname")
+                .unwrap_or(&host.to_string())
+                .to_string(),
+        );
         token_map.insert("%n".to_string(), host.to_string());
         token_map.insert("%r".to_string(), target_user.to_string());
         token_map.insert(
@@ -612,7 +618,6 @@ impl Config {
             | "localforward" | "remotecommand" | "remoteforward" | "userknownkostsfile" => {
                 Some(&["%C", "%d", "%h", "%i", "%L", "%l", "%n", "%p", "%r", "%u"])
             }
-            "hostname" => Some(&["%h"]),
             "localcommand" => Some(&[
                 "%C", "%d", "%h", "%i", "%k", "%L", "%l", "%n", "%p", "%r", "%T", "%u",
             ]),
--- a/wezterm-ssh/src/config.rs
+++ b/wezterm-ssh/src/config.rs
@@ -532,13 +532,7 @@ impl Config {
         }
 
         let mut token_map = self.tokens.clone();
-        token_map.insert(
-            "%h".to_string(),
-            result
-                .get("hostname")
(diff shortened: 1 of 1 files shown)


wez:
Thanks for this; please add a test case, because it isn't obvious why part of this PR is removing some code, which is a great sign that there needs to be more test coverage!

emc2314:
Sorry I didn't check the test case and I didn't realize that to fix issue #4503 actually requires two-phase parsing which is not supported yet. I tried to modify the ```token_map``` during single-phase parsing but I don't know whether it's appropriate. Thanks for reviewing.

wheatdog:
I think we still need to update `token_map["%h"]` right after resolving hostname. How about this implementation?

```
diff --git a/wezterm-ssh/src/config.rs b/wezterm-ssh/src/config.rs
index 0ccba7572..81e4e95b8 100644
--- a/wezterm-ssh/src/config.rs
+++ b/wezterm-ssh/src/config.rs
@@ -532,7 +532,18 @@ impl Config {
         }
 
         let mut token_map = self.tokens.clone();
+
         token_map.insert("%h".to_string(), host.to_string());
+        result
+            .entry("hostname".to_string())
+            .and_modify(|curr| {
+                if let Some(tokens) = self.should_expand_tokens("hostname") {
+                    self.expand_tokens(curr, tokens, &token_map);
+                }
+            })
+            .or_insert_with(|| host.to_string());
+        token_map.insert("%h".to_string(), result["hostname"].to_string());
+
         token_map.insert("%n".to_string(), host.to_string());
         token_map.insert("%r".to_string(), target_user.to_string());
         token_map.insert(
@@ -553,10 +564,6 @@ impl Config {
             }
         }
 
-        result
-            .entry("hostname".to_string())
-            .or_insert_with(|| host.to_string());
-
         result
             .entry("port".to_string())
             .or_insert_with(|| "22".to_string());

```

emc2314:
> I think we still need to update `token_map["%h"]` as early as we can. How about this implementation?
> 
> ```
> diff --git a/wezterm-ssh/src/config.rs b/wezterm-ssh/src/config.rs
> index 0ccba7572..81e4e95b8 100644
> --- a/wezterm-ssh/src/config.rs
> +++ b/wezterm-ssh/src/config.rs
> @@ -532,7 +532,18 @@ impl Config {
>          }
>  
>          let mut token_map = self.tokens.clone();
> +
>          token_map.insert("%h".to_string(), host.to_string());
> +        result
> +            .entry("hostname".to_string())
> +            .and_modify(|curr| {
> +                if let Some(tokens) = self.should_expand_tokens("hostname") {
> +                    self.expand_tokens(curr, tokens, &token_map);
> +                }
> +            })
> +            .or_insert_with(|| host.to_string());
> +        token_map.insert("%h".to_string(), result["hostname"].to_string());
> +
>          token_map.insert("%n".to_string(), host.to_string());
>          token_map.insert("%r".to_string(), target_user.to_string());
>          token_map.insert(
> @@ -553,10 +564,6 @@ impl Config {
>              }
>          }
>  
> -        result
> -            .entry("hostname".to_string())
> -            .or_insert_with(|| host.to_string());
> -
>          result
>              .entry("port".to_string())
>              .or_insert_with(|| "22".to_string());
> ```

Thanks for your guidance! I totally agree with you.

wez:
Thanks!

PR #5173, opened 2024-03-15
Add simple implementation to remove_capabilities

mangas:
Prevents a crash when the only keyboard available is removed which would hit the todo()! and panic

### DIFF
--- a/window/src/os/wayland/seat.rs
+++ b/window/src/os/wayland/seat.rs
@@ -65,9 +65,11 @@ impl SeatHandler for WaylandState {
         _conn: &Connection,
         _qh: &QueueHandle<Self>,
         _seat: WlSeat,
-        _capability: smithay_client_toolkit::seat::Capability,
+        capability: smithay_client_toolkit::seat::Capability,
     ) {
-        todo!()
+        if capability == Capability::Keyboard && self.keyboard.is_some() {
+            self.keyboard = None;
+        }
     }
 
     fn remove_seat(&mut self, _conn: &Connection, _qh: &QueueHandle<Self>, _seat: WlSeat) {


jokeyrhyme:
Should we keep the TODO in there for when the capability is not a keyboard? Or do we only ever deal with keyboard capabilities?

window/src/os/wayland/seat.rs
```diff
@@ -65,9 +65,11 @@ impl SeatHandler for WaylandState {
         _conn: &Connection,
         _qh: &QueueHandle<Self>,
         _seat: WlSeat,
-        _capability: smithay_client_toolkit::seat::Capability,
+        capability: smithay_client_toolkit::seat::Capability,
     ) {
-        todo!()
+        if capability == Capability::Keyboard && self.keyboard.is_some() {
+            self.keyboard = None;
+        }
```
jokeyrhyme, line 72: Should we keep the TODO in there for when the capability is not a keyboard? Or do we only ever deal with keyboard capabilities?

e.g.

```
if capability == Capability::Keyboard {
    if self.keyboard.is_some() {
        self.keyboard = None;
    }
} else {
    todo!();
}
```

mangas: I don't know if there are more, according to the wayland spec you should use this hook to destroy and resources, from what I can tell, this was the only one we set that's why I added the removal as well. I don't think we should keep the todo! since that actually will crash the app. We could add some logging or something for an unhandled type of removal event.

woojiq:
I applied this patch and it seems to work. Thank you very much for taking this❤️

woojiq:
I noticed that I cannot resize or scroll inside the terminal after suspend. After I close wezterm and open a new instance it works again.

wez:
I think this is resolved by https://github.com/wez/wezterm/pull/5276, so closing this out.
Thanks for taking a run at this!

PR #5184, opened 2024-03-18
feat(ssh/config): handle absolute path includes in ssh config

xxxbrian:
Absolute path includes are currently not supported by the ssh config. For example, an Include like `Include ~/.orbstack/ssh/config` would be ignored. This PR adds support for absolute path processing while ensuring that wildcard support is still maintained.

### DIFF
--- a/wezterm-ssh/src/config.rs
+++ b/wezterm-ssh/src/config.rs
@@ -175,7 +175,13 @@ impl ParsedConfigFile {
         groups: &mut Vec<MatchGroup>,
         loaded_files: &mut Vec<PathBuf>,
     ) {
-        match filenamegen::Glob::new(&pattern) {
+        let home = dirs_next::home_dir().unwrap_or_default();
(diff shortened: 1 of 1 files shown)


wezterm-ssh/src/config.rs
```diff
@@ -175,7 +175,13 @@ impl ParsedConfigFile {
         groups: &mut Vec<MatchGroup>,
         loaded_files: &mut Vec<PathBuf>,
     ) {
-        match filenamegen::Glob::new(&pattern) {
+        let home = dirs_next::home_dir().unwrap_or_default();
```
wez, line 178: Thanks for diving in!

* Please make the `home_dir` resolution only trigger when we observe the `~/` prefix.
* I'm not clear on why the cwd overrides the caller-supplied value for either the `~` or `/` case; that's something that would benefit from some explanation in the code with comments.
* Please add unit tests for these new cases! You can find various examples towards the bottom of this file

PR #5206, opened 2024-03-23
Fix various cargo audit errors

Pi-Cla:
Compare the cargo audit output [before](https://github.com/wez/wezterm/files/14730281/cargo-audit-old.txt) and [after](https://github.com/wez/wezterm/files/14730282/cargo-audit.txt). (The remaining errors can't be fixed until the authors of these dependencies update their own dependencies and release a new version)

I ran `cargo update` which on it's own fixed [RUSTSEC-2024-0019](https://rustsec.org/advisories/RUSTSEC-2024-0019) and [RUSTSEC-2024-0020](https://rustsec.org/advisories/RUSTSEC-2024-0020) which are serious vulnerabilities.

I bumped criterion to 0.5 so that we stop transitively depending [atty which is unmaintained and has an unmerged fix for its vulnerability](https://rustsec.org/advisories/RUSTSEC-2021-0145).

I bumped k9 to 0.12 so that we stop transitively depending on [term_size which is unmaintained](https://rustsec.org/advisories/RUSTSEC-2020-0163).

Finally I bumped starship-battery to 0.8 so that we stop transitively depending on [mach which is also unmaintained](https://rustsec.org/advisories/RUSTSEC-2020-0168).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -36,9 +36,9 @@ dependencies = [
 
 [[package]]
 name = "ahash"
-version = "0.8.8"
+version = "0.8.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "42cd52102d3df161c77a887b608d7a4897d7cc112886a9537b738a887a03aaff"
+checksum = "e89da841a80418a9b391ebaea17f5c112ffaaa96f621d2c285b5174da76b9011"
 dependencies = [
  "cfg-if",
  "getrandom",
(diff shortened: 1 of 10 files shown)


wez:
Thanks for this!

Please undo the whitespace changes to files that have no functional effect. They make it harder to merge (and importantly: revert later if needed!) independent commits.

I also want to note that that none of the rustsec advisories you listed are practical vulnerabilities to wezterm itself; we don't use any of the affected functionality.

wez:
Thanks; I rebased this and pushed it to main!

PR #5229, opened 2024-03-28
equivalent tab bar styling example with current api

### LINKED ISSUE #5219: tab_bar_style not applied

Linux Wayland

sway

### WezTerm version

### Describe the bug

The tab_bar_style isn't being applied, other things such as background color and font do work though.

The nightly version shows different behaviour:

```
Configuration Error: `active_tab_left` is not a valid TabBarStyle field.  There are too many alternatives to list here; consult the documentation!
`active_tab_right` is not a valid TabBarStyle field.  There are too many alternatives to list here; consult the documentation!
`inactive_tab_left` is not a valid TabBarStyle field.  There are too many alternatives to list here; consult the documentation!
`inactive_tab_right` is not a valid TabBarStyle field ...
```

### To Reproduce

_No response_

### Configuration

I have mostly used the config from the example in the docs:

```
    tab_bar_at_bottom = true,
    use_fancy_tab_bar = false,
    tab_max_width = 32,
    tab_bar_style = {
        active_tab_left = wezterm.format {
            { Background = { Color = '#0b0022' } },
            { Foreground = { Color = '#2b2042' } },
            { Text = SOLID_LEFT_ARROW },
        },
        active_tab_right = wezterm.format {
            { Background = { Color = '#0b0022' } },
            { Foreground = { Color = '#2b2042' } },
            { Text = SOLID_RIGHT_ARROW },
        },
        inactive_tab_left = wezterm.format {
            { Background = { Color = '#0b0022' } },
            { Foreground = { Color = '#1b1032' } },
            { Text = SOLID_LEFT_ARROW },
        },
        inactive_tab_right = wezterm.format {
            { Background = { Color = '#0b0022' } },
            { Foreground = { Color = '#1b1032' } },
            { Text = SOLID_RIGHT_ARROW },
        },
    },
```

### Expected Behavior

_No response_

### Logs

I can't see any logfiles or any output in the debug view.

### Anything else?

_No response_

saep:
Fixes #5219 

I've also been a little confused when I tried out the example from the docs. So, I've added an equivalent definition for the tab bar that works for the current version.


wez:
Thanks for this!

If the example in the docs doesn't work for the currently released version, please update the PR to update the example rather than adding a new one; that will help keep things less confusing overall in the docs.

saep:
I've moved the example to the top, so that it is outside the versioned sections and I've deleted the old one.

I've also tested the code again on version `20240203-110809-5046fc22` (relatively current nixos-unstable version).

Issue #5209, opened 2024-03-24
support for server-side cursors(i.e. hyprcursor)


deviantsemicolon:
**Is your feature request related to a problem? Please describe.**
I like my cursor on my system to be themed uniformly, and that doesn't seem to work with Wezterm, at least with Wayland disabled.

**Describe the solution you'd like**
I would like Wezterm to work with hyprcursor, as well as other future server-side cursor things.

**Describe alternatives you've considered**
An alternative could be making WezTerm work with xcursor or gtk theming

**Additional context**
https://blog.vaxry.net/articles/2024-cursors

wez:
Presumed fixed in `main` by https://github.com/wez/wezterm/pull/5276
Please re-open if it isn't.

Issue #4604, opened 2023-11-21
bug: `ERROR  window::os::wayland::frame > Unable to set cursor to left_ptr`  on start


gabyx:
### What Operating System(s) are you seeing this problem on?

Linux Wayland

### Which Wayland compositor or X11 Window manager(s) are you using?

sway 1.8.1

### WezTerm version

20230712-072601-f4abf8fd

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

Launching 
```shell
wezterm start -- zsh -c top
```
from an already opened `wezterm` inside sway starts `wezterm` in a new window but does not launch `top`.
It just enters the shell...

### To Reproduce

```
wezterm start --always-new-process  -- zsh -c top
```

### Configuration

```shell
-- Pull in the wezterm API
local wezterm = require("wezterm")
local act = wezterm.action
local mux = wezterm.mux

-- This table will hold the configuration.
local config = {}

-- In newer versions of wezterm, use the config_builder which will
-- help provide clearer error messages
if wezterm.config_builder then
    config = wezterm.config_builder()
end

wezterm.on("gui-startup", function()
    local tab, pane, window = mux.spawn_window({})
    window:gui_window():maximize()
end)

-- This is where you actually apply your config choices
config.color_scheme = "Cobalt2"
config.font = wezterm.font_with_fallback({
    { family = "JetBrainsMono Nerd Font", weight = "Bold" },
})
config.font_size = 12
config.window_decorations = "RESIZE"
config.hide_tab_bar_if_only_one_tab = true

config.enable_kitty_keyboard = true
config.debug_key_events = true -- Start `wezterm start --always-new-process` to see the keys
config.disable_default_key_bindings = true

config.leader = { key = "n", mods = "CTRL", timeout_milliseconds = 1000 }
config.keys = {
    -- Copy
    { key = "c", mods = "SHIFT|CTRL", action = act.CopyTo("Clipboard") },
    -- { key = "c", mods = "SUPER", action = act.CopyTo("Clipboard") },

    -- Paste
    { key = "v", mods = "SHIFT|CTRL", action = act.PasteFrom("Clipboard") },
    -- { key = "v", mods = "SUPER", action = act.PasteFrom("Clipboard") },

    -- Leader stuff
    { key = "p", mods = "LEADER",     action = act.ActivateCommandPalette },
[...]

Issue #3106, opened 2023-02-13
Cursor is either wrong or invisible


paveloom:
### What Operating System(s) are you seeing this problem on?

Linux Wayland

### Which Wayland compositor or X11 Window manager(s) are you using?

Mutter 43.2.

### WezTerm version

wezterm a5c2b1f3adb06054bf522cb3d350697938d6f8e9

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

Upon entering the window, the cursor either has a wrong style or is invisible. 

### To Reproduce

Put your cursor in the window.

### Configuration

I can reproduce the issue without a config by running `wezterm -n`.

### Expected Behavior

The cursor should be present and have a correct style. 

### Logs

```
Debug Overlay
wezterm version: a5c2b1f3adb06054bf522cb3d350697938d6f8e9 x86_64-unknown-linux-gnu
Window Environment: Wayland
OpenGL version: AMD Radeon Graphics (renoir, LLVM 15.0.7, DRM 3.49, 6.1.11) 4.6 (Compatibility Profile) Mesa 22.3.4
Enter lua statements or expressions and hit Enter.
Press ESC or CTRL-D to exit
23:08:52.226 ERROR window::os::wayland::pointer > Unable to set cursor to xterm: cursor not found
23:08:52.384 ERROR window::os::wayland::pointer > Unable to set cursor to xterm: cursor not found
```

Also getting the following errors when running `wezterm -n`:

```
23:08:57.021  ERROR  window::os::wayland::frame   > Unable to set cursor to left_ptr: cursor not found
23:08:57.021  ERROR  window::os::wayland::frame   > Unable to set cursor to right_side: cursor not found
23:08:57.022  ERROR  window::os::wayland::pointer > Unable to set cursor to arrow: cursor not found
```

### Anything else?

Here's a recording:

https://user-images.githubusercontent.com/49961859/218564572-e2b9f1b7-3bf0-4936-9c59-7f5972dfef82.mp4

Interesting that the cursor is visible in the last part of the recording, even though I haven't seen it on my screen.

wez:
How did you install wezterm? The version information seems a bit unusual.
Is this a duplicate of https://github.com/wez/wezterm/issues/2687 ?

paveloom:
[...]

Issue #2687, opened 2022-10-31
flatpak release: Unable to set cursor to left_ptr: cursor not found


sdroege:
### What Operating System(s) are you seeing this problem on?

Linux Wayland

### Which Wayland compositor or X11 Window manager(s) are you using?

gnome-shell

### WezTerm version

20220905-102802-7d4b8249

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

Running the current release from flathub as of today gives lots of warnings as follows

```
17:08:48.768  ERROR  window::os::wayland::frame > Unable to set cursor to bottom_side: cursor not found
17:08:48.772  ERROR  window::os::wayland::frame > Unable to set cursor to left_ptr: cursor not found
17:08:48.772  ERROR  window::os::wayland::frame > Unable to set cursor to left_ptr: cursor not found
17:08:48.772  ERROR  window::os::wayland::pointer > Unable to set cursor to xterm: cursor not found
17:08:48.773  ERROR  window::os::wayland::pointer > Unable to set cursor to xterm: cursor not found
17:08:48.774  ERROR  window::os::wayland::pointer > Unable to set cursor to xterm: cursor not found
17:08:48.775  ERROR  window::os::wayland::pointer > Unable to set cursor to xterm: cursor not found
17:08:48.776  ERROR  window::os::wayland::pointer > Unable to set cursor to xterm: cursor not found
```

The cursors are also indeed not changing correctly, e.g. when going to the edges of the window for resizing.

The same works correctly when building the current version locally from git.

### To Reproduce

Just run it via `flatpak run org.wezfurlong.wezterm`

### Configuration

no config

### Expected Behavior

Not hundreds of lines of errors and cursors are changing correctly.

wez:
To troubleshoot, see: https://wezfurlong.org/wezterm/faq.html#i-use-x11-or-wayland-and-my-mouse-cursor-theme-doesnt-seem-to-work

I think you probably want to try running something like this to see more diagnostics:

```
flatpak run --env=WEZTERM_LOG=window::os::x11::cursor=trace,window::os::wayland::pointer=trace,info org.wezfurlong.wezterm
```

wez:
This should be fixed now in `main`.

It typically takes about an hour before commits are available as nightly builds for all platforms.  Linux builds are the fastest to build and are often available within about 20 minutes.  Windows and macOS builds take a bit longer.

[...]

PR #5276, opened 2024-04-08
Update smithay-client-toolkit to v0.18

### LINKED ISSUE #5071: Cursor hiding on broken on Wayland

Linux Wayland

KDE/Plasma

### WezTerm version

20240221_075452_95581d86

### Describe the bug

Cursor hiding does not work.

In addition, every keypress generates the following log entry:

> ERROR  window::os::wayland::window > set_cursor: Cursor not found

Disabling cursor hiding makes the error go away.

### To Reproduce

Run Wezterm, press a key

### Configuration

This happens with an empty configuration.

### Expected Behavior

_No response_

### Logs

_No response_

### Anything else?

This is from a freshly installed Wezterm from the COPR on Fedora 39. KDE/Plasma, AMD graphics card, Wayland.

#4948 looks interesting (since it's the only open issue I can find that mentions the same log entry) so maybe this is a dupe and I'm lucky that it doesn't panic?

Anyway, happy to help debug if I can.

deviant:
SCTK v0.18 includes some changes to how cursors are set, which fixes not being able to change the cursor when `XCURSOR_THEME` is unset (and emitting an error message about this once every frame). Probably fixes #5071 and #4604, and, I think, #5209. Needs testing on GNOME to see if #3106 or #2687 are fixed by this (GNOME unfortunately does not yet implement the [cursor-shape-v1](https://wayland.app/protocols/cursor-shape-v1) protocol), but that's obviously not a blocker.


deviant:
This appears to break the primary selection, and I'm not sure why. However, I've noticed that the update also has a new `delegate_primary_selection` macro, which I'm looking into currently.

deviant:
This is now complete, and the primary selection works once again. 😊 Pipe handling is a little cleaner as well, and there aren't quite as many `unsafe` blocks dotted around.

deviant:
Fixed an existing crash that occurred when switching TTYs. Looks like someone else has found and tried to fix this (#5173), but this implementation is more complete (handles loss of each different capability type, not just keyboard) and more correct (releases the associated resources).

deviant:
Fixed a rare, latent crash in code which assumed that capabilities couldn't be lost, uncovered by the previous fix.

deviant:
@wez Would it be possible to take a look at this? I've been running this (specifically, cce0706b1f2a9e2d1f02c57f2d1cd367c91df1ae plus this patchset) daily for the past few weeks and not run into any issues (besides the ones I've provided fixes for).

window/src/os/wayland/seat.rs
```diff
@@ -65,9 +75,23 @@ impl SeatHandler for WaylandState {
...
         _qh: &QueueHandle<Self>,
         _seat: WlSeat,
-        _capability: smithay_client_toolkit::seat::Capability,
+        capability: Capability,
     ) {
-        todo!()
+        match capability {
+            Capability::Keyboard => {
+                log::trace!("Lost keyboard capability");
+                self.keyboard.take().unwrap().release();
```
wez, line 83: Please avoid the unwrap for this; I'm assuming that this will work:

```suggestion
                self.keyboard.take().map(|p| p.release());
```

wez (commented):
Thanks for this!
Just one comment about an unwrap that would be good to avoid.
I did quickly test this against weston running in X and it seemed to startup and be basically functional, so it looks good!

wez:
I'll fixup the unwrap and merge this, otherwise it may be a while before I get back to it!

wez:
I rebased and pushed to main, thanks!

deviant:
Some of the issues I linked in the initial comment can be closed as well now, I don't think they get auto-closed when the commits are merged in manually and the enclosing PR closed.

deviant:
Same with the issues/PR linked in https://github.com/wez/wezterm/pull/5276#issuecomment-2051001131

woojiq:
I think these issues can be closed too:
#5363 #5311 #5299 #5258 #5255
They were created recently and have similar description (but I haven't tested each separately).

wez:
I closed those as duplicates, thanks for pointing me to them!

deviant:
> I think these issues can be closed too: #5363 #5311 #5299 #5258 #5255 They were created recently and have similar description (but I haven't tested each separately). @deviant , am I right?

I've taken a look and am unsure: it's possible that hyprland and river support will have been improved with the dependency bump, but I also don't use either of these compositors— and the errors reported don't seem familiar to me. I guess we'll see if people report seeing further issues.

Cu3PO42:
I do use Hyprland. The reported issues are familiar and were resolved by this PR :-)

PR #5316, opened 2024-04-18
Implement Drag and Drop for X11

### LINKED ISSUE #640: Drag & Drop files/folders
Not really a problem, but it is sometimes easier to cd into folder by dragging and dropping it or paste the path of file by dragging and dropping into the terminal.

Drag & Drop file/folders, like in the regular terminal, once you drag something it copies the path of file/folder.

ssiegel:
This implements accepting files/folders via drag and drop as on the
other platforms. It adds receiving URLs dragged e.g. from a browser and
plain text.

For filenames and urls an additional space is inserted after the last
item to enable adding more files and urls with another drag-and-drop
operation without the need to manually enter the space in between.

This is the last platform to resolve #640

### DIFF
--- a/wezterm-gui/src/termwindow/mod.rs
+++ b/wezterm-gui/src/termwindow/mod.rs
@@ -1011,6 +1011,28 @@ impl TermWindow {
                 }
                 Ok(true)
             }
+            WindowEvent::DroppedString(text) => {
+                let pane = match self.get_active_pane_or_overlay() {
+                    Some(pane) => pane,
+                    None => return Ok(true),
+                };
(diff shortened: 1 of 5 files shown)


ssiegel:
The change to add the space after the dropped files/urls might be up for debate, but it is what I'm used to from Konsole. I don't see how an additional space would be undesirable anywhere where I might want to drop filenames. Even if there might be a situation, then deleting it with a single backspace seems easier to me than having to change focus and press space between multiple drag-and-drop operations.

wez:
Thank you!

PR #5343, opened 2024-04-27
Documentation naming fixes

MLFlexer:
Reading though the docs i found an error in the action for MoveToSelectionOtherEndHoriz as the code calls MoveToSelectionOtherEnd not MoveToSelectionOtherEndHoriz as expected.

The docs also has MoveFor[**e**]wardSemanticZone instead of MoveForwardSemanticZone


wez:
Thanks!

PR #5345, opened 2024-04-27
Wire up agent forward for libssh backend

Riatre:
TSIA.

Depends on https://github.com/wez/libssh-rs/pull/24

There's a drive-by change in sftpwrap.rs for bumping to new libssh-rs.

### DIFF
--- a/wezterm-ssh/Cargo.toml
+++ b/wezterm-ssh/Cargo.toml
@@ -31,10 +31,11 @@ portable-pty = { version="0.8", path = "../pty" }
 regex = "1"
 smol = "1.2"
 ssh2 = {version="0.9.3", features=["openssl-on-win32"], optional = true}
-libssh-rs = {version="0.2.1", features=["vendored"], optional = true}
+libssh-rs = {version="0.3.1", features=["vendored"], optional = true}
 #libssh-rs = {path="../../libssh-rs/libssh-rs", features=["vendored"], optional = true}
 thiserror = "1.0"
 socket2 = "0.5"
+uds_windows = "1.1.0"
 
 # Not used directly, but is used to centralize the openssl vendor feature selection
 async_ossl = { path = "../async_ossl" }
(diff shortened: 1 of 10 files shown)


wezterm-ssh/src/sessioninner.rs
```diff
@@ -517,8 +525,16 @@ impl SessionInner {
 
             let stdin = &mut chan.descriptors[0];
             if stdin.fd.is_some() && !stdin.buf.is_empty() {
-                write_from_buf(&mut chan.channel.writer(), &mut stdin.buf)
```
Riatre, line 520: That's because I'm seeing weird error returns from here, causing wezterm to give up shortly after a ssh-agent forward sessions ended, even if there's still other channels (mainly, shell) connected.

I tried to write a e2e test for this, but unfortunately I can only reproduce on Windows against KeeAgent, and the testing infra seems to support Linux/Mac only.

wezterm-ssh/Cargo.toml
```diff
@@ -31,10 +31,11 @@ portable-pty = { version="0.8", path = "../pty" }
 regex = "1"
 smol = "1.2"
 ssh2 = {version="0.9.3", features=["openssl-on-win32"], optional = true}
-libssh-rs = {version="0.2.1", features=["vendored"], optional = true}
-#libssh-rs = {path="../../libssh-rs/libssh-rs", features=["vendored"], optional = true}
+# libssh-rs = {version="0.2.1", features=["vendored"], optional = true}
+libssh-rs = {path="../../libssh-rs/libssh-rs", features=["vendored"], optional = true}
```
wez, line 35: Should be able to grab this from crates.io now!

```suggestion
libssh-rs = {version="0.3.1", features=["vendored"], optional = true
```

wez (commented):
I think this looks good; let's update to point at crates.io and I think we're good!
Thank you!

wezterm-ssh/Cargo.toml
Riatre: Done.

wez:
Thanks!

PR #5362, opened 2024-05-02
fixed the issue #5348 and a typo

### LINKED ISSUE #5348: unusable windows buttons appears when win32_system_backdrop is set to acrylic and tab bar opacity is less than 100

Windows

_No response_

### WezTerm version

20240405-180910-cce0706b

### Describe the bug

unusable windows buttons appear in the tab bar when win32_system_backdrop is set to acrylic or mica and the tab bar transparency is below 100% and window_decorations is set to resize
Screenshot 2024-04-29 040329

### To Reproduce

Copy the config below

### Configuration

config.window_decorations = "RESIZE"
config.window_background_opacity = 0.6
config.win32_system_backdrop = "Acrylic"
config.colors = {
  tab_bar = {
    background = "rgb(22, 24, 26 / 30%)",
  },
}

### Expected Behavior

_No response_

### Logs

_No response_

### Anything else?

_No response_

Kushagra2569:
Fixed  #5348 for windows by setting margins check for window decoration for DwmExtendFrameIntoClientArea instead and using value -1 margin for only the TITLE style

### DIFF
--- a/window/src/os/windows/window.rs
+++ b/window/src/os/windows/window.rs
@@ -1424,7 +1424,7 @@ fn apply_theme(hwnd: HWND) -> Option<LRESULT> {
         if let Some(inner) = rc_from_hwnd(hwnd) {
             let mut inner = inner.borrow_mut();
 
-            // Set Arylic or Mica system Backdrop
+            // Set Acrylic or Mica system Backdrop
             let pv_attribute = match inner.config.win32_system_backdrop {
                 SystemBackdrop::Auto => DWM_SYSTEMBACKDROP_TYPE::DWMSBT_AUTO,
                 SystemBackdrop::Disable => DWM_SYSTEMBACKDROP_TYPE::DWMSBT_NONE,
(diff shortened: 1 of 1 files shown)


Taaqif:
Keen on seeing this merged!

wez:
Thanks for diving in! Could you run through the various combinations of title and system backdrop styles and verify that they all work correctly and/or share some screenshots of these?
I think the change you're proposing sounds plausible, but I'd it's not immediately obviously from the diff which combinations should result in which outcomes, and I don't currently have time to dig into this myself on a Windows machine.

Thanks!

Kushagra2569:
HI @wez , my bad , forgot to mention I did test before the pr and saw that only the TITLE decoration was having the issue where the acrylic was not being applied, and since TITLE by default has the windows buttons therefore it worked perfectly setting the margins for title to -1 while every other setting worked perfectly with these setting.
Essentially I have compared that all the settings looks exact same as without my change just without the bug
Tested on a Windows 11 machine
Adding the screenshots with my change as well
resize acrylic
resize disabled
title acrylic 0 6 opacity
title disable 0 6 opacity
title mica
title tabbed

wez:
Thanks for confirming!

PR #5377, opened 2024-05-06
Handle missing data device manager more gracefully

tmccombs:
I'm not sure how necessary this is, but it follows example code from wlroots on how to use these events.


wez:
Seems reasonable, thanks!

PR #5388, opened 2024-05-07
chore(nix): Update refs to submodules & explain why this is needed

bew:
I saw that you updated harfbuzz' version in ec438e714c25f24783e2a23df5d3fb8eef192b55, and noticed last time that the Nix code currently needs to get the submodules manually, so here you go.
I added some explanation why this whole submodule duplication thing is needed at the moment.

I also preferred to use tags instead of commit hash to be much more readable and maintainable.

cc: @happenslol @gabyx is that okay for you?

### DIFF
--- a/nix/flake.nix
+++ b/nix/flake.nix
@@ -12,23 +12,27 @@
       };
     };
 
+    # NOTE: @2024-05 Nix flakes does not support getting git submodules of 'self'.
+    # refs:
+    # - https://discourse.nixos.org/t/get-nix-flake-to-include-git-submodule/30324
+    # - https://github.com/NixOS/nix/pull/7862
+    #
+    # ... In the meantime we kinda duplicate the dependencies here then replace the submodules with
+    # links to each repo in package sources.
     freetype2 = {
-      url = "github:wez/freetype2/de8b92dd7ec634e9e2b25ef534c54a3537555c11";
+      url = "github:wez/freetype2/VER-2-13-0";
       flake = false;
     };
-
     harfbuzz = {
-      url = "github:harfbuzz/harfbuzz/60841e26187576bff477c1a09ee2ffe544844abc";
+      url = "github:harfbuzz/harfbuzz/8.4.0";
       flake = false;
     };
-
     libpng = {
       url = "github:glennrp/libpng/8439534daa1d3a5705ba92e653eda9251246dd61";
       flake = false;
     };
-
     zlib = {
-      url = "github:madler/zlib/cacf7f1d4e3d44d871b605da3b647f07d718623f";
+      url = "github:madler/zlib/v1.2.11";
       flake = false;
     };
   };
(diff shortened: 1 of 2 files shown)


nix/flake.nix
```diff
@@ -12,23 +12,27 @@
...
     };
-
     harfbuzz = {
-      url = "github:harfbuzz/harfbuzz/60841e26187576bff477c1a09ee2ffe544844abc";
+      url = "github:harfbuzz/harfbuzz/8.4.0";
       flake = false;
     };
-
     libpng = {
       url = "github:glennrp/libpng/8439534daa1d3a5705ba92e653eda9251246dd61";
```
bew, line 31: I didn't find a relevant tag for this commit, is that correct @wez 🤔

(I'm curious)
Is there something specific about that commit? or it was the last when you included that submodule and never needed to update?

happenslol:
Looks good to me, I went through the threads you linked when I created the flake 😅

Linking the tags instead of commit hashes makes sense for readability. I'm worried that the manual upgrade process becomes harder/more tedious though, since now we need to look through the repos and find the tag instead of just using the ref the submodule points to.

bew:
> I went through the threads you linked when I created the flake 😅


yeah I figured, now it's explicit and we can subscribe to something to remove this 'hack' some time in the future (lazy trees is actively being worked on atm, according to recent nix team meeting summaries 👌)

> I'm worried that the manual upgrade process becomes harder/more tedious though, since now we need to look through the repos and find the tag instead of just using the ref the submodule points to.

Well, running `git submodule status` seems to show the result of `git describe` for each submodules, so it helps a lot:
image

nix/flake.nix
wez: I have no specific recollection here. It's possible that updating to a newer rev would be a good idea.

nix/flake.nix
```diff
@@ -12,23 +12,31 @@
...
+    #
+    # ... In the meantime we kinda duplicate the dependencies here then replace the submodules with
+    # links to each repo in package sources.
+    #
+    # Try to use tags when possible to increase readability
+    # (note: `git submodule status` in wezterm repo will show the `git describe` result for each
+    # submodule, can help finding a tag if any)
     freetype2 = {
-      url = "github:wez/freetype2/de8b92dd7ec634e9e2b25ef534c54a3537555c11";
+      url = "github:wez/freetype2/e4586d960f339cf75e2e0b34aee30a0ed8353c0d";
```
wez, line 27: FWIW, you could use the tag for this too:

```
 e4586d960f339cf75e2e0b34aee30a0ed8353c0d deps/freetype/freetype2 (VER-2-13-1)
```

bew: hmm I don't see that tag 👀
https://github.com/wez/freetype2/tags

wez:
Thanks!

PR #5396, opened 2024-05-09
fix wrapped long line copy problem on tmux

### LINKED ISSUE #5283: Investigation "Wrapped long line copy under tmux"

Windows

_No response_

### WezTerm version

wezterm 20231228-084719-ff274374

### Describe the bug

Hi.
First, I really liked to use wezterm on my daily work and I satisfied most of features.
One thing I am having a problem is that copying single logical line on multiple physical line on tmux.
I also searched issues on the board and I found already closed issues such as https://github.com/wez/wezterm/issues/4706 and https://github.com/wez/wezterm/issues/5016.

I spent some time to debug why this is happening on tmux and I found the reason should_mark_wrapped is set to false when alt screen is on.
https://github.com/wez/wezterm/blob/cce0706b1f2a9e2d1f02c57f2d1cd367c91df1ae/term/src/terminalstate/performer.rs#L165

I confirmed that it worked as expected after removing '!is_alt'

Can we remove this condition ? Or is there any reason for checking is_alt ?

### To Reproduce

_No response_

### Configuration

no config

### Expected Behavior

logical single line should be copied as single line.

### Logs

_No response_

### Anything else?

_No response_

eternity74:
wezterm ignores wrapped long line when alt screen is used and tmux uses alt screen that makes copying wrapped long line into multiple lines.

After testing on sevaral tui apps like tmux, w3m, htop it seems to show reasonable performance on resizing.

Fixes #5283

### DIFF
--- a/term/src/terminalstate/performer.rs
+++ b/term/src/terminalstate/performer.rs
@@ -162,14 +162,13 @@ impl<'a> Performer<'a> {
                         }
                     }
 
-                    let should_mark_wrapped = !is_alt
-                        && (!is_conpty
-                            || screen
-                                .line_mut(y)
-                                .visible_cells()
-                                .last()
-                                .map(|cell| makes_sense_to_wrap(cell.str()))
-                                .unwrap_or(false));
+                    let should_mark_wrapped = !is_conpty
+                        || screen
+                            .line_mut(y)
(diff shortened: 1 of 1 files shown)


wez:
Thanks!

PR #5397, opened 2024-05-09
Update wayland keymap handling for protocol 7+.

### LINKED ISSUE #5393: Wezterm windows hang when launched under chromeos's crostini

Linux Wayland

ChromeOS implements a wayland proxy compositor known as somellier

### WezTerm version

latest 

### Describe the bug

So when launched under chromeos's linux containers, most wayland apps work fine.
However, wezterm hangs.  

What is happening under the covers is that it is waiting for a configure event after creating the window, but the event never comes.
This is true regardless of backend (Software, OpenGL, WebGpu).
This is true regardless of window decorations between server or client or none.
enable_wayland = false works fine.

I actually implemented support for wayland's activation protocol  (since it was like 5 lines of code) but it turns out the proxy compositor doesn't use it.

I also played around with where commit is called, no luck.

Note that all of the smithay client toolkit examples work fine and happily show windows.

Logs at the trace level show nothing interesting.
Happy to debug on my own, produce logs, etc, if anyone has any ideas on where this might be going wrong.

### To Reproduce

_No response_

### Configuration

no config

### Expected Behavior

Wezterm is visible

### Logs

_No response_

### Anything else?

_No response_

dberlin:
Newer wayland protocol expects us to mmap the keymap file
descriptor (and do so with certain mmap flags).

The current code tries to handle both normal files and pipes,
but neither will work in some cases.

This updates the code to use xkbcommon's new_from_fd,
which is what the toolkit uses, and mmap's the file.

This fixes #5393, which was caused by hanging when
trying to read the file as if it was a pipe.

### DIFF
--- a/window/src/os/wayland/keyboard.rs
+++ b/window/src/os/wayland/keyboard.rs
@@ -1,10 +1,9 @@
 use std::borrow::BorrowMut;
-use std::io::Read;
-use std::os::fd::{AsRawFd, FromRawFd};
-use std::os::unix::fs::FileExt;
 
 use wayland_client::protocol::wl_keyboard::{Event as WlKeyboardEvent, KeymapFormat, WlKeyboard};
 use wayland_client::{Dispatch, Proxy};
+use xkbcommon::xkb;
+use xkbcommon::xkb::CONTEXT_NO_FLAGS;
 
 use crate::x11::KeyboardWithFallback;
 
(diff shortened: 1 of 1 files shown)


dberlin:
I'm also happy to add new_from_fd to Keyboard/KeyboardWithFallback instead of converting to string and back if we want.

I'm just about to submit a PR to rewrite the keyboard handling to use the toolkit, so i didn't bother.

(The API guarantees that converting it to a string will result in a valid keymap, so it either fails during new_from_fd, or it will successfully convert to a string and back)

window/src/os/wayland/keyboard.rs
```diff
@@ -59,39 +58,39 @@ impl Dispatch<WlKeyboard, KeyboardData> for WaylandState {
...
-                        match KeyboardWithFallback::new_from_string(s) {
-                            Ok(k) => {
-                                state.keyboard_mapper.replace(k);
+                            Ok(None) => {
+                                log::error!("invalid keymap");
                             }
+
                             Err(err) => {
-                                log::error!("Error processing keymap change: {:#}", err);
+                                log::error!("{}", err);
```
wez, line 93: Please restore the original error formatting here; the additional context is useful and error messages tend to render better as `{err:#}`.

wez (commented):
Thanks for diving in!

> I'm just about to submit a PR to rewrite the keyboard handling to use the toolkit, so i didn't bother.

Let's talk about that before you put in any significant effort on it, as we are deliberately not using sctk's handling because it, once upon a time, had significant issues with repeat handling.

Otherwise, this PR looks OK to me; just one minor comment about preserving more context in an error message.

window/src/os/wayland/keyboard.rs
dberlin: Whoops, though i reverted that, but must not have staged it, will fix again and push.

dberlin:
> Thanks for diving in!
> 
> > I'm just about to submit a PR to rewrite the keyboard handling to use the toolkit, so i didn't bother.
> 
> Let's talk about that before you put in any significant effort on it, as we are deliberately not using sctk's handling because it, once upon a time, had significant issues with repeat handling.

Sure. It was like 1-2 hours of work, and was a good intro to the wayland internals of both anyway, so if it ends up not being usable, that's fine - it was a good learning experience.

window/src/os/wayland/keyboard.rs
dberlin: Should be fixed now

wez:
Thanks!

wez:
re: toolkit keyboard, if you think that it has some advantages over what we're doing currently and already have something we can discuss in a PR, that's great. Otherwise, I'd like to discuss what you think the advantages are before you spend too much time on it--I don't want you to feel like you've wasted your own time!

Thank you!

dberlin:
> re: toolkit keyboard, if you think that it has some advantages over what we're doing currently and already have something we can discuss in a PR, that's great. Otherwise, I'd like to discuss what you think the advantages are before you spend too much time on it--I don't want you to feel like you've wasted your own time!
> 
> Thank you!

I'll create a PR and we can discuss.  I did it more to try to see what was going on than anything else, so if we decide it's not worth it, or won't support what we need,  it really is no big thing.

If I was going to spend days on it, I would have started a discussion beforehand so i didn't throw away a lot of work :)

PR #5408, opened 2024-05-10
x11: fix exposures on child window

### LINKED ISSUE #5405: Unfocused Wezterm window content doesn't get drawn until "damage" occurs

Linux X11

Xfce4 4.18 / Xorg 21.1.13

### WezTerm version

20240509-223025-91a16e52

### Describe the bug

When switching to a virtual desktop unfocused Wezterm windows don't get drawn until some sort of "damage" (mouse over, focusing, etc.) occurs.

wezterm-unfocused

This happens with or without an active compositor.
Default or front_end = "OpenGL" makes no difference either

### To Reproduce

Switch away and back to a virtual desktop that has unfocused Wezterm windows on it.

### Configuration

no config

Tested with:  WEZTERM_CONFIG_FILE=/dev/null wezterm

### Expected Behavior

Windows should get drawn as soon as they become visible

### Logs

_No response_

### Anything else?

Commit 0184e88 (29. April 2024) works fine. This issue started happening after that release.

jknockel:
We should watch the child window for exposure events, not the parent window (the parent window is just a background color and we never paint to it).

This was a regression from commit 809bcc55.

Fixes #5405.

### DIFF
--- a/window/src/os/x11/connection.rs
+++ b/window/src/os/x11/connection.rs
@@ -70,6 +70,7 @@ pub struct XConnection {
     pub atom_net_active_window: Atom,
     pub(crate) xrm: RefCell<HashMap<String, String>>,
     pub(crate) windows: RefCell<HashMap<xcb::x::Window, Arc<Mutex<XWindowInner>>>>,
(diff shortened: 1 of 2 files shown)


wez:
Thanks!

PR #5413, opened 2024-05-13
nix: allowBuiltinFetchGit = true

mirror-kt:
implement https://github.com/wez/wezterm/pull/5373#issuecomment-2106088660

This will reduce the frequency of nix-related maintenance.
I tested this change with `nix run ./nix`.

### DIFF
--- a/nix/flake.nix
+++ b/nix/flake.nix
@@ -105,10 +105,7 @@
 
         cargoLock = {
           lockFile = ../Cargo.lock;
-          outputHashes = {
-            "xcb-imdkit-0.3.0" = "sha256-fTpJ6uNhjmCWv7dZqVgYuS2Uic36XNYTbqlaly5QBjI=";
-            "sqlite-cache-0.1.3" = "sha256-sBAC8MsQZgH+dcWpoxzq9iw5078vwzCijgyQnMOWIkk";
-          };
+          allowBuiltinFetchGit = true;
         };
 
         preConfigure = ''


bew:
Hello o/

The nixpkgs manual mentions this:
> For usage outside nixpkgs, `allowBuiltinFetchGit` could be used to avoid having to specify `outputHashes`.

Do you know why it mentions "outside of nixpkgs" ? Are there limitations or things to be aware of when using that option? 🤔

wez:
Thanks!

Cu3PO42:
> Do you know why it mentions "outside of nixpkgs" ? Are there limitations or things to be aware of when using that option? 🤔

I can explain this. In Nix, you have two fundamental ways to fetch a Git repository:

* via the builtin operation fetchGit
* via a fixed-output derivation that "just" invokes git in its build phase

The problem with the built-in is that it happens during evaluation time of the Nix expression and blocks the process from doing anything else. Nothing else can be computed, downloaded, or built during that time. Hence, nixpkgs disallows usage of the built-in and only permits the second option because CI would just slow down to a crawl.

A fixed output derivation is allowed network access and can simply use git to clone a repository in its build step. This can be executed in parallel with a number of other fetchers, builders, etc. The cost for getting network access is that you must specify a hash of the output of your result beforehand so that you cannot introduce any impurities. These hashes were what was previously specified in the Nix file.

You may ask: "How does the built-in get away without a hash?" Well, the built-in requires you to specify a commit SHA, which also means you are locking down the exact results you are getting. The built-in being specific to Git knows this, but unfortunately there's no way to use that same commit hash in the fixed output derivation.

In the context of this Flake, using the built-in is probably fine.

PR #5424, opened 2024-05-15
Windows: fix warnings in `portable-pty`

rgwood:
Hi! Tiny PR, just fixes up a few build warnings when building `portable-pty` on Windows:

```
warning: the item `Path` is imported redundantly
   --> pty\src\cmdbuilder.rs:608:13
    |
9   | use std::path::{Component, Path};
    |                            ---- the item `Path` is already imported here
...
608 |         use std::path::Path;
    |             ^^^^^^^^^^^^^^^
    |
    = note: `#[warn(unused_imports)]` on by default

warning: function `is_cwd_relative_path` is never used
   --> pty\src\cmdbuilder.rs:749:4
    |
749 | fn is_cwd_relative_path<P: AsRef<Path>>(p: P) -> bool {
    |    ^^^^^^^^^^^^^^^^^^^^
    |
    = note: `#[warn(dead_code)]` on by default
 ```
 
I'm hoping to contribute more substantial Windows fixes soon. I've been noticing that every few days WezTerm pegs a CPU core on Windows but haven't been able to pinpoint why. I now have a build with debug symbols and I'll profile it next time the issue occurs.

### DIFF
--- a/pty/src/cmdbuilder.rs
+++ b/pty/src/cmdbuilder.rs
@@ -6,7 +6,9 @@ use std::collections::BTreeMap;
 use std::ffi::{OsStr, OsString};
 #[cfg(windows)]
 use std::os::windows::ffi::OsStrExt;
-use std::path::{Component, Path};
+#[cfg(unix)]
+use std::path::Component;
+use std::path::Path;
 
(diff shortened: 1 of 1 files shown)


wez:
Thanks!

PR #5430, opened 2024-05-16
Check for the presence of CARGO_BUILD_TARGET_DIR env variable 

itkovian:
When building in the presence of a CARGO_BUILD_TARGET_DIR, there is no target dir created inside the repo, so copying to a location within the local repo dir fails, hence we need to copy to the actual target dir that is being used.

### DIFF
--- a/wezterm-gui/build.rs
+++ b/wezterm-gui/build.rs
@@ -168,7 +168,7 @@ END
             .join("WezTerm.app")
             .join("Contents")
             .join("Info.plist");
-        let build_target_dir = std::env::var("CARGO_BUILD_TARGET_DIR")
+        let build_target_dir = std::env::var("CARGO_TARGET_DIR")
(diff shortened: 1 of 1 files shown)


wezterm-gui/build.rs
```diff
@@ -168,7 +168,10 @@ END
             .join("WezTerm.app")
             .join("Contents")
             .join("Info.plist");
-        let dest_plist = repo_dir.join("target").join(profile).join("Info.plist");
+        let build_target_dir = std::env::var("CARGO_BUILD_TARGET_DIR")
```
wez, line 171: Should this be `CARGO_TARGET_DIR`? When is `CARGO_BUILD_TARGET_DIR` set?

https://doc.rust-lang.org/cargo/appendix/glossary.html?highlight=CARGO_TARGET_DIR#target

wez:
Thanks!

PR #5436, opened 2024-05-17
use correct flake dir in update-flake workflow

michaelBelsanti:
Will fix the update-flake workflow, which is currently failing each time. Will also remove the warning of an old nixpkgs revision that is currently shown on each Nix build workflow.

Fix was posted here: https://github.com/wez/wezterm/pull/5373#issuecomment-2103484296


wez:
Thanks!

PR #5441, opened 2024-05-19
Fix: slow close non last tab

Mrreadiness:
Fix issue: https://github.com/wez/wezterm/issues/5304
Problem: after closing a non-last tab [invalidate](https://github.com/wez/wezterm/blob/396c8d9a886d06c4f839a24d3942f8612b6b84d0/mux/src/window.rs#L74) wasn't called, which is important for rendering tabs.

### DIFF
--- a/mux/src/window.rs
+++ b/mux/src/window.rs
@@ -131,6 +131,8 @@ impl Window {
 
         if len > 0 && self.active >= len {
             self.set_active_without_saving(len - 1);
+        } else if len > 0 {
+            self.set_active_without_saving(self.active);
         }
     }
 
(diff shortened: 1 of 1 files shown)


mux/src/window.rs
```diff
@@ -131,6 +131,8 @@ impl Window {
 
         if len > 0 && self.active >= len {
             self.set_active_without_saving(len - 1);
+        } else if len > 0 {
+            self.set_active_without_saving(self.active);
```
wez, line 135: Seems simpler to invalidate in the else branch?
 
```suggestion
        } else {
            self.invalidate();
```

wez (commented):
Thanks for diving in!

mux/src/window.rs
Mrreadiness: I thought call `set_active_without_saving` would be better, because it's exactly what we need - set a new active tab which index = prev active tab index.

PR #5444, opened 2024-05-20
Fix Windows CI.

bogdan2412:
It's currently failing when running tests with errors of the form:

> error: creating test list failed
>
> Caused by:
>   for `wezterm-config-derive`, command
>   `'D:\a\wezterm\wezterm\target\debug\deps\wezterm_config_derive-f1c7f0f2de220b6e.exe'
>   --list --format terse` exited with code 0xc0000135: The specified
>   module could not be found. (os error 126)

This seems to have been caused by nextest-rs/nextest#1493, with this comment as a workaround fix: https://github.com/nextest-rs/nextest/issues/1493#issuecomment-2106331574


wez:
Thank you!

PR #5445, opened 2024-05-20
ci: Update Fedora versions.

bogdan2412:
Fedora 40 was released, Fedora 37 is EOL.

### DIFF
--- a/.github/workflows/gen_fedora40.yml
+++ b/.github/workflows/gen_fedora40.yml
@@ -1,14 +1,31 @@
-name: fedora37_tag
+name: fedora40
 
 on:
-  push:
-    tags:
-      - "20*"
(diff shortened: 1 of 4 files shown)


wez:
Thanks!
As a follow-up PR (it can't be part of this one, because this needs to be merged first), please update the docs:

https://github.com/wez/wezterm/blob/993eb9f3a961dd3ee1594521477162d3cc3c0034/docs/install/linux.md?plain=1#L204-L206

what I usually do for this is keep the removed version around until the next stable release, so the change would be to add just the nightly asset columns for F40; the stable columns can say "nightly only" until we make the next release.

You can locally run the docs build via `./ci/build-docs.sh serve`

PR #5447, opened 2024-05-20
Update docs to reflect CI nightly builds being available for Fedora 40.

bogdan2412:
Mention that nightly rpms are also available in Copr while I'm at it.


wez:
Thanks!

Issue #5479, opened 2024-05-27
ProxyCommand zombie process


daaku:
### What Operating System(s) are you seeing this problem on?

Linux Wayland

### Which Wayland compositor or X11 Window manager(s) are you using?

Sway 1.9

### WezTerm version

20240203-110809-5046fc22

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

No, and I'll explain why below

### Describe the bug

When a host is configured to use a `ProxyCommand`, and the proxy command exits with a non-zero exit code, for example when there's some connection error, a zombie process is left around.

### To Reproduce

`wezterm ssh -o ProxyCommand=/usr/bin/false example.com` should start a new wezterm instance, and you'll see a `[false] <defunct>` process until you kill that wezterm instance.

### Configuration

no config

### Expected Behavior

The ProxyCommand process should be reaped.

PR #5494, opened 2024-05-31
kill and reap ProxyCommand

daaku:
Using the scopeguard crate ensures the child process (if any) spawned for the ProxyCommand is cleaned up under all the various error scenarios (such as auth failures etc), as well as in the normal successful use case. This ensures killing it if necessary, and then waiting for it.

This is a fix for #5479.

### DIFF
--- a/wezterm-ssh/src/sessioninner.rs
+++ b/wezterm-ssh/src/sessioninner.rs
@@ -205,7 +205,7 @@ impl SessionInner {
             sess.set_option(libssh_rs::SshOption::HostKeys(host_key.to_string()))?;
         }
 
-        let sock =
+        let (sock, _child) =
             self.connect_to_host(&hostname, port, verbose, self.config.get("proxycommand"))?;
         let raw = {
             #[cfg(unix)]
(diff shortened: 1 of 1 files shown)


wezterm-ssh/Cargo.toml
```diff
@@ -39,6 +39,7 @@ wezterm-uds = { path = "../wezterm-uds" }
 
 # Not used directly, but is used to centralize the openssl vendor feature selection
 async_ossl = { path = "../async_ossl" }
+scopeguard = "1.2.0"
```
wez, line 42: while scopeguard is handy in a pinch, I'm not in favor of it here for a couple of reasons:

* It is an additional external dependency
* You still have to remember to initiate the scope guard, and that is required in multiple places in this PR

What I'd prefer to see as a solution for this is for `connect_to_host` to return a little wrapper around the Child:

```rust
struct KillOnDropChild(Child);

impl std::ops::Deref for KillOnDropChild { ... }
impl std::ops::DerefMut for KillOnDropChild { ... }
impl Drop for KillOnDropChild {
   fn drop(&mut self) {
       self.0.kill();
       self.0.wait();
   }
}
```

wez (changes requested):
Thanks for this!
I have a slightly different proposal; would you mind updating to match it?

daaku:
Sounds reasonable, I'll update with the requested changes.

wez:
Thank you!

Issue #1737, opened 2022-03-19
Feature request: adjustable height of tab bar to help avoid terminal being covered by Macbook Pro notch


mike-hearn:
**Is your feature request related to a problem? Please describe.**
This is a pretty straight forward problem - new Apple laptops have a 64px notch right at the top of the screen, and if you fullscreen wezterm, a part of the terminal is completely hidden.

This only happens w/ the old tab bar style *and* non-native MacOS fullscreen on (both of those are, by far, my preferred way to use wezterm).

Here's a photo to demonstrate, where the notch covers the top corner of a split pane tmux window:

<img width="1382" alt="Screen Shot 2022-03-19 at 3 25 58 PM" src="https://user-images.githubusercontent.com/1016999/159135566-f2a56840-22b2-421e-ac4c-56e420bafcbb.png">

**Describe the solution you'd like**
Perhaps some kind of ability to add top, black-color padding in only full screen? Or just the ability to adjust the tab bar height (or tab bar padding)?

**Describe alternatives you've considered**
Adding padding to the top of the terminal itself works, but it's clunky because the padding isn't necessary anywhere other than fullscreen (so windowed it looks silly), and it's also not necessary on non Apple laptops, so my shared config makes the terminal look wonky on other machines.

**Additional context**
This bug is obviously not a dealbreaker, but it's just one of those QoL type of things.

Also, this terminal is great. Thank you for creating it and the work you've put into documenting it and maintaining it.

wez:
Ah, the notch.  I think handling this overlaps a bit with https://github.com/wez/wezterm/pull/1675 
That PR is adding support for OS-dependent borders, but could be extended to add OS-dependent padding.
It sounds like https://developer.apple.com/documentation/appkit/nsscreen/3882821-safeareainsets?changes=latest_major&language=objc could be used to populate that information on macOS.

I'm leaning towards a fairly basic approach to this: in non-native fullscreen we'll use safeAreaInserts to populate the os padding parameters, then when we render, factor that in to the configured window padding.

wez:
I pushed a commit that I think will do the job here. It should show up in a nightly build within an hour of this comment posting.

Please give it a try and let me know if it works!

If it doesn't work, please launch wezterm like this:

```
WEZTERM_LOG=window::os::macos::window=trace,info wezterm
```

and share the log output that looks like:

```
NSEdgeInsets { top: 0.0, left: 0.0, bottom: 0.0, right: 0.0 }
```

[...]

PR #5515, opened 2024-06-04
Fix macOS notch avoidance padding in full screen mode

mbaird:
Fixes https://github.com/wez/wezterm/issues/3807

The values returned by [`safeAreaInsets`][0] need to be multipled by the scale factor of the display (in case of the MacBook Pro 14-inch, 2021, a value of 2.0).

There's a code comment referencing [#1737 (comment)][1], where an extra `2` was added to this number, but something must have changed since then because I cannot see a need for this now.

(Not familiar with Rust, so let me know if there's a better way to write this!)

[0]: https://developer.apple.com/documentation/appkit/nsscreen/3882821-safeareainsets
[1]: https://github.com/wez/wezterm/issues/1737#issuecomment-1085923867


wez:
Thanks!

PR #5528, opened 2024-06-08
Update source.md to include how to build from source on windows

cleemesser:
based upon discussion [5521](https://github.com/wez/wezterm/discussions/5521) adding detailed instructions for new rust/windows programmer to build wezterm on windows from source.


wez:
Thanks for this, however, I opted to cover this in a different way that better matches the existing "voice" of the install info, and explicitly mentions that the MSVC flavor of rust is required.

PR #5564, opened 2024-06-14
Feature: The search overlay accepts IME composed input

kenchou:
-  The search overlay accepts IME composed input

https://github.com/wez/wezterm/issues/5333#issuecomment-2166379992
Based on other Writers, I wrote a writer for the search overlay. 
The good news is that it appears to be working well. 
However, as I am relatively new to Rust, there might be some bad code. I would greatly appreciate it if you could review the code for me. 
Thank you. @wez

### DIFF
--- a/termwiz/src/lineedit/buffer.rs
+++ b/termwiz/src/lineedit/buffer.rs
@@ -174,15 +174,7 @@ impl LineEditBuffer {
                 position
             }
             Movement::StartOfLine => 0,
-            Movement::EndOfLine => {
-                let mut cursor =
-                    GraphemeCursor::new(self.line.len().saturating_sub(1), self.line.len(), false);
-                if let Ok(Some(pos)) = cursor.next_boundary(&self.line, 0) {
-                    pos
-                } else {
-                    self.cursor
-                }
-            }
+            Movement::EndOfLine => self.line.len(),
             Movement::None => self.cursor,
         }
     }
--- a/termwiz/src/lineedit/buffer.rs
+++ b/termwiz/src/lineedit/buffer.rs
@@ -174,7 +174,15 @@ impl LineEditBuffer {
                 position
             }
             Movement::StartOfLine => 0,
-            Movement::EndOfLine => self.line.len(),
+            Movement::EndOfLine => {
+                let mut cursor =
+                    GraphemeCursor::new(self.line.len().saturating_sub(1), self.line.len(), false);
+                if let Ok(Some(pos)) = cursor.next_boundary(&self.line, 0) {
+                    pos
+                } else {
+                    self.cursor
+                }
+            }
             Movement::None => self.cursor,
         }
     }
--- a/wezterm-gui/src/overlay/copy.rs
+++ b/wezterm-gui/src/overlay/copy.rs
@@ -163,7 +163,9 @@ impl CopyOverlay {
         render.update_search();
 
         let shared_render = Arc::new(Mutex::new(render));
-        let writer = SearchOverlayPatternWriter {render: Arc::clone(&shared_render)};
+        let writer = SearchOverlayPatternWriter {
+            render: Arc::clone(&shared_render),
+        };
 
         Ok(Arc::new(CopyOverlay {
             delegate: Arc::clone(pane),
@@ -1522,7 +1524,9 @@ pub struct SearchOverlayPatternWriter {
 impl std::io::Write for SearchOverlayPatternWriter {
     fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
         let mut render = self.render.lock();
-        let s = std::str::from_utf8(buf).map_err(|err| std::io::Error::new(std::io::ErrorKind::Other, format!("invalid UTF-8: {err:#}")))?;
+        let s = std::str::from_utf8(buf).map_err(|err| {
+            std::io::Error::new(std::io::ErrorKind::Other, format!("invalid UTF-8: {err:#}"))
+        })?;
         render.pattern.push_str(s);
         render.schedule_update_search();
         Ok(buf.len())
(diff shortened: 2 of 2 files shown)


wezterm-gui/src/overlay/copy.rs
```diff
@@ -1507,6 +1515,23 @@ impl Pane for CopyOverlay {
...
 }
 
+pub struct SearchOverlayPatternWriter {
+    render: Arc<Mutex<CopyRenderable>>,
+}
+
+impl std::io::Write for SearchOverlayPatternWriter {
+    fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
+        let mut render = self.render.lock();
+        render.pattern.push_str(std::str::from_utf8(buf).unwrap());
```
wez, line 1525: Even though in this particular use case, the incoming bytes are produced from writing a `String`, which is guaranteed to be UTF-8, the `Write` trait could be triggered from other situations and arbitrary byte sequences are potentially possible.

The `unwrap` here is undesirable, as it can cause the program to terminate with no good opportunity to catch and handle the error.

Since `write` can return `Result`, we can propage the error instead, something like this:

```suggestion
        let s = std::str::from_utf8(buf).map_err(|err| std::io::Error::new(std::io::ErrorKind::Other, format!("invalid UTF-8: {err:#}")))?;
        render.pattern.push_str(s);
```

wezterm-gui/src/overlay/copy.rs
```diff
@@ -161,9 +162,13 @@ impl CopyOverlay {
         render.dirty_results.add(search_row);
         render.update_search();
 
+        let shared_render: Arc<Mutex<CopyRenderable>> = Arc::new(Mutex::new(render));
+        let writer = SearchOverlayPatternWriter {render: Arc::clone(&shared_render)};
```
wez, line 166: I think the CI will flag this as not being formatted correctly.
You can use `make fmt` to automatically format (or, if you don't have `make`, you can more directly run: `cargo +nightly fmt`).
You will need nightly rustfmt to apply the preferred formatting.

wezterm-gui/src/overlay/copy.rs
```diff
@@ -161,9 +162,13 @@ impl CopyOverlay {
         render.dirty_results.add(search_row);
         render.update_search();
 
+        let shared_render: Arc<Mutex<CopyRenderable>> = Arc::new(Mutex::new(render));
```
wez, line 165: unless it is required by the compiler, I generally prefer to avoid using explicit type annotations, as it results in less code to read, which makes it easier, for me at least, to grok more code at a glance:

```suggestion
        let shared_render = Arc::new(Mutex::new(render));
```

wez (commented):
Thanks for this! A couple of minor comments!

wezterm-gui/src/overlay/copy.rs
```diff
@@ -1507,6 +1517,26 @@ impl Pane for CopyOverlay {
...
+    render: Arc<Mutex<CopyRenderable>>,
+}
+
+impl std::io::Write for SearchOverlayPatternWriter {
+    fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
+        let mut render = self.render.lock();
+        let s = std::str::from_utf8(buf).map_err(|err| {
+            std::io::Error::new(std::io::ErrorKind::Other, format!("invalid UTF-8: {err:#}"))
+        })?;
+        render.pattern.push_str(s);
```
wez, line 1530: actually, it might be possible to do this:

```rust
  render.pattern.write(buf)?;
```

kenchou: emm, IDE hint 
```Type mismatch [E0308] expected `&mut TiffWriter<_>`, but found `&[u8]` ```

Comment (kenchou) [reply wezterm-gui/src/overlay/copy.rs:1530]: [MEMO]: After accepting PR #5416, `render.search_line.insert_text(s)` should be used here instead.

Comment (wez) [reply wezterm-gui/src/overlay/copy.rs:1530]: Would you like to rebase and update this PR to do that now that #5416 is merged?

File: termwiz/src/lineedit/buffer.rs [hunk 5]
```diff
@@ -174,15 +174,7 @@ impl LineEditBuffer {
...
-            Movement::EndOfLine => {
-                let mut cursor =
-                    GraphemeCursor::new(self.line.len().saturating_sub(1), self.line.len(), false);
-                if let Ok(Some(pos)) = cursor.next_boundary(&self.line, 0) {
-                    pos
-                } else {
-                    self.cursor
-                }
-            }
+            Movement::EndOfLine => self.line.len(),
```
kenchou, line 177: `EndOfLine` is simply set to `self.line.len()`

BTW: `self.cursor` is a **byte index**, and the index(storage) is not related to graphemes(visual), so perhaps calling it `cursor_index` would be clearer.
@Mrreadiness @wez

kenchou: emm... I think the crash might be caused by `self.line.len().saturating_sub(1)`, as the offset is not at the boundary of a Unicode character, which causes `cursor.next_boundary(&self.line, 0)` to panic. Just like `StartOfLine`, `EndOfLine` doesn't need any extra calculation; simply using `self.line.len()` is the most straightforward approach.

wez: Hmm, what I think we should do here is:
* Remove this change from this PR, as it has a different thesis from the rest of the PR
* Make a new PR that adds a couple of unit tests around is EndOfLine operation, and fixes the problematic case

kenchou: reverted

wez:
Thank you!

PR #5576, opened 2024-06-17
Float Pane

e82eric:
Hi,  I wanted to make an attempt at https://github.com/wez/wezterm/issues/270

This probably needs quite a bit of iteration (this is my first attempt at rust), but wanted to create the pull request to see if this is something you would be open to adding and to see if I am on the right track.

float_pane


derekthecool:
Nice work! Demo video looks good.

mathjiajia:
it would be nice to have a border for this floating pane

e82eric:
Updated with border.

float_pane_with_border

Configs:
````lua
config.float_pane_padding = {
  left = '10%',
  right = '10%',
  top = '10%',
  bottom = '10%',
}
````
````lua
config.float_pane_border = {
  left_width = '0.5cell',
  right_width = '0.5cell',
  bottom_height = '0.25cell',
  top_height = '0.25cell',
  left_color = '#665c54',
  right_color = '#665c54',
  bottom_color = '#665c54',
  top_color = '#665c54',
}
````

**Comment by e82eric** [comment]
I think this is ready to be reviewed now.

**Comment by Pajn** [comment]
Thanks a lot, this is the one feature I've been missing since transition from tmux to wezterm mux 🙏🏼 

Tested this and it works really well, just commenting with some small issues I've seen:
- The float panes all open in my home directory instead of the current directory like split panes and new tabs do
- The pane selection UI is hard to read if having no split panes and one floating since the underlying pane and the float panes letter overlaps
- If moving all non floating panes to a new tab or window the floating pane is shown on an otherwise invisible window, would be nice to either disallow this or de-float the floating pane

Screenshot from 2024-07-13 10-26-29
Screenshot from 2024-07-13 10-26-51

A very nice action to have but definitely not required for this to be super useful on its own, is to be able to toggle the floating status of a pane so you can transition a floating pane to a split pane and vice verse.

**Comment by e82eric** [comment]
@Pajn I think I have the working directory issue fixed now.

For the pane selection my intention was to treat the float like a overlay and turn any selection operations into no-ops and prevent the mouse from selecting a pane (I think there were some scenarios that I missed like the selection UI).  I pushed an update that I think turn those into nops now.  This may not be the right strategy,  is there a scenario where you would want to be able to select/interact with the panes underneath the float?

I think I have it updated so that closing the pane underneath the float is blocked now.  Which I think will prevent the scenario where only the float is showing.

I tried out being able to move the float to a split today.  It seemed to work,  I pushed it to this branch if you want to give it a try. I didn't want to include it in this pull request to keep the scope down. 
https://github.com/e82eric/wezterm/tree/move-float-to-split

This is what it looked like.
move_float_to_split

**Comment by Pajn** [comment]
Wow! I was just starting to familiarize myself with the code to see if I could help but not at that speed :)
Very quick testing have all the issues I experienced fixed/

>This may not be the right strategy, is there a scenario where you would want to be able to select/interact with the panes underneath the float?

I was mostly just trying to break it to see where the limitations where. The one possible usecase I can see is to move one background pane to another tab/window if you need access to it after realizing the float pane was longer lived than you intended. However I think that's better solved by being able to transition the float pane to a split anyway. And disabling those actions when in a float definitely avoids a whole class of potential problems.

>I didn't want to include it in this pull request to keep the scope down.

Totes! I was just excited. Thanks a lot for the branch though, will definitely start playing with it.

**Comment by Suri312006** [comment]
i really like the work man, good stuff been wanting this feature, as im coming from zellij that has a really nice implementation of floating panes

File: wezterm-gui/src/commands.rs [hunk 1]
```diff
@@ -1445,6 +1445,23 @@ pub fn derive_command_from_key_assignment(action: &KeyAssignment) -> Option<Comm
...
+                .into(),
+            keys: vec![(
+                Modifiers::CTRL
+                    .union(Modifiers::ALT)
+                    .union(Modifiers::SHIFT),
+                "p".into(),
+            )],
+            args: &[ArgType::ActivePane],
+            menubar: &["Shell"],
+            icon: Some("cod_split_vertical"),
```
MLFlexer, line 1463: Awesome that you created this PR! Trying to do my part in getting this merged by going though the code and thought i would add this:

This icon should probably be changed to something resembling a floating window, my suggestion would be: `cod-primitive_square`, `nf-cod-chrome_restore` or `cod-layout_panel_center`

wezterm-gui/src/commands.rs
```diff
@@ -1501,6 +1518,15 @@ pub fn derive_command_from_key_assignment(action: &KeyAssignment) -> Option<Comm
...
             icon: Some("cod_split_vertical"),
         },
+        FloatPane(_) => CommandDef {
+            brief: label_string(action, "Create a float pane".to_string()).into(),
+            doc: "Create a float pane"
+                .into(),
+            keys: vec![],
+            args: &[ArgType::ActivePane],
+            menubar: &[],
+            icon: Some("cod_split_vertical"),
```
MLFlexer, line 1528: See comment https://github.com/wez/wezterm/pull/5576/files#r1709748973

e82eric:
Thanks! updated

dev-lop36:
What is the status of this PR? Really great feature&work btw!

e82eric:
Thanks!  I think this is more or less still pending review.  I am happy to put in the work to get this across the the finish line, but don't want bug Wez about it,  I am sure he has a million priorities.  In the mean time if anyone wants to give it a try and notice any issues or want to review the code, I can work through those.

mux/src/tab.rs
```diff
@@ -1861,6 +1945,42 @@ impl TabInner {
         None
     }
 
+    fn get_float_size(&self) -> TerminalSize {
```
wez, line 1948: To me, `get_float_size` implies getting the size of the current float, but this is computing the desired size, so I think it should be renamed to make it more clear

```suggestion
    fn compute_float_size(&self) -> TerminalSize {
```

mux/src/tab.rs
```diff
@@ -1952,6 +2072,27 @@ impl TabInner {
...
+        if self.float_pane.is_some() {
+            anyhow::bail!("cannot add float while another float is active")
+        }
+
+        self.float_pane = Some(Arc::clone(&pane));
+        {
+            pane.resize(float_size)?;
+        }
+
+        Ok(())
```
wez, line 2093:
```suggestion
        pane.resize(float_size)?;
        Ok(())
```

wezterm-gui/src/commands.rs
```diff
@@ -1445,6 +1445,23 @@ pub fn derive_command_from_key_assignment(action: &KeyAssignment) -> Option<Comm
             menubar: &["Edit"],
             icon: Some("md_content_copy"),
         },
+        FloatPane(SpawnCommand {
+                          domain: SpawnTabDomain::CurrentPaneDomain,
+                          ..
+                      }) => CommandDef {
+            brief: label_string(action, "Create a float pane".to_string()).into(),
```
wez, line 1452:
```suggestion
            brief: label_string(action, "Create a floating pane".to_string()).into(),
```

wezterm-gui/src/commands.rs
```diff
@@ -1445,6 +1445,23 @@ pub fn derive_command_from_key_assignment(action: &KeyAssignment) -> Option<Comm
...
+            doc: "Create a float pane"
```
wez, line 1453:
```suggestion
            doc: "Create a floating pane"
```

wezterm-gui/src/commands.rs
```diff
@@ -1501,6 +1518,15 @@ pub fn derive_command_from_key_assignment(action: &KeyAssignment) -> Option<Comm
             menubar: &[],
             icon: Some("cod_split_vertical"),
         },
+        FloatPane(_) => CommandDef {
+            brief: label_string(action, "Create a float pane".to_string()).into(),
```
wez, line 1522:
```suggestion
            brief: label_string(action, "Create a floating pane".to_string()).into(),
```

wezterm-gui/src/commands.rs
```diff
@@ -1501,6 +1518,15 @@ pub fn derive_command_from_key_assignment(action: &KeyAssignment) -> Option<Comm
...
+            doc: "Create a float pane"
```
wez, line 1523:
```suggestion
            doc: "Create a floating pane"
```

wezterm-gui/src/termwindow/mod.rs
```diff
@@ -541,10 +538,6 @@ impl TermWindow {
         // force cursor to be repainted
         window.invalidate();
 
-        if let Some(pane) = self.get_active_pane_or_overlay() {
-            pane.focus_changed(focused);
-        }
-
```
wez, line 547: removing this logic is probably a regression?

wezterm-gui/src/termwindow/mouseevent.rs
```diff
@@ -667,7 +667,18 @@ impl super::TermWindow {
             Some(MouseCapture::TerminalPane(_))
         );
 
-        for pos in self.get_panes_to_render() {
+        let panes = if let Some(float_pane) = self.get_float_pane_to_render() {
+            if position.column >= float_pane.left && position.column <= float_pane.left + float_pane.width
+                && position.row as usize >= float_pane.top && position.row as usize <= float_pane.top + float_pane.height {
+                vec![float_pane]
+            } else {
+                return
```
wez, line 675: does this return prevent mouse handling from working in the other panes when a float is present?
Is that intentional? If so, please add a comment around this block to explain this, as it isn't clear from the code.

Also, I prefer an early return style to avoid lots of nested blocks, something like this:

```rust
let mouse_in_float = position.column >= float_pane.left && 
    position.column <= float_pane.left + float_pane.width &&
    position.row as usize >= float_pane.top &&
    position.row as usize <= float_pane.top + float_pane.height;

if !mouse_in_float {
  // Mouse events are not dispatched to the other panes when
  // a floating pane is active, because REASON
  return;
}
vec![float_pane]
```

wezterm/src/cli/mod.rs
```diff
@@ -103,6 +103,14 @@ Outputs the pane-id for the newly created pane on success"
     )]
     SplitPane(split_pane::SplitPane),
 
+    #[command(
+        name = "float-pane",
+        rename_all = "kebab",
+        trailing_var_arg = true,
+        about = "spawn a float pane"
```
wez, line 110:
```suggestion
        about = "spawn a floating pane"
```

wezterm/src/cli/split_pane.rs
```diff
@@ -113,3 +113,46 @@ impl SplitPane {
         Ok(())
     }
 }
+
+#[derive(Debug, Parser, Clone)]
+pub struct FloatPane {
+    /// Specify the pane that should be split.
```
wez, line 119: this comment doesn't seem to fit the action

wez (commented):
Many thanks for diving into this, and doing so deeply!

I haven't had the bandwidth to look at this in a great deal of detail, just skimmed here on GH without trying to run it, but have a few comments/suggestions.

There are a number of places with TODOs and comments about refactoring to reduce duplication: I would like to avoid having duplicate or almost identical code for the floating pane stuff, so I would appreciate seeing that tackled before we get to merging this.

I think there are perhaps some other bigger picture or weird edge-case questions around this that I think it would be good to have answers for; there may be more of these in the long run, but these are the ones more or less off of the top of my head:

* How does and how should the floating pane be reflected in `wezterm cli list`?
* If you create a floating pane against a multiplexer and detach, does re-attaching preserve the float?
* If you have one wezterm connected to a mux, then attach a second wezterm instance to that same mux and spawn a float in one, does the float instantly appear in the other one?
* How does the scrollbar UI/UX work with the floating pane? I'm OK if "it doesn't" is the answer if we document that it doesn't, and ensure that it is well-behaved.
* Are there any special gotchas when it comes to using the floating pane?
* Is it necessary to add an explicit kill-floating-pane action, or are the existing actions sufficient?
* How does the floating pane get exposed to the various window/tab styling events in lua, and is it likely to cause problems with existing setups?

wezterm-gui/src/termwindow/mod.rs
e82eric: reverted.  (sorry, I should have caught this).

e82eric:
Thanks Wez!  I pushed a commit to address the initial feedback.

I think the multiplexing, scrollbar and lua events are probably all gaps.  Will start working through those and will @ you when I get through them.  Marking the PR as draft until I get through them.

quantonganh:
@e82eric Thank you for taking your time to implement this.

> For the pane selection my intention was to treat the float like a overlay and turn any selection operations into no-ops and prevent the mouse from selecting a pane (I think there were some scenarios that I missed like the selection UI). I pushed an update that I think turn those into nops now. This may not be the right strategy, is there a scenario where you would want to be able to select/interact with the panes underneath the float?

I'm using [Helix](https://helix-editor.com/) as my IDE with the help of WezTerm and other CLI tools (lazygit, tig, fzf, gh, ...). For example, I've set up key bindings to run my code in the bottom pane, or connect to a database in the right pane using `lazysql`, ... Actually, I've been trying your PR to move some of these actions to a floating pane. However, I find myself needing to frequently switch between the floating pane and the one beneath it. It would be really helpful if we can do that by using some shortcut keys.

Thanks,

e82eric:
@quantonganh right now it has a default of CTRL|ALT|SHIFT(P) for creating the float pane, and can be closed using the normal binding for CloseCurrentPane.  It can be updated in the config with something like this.

```
config.keys = {
    {
      key = 'e',
      mods = 'CTRL|SHIFT',
      action = wezterm.action.FloatPane,
    },
    key = 'w',
      mods = 'CTRL|SHIFT',
      action = wezterm.action.CloseCurrentPane { confirm = false },
    }
}
```

Does that help or am I misunderstanding the problem?  Are you looking for something like being able to hide the floating pane instead of closing it when switching between the floating pane and the one beneath it?

quantonganh:
> Are you looking for something like being able to hide the floating pane instead of closing it when switching between the floating pane and the one beneath it?

Exactly! For example: while coding, I might want to check something in the database, I can bind a key to send `lazysql` to the floating pane. Afterward, I want to switch back to my main pane (Helix editor) without closing the floating pane - just hiding it.

I realized that to achieve this, we might need to add additional values, such as `floating` and `beneath` to the `direction` argument for the [get-pane-direction](https://wezfurlong.org/wezterm/cli/cli/get-pane-direction.html) command. This would allow checking whether a floating pane already exists.

e82eric:
@quantonganh if you get a chance can you try the latest commit.  I made an attempt at it and wanted to see if you notice anything weird and if it matches what you were looking for,  I think being able to hide the float may make more sense ux wise (I can now just hide the floating pane if someone clicks outside of it, etc).

ctrl+shift+e is the default for ToggleFloatingPane

aleksandersumowski:
Do you expect to be able to have more than one floating pane at a time? My use case would be to have short lived panes for interactive use, eg to implement [extracto](https://github.com/laktak/extrakto) like functionality. I was wondering if long lived pane and short lived ones could co-exist

quantonganh:
@e82eric It's working fine now and I can use Ctrl+Shift+E to toggle the floating pane. Thank you!

Regarding the `wezterm cli list --format json`, I think it would be useful to add a field like `is_floating`, similar to `is_active` and `is_zoomed`, to indicate whether a pane is floating.

My workflow would be something like this:
- Check if there's a floating pane in the current tab
- If there is, send the command directly to the floating pane
- If not, spawn a floating pane and send the command to it

To support this, we'd also need to enable `wezterm cli activate-pane` to work with floating panes.

What do you think?

e82eric:
@quantonganh have been trying to get multiple floating panes working and it has taken a lot longer than expected.  

For the cli list,  it is probably going to take some iteration to get the data model right.  

The float floating panes are in their own storage outside of the regular panes which are a binary tree and the floating panes are a regular list.  Would it work in your use case if cli list just had the clients in the list but not an indication of it is a floating pane or not and then separate command for list-floating-panes?  Was thinking it would be useful to have a sperate command with floating pane specific information like what index the pane is so you toggle floating panes by index ex leader 1 toggles lazygit and leader 2 toggles htop

quantonganh:
@e82eric I don't currently need multiple floating panes, but I do find the feature useful. That was the question from @aleksandersumowski.

> Would it work in your use case if cli list just had the clients in the list but not an indication of it is a floating pane or not and then separate command for list-floating-panes?

In my case, that's fine. But I'm not sure if we should have a separate command for listing floating panes. @wez might have more thoughts on this.

e82eric:
@quantonganh  That makes sense.  I will try to add the is_floating property and see why activate-pane isn't working.

It looks like Zellij will create the floating pane if one doesn't exist when you call toggle-floating-pane, if this did the same would it simplify the workflow at all?

e82eric:
@quantonganh I just pushed the changes for the is_floating property and active-pane.  If you get a chance could you try it out?

quantonganh:
@e82eric I've pulled your latest commit and tested it, but there seems to be an issue with listing panes:

```
$ ./target/release/wezterm cli list --format json
14:56:30.039  ERROR  wezterm_client::client > Error while decoding response pdu: Invalid Bool Encoding
14:56:30.039  ERROR  wezterm                > Error while decoding response pdu: Invalid Bool Encoding; terminating
```

Additionally, I'm unable to switch between the normal and floating panes using `activate-pane` (though switching between normal panes works fine).

e82eric:
@quantonganh can you try to restart wezterm-mux-server, think that error happens when the mux server is still running on the previous version that doesn't know to send the new is_floating property.

quantonganh:
@e82eric The issue still persists when starting `wezterm-mux-server` first (as a separate processs):

```
$ ps -ef | grep wezterm
  501 26436     1   0 10:45AM ??         0:00.84 ./e82eric/wezterm/target/release/wezterm-mux-server --pid-file-fd 3
  501 46561 83615   0 11:00AM ttys000    0:01.45 ./e82eric/wezterm/target/release/wezterm-gui connect unix
  ```
  
  ```
  $ ./target/release/wezterm cli list --format json
11:02:12.487  ERROR  wezterm_client::client > Error while decoding response pdu: Invalid Bool Encoding
11:02:12.488  ERROR  wezterm                > Error while decoding response pdu: Invalid Bool Encoding; terminating
```

Please note that, in some cases, the use might start `wezterm-gui` directly without first launching `wezterm-mux-server`.

e82eric:
@quantonganh I bumped the codec version, can you try a clean and build, make sure there aren't any running wez* processes and try again?

quantonganh:
@e82eric Thank you!

The listing of panes and switching between the floating pane and the pane beneath it are working as expected. However, it seems the `is_active` field of the floating pane does not update when switching back to the normal pane:

```
$ echo $WEZTERM_PANE
0

$ ./target/release/wezterm cli list --format json
[
  {
    "window_id": 0,
    "tab_id": 0,
    "pane_id": 0,
    "window_title": "./target/release/wez ~/C/g/e/wezterm",
    "is_active": true,
    "is_zoomed": false,
    "is_floating": false,
    "tty_name": "/dev/ttys001"
  },
  {
    "window_id": 0,
    "tab_id": 0,
    "pane_id": 2,
    "window_title": "./target/release/wez ~/C/g/e/wezterm",
    "is_active": true,
    "is_zoomed": false,
    "is_floating": true,
    "tty_name": "/dev/ttys002"
  }
]
```

e82eric:
@quantonganh just pushed a fix.

quantonganh:
@e82eric It's ok now:

```
$ echo $WEZTERM_PANE
0

$ ./target/release/wezterm cli list --format json
[
  {
    "window_id": 0,
    "tab_id": 0,
    "pane_id": 0,
    "window_title": "./target/release/wez ~/C/g/e/wezterm",
    "is_active": true,
    "is_zoomed": false,
    "is_floating": false,
    "tty_name": "/dev/ttys001"
  },
  {
    "window_id": 0,
    "tab_id": 0,
    "pane_id": 2,
    "window_title": "./target/release/wez ~/C/g/e/wezterm",
    "is_active": false,
    "is_zoomed": false,
    "is_floating": true,
    "tty_name": "/dev/ttys002"
  }
]
```

However, for some reasons, `wezterm cli activate-pane --pane-id 0` does not switch back from the floating pane to the normal pane, although it works perfectly in the reverse direction (from the normal pane to the floating pane).

e82eric:
@quantonganh just pushed fix for activate-pane --pane-id 0

akthe-at:
I've been watching this PR for a while, just wanted to cheer you on @e82eric  and say great job, keep it up.

e82eric:
Quick update on this

New stuff.
- Added support for hiding/showing the floating pane.
- Added support for multiple floating panes.
  - Switching between these uses the left and right pane navigation
  - @aleksandersumowski does this match what you were looking for with temporary panes?
- Added support for moving an existing pane to a floating pane
- Added support for moving a floating pane to a vertical or horizontal split

Items from PR review
- The floating panes now shows in the results of `wezterm cli list`
  - The json output has an additional is_floating property
- `cli activate-pane --pane-id` now works with floating panes
- `cli kill-pane --pane-id` works with floating panes
- Re-attaching multiplexer now retains floating pane visibility and active index.
- When multiple clients are attached to multiplexer, changes get updated on both screens now
- The floating pane does not currently work with the scrollbar
  - This should be possible.  (I spent some time trying to get this to work and I think I just need to spend some more time to figure out how to get it to draw in the correct place)
    - This one may make sense as a separate PR
- Special Gotchas.
  - Working on compiling a list
- Lua Window/Tab Styling events
  - Need to figure out how to test/validate these.

I think it will take me a week or 2 more to get this in a state to move out of draft.

If anyone has a chance please try the latest and let me know if you find any issues.

Multiple floating panes:
multiple-floating-panes

Re-attach multiplexer:
reattach

Multiple Clients:
multiple-clients

aleksandersumowski:
looks great from my perspective!

one minor question: is it be possible to pass a command to execute in the pane as an optional parameter? or does it always launch system shell?

e82eric:
@aleksandersumowski are you thinking of something like being able to run fzf in a floating pane?

This is my attempt at being able to run fzf and use the result in the current pane (not sure if the send-text is the best way to do this).
```lua
local function spawn_new_tab_with_pane_info(window, pane)
  local current_pane_id = pane:pane_id()

  local command = string.format([[
    #!/bin/bash
    current_pane_id=%s
    selected_commit=$(git log --oneline | fzf | awk '{print $1}')
    if [ -n "$selected_commit" ]; then
      wezterm cli activate-pane --pane-id "$current_pane_id"
      wezterm cli send-text --pane-id "$current_pane_id" -- "$selected_commit"
    fi
    exec $SHELL
  ]], current_pane_id)

  window:perform_action(
    wezterm.action.SpawnCommandInNewFloatingPane {
      args = { 'bash', '-c', command },
    },
    pane
  )
end

config.keys = {
  {
    key = 'y',
    mods = 'CTRL|SHIFT',
    label = 'floating fzf',
    action = wezterm.action_callback(spawn_new_tab_with_pane_info),
  },
}
```

Yukaii:
I hope `wezterm cli spawn` will also have the ability to start commands in a floating panel, so I can be used like `tmux popup` command!

quantonganh:
@Yukaii Have you tried something like this?

```
$ wezterm cli spawn --floating-pane lazygit
```

aleksandersumowski:
😍😍😍
this is exactly the kind of possibility I was thinking about - I can also see that the floating panel is closing once the fzf is done. Ideally I would like more possibilities than fzf without additional wrapper can provide, but that seems very much doable with the floating pane as is. 

floating-pane

tbh I've tried compiling your branch and testing it myself but `./target/debug/wezterm` was still complaining that SpawnCommandInNewFloatingPane - I must be doing something incorrectly

e82eric:
@aleksandersumowski just noticed in the screenshot that you were on the main branch from my fork, I think that might be why SpawnCommandInNewFloatingPane isn't working,  the floating pane work is on the float-pane branch.

e82eric:
@Wez quick heads up, this is ready to be reviewed again.  (No rush)

New stuff.
- Added support for multiple floating panes.
- Added support for hiding/showing the floating panes.
  - Switching between these uses the left and right pane navigation.  Up and down are currently no-op
- Added support for moving an existing pane to a floating pane
- Added support for moving a floating pane to a vertical or horizontal split

Items from PR review
- The floating panes now show in the results of wezterm cli list
  - The json output has an additional is_floating property
- cli activate-pane --pane-id now works with floating panes
- cli kill-pane --pane-id now works with floating panes
- Re-attaching multiplexer now retains floating pane visibility and active index.
- When multiple clients are attached to multiplexer, changes now get updated on both screens
- The floating pane does not currently work with the scrollbar
  - This should be possible. (I spent some time trying to get this to work and I think I just need to spend some more time to figure out how to get it to draw in the correct place)
  - Does it work if I add this as a follow-up PR?
- Special Gotchas.
  - If you close the last non floating pane on a tab it also closes the floating panes
  - If the floating panes are hidden it is kind of hard to know that they are there.
  - Pane selection UI is no-op when floating panes are visible
  - If you click outside of the floating pane it will activate the clicked pane and hide the floating pane.  It may be hard for the user to know that they need to toggle the floating panes to unhide them.
  - Pane selection by index only applies to non floating panes.  if a floating pane is visibile when select by index is called the floating panes will be hidden
- Lua Window/Tab Styling events
  - Need to figure out how to test/validate these.  Do you happen to have an example?

Stuff that works with floating panes
- cli list
- cli spawn --floating
- cli activate-pane --pane-id
- cli move-pane-to-new-tab
- cli kill-pane
- cli zoom-pane
- pane level overlays work (quick select, search, copy, etc)
- tab/window overlays work
  - floating panes will be covered while the overlay is active

Stuff that doesn't work
- split pane when floating pane is active
  - This seems possible but I am not sure if it would be useful
- get-pane-direction
  - This may take some design.  (Maybe up or down or a new floating direction etc)
- Adjust pane size
  - Should be possible.  Right now my thought is to add this in a follow up PR
- Adjust position.
  - Should be possible.  Right now my thought is to add this in a follow up PR
- Scrollbar
  - Should be possible.  Right now my thought is to add this in a follow up PR
- Activate-pane-direction
  - This may take some design.  (Maybe up or down or a new floating direction etc)
- pane-select
  - This would take some design to figure out how to figure out what parts of the underlying panes are not hidden by the floating pane.

darianmorat:
@aleksandersumowski I know this is a bit out of the topic (topic that I love btw, floating panes is perfect). But may I know how did u got the padding for the tabs? (I know u can add padding, but to the content in the window, not to the bottom tabs as you. Thanks

mikiadev:
Is there something preventing this from being merged? As I really need this feature I'm now using on multiple systems @e82eric 's version and it works perfectly...

jacobb:
I've been using @e82eric's branch for almost two months now with no issues. Rebased against main a few days ago to get the screen redraw fixes and continues to be rock solid.

mgpinf:
Hi @e82eric. Encountering issue where the text is rendering beyond the pane when a program like htop is running in a split and moving the pane to floating window.
- For an interactive program running in horizontal split, when it is converted to floating pane, the overflow happens towards the bottom.
- For an interactive program running in vertical split, when it is converted to floating pane, the overflow happens towards the right.

Attaching images for reference
image
image

e82eric:
@mgpinf taking a look,  quick question are you running in the multiplexer mode?

e82eric:
@mgpinf nevermind the question about the multiplexer,  I was able to repro it, and I think I found the issue, if you get a chance pull the latest and check if it works now.

RoloEdits:
Ive noticed that the pane doesn't seem to fill up the border, at least on the left side. Not sure of the right, but top and bottom are fine.
image

e82eric:
@RoloEdits trying to repro the issue,  any chance you can share your config?

RoloEdits:
Sure. I am on windows as well, by the way.

```lua
config.floating_pane_padding = {
  left = '5%',
  right = '5%',
  top = '10%',
  bottom = '10%',
}

config.floating_pane_border = {
  left_width = '0.2cell',
  right_width = '0.2cell',
  bottom_height = '0.10cell',
  top_height = '0.10cell',
  left_color = '#cdd6f4',
  right_color = '#cdd6f4',
  bottom_color = '#cdd6f4',
  top_color = '#cdd6f4',
}
```

e82eric:
@RoloEdits I think I found the issue,  if you get a chance can you try the latest commit?

I think this is what you are looking for.

```lua
{ key = '`', mods = 'CTRL', action = act.ToggleFloatingPane }
```

RoloEdits:
@e82eric Yep, that fixed it!
image

And that key toggle is exactly what I was looking for.

Thanks for the help and this feature, its working great.

RoloEdits:
One thing im not sure is functioning correctly or if its a bug is what pane is shown after a floating pane is closed. For example, if I have a "persistent"  floating terminal hidden, for example to edit code but want to keep the state around, and then bring up lazygit in a newly spawned floating window, when I quit out of that, the expectation is that I will be back in my text editor, the last pane I was in. But when lazygit closes, the supposed to be hidden floating terminal pane shows up and captures the cursor, needing me to hide it again.

e82eric:
@RoloEdits I just pushed an update so that when you close a floating pane it will show the regular pane beneath instead of showing the next floating pane.

quantonganh:
Is anyone here using [Helix](https://github.com/helix-editor/helix)?

It has an option to enable [auto save](https://docs.helix-editor.com/editor.html#editorauto-save-section):

```
[editor]
auto-save=true
```

I've just noticed that the auto-save feature doesn't trigger when switching to a floating pane. However, it works correctly when switching to a normal pane.

Could it be that the focus events aren't emitted in this case?

krskrs:
@e82eric my sincere congratulations and thank you for your significant work! I hope it gets reviewed soon :)

e82eric:
> Is anyone here using [Helix](https://github.com/helix-editor/helix)?
> 
> It has an option to enable [auto save](https://docs.helix-editor.com/editor.html#editorauto-save-section):
> 
> ```
> [editor]
> auto-save=true
> ```
> 
> I've just noticed that the auto-save feature doesn't trigger when switching to a floating pane. However, it works correctly when switching to a normal pane.
> 
> Could it be that the focus events aren't emitted in this case?

@quantonganh if you get a chance could try the latest commit.

quantonganh:
> @quantonganh if you get a chance could try the latest commit. I think I found the issue

@e82eric Thank you for your quick fix. I confirm that it's working fine now.

Icvza:
Hey ya'll I'm probably WAY off but besides the floating part I think this can be done with 

```
--mini term
	{
		key = "k",
		mods = "CMD",
		action = wezterm.action_callback(function(_, pane)
			local tab = pane:tab()
			local panes = tab:panes_with_info()
			if #panes == 1 then
				pane:split({
					direction = "Right",
					size = 0.4,
				})
			elseif not panes[1].is_zoomed then
				panes[1].pane:activate()
				tab:set_zoomed(true)
			elseif panes[1].is_zoomed then
				tab:set_zoomed(false)
				panes[2].pane:activate()
			end
		end),
	}
```
along with Dim inactive panes
```
config.inactive_pane_hsb = {
	saturation = 1,
	brightness = 0.24,
}
```

https://github.com/user-attachments/assets/f662a785-57d7-4895-b890-1f43139da4b8

Rahmyeon:
Hey thanks for your hard work. But do you think it would be possible to make this a plugin for wezterm, while your waiting?

oskarkook:
@Icvza I've used a very similar approach for months, but pane zoom messes up the sizing of my neovim splits. As I usually use just vertical splits, it can be countered by spawning a pane on the bottom and zooming that, but it's still not ideal.

I've been test-driving this branch and it's great. Great work, @e82eric!

Rabeez:
> ... but pane zoom messes up the sizing of my neovim splits, because the panes' sizes change. As I usually use just vertical splits, it can be countered by spawning a pane on the bottom and zooming that, but it's still not ideal.

@oskarkook `kwkarlwang/bufresize.nvim` might solve the nvim pane sizing issue you're having. Have been using this with wezterm splits for a while.

mgpinf:
@e82eric Since there are multiple floating panes in each tab, but only one can be rendered (if we choose to show), I think it would be useful to have an option to switch between floating panes or create a new floating pane.

This can be done by adding `FLOATING_PANES` in `ShowLauncherArgs` flags, similar to `WORKSPACES`

quantonganh:
@mgpinf https://github.com/wezterm/wezterm/pull/5576#issuecomment-2415738290

You can switch between floating panes by using the same keys that are used to activate the left/right pane:

```lua
  {
    key = 'h',
    mods = 'ALT',
    action = act.ActivatePaneDirection 'Left',
  },
  {
    key = 'l',
    mods = 'ALT',
    action = act.ActivatePaneDirection 'Right',
  },
```

pablospe:
Wait... closed :( This was an amazing piece of work, thanks @e82eric!

rachartier:
Yes, that could be a huge loss... :(

jacob-bayer:
What happened?

e82eric:
Hi!  I am hoping to be able to re-open this PR now that activity is picking up on the repository again.  But want to wait until it a good time (Bew mentioned after the next release would make sense).

PR #5759, opened 2024-06-26
Add option to extend the osx non-native fullscreen behind the notch to the top of the screen

wryanzimmerman:
This just adds a boolean option: `macos_fullscreen_extend_behind_notch` which is checked before the osx window calculates the borders to make non-native fullscreen  go all the way to the top of a mac laptop with a notch, for use-cases (like mine) where that is useful. 

I was also having issues with multiple monitors making the notch-avoiding window jump up and down when the active monitor changes on the release version, but I haven't been able to check if that issue still exists on `main` and whether or not this option fixes it. I will once I'm back with my external monitor!


wryanzimmerman:
This is how it looks with the option enabled:

IMG_3418

window/src/os/macos/window.rs
```diff
@@ -879,42 +879,44 @@ impl WindowOps for Window {
...
-                        left: ULength::new(insets.left.ceil() as usize),
-                        right: ULength::new(insets.right.ceil() as usize),
-                        bottom: ULength::new(insets.bottom.ceil() as usize),
-                        color: crate::color::LinearRgba::with_components(0., 0., 0., 1.),
-                    })
-                } else {
-                    None
+        let border_dimensions = if window_state.contains(WindowState::FULL_SCREEN)
+            && !native_full_screen
+            && !_config.macos_fullscreen_extend_behind_notch
```
wez, line 884: Please rename the `_config` function parameter to `config`; the leading underscore is a rust convention to indicate that a variable/binding is unused. Since it is now used, that prefix should be removed.

```suggestion
            && !config.macos_fullscreen_extend_behind_notch
```

wez (commented):
Thanks for this!
Please also add a page to the docs; you can use https://github.com/wez/wezterm/blob/main/docs/config/lua/config/macos_window_background_blur.md?plain=1 as a basis for it, creating a page for this new option alongside that one, but use `{{since('nightly')}}` for the version tag.

You can run `./ci/build-docs.sh serve` to run the docs locally and preview how it looks.

wez:
Thank you!

PR #5767, opened 2024-06-29
nix/flake.nix: remove invalid follows

khaneliman:
Rust overlay no longer uses flake-utls as of https://github.com/oxalica/rust-overlay/pull/180


wez:
Thanks!

PR #5775, opened 2024-07-02
nix/flake.nix: add a build step for a macos application bundle

x0ba:
Builds an Applications/WezTerm.app with the flake


x0ba:
seems to be the same pull request as #5073

PR #5779, opened 2024-07-03
wezterm-ssh: add support for comments in `Match` statements

### LINKED ISSUE #5498: Trips on inline comment in SSH config

Linux X11

xfwm4 4.18.0 on Fedora 40

### WezTerm version

20240520-135708-b8f94c47

### Describe the bug

Trying to connect to my other machine on the home network by choosing "Shell: New Tab (Domain SSH:machine)" causes this error:
```
Using libssh-rs to connect to user@192.168.1.122 # 219 on wifi:22
Error: Fatal: failed to set option

⚠️  Process RemoteSshDomain didn't exit cleanly
Exited with code 1.
This message is shown because exit_behavior="CloseOnCleanExit"
```
The corresponding config in `~/.ssh/config` is:
```
Host    machine
        User            user
        HostName        192.168.1.122 # 219 on wifi
```

I think it's failing to parse the `HostName` line. It includes the inline comment as the hostname.  When I move the comment to the next line, I can connect.

### To Reproduce

Try to connect to a host where the corresponding config has an inline comment in the line with `HostName`
```
Host    machine
        User            user
        HostName        192.168.1.122 # 219 on wifi
```

### Configuration

no config

### Expected Behavior

I can connect to the remote host.

### Logs

_No response_

### Anything else?

_No response_

hdonnay:
Looking at the openssh-portable code, it seems that inline comments are only handled specially on `Match` and `ProxyJump` statements. This change adds inline comment handling.

Closes: #5498
See-also: https://github.com/openssh/openssh-portable/blob/ee6b9e661633fcefd29dba0c811cecbc4d027f6f/readconf.c#L652

### DIFF
--- a/wezterm-ssh/src/config.rs
+++ b/wezterm-ssh/src/config.rs
@@ -237,6 +237,8 @@ impl ParsedConfigFile {
             if line.is_empty() || line.starts_with('#') {
                 continue;
             }
+            // Only `Match` and `ProxyJump` options seem to allow inline comments.
+            // The latter is not handled.
 
             if let Some(sep) = line.find(|c: char| c == '=' || c.is_whitespace()) {
                 let (k, v) = line.split_at(sep);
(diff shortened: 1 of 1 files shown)


wezterm-ssh/src/config.rs
```diff
@@ -293,7 +295,14 @@ impl ParsedConfigFile {
                     let mut criteria = vec![];
                     let mut context = Context::FirstPass;
 
-                    let mut tokens = v.split_ascii_whitespace();
+                    let mut tokens = v.split_ascii_whitespace().map_while(|tok| {
```
wez, line 298: Hmm, I don't think this works as you intend, because I think this results in ignoring only the `# fun` token but not the `games` token from your expanded test case?

A more robust alternative would be to do something like:

```rust
let v, _commented = v.split_once('#').unwrap_or(v);
let mut tokens = v.split_ascii_whitespace();
```

Please expand the test case to explicitly verify the effective values for the `games` host so that we can prove/disprove this.

wez (commented):
Thanks for this!

PR #5780, opened 2024-07-04
Simplify wayland configure event

tmccombs:
Remove the unnecessary PendingEvent struct, and just call functions directly. I think this is easier to reason about, and fits better with the current architecture of SCTK 0.17

### DIFF
--- a/window/src/os/wayland/window.rs
+++ b/window/src/os/wayland/window.rs
@@ -147,11 +147,6 @@ impl KeyRepeatState {
     }
 }
 
-enum WaylandWindowEvent {
-    Close,
-    Request(WindowConfigure),
-}
-
 #[derive(Clone, Debug, PartialEq, Eq, Hash, Ord, PartialOrd)]
 pub struct WaylandWindow(usize);
 
@@ -181,7 +176,6 @@ impl WaylandWindow {
             .wayland();
 
         let window_id = conn.next_window_id();
-        let pending_event = Arc::new(Mutex::new(PendingEvent::default()));
 
         let (pending_first_configure, wait_configure) = async_channel::bounded(1);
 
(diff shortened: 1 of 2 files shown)


window/src/os/wayland/window.rs
```diff
@@ -1323,18 +1184,58 @@ impl WindowHandler for WaylandState {
...
+            state
+        );
+        window_inner.window_state = state;
+
+        window_inner.window_frame.update_state(configure.state);
+        window_inner
+            .window_frame
+            .update_wm_capabilities(configure.capabilities);
+
+        // TODO: should we resize if only one of width or height is specified?
```
wez, line 1230: I feel like we should handle that case

wez (commented):
Thanks for this!

On startup, the border frame now matches the window dimensions under weston (which may be a more general side effect of the sctk upgrade, I didn't try to run down when that happened).

However, when I resize the window, it appears to be 1 cell out of sync:

image

is that something that is logically part of this PR? weston only seems to allow me to resize either horizontally or vertically, so I wonder if this is related to the TODO comment about width+height only triggering some work somehow?

tmccombs:
> is that something that is logically part of this PR?

that is definitely not intentional. 

My commit to handle the case where only one dimension is given doesn't seem to help.  I'm not sure what the actual problem is yet. Interstingly, I _don't_ get the same problem if I run it in floating mode in sway, and resize it. I'm not sure what the difference is.  Maybe sway doesn't let the client resize the window itself?

wez:
This PR doesn't rebase easily on current `main`. What would you like to do about it? Should we close it, or do you plan to pick it up again in the future?

tmccombs:
I fixed the conflicts. However, while going through it I noticed https://github.com/wezterm/wezterm/pull/5971#issuecomment-2650119313. 

I've done a little testing in sway, but haven't tested with other compositors.

PR #5814, opened 2024-07-14
wgpu -> 0.20

wez:
This compiles, but when I run it on X11 I get a fully transparent window.

I have not tried to debug this yet.
I'm throwing this up here to see if it compile on !linux.

This is based on a revert of bf07f6de0f73c15149538619588b78dc53f1a961 which was where wgpu 0.19 was rolled back previously.

### DIFF
--- a/window/src/os/x11/window.rs
+++ b/window/src/os/x11/window.rs
@@ -13,12 +13,14 @@ use async_trait::async_trait;
 use config::ConfigHandle;
 use promise::{Future, Promise};
 use raw_window_handle::{
-    HasRawDisplayHandle, HasRawWindowHandle, RawDisplayHandle, RawWindowHandle, XcbDisplayHandle,
-    XcbWindowHandle,
+    DisplayHandle, HandleError, HasDisplayHandle, HasWindowHandle, RawDisplayHandle,
+    RawWindowHandle, WindowHandle, XcbDisplayHandle, XcbWindowHandle,
 };
 use std::any::Any;
 use std::convert::TryInto;
+use std::num::NonZeroU32;
 use std::path::PathBuf;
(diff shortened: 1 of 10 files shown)


wez:
Updated to wgpu 22.1.

Still gives me a transparent window under X11. I don't have time to dig into this right now.
I do want to note that you can see wgpu's trace stuff by explicitly enabling it:

```
 WEZTERM_LOG=wgpu_core=trace,wgpu_hal=trace ./target/debug/wezterm start sleep 1
```

There's no glaringly obvious reason why this might be broken in this way, probably wouldn't hurt to engage with the wgpu folks for suggestions, but I have no bandwidth for this right now.

window/src/os/x11/window.rs
```diff
@@ -134,24 +136,24 @@ impl Drop for XWindowInner {
...
-unsafe impl HasRawWindowHandle for XWindowInner {
-    fn raw_window_handle(&self) -> RawWindowHandle {
-        let mut handle = XcbWindowHandle::empty();
-        handle.window = self.child_id.resource_id();
-        handle.visual_id = self.conn.upgrade().unwrap().visual.visual_id();
-        RawWindowHandle::Xcb(handle)
+impl HasWindowHandle for XWindowInner {
+    fn window_handle(&self) -> Result<WindowHandle<'_>, HandleError> {
+        let mut handle =
+            XcbWindowHandle::new(NonZeroU32::new(self.window_id.resource_id()).expect("non-zero"));
```
wez, line 154: This resolves my blank window issue:

```suggestion
            XcbWindowHandle::new(NonZeroU32::new(self.child_id.resource_id()).expect("non-zero"));
```

Issue #5812, opened 2024-07-14
Plugin aliases to enable plugin sub modules.


MLFlexer:
**Is your feature request related to a problem? Please describe.**
As a plugin developer I would like to be able to use multiple files. This is currently not well supported, and the plugin author must do some tricks to enable this. See #4248 and #4150.

**Describe the solution you'd like**
The ideal solution would be that I as a plugin developer or a user of the plugin is able to call require("<plugin_identifier.file_name_or_sub_module>") as i would when requiring other lua files. 
One way to achive this without breaking current plugins would be to add the option to alias plugins. That is, there is a directory where plugins are stored in subdirectories, with each subdirectory having the same name as the alias (plugin_identifier). Then the alias directory is added to the lua searchpath making it possible to have the descriped ideal solution, while also not breaking any existing plugins.

To add this functionality it would require:
1. adding the alias directory
I would suggest adding it as a parallel or a subdirectory of the current `data/wezterm/plugins` directory
2. Adding a function call to alias a plugin
This could be something like `wezterm.plugin.require_aliased(<url_str>, <alias_str>)`. This would function closely to the other require function call, but clone the repository into the aliased directory.
3. Update the `wezterm.plugin.update_all()` and `wezterm.plugin.list()` function calls to include aliased functions.
4. Add the aliased directory to the lua search path.

If my thought process is correct, then I belive that I am able to implement this and I am willing to give it a shot. I could also make documentation for this, as there is no documentation for the `wezterm.plugin` module.

**Describe alternatives you've considered**
I don't have any great alternatives at the moment.

**Additional context**
Sub directory approach
```
data/
└── wezterm/
    └── plugins/
        └── aliases/
            ├── plugin_a
            ├── plugin_b
            ├── ...
            └── plugin_z
```
Parallel approach
```
data/
└── wezterm/
    ├── plugins
    └── aliases/
        ├── plugin_a
        ├── plugin_b
        ├── ...
        └── plugin_z
```

MLFlexer:
A third option would be to add the path `<HOME_DIR>/.local/share/wezterm/plugins/?.lua;` and then have `wezterm.plugin.ensure_installed(<url>)` and `wezterm.plugin.ensure_install_named(<url>, <name>)` which would just ensure that the plugin is installed. Then the plugin author and user can then use `require(<name>)` as they like. 
[...]

PR #5820, opened 2024-07-15
Add plugin aliases

MLFlexer:
This PR should solve #5812 and #4248 when complete. 

I have changed the lua path to be able to call `require("<plugin_name>.<some_lua_file>")` as described [here](https://github.com/wez/wezterm/issues/5812#issuecomment-2229451131).
I have also added documentation for the `wezterm.plugin` and also added the functions `wezterm.plugin.require_as_alias` to clone repositories into a named directory.
Lastly i have added the `wezterm.plugin.plugin_dir` function to return the plugin directory root


MLFlexer:
Should also solve #4248

docs/config/lua/wezterm.plugin/index.markdown
```diff
@@ -0,0 +1,9 @@
+# `wezterm.plugin` module
+
+{{since('???')}}
```
MLFlexer, line 3: I am unsure of how to get the correct version for these.

yriveiro:
Thanks for this PR. I'm experimenting with Wezterm plugins and the experience is not too clean

It's impossible (as far as I know) to structure a plugin into different submodules. The submodules aren't found because the plugin is installed using the full URL path rather than a 'human-readable' name.

One question, Does this PR allow a Wezterm plugin to include another Wezterm plugin directly?

MLFlexer:
> Thanks for this PR. I'm experimenting with Wezterm plugins and the experience is not too clean
> 
> It's impossible (as far as I know) to structure a plugin into different submodules. The submodules aren't found because the plugin is installed using the full URL path rather than a 'human-readable' name.
> 
> One question, Does this PR allow a Wezterm plugin to include another Wezterm plugin directly?

It would allow you to be able to do the following:

```lua
wezterm.plugin.require_as_alias("github.com/user/repo", "cool_plugin_alias")

cool_sub_module = require("cool_plugin_alias.sub_module")
```

Then the plugin author can specify that a user should install it with a specific alias, or alternatively the plugin could call the `require_as_alias` function itself and then use that to ensure that the plugin submodules are loaded. This would then allow a plugin to use another if needed as the plugins are just an extention of the configuration

PR #5850, opened 2024-07-24
feat: Identify modifiers on Wayland/X11 (new)

### LINKED ISSUE #4626: fix: Implement modifier detection on wayland/x11

gabyx:
Fixes: #4626

- On Wayland the workaround goes by probing the keyboard for all keys and rebuilding the missing API which is not available as in X11.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6752,6 +6752,7 @@ dependencies = [
  "promise",
  "raw-window-handle",
  "resize",
+ "scopeguard",
  "serde",
  "shared_library",
  "shlex",
--- a/termwiz/src/escape/csi.rs
+++ b/termwiz/src/escape/csi.rs
@@ -46,7 +46,7 @@ fn csi_size() {
     assert_eq!(std::mem::size_of::<Cursor>(), 12);
     assert_eq!(std::mem::size_of::<Edit>(), 8);
     assert_eq!(std::mem::size_of::<Mode>(), 24);
-    assert_eq!(std::mem::size_of::<MouseReport>(), 8);
+    assert_eq!(std::mem::size_of::<MouseReport>(), 12);
     assert_eq!(std::mem::size_of::<Window>(), 40);
     assert_eq!(std::mem::size_of::<Keyboard>(), 8);
     assert_eq!(std::mem::size_of::<CSI>(), 32);
--- a/wezterm-input-types/src/lib.rs
+++ b/wezterm-input-types/src/lib.rs
@@ -482,19 +482,18 @@ bitflags! {
     #[dynamic(into="String", try_from="String")]
     pub struct Modifiers: u32 {
         const NONE = 0;
-
         const SHIFT = 1<<1;
-        const LEFT_SHIFT = 1<<10;
-        const RIGHT_SHIFT = 1<<11;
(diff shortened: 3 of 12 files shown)


gabyx:
@wez : not sure if I should give this PR some love still, it might fix some modifier issues or at least improve the status quo ? Do you know significant changes which happened which I need to be aware?

tuxmark5:
Would love to see this implemented, as I use both Hyper and Meta (in addition to "normal" Alt) to avoid key binding collisions, that way custom key bindings are easier to setup without interfering with anything that might run in the terminal itself.

gabyx:
Huch, this PR is really old, glad if somebody revives it. It clearly needs a rebase etc...

Issue #5750, opened 2024-06-24
wezterm terminfo leads to corrupted buffer text when scrolling in zellij + neovim vertical split


CarbonChauvinist:
### What Operating System(s) are you seeing this problem on?

Linux Wayland

### Which Wayland compositor or X11 Window manager(s) are you using?

Mutter + Zellij

### WezTerm version

20240203-110809-5046fc22

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

When setting  `term = wezterm` in config and using Zellij as multiplexer if one creates a vertical split in neovim scrolling in one buffer causes the other buffer's text to become garbled and corrupted and eventually disappear.

I realize there are a lot of potential culprits here and I've tried as best I can to isolate what may be contributing the unexpected behavior. To wit, not setting the term to wezterm and instead using the default xterm-256color does not cause the reported issue in Zellij + neovim. In fact Wezterm with wezterm terminfo without Zellij does not cause the reported behavior. My impression is that this may have something to do with the wezterm terminfo and how it interacts with Zellij and/or neovim. What gives me hope that it may be able to be addressed is that foot for instance has it's own custom terminfo foot-extra that works as expected with Zellij + neovim.

So far, I've tested the following terminals with the listed TERM settings and also did not experience the corrupted buffer text.

The terminals I tested where the issue did not present itself: (terminal/TERM)

- Foot with foot-extra TERM env var
- Sakura with xterm-256color TERM env var
- Kitty with xterm-kitty env var
- Wezterm with xterm-256color TERM env var

### To Reproduce

To reproduce:

1. run wezterm with term set to wezterm in [config](https://wezfurlong.org/wezterm/config/lua/config/term.html?h=term)
2. run zellij with default settings.
3. run development version of neovim also in a clean state with no plugins loaded (i.e. nvim --clean -u NONE -i NONE)
4. open any file
5. create a vertical split
6. scroll down in the current window and observe the reported issue in the neighboring window within neovim

Attached a short screen capture of the behavior for reference.

https://github.com/wez/wezterm/assets/21365869/9bd6e933-bac7-4ae2-8128-744d7a12f8ac

Also note, in Wezterm with wezterm TERM env this issue does not present itself when running latest version of vim available in the standard Arch repos, nor does it present itself when running neovim without zellij.

### Configuration

```
local wezterm = require("wezterm")
local mux = wezterm.mux

[...]

PR #5871, opened 2024-07-30
fix: Use cols not rows for left margin bound

tmccombs:
The location of the left margin should depend on the width, not the height.

Fixes: #5750


tmccombs:
It would probably be good to have tests for this, but I'm not familiar enough to feel comfortable doing that.

wez:
Ah, great find! Thanks for this, and to @j4james for spotting it!
Unsure if this was a brainfart on my part or copypasta gone wrong.

PR #5872, opened 2024-07-30
Update linux.md

KloolK:
Add command to set correct permission for `/usr/share/keyrings/wezterm-fury.gpg`

Following the instructions, the file was created with 640 on my machine. This lead to an error with `sudo apt update`:
```
Get:12 https://apt.fury.io/wez * InRelease
Err:12 https://apt.fury.io/wez * InRelease
  The following signatures couldn't be verified because the public key is not available: NO_PUBKEY D7BA31CF90C4B319
[...]
W: GPG error: https://apt.fury.io/wez * InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY D7BA31CF90C4B319
E: The repository 'https://apt.fury.io/wez * InRelease' is not signed.
```


wez:
Thanks, but I'm pretty sure that you need to run `sudo apt update` instead.

KloolK:
I did run it as root, I just forgot it in my description above (changed now).

All other `*.gpg` files in `/usr/share/keyrings` have 644. Changing it manually back to 640 will make the error occur again.

On my system, the umask for root is 027, causing the file to be created with 640 instead of 644. I don't know if this is the default for Ubuntu, but I'd be surprised if I'm the only one having this configuration.

Maybe a hint is more appropriate then making it part of the default instructions? I just want to spare others the time.

wez:
Thanks!

PR #5894, opened 2024-08-03
Create dirs before writing file.

aldhsu:
Loving wezterm! I was attempting to add new lua hooks and ran into this problem on MacOs.

# Expected
Clean build succeeds on MacOS

# Actual
Fails to build.

```
 Caused by:
      No such file or directory (os error 2)
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```


wez:
Thanks, but this shouldn't be needed if you follow the https://wezfurlong.org/wezterm/install/source.html instructions.
What's different about your environment/setup that results in you needing this?

aldhsu:
Ah I see the problem. I just blindly made the folder without looking at what it's doing.
I should have read the error more closely.

I use a custom CARGO_TARGET_DIR:
```toml
[build]
target-dir = "/Users/ahsu/.cargo/builds/"
```
This is so deps are shared.

Might just be my configuration but I tried `cargo build -vv --target-dir=target` and printing out the "CARGO_TARGET_DIR" env var. "CARG_TARGET_DIR" seems to always be empty?  This comment https://github.com/rust-lang/cargo/issues/9661#issuecomment-1769481293 seems to suggest that not passing from Cargo "CARGO_TARGET_DIR" to build scripts is intentional currently. 

A user can set it themselves though eg. `CARGO_TARGET_DIR=target cargo build -vv`.

The script currently falls back to the default "target".

A solution would be to walk back from "OUT_DIR". 
I'm not sure I know enough to suggest a good solution but if the above will suffice I can amend the PR.

aldhsu:
I've updated the PR to use OUT_DIR.

wez:
Can you rebase this PR so that I can tell the CI to run? Thanks!

aldhsu:
No problems! Done.

PR #5923, opened 2024-08-10
improve x11 startup times

blukai:
related issue: https://github.com/wez/wezterm/issues/5802

requesting `GetScreenResourcesCurrent` instead of `GetScreenResources` brings startup times down from 1.5 seconds to 30 microseconds on my machine (measured with dumb printlns).

according to some findings `GetScreenResourcesCurrent` may return nothing, so i kept `GetScreenResources` there as a fallback, just in case, although I'm not 100% sure that it's necessary. see:
- https://github.com/qt/qtbase/blob/c234700c836777d08db6229fdc997cc7c99e45fb/src/plugins/platforms/xcb/qxcbscreen.cpp#L963
- https://github.com/qt/qtbase/blob/c234700c836777d08db6229fdc997cc7c99e45fb/src/plugins/platforms/xcb/qxcbconnection_screens.cpp#L390

also worth to note that i3 does not rely on `xcb_randr_get_screen_resources`, but only on `xcb_randr_get_screen_resources_current`, see:
- https://github.com/search?q=repo%3Ai3%2Fi3%20get_screen_resources&type=code

### DIFF
--- a/window/src/os/x11/connection.rs
+++ b/window/src/os/x11/connection.rs
@@ -19,6 +19,34 @@ use x11::xlib;
 use xcb::x::Atom;
 use xcb::{dri2, Raw, Xid};
 
+enum ScreenResources {
+    Current(xcb::randr::GetScreenResourcesCurrentReply),
+    All(xcb::randr::GetScreenResourcesReply),
+}
+
+impl ScreenResources {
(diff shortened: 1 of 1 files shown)


wez:
Thank you!

PR #5940, opened 2024-08-13
chore: typo psuedo -> pseudo

FrancescElies:
In https://github.com/microsoft/terminal/issues/17688 among other things I have been pointed out that wezterm has `psuedo` instead of `pseudo` in the codebase.

## This pr
Just fixes a typo, this makes `rg -t rust pseudo` show what you expect.

I hope you don't find this pr impolite, as I am just renaming a file in a codebase I don't mantain.

## NOTEs
In https://github.com/microsoft/terminal/issues/17688 I was told that there might be some problems the way wezterm uses `conpty.dll`, a separate pr will follow on that.


FrancescElies:
closing in favor of #5977

Issue #5928, opened 2024-08-11
tab:set_zoomed not working when active pane is manually set


SpyMachine:
### What Operating System(s) are you seeing this problem on?

macOS

### WezTerm version

20240811-112835-a0a415d1

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

I'm trying to setup a workspace initializer where I have my editor and a terminal that is hidden by default. I.e. I want to zoom my editor by default. However the configuration (see below) doesn't work. The `editor_pane` is not zoomed. Interestingly, if I comment out `editor_pane:activate()` it does zoom the terminal pane which I'd expect so I think perhaps there's an issue with the zoom when the pane is "manually" activated.

### To Reproduce

For simplicity, I was able to reproduce the issue on "gui-startup" just to make it easy to test. So just copying that gui-startup event handler should show the issue.

### Configuration

```
wezterm.on("gui-startup", function()
	local _, _, window = wezterm.mux.spawn_window({})
	window:gui_window():maximize()
	local editor_tab = window:active_tab()
	editor_tab:set_title("editor")
	local editor_pane = editor_tab:active_pane()
	editor_pane:send_text("nvim -c ':Telescope find_files' .\n")
	local _terminal_pane = editor_pane:split({ direction = "Bottom", size = 0.15 })
	editor_pane:activate()
	editor_tab:set_zoomed(true)
end)
```

### Expected Behavior

I expected my "editor_pane" to be zoomed using the above configuration.

SpyMachine:
oh interesting.... printing out `tab:pane_with_info`, wezterm seems to think the pane is zoomed but it's not.

```
20:34:58.272  INFO   logging > lua: [
    {
        "height": 20,
        "index": 0,
        "is_active": true,
        "is_zoomed": true,
        "left": 0,
        "pane": MuxPane(pane_id:0, pid:85366),
        "pixel_height": 680,
        "pixel_width": 1280,
        "top": 0,
        "width": 80,
    },
    {
        "height": 3,
        "index": 1,
        "is_active": false,
        "is_zoomed": false,
        "left": 0,
        "pane": MuxPane(pane_id:1, pid:85366),
        "pixel_height": 102,
        "pixel_width": 1280,
        "top": 21,
        "width": 80,
    },
]
```

SpyMachine:
[...]

PR #5964, opened 2024-08-14
fix: add short circuit to set active pane

SpyMachine:
This is a go at addressing #5928. It solves the problem by short circuiting setting the active pane by checking if the requested pane is indeed already the active pane.

The end result is mainly the same with one caveat. The recency table is not getting updated if we short circuit...I am under the impression this won't change anything in a meaningful way as the active pane should already have the highest recency score. This might sort of be a bug as is...

If there's a good reason to keep the recency tag this can be refactored to ensure it gets set in the short circuit.


wez:
Thank you!

Issue #4984, opened 2024-02-08
SplitPane using `top_level = true` leaves behind ghost space


amanjitsk:
### What Operating System(s) are you seeing this problem on?

Linux X11

### Which Wayland compositor or X11 Window manager(s) are you using?

bspwm w/ picom

### WezTerm version

wezterm 20240208-085947-f671a9ab

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

When splitting across the full width (/length) using `SplitPane({ top_level = true })`, it causes the next call to `SplitPane` to fail, and even with the pane closed (this is the same bug reported in #2579), it seems that the space is still occupied by the now closed pane (created using top_level = true).

### To Reproduce

1. Start with a single pane in a `tab`
```
|       |
    1
|       |
```
2. Create a vertical split (split it bottom)
```
|   1   |
---------
|   2   |
```
3. Split Pane 2 horizontally to the right using the full height i.e.
   `SplitPane({ top_level = true, direction = "Right" }) 
```
| 1 |   |
----| 3
| 2 |   |
```
4. Trying to split Pane 3 does not work anymore. Furthermore, even 
   if Pane 3 is closed, splitting Pane 1 or 2 does not work either
   (see #2579)
5. Closing Pane 3 and 2 now, and repeating Step 2. now results in the
   following layout (unexpected)
```
| 1 |   |
----|  
| 2 |   |
```
with the space previously occupied by Pane 3 still empty.

### Configuration

no config

### Expected Behavior

Step 5. should result in a layout like Step 2., but the space previously occupied by a full-width (/height) pane remains occupied.

### Logs

```
Debug Overlay
wezterm version: 20240208-085947-f671a9ab x86_64-unknown-linux-gnu
Window Environment: X11 bspwm
Lua Version: Lua 5.4
OpenGL: Mesa Intel(R) HD Graphics 520 (SKL GT2) 4.6 (Compatibility Profile) Mesa 23.0.4-0ubuntu1~22.04.1
Enter lua statements or expressions and hit Enter.
Press ESC or CTRL-D to exit
14:20:34.542 ERROR wezterm_gui::spawn > Failed to spawn: split_pane: No space for split!
14:20:35.819 ERROR mux::tab > No space for split!!! SplitDirectionAndSize {
    direction: Vertical,
    first: TerminalSize {
        rows: 11,
        cols: 210,
[...]

PR #5969, opened 2024-08-15
Fix ghost space left by SplitPane with top_level=true

### LINKED ISSUE #2579: Split created with top_level=true cannot be split any further if started from "non top-level split"

Linux Wayland

Gnome 42

### WezTerm version

20221001-071223-478109de

### Describe the bug

When I have a window that is already split horizontally, and I create a top-level vertical split from it, I cannot split the new tab horizontally, the log shows "No splace for split!!!"

### To Reproduce

Note, I have shortcuts in my config to reproduce this behaviour, but I'm using cli commands here, so no dedicated shortcuts are necessary

1. Start wezterm
2. Create a new "bottom" split: `wezterm cli split-pane --bottom`
3. Create a new top-level "right" split: `wezterm cli split-pane --right --top-level`
4. Move to the right pane (which occupies the full height)
5. Try to create a new bottom split:  `wezterm cli split-pane --bottom`
6. The command fails with "No splace for split!!!" (while there is enough room, as evidenced by the "left part" being split in two)

### Configuration

no config

### Expected Behavior

A new bottom split is opened

### Logs

16:35:45.711 ERROR mux::tab > No splace for split!!! SplitDirectionAndSize {
    direction: Vertical,
    first: TerminalSize {
        rows: 29,
        cols: 127,
        pixel_width: 1270,
        pixel_height: 638,
        dpi: 96,
    },
    second: TerminalSize {
        rows: 30,
        cols: 127,
        pixel_width: 1270,
        pixel_height: 660,
        dpi: 96,
    },
} height=60 width=127 top_of_second=30 left_of_second=0 tab_size=TerminalSize { rows: 60, cols: 126, pixel_width: 1260, pixel_height: 1320, dpi: 96 }

### Anything else?

_No response_

boazy:
Fixes  #2579, #4984

When there is more than one pane in the tab, the tab.resize() is
called and the tab resize logic is reused to correctly resize all the
panes in the tab before adding the new pane split, so the new split
pane will not overwrite any existing pane.

The issue seems to be that after this resize, the tab "remembers" its
new size, and when the new split pane is closed, the tab is left with
ghost space which cannot be reused.

The fix is quite simple: we just need to immediately resize the tab
back to its original proportions. Since the underlying panes have
already been resized by tab.resize(), we don't need to keep the
restricted tab size.

I may have missed something, but this seems to work correctly.

### DIFF
--- a/mux/src/tab.rs
+++ b/mux/src/tab.rs
@@ -2004,6 +2004,9 @@ impl TabInner {
                 } else {
                     self.resize(split_info.second.clone());
                 }
+                // Resize the tab back to the original tab size
+                // to avoid leaving unusable space within the tab
+                self.resize(tab_size)
             }
 
             let mut cursor = self.pane.take().unwrap().cursor();


mux/src/tab.rs
```diff
@@ -2004,6 +2004,9 @@ impl TabInner {
                 } else {
                     self.resize(split_info.second.clone());
                 }
+                // Resize the tab back to the original tab size
+                // to avoid leaving unusable space within the tab
+                self.resize(tab_size)
```
wez, line 2009: Thanks for diving in. I'm not sure about this though; if this results in the right outcome, should the above logic to resize be retained?
Can you provide more details on the precise problem scenario, and explain what you meant by the size being remembered?

boazy: I didn't write this logic, so I wasn't sure about why it exists in the first place. So of course I tried to remove it. But when I removed it completely, something else broke: The new pane text just wrote over the old pane's text without removing it. For instance, If I had a situation like this:

Before Split:
```
┌─────────────┬──────┐            
│> ls -1      │ P    │
│file.txt     │ a    │
│file2.txt    │ n    │
│file3.txt    │ e    │
│             │      │
│> echo Hi    │      │
│Hi           │      │
│             │      │
│> vim file.t │      │
└─────────────┴──────┘            
```

After Split (with `direction=Down, top_level=true`):
```
┌──────────────┬─────┐
│> ls -1       | P   │
│file.txt      | a   │
│file2.txt     | n   │
│file3.txt     | e   │
├──────────────┴─────┤
│mysh 1.0 Hi         │
│>                   │
│                    │
│> vim file.t        │
└────────────────────┘
```

It was a lot uglier in practice, but that's the basic gist. I'm not 100% what is the underlying issue, but I guess the existing panes' visible area doesn't get redrawn as necessary without calling resize if there is more than one pane.

So the `resize()` call does solve an actual issue. The main problem is that after the new pane you added is closed, the pane(s) that you resized won't refill the area of the closed pane, because the tab size is now smaller than full window size. This is exactly what issue #4984 refers to. Adding this extra `resize()` call resizes the tab to the full window size again, but the original (pre-split) panes' contents remain constrained to their new size.

boazy: An example of the issue that happens with the current code (without my changes) is provided in #4984 (reproduction of issue).

Due to this bug, `top_level` splits are practically useless. They are not needed to begin with if you just have 1 pane before the split, but if you have more than 1 pane they always leave unusable space behind. I don't think this fix can possibly make anything worse.

nacl-gb3:
Hey, I tried applying this change on the current wezterm codebase and it didn't seem to solve the issue for me as of now. 
Here are screenshots of where my session ended up after applying the change and opening a few panes:

Screenshot_2026-01-18_15-32-56

Screenshot_2026-01-18_15-34-04

(i don't recall if this last screenshot was with or without this change) 
Screenshot_2026-01-18_15-35-38

I also have two screenshots prepared of how wezterm works with top_level pane splits currently (ie without the change from this PR) that I will also shortly put as a comment on issue #4984 

Screenshot_2026-01-18_15-38-22
(this one has annotations of what exactly I did in hopes it may help with finding a solution)

Screenshot_2026-01-18_15-39-10

For reference, my system is currently running the COSMIC Wayland Desktop version 1.00 on Fedora 43 and I am currently using the following wezterm version: wezterm 20260117_154428_05343b38

I will also look the codebase myself to see what the issue could be when I have the time

nacl-gb3:
I also forked the project just now in case you wanted to test this PR on a more current codebase real quick
https://github.com/nacl-gb3/wezterm

PR #5971, opened 2024-08-15
fix(wayland): match inner window and window frame sizes

aliaksandr-trush:
If we subtract_borders for frame size AND add_borders for inner window, we are receiving inner window bigger than frame:
image

### DIFF
--- a/window/src/os/wayland/window.rs
+++ b/window/src/os/wayland/window.rs
@@ -250,14 +250,10 @@ impl WaylandWindow {
         }
 
         window.set_min_size(Some((32, 32)));
-        let (w, h) = window_frame.add_borders(
-            dimensions.pixel_width as u32,
-            dimensions.pixel_height as u32,
-        );
         let (x, y) = window_frame.location();
         window
             .xdg_surface()
-            .set_window_geometry(x, y, w as i32, h as i32);
+            .set_window_geometry(x, y, dimensions.pixel_width as i32, dimensions.pixel_height as i32);
         window.commit();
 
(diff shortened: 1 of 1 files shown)


aliaksandr-trush:
After some testing, fount out that frame size and window size should be the same, in this case it will match and draw without any issues.

loops:
Hi Aliaksandr,

Was going to try this out, but it has a compile error since your forced push.

aliaksandr-trush:
> Hi Aliaksandr,
> 
> Was going to try this out, but it has a compile error since your forced push.

Thank you for the information.
It should work now.

wez:
Thank you!!

aliaksandr-trush:
> > foun[d] out that frame size and window size should be the same
> 
> Do you have a source for that?
> 

The solution was discovered during debugging, and it fixed the visual issue for me.

> > The size must be without the borders, as in [Self::subtract_borders] were used on it.
> 
> But maybe that is a bug in SCTK?
> 
> Or maybe wezterm was using the outer width instead of the inner width when drawing?
> 
> Either way, I think this introduces a problem where we don't account for the size of the borders when we receive a configure event. That is, we receive a configure event, then set the _inner_ size of the window to the width and height we receive, without subtracting the size of the borders.

It is something wrong with current implementation for window and frame sizes calculation in Wezterm.
While it works fine in windowed mode right now, it still has issue in Full-screen with borders. I still have this issue in my personal TODO, but if your PR will refactor this implementation, probably I need to wait.

Issue #4238, opened 2023-09-02
Add a way to check current config value with wezterm CLI


Destroy666x:
**Is your feature request related to a problem? Please describe.**
I just lost like 10 minutes trying to set `cursor_thickness` option which didn't work with the block one apparently (not documented) and I was thinking it's wrong config location. Then I lost another few minutes because the 2nd setting I tried - `windows_close_coinfirmation` - apparently doesn't work either, at least on Windows (also not documented, if not a bug).

So a way to check config values would be nice to avoid such situations and not waste time on trying the config in 3 other locations.

**Describe the solution you'd like**
A command like `wezterm --config cursor_thickness` or `wezterm config cursor_thickness` could tell you the current version of that option.

**Describe alternatives you've considered**
-

**Additional context**
Windows 10

hgkamath:
This would be useful . 
Like ```wezterm --config color_scheme``` could tell you what the current color scheme is without trying to go looking for it in the config file, especially, give that things can be set programmatically. 
ref  
- https://www.reddit.com/r/wezterm/comments/1bbq6ro/i_implemented_a_theme_switcher/

smujaddid:
I would love to have this feature. It would be a big help to debug and trying new configuration and everything else we can do with this feature. Any follow up on this request?

Apreche:
New to wezterm. I'm also immediately frustrated by this issue. The documentation doesn't list default values for all configuration options. How can I know if I want to change the value if I don't know the current value? There should at least be a way to print out the entire current configuration.

mattmc3:
The feature exists, it's just frustratingly not obvious for a new user to find. After combing through the docs, I figured it out.

From your WezTerm [CTRL-SHIFT-L](https://wezfurlong.org/wezterm/config/lua/keyassignment/ShowDebugOverlay.html) gives you a "Debug Overlay". From this Lua debugger, you can enter all sorts of Lua commands to interact with WezTerm, and the one you want is `window:effective_config()` which will print out all the in force configuration options and their current values.

You can also store the config in a variable and test values:

```lua
> fullconfig = window:effective_config()
> fullconfig["window_close_confirmation"]
"AlwaysPrompt"
```

Or print the whole thing:

```
> fullconfig
{
    "allow_download_protocols": true,
    "allow_square_glyphs_to_overflow_width": "WhenFollowedBySpace",
    "allow_win32_input_mode": true,
    "alternate_buffer_wheel_scroll_speed": 3,
[...]

PR #5972, opened 2024-08-16
Add command to check config values

rstefanic:
First off, thanks for wezterm. I just started using it this week and it's fantastic!

While getting setup with wezterm, I tried to see if there's a way to check an existing config value and I came across issue #4238. I wasn't sure if you wanted to add a feature like this, but I took a crack at it.

This PR adds a new `ls-config` subcommand to list configuration values. If no options are passed in, it will list all the configuration values. The `--setting` flag can be passed with the name of a config to show its current value.

```console
wezterm ls-config --setting font_size
# "font_size" = 16.0
```

Implementation wise, I figured that, since the config came in as a Lua table, I should be able to iterate over it like a Lua table. I came across the `ToDynamic` trait in the project and used it to convert the `Config` into a dynamic `Value` in order to iterate through it. While converting the `Config` into a dynamic is probably not the most performant thing, the advantage to this approach is that any future changes made to `Config` will automatically be reflected in this feature. Since this `Config` conversion only occurs once and then the program exits, I figured it would be alright.

An alternative approach I came up with was to create a lookup table with all the possible config values mapped to the `Config` struct but I figured that having to maintain a separate lookup table like that would be a burden.

As a side note, the issue referenced above suggested using something like `wezterm --config` but I didn’t want to conflict with the existing flag to override a specific configuration value. I named the subcommand `ls-config` since it seemed similar to `ls-fonts`. In the same vein, I added the implementation in `wezterm-gui` simply because that's where `ls-fonts` was implemented. Let me know if they belong somewhere else or if the subcommand needs to be renamed.

If this is not a feature that you want to add, then feel free to close this this PR. Thanks again for wezterm. I'm loving it!

### DIFF
--- a/docs/cli/ls-config.md
+++ b/docs/cli/ls-config.md
@@ -1,0 +1,6 @@
+# `wezterm ls-config`
+
+```console
+{% include "../examples/cmd-synopsis-wezterm-ls-config--help.txt" %}
+```
+
--- a/docs/examples/cmd-synopsis-wezterm-ls-config--help.txt
+++ b/docs/examples/cmd-synopsis-wezterm-ls-config--help.txt
@@ -1,0 +1,8 @@
+Show current configuration values
+
+Usage: wezterm-gui ls-config [OPTIONS]
+
+Options:
+      --setting <SETTING>  Show the value of one specific configuration
+                           setting
+  -h, --help               Print help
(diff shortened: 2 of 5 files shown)


mattmc3:
See also - there's a method via the debugger described here: https://github.com/wez/wezterm/issues/4238#issuecomment-2295092917

PR #5978, opened 2024-08-16
Add links to Ubuntu24 nightly .deb images

nburlett:
The nightly builds include Ubuntu24.04 .deb files, but it's not linked on the downloads page. This is an attempt to add links to the downloads. I'm not sure how the link url variables get filled in though, so maybe something more is needed?


wez:
Thanks; the variables are substituted by `ci/subst-release-info.py` based on available packages in the release info.
The arm packages weren't actually built out by CI, so I've pushed a commit that should do that now too.

PR #5979, opened 2024-08-16
Nix "nightly" build

thomascft:
## Summary

I've created an action that does a scheduled build of WezTerm with the Nix flake, and pushes the outputs to a binary cache.

## The Problem

I find myself having to build WezTerm from source on occasion when Hyprland releases updates. The version of WezTerm currently in nixpkgs doesn't include various Hyprland/Wayland bugfixes because there hasn't been a release in a while.

## My Solution

We can push to a Cachix with a scheduled build of WezTerm using the flake. People can add this binary cache to their configurations to avoid building WezTerm from source if they're using the flake. There's a couple more things that need to be done before this action will work though.

- Setup [https://wezterm.cachix.org](https://wezterm.cachix.org) (This is a super quick process, and the login is just read email access for your GitHub account.)
- Generate an auth token for that cache and add it to WezTerm's repo secrets so it can be used in the GitHub action 
 
I've got a working version here https://github.com/thomascft/wezterm-nightly-flake/blob/main/flake.nix. The actions are a touch more complicated because I wanted nightly lockfile updates as well. 

Here are some examples of Cachix being used for other projects

- [Neovim](https://github.com/nix-community/neovim-nightly-overlay)
- [Firefox](https://github.com/nix-community/flake-firefox-nightly)
- [Hyprland](https://wiki.hyprland.org/Nix/Cachix/)

I've also added some basic install documentation for Nix/NixOS which can be improved on if you'd like.

### DIFF
--- a/docs/install/linux.md
+++ b/docs/install/linux.md
@@ -286,6 +286,91 @@ hide:
     $ brew rm wezterm
     $ brew install --HEAD wezterm
     ```
+=== "Nix/NixOS"
+
+    ## Nix
+    
+    WezTerm is available in nixpkgs as `wezterm`.
+
+    ```nix
+    {
+        # configuration.nix
+
+        environment.systemPackages = [
+            pkgs.wezterm
+        ]
+    }
+    ```
+
+    ### Flake
+    
+    If you need a newer version use the flake. Use the cachix if you want to avoid building WezTerm from source.
+
+    The flake is in the `nix` directory, so the url will be something like `github:wez/wezterm?dir=nix`
+
+    Here's an example for NixOS configurations:
+    
+    ```nix
+    {
+        inputs.wezterm.url = "github:wez/wezterm?dir=nix";
+        # ...
+
+        outputs = inputs @ {nixpkgs, ...}:{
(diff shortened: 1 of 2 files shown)


docs/install/linux.md
```diff
@@ -286,6 +286,91 @@ hide:
...
+    ### Cachix
+
+    Successful builds of the nightly nix action are pushed to this binary cache.
+
+    ```nix
+    # nixosConfiguration module
+    {
+        nix.settings = {
+            substituters = ["https://wezterm.cachix.org"];
+            trusted-public-keys = ["wezterm.cachix.org-<UPDATE_WITH_PUBLIC_KEY>"];
```
wez, line 369:
```suggestion
            trusted-public-keys = ["wezterm.cachix.org-1:kAbhjYUC9qvblTE+s7S+kl5XM1zVa4skO+E/1IDWdH0="];
```

wez (commented):
Thanks for this; I've set up a cache and added the auth token to the repo secrets.
I've added a suggestion for the public key portion.
Is there anything else we need for this before I merge?

thomascft:
Everything looks great. Thanks!

wez:
Thank you!

PR #5981, opened 2024-08-17
Remove mux-server race condition.

loops:
New changes were being lost, and not sent to the
client, if they occurred during the time previous
updates were being processed.


wez:
Thank you!

Issue #913, opened 2021-06-30
Accessibility for the visually impaired


wez:
My intent with this issue is to understand better what first-class support for the visually impaired looks like in a terminal emulator, and then figure out how to implement it.

My reading so far has found:

* [TDSR - A console-based screen reader](https://github.com/tspivey/tdsr) wherein I posted https://github.com/tspivey/tdsr/issues/19
* [emacspeak](http://emacspeak.sourceforge.net/) which has some interesting concepts
* [Slint](https://slint.fr/) an accessible linux distro

A discussion with Didier of the slint project gave me some leads:

* mlterm has support for brltty: https://github.com/arakiken/mlterm/blob/master/doc/en/README.brltty
* The `#a11y` IRC channel on `irc.linux-a11y.org` is a good place to ask questions

# Required Features

* For low-vision users, it may be sufficient to have a quick and easy way to activate a color scheme with high contrast and large text
* Some kind of screen reader and/or braille display support

## Screen Reader

Based on my reading of the TDSR and emacspeak documentation, it seems like a key feature of a good terminal screen reading experience is being able to manage what is being read. I think of the terminal as having a few distinct interface elements:

* The scrollback, which is the entirety of the text that is navigable/viewable
* The viewport, which is the section of the scrollback that is visible in the window.  This defines a relatively coarse visual "cursor" representing what the user is looking at.
* The mouse cursor, which is used for selecting and interacting with text
* The text cursor, which is used primarily by the application

WezTerm has [CopyMode](https://wezfurlong.org/wezterm/copymode.html) and [Quick Select Mode](https://wezfurlong.org/wezterm/quickselect.html) for mouse-less selection capabilities.

For a screen reader, I think it may make sense to introduce an explicit *read cursor* that is conceptually similar to the viewport but much finer grained; based on the TDSR docs, it seems desirable to be able to specify the position based on character, word or line. To me, controlling this feels similar to the navigation functions available in mouseless copy mode, except that the data is read aloud instead of being copied.

In a situation where the rate of output is high, it seems like it might be useful to have the *read cursor* move independently from the viewport so that it can lag behind the output. Having a notification to inform the user that the output rate is high seems like it might be valuable; eg: "NOTE: one megabyte of text has been output in the past ten seconds", or "The next line is no longer in the scrollback" if it got scrolled away.  Notifications will need to trigger some audio cue (perhaps an adjustment to the speed/tone parameters) to disambiguate the notification from what was being read.

[...]

PR #5995, opened 2024-08-19
Attempt to implement AccessKit inside the window crate

DataTriny:
As the PR title suggests, this is neither fully working nor complete. I think it can be a starting point nonetheless.

Relates to #913 

## Current status

The `async` example is working correctly on Windows. I tested it on Windows 11 with NVDA and Narrator (two free assistive technologies).

## Issues encountered

- The `WindowEventSender` cannot be shared so when accessibility-related events come from the system, I had to place a message into the window loop so the event could be dispatched to the app later. Although not elegant, this approach can work on Windows, and probably on X11 as well, but I think there would be no way to replicate it under Wayland. I haven't really checked on macOS.
- There is no way for the app to respond to a `WindowEvent`: when accessibility is first requested, I had to instruct AccessKit to generate a placeholder accessibility tree, until I can push the real one later. Another option would be to have a dedicated callback on `Window` to build the initial accessibility tree, but then the entire app state must be shared.
- `WindowOp::update_accesskit_if_active` cannot take the app state in its closure because its signature impose new restrictions that clashes with, for instance, the `glium::Context` in the `async` example. This also mean that the `WindowOp` trait is not object safe anymore, which will be an issue with the real wezterm app.

Overall, adding accessibility support to the windowing crate of wezterm would first require some architectural changes. I hope that this proof of concept can help shape the future of a nice, accessible, terminal emulator.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,12 @@
 version = 3
 
 [[package]]
+name = "accesskit"
+version = "0.16.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e4700bdc115b306d6c43381c344dc307f03b7f0460c304e4892c309930322bd7"
+
+[[package]]
 name = "addr2line"
 version = "0.22.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -6721,6 +6727,7 @@ checksum = "712e227841d057c1ee1cd2fb22fa7e5a5461ae8e48fa2ca79ec42cfc1931183f"
 name = "window"
 version = "0.1.0"
 dependencies = [
+ "accesskit",
  "anyhow",
  "async-channel",
  "async-io",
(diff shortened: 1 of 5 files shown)


wez:
I don't have any context on accesskit and no bandwidth to look right at this moment.
I will note that we currently push things into the window that accessibility adjacent functions like the IME need: 

https://github.com/wez/wezterm/blob/1c4fdd3f7131785d0afaed08bed6cf1856f84328/window/src/lib.rs#L311-L314
https://github.com/wez/wezterm/blob/1c4fdd3f7131785d0afaed08bed6cf1856f84328/wezterm-gui/src/termwindow/mod.rs#L2118

Could we do something similar for accesskit?

alex19EP:
hello @DataTriny and @wez I am very interested in the current status of this pr and wezterm's work on accessibility for the blind in general.
my interest is mainly in the linux platform as I am currently looking for a terminal that is accessible to the blind and has more functionality than the vte-based terminals.

DataTriny:
It does not seem that any of the points I mentioned have been addressed since this PR was opened. I currently don't plan to work on this until the windowing layer is re-architectured to accommodate accessibility events.

Issue #1342, opened 2021-11-25
Better scrollback behavior


junkblocker:
**Is your feature request related to a problem? Please describe.**

I am surprised to find that wezterm holds the whole scrollback in memory and doesn't attempt to swap out to disk or even just use a circular buffer or something even in memory. This causes wezterm to run out of memory and get killed. 

**Describe the solution you'd like**

Swap lines beyond the scrollback_lines out to disk still keeping them available via `pane:get_lines_as_text(...)` .

**Describe alternatives you've considered**

Reduce the scrollback_lines size which is inadequate since that then requires changing your tools/habits.

**Additional context**

.

cowens:
The issues identified in https://github.com/microsoft/terminal/issues/1410#issuecomment-504942872 seem to be:

1. implementation limitations on the size of the buffer and the integers pointing the to the current place in the buffer
1. security of the file on disk

Looking at the code, it appears as if scrollback_size is already usize which has a max of 2<sup>64</sup> − 1 which is much better than Windows Terminal's use of 16 bit integers, so item 1 isn't much of an issue.

Item 2 is more of an issue, but if the file is owned by the user and the file system itself is encrypted (which it should be these days), then this is only an issue if an attacker has access to a logged in account with permission to the file. That is a level of risk I am willing to take on a single user machine (if an attacker has permission to the file, they can probably get the key from the wezterm and decrypt it anyway).

junkblocker:
Just want to say that between this and the OOM kill related bug #3907 I keep needing to (remember to) switch to Konsole for memory heavy operations and that mars the experience/habit.

( Good to run into you on the interwebs, @cowens ! )

cowens:
Looking a bit deeper, it looks like the practical limit to the scrollback size is actually constrained by isize not usize, so the maximum appears to be 2<sup>63</sup>-1 not 2<sup>64</sup>-1:
```
    pub fn stable_row_to_phys(&self, stable: StableRowIndex) -> Option<PhysRowIndex> {
        let idx = stable - self.stable_row_index_offset as isize;
        if idx < 0 || idx >= self.lines.len() as isize {
            // Index is no longer valid
            None
        } else {
            Some(idx as PhysRowIndex)
        }
    }
```

Sadly, I am not finding anything that is a drop-in replacement for std::collections::VecDeque that is backed by a file. If I have enough time, I am going to play around with implementing all of the methods term/src/screen.rs is using for the scrollback with a hybrid in-memory/on-disk data structure.

[...]

PR #5996, opened 2024-08-19
Validate scrollback_lines to avoid crashes

x3ro:
I started using wezterm today, and I immediately tried to configure it to use "infinite scrollback", as I use in iTerm2. From the configuration I couldn't tell if there was a way to do this, so I just set it to a really large number, hoping that would work. Interestingly this works for very large numbers when the config is just being reloaded while the terminal is running, but if you then try to restart the application it crashes (tried to allocate like 100PB or something).

I then came across #1342 and thought "that's seems a bit too involved", and decided that I probably don't need infinite scrollback, but just kind of a large number. Through fair dice roll I determined `one billion - 1` will probably suffice.

Now, this might not be the best solution, so I'm happy to get some feedback. I was also thinking that it would be nice if one could just set it to `0`, and then the applicatio determines a suitably large number for the amount of RAM available, but a) I wasn't sure how this would be best implemented in the confines of the current architecture, and b) I wasn't sure if it would be well received.

Long story short, happy to hear your feedback.

PS: I'm happy to add documentation, etc. as needed, just wanted to get some feedback first before I invest that time 🙂

### DIFF
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -185,7 +185,10 @@ pub struct Config {
     pub color_schemes: HashMap<String, Palette>,
 
     /// How many lines of scrollback you want to retain
-    #[dynamic(default = "default_scrollback_lines")]
+    #[dynamic(
+        default = "default_scrollback_lines",
+        validate = "validate_scrollback_lines"
+    )]
     pub scrollback_lines: usize,
 
     /// If no `prog` is specified on the command line, use this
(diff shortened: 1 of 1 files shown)


config/src/config.rs
```diff
@@ -1638,6 +1641,17 @@ fn default_scrollback_lines() -> usize {
...
 
+const MAX_SCROLLBACK_LINES: usize = 999_999_999;
+fn validate_scrollback_lines(value: &usize) -> Result<(), String> {
+    if *value > MAX_SCROLLBACK_LINES {
+        Err(format!(
+            "Illegal value {value} for scrollback_size; it must be <= {MAX_SCROLLBACK_LINES}!"
+        ))
+    } else {
+        Ok(())
+    }
```
wez, line 1652:
```suggestion
    if *value > MAX_SCROLLBACK_LINES {
        return Err(format!(
            "Illegal value {value} for scrollback_size; it must be <= {MAX_SCROLLBACK_LINES}!"
        ));
    }
    Ok(())
```

wez:
Thanks!

Issue #4471, opened 2023-10-21
Additional vi-like keybindings in copy-mode


agostonbarna:
WezTerm already has the basic vi-like key bindings in copy-mode, but there a couple of useful keybindings in vim that would be awesome to have in WezTerm too:

1. jump / select to matching bracket (vim docs: `:h various-motions`):
  - `%`: when the cursor is over a bracket (`{`, `[`, `(`) and we press `%` it jumps to the matching closing/opening bracket (`)`, `]`, `}`) taking nested brackets into consideration (this is also implemented in tmux and I use it quite often to extract a json from log files when scrolling trough them, without having to manually find the matching brace, which can be especially painful to manually do so if it's a jsonl / compress json)
  - `v%`: similarly, select the text to the matching bracket by typing `v%`

2. word motions / selection (vim docs: `:h word-motions`):
  - `vE` / `ve`: WezTerm already have the `ve` binding, but it actually works like `vE` in vim
    - `ve`: select the next `word` forward
    - `vE`: select the next `WORD` forward
  - `vB` / `vb`: WezTerm already have the `vb` binding, and it works the same as in vim (unlike `ve`)
    - `vb`: select the previous `word` backward
    - `vB`: select the previous `WORD` backward

3. object selection (vim docs: `:h object-select`):
  - `viw`: select inside `word`
  - `vaw`: select around `word`
  - `viW`: select inside `WORD`
  - `vaW`: select around `WORD`

`word` / `WORD` definitions referenced above:
  - `word`: consists of letters/numbers/underscores, or non-blank chars separated by whitespace (vim docs: `:h word`)
  - `WORD`: non-blank characters, separated by whitespace (vim docs: `:h WORD`)

1. and 2. are already available in tmux (with this config: `bind -T copy-mode-vi v send -X begin-selection`), 3. I always missed from tmux too.

jackielii:
Can I add <C-e> and <C-y> to scroll up and down without moving the cursor position?

grantwforsythe:
May I also suggest the ability to yank inside a pair of characters, e.g. `yi"`.

PR #5997, opened 2024-08-19
Enhanced vi-like word motions in copy mode

eduardorittner:
Item 2 from #4471, which are more complete vi motions. Before the PR, copy-mode already had:
- w for moving to the next word
- e for moving to the end of  WORD (not word)
- b for moving to the start of word

And now it has
- W for moving to the next WORD
- e for moving to the end of word (fixed)
- E for moving to the end of WORD
- B for moving to the start of  WORD

In the code, word and WORD are called word and non_blank_word, respectively, since vim's definition of WORD is a sequence of non-blank characters separated by whitespace.
These bindings were manually tested and seem to work as intended, however I thought of adding tests to validate the behavior but found no examples of how to do so, any tips would be greatly appreciated.


loops:
Tested here, and seems to work well.  Is a nice addition . +1

wezterm-gui/src/overlay/copy.rs
```diff
@@ -877,6 +878,61 @@ impl CopyRenderable {
...
+
+                self.cursor.x = self.cursor.x.saturating_sub(width);
+                passed_word = true;
+            }
+
+            if last_was_whitespace && self.cursor.y > 0 {
+                // The line begins with whitespace
+                self.cursor.x = usize::max_value();
+                self.cursor.y -= 1;
+                return self.move_backward_one_word();
```
mgpinf: I think this should be `self.move_backward_one_non_blank_word()`
```suggestion
                return self.move_backward_one_non_blank_word();
```

mgpinf: I think you also need to update call to `self.move_to_end_of_non_blank_word` in the definition of `move_to_end_of_non_blank_word` similar to the above mentioned change

mgpinf:
I think you need to update the call from `move_to_end_of_word` to `move_to_end_of_non_blank_word` in the definition of `move_to_end_of_non_blank_word`

mgpinf:
IMO, there is a lot of common logic in the `word` and `non_blank_word` versions of the word movement functions. It's probably better to add `non_blank_word` as a boolean parameter and add  conditional logic inside the function. This helps with lesser diff. There is a similar PR where this point was mentioned https://github.com/wezterm/wezterm/pull/6445/files#r1947946777.

mgpinf:
Can you add documentation for `MoveBackwardNonBlankWord`, `MoveForwardNonBlankWord`, and `MoveForwardNonBlankWordEnd`?

eduardorittner:
> IMO, there is a lot of common logic in the `word` and `non_blank_word` versions of the word movement functions. It's probably better to add `non_blank_word` as a boolean parameter and add conditional logic inside the function. This helps with lesser diff. There is a similar PR where this point was mentioned https://github.com/wezterm/wezterm/pull/6445/files#r1947946777.

Yeah that makes sense! Will do.

eduardorittner:
Just made the changes you requested, but haven't been able to test them manually since wezterm's crashing on me (something to do with asahi, maybe I'll open an issue on that), so in the meantime if anybody could check that it stills works as intended it would be great!

docs/config/lua/keyassignment/CopyMode/MoveForwardNonBlankWord.md
```diff
@@ -0,0 +1,19 @@
+# CopyMode `MoveForwardNonBlankWord`
+
+{{since('20220624-141144-bd1b7c5d')}}
```
mgpinf: Can you update this line to `{{since('nightly')}}`?
```suggestion
{{since('nightly')}}
```

docs/config/lua/keyassignment/CopyMode/MoveForwardNonBlankWordEnd.md
```diff
@@ -0,0 +1,23 @@
+# CopyMode `MoveForwardNonBlankWord`
+
+{{since('20230320-124340-559cb7b0')}}
```
mgpinf: Here as well
```suggestion
{{since('nightly')}}
```

docs/config/lua/keyassignment/CopyMode/MoveBackwardNonBlankWord.md
```diff
@@ -0,0 +1,18 @@
+# CopyMode `MoveBackwardNonBlankWord`
+
+{{since('20220624-141144-bd1b7c5d')}}
```
mgpinf: Can you update this line to `{{since('nightly')}}`?
```suggestion
{{since('nightly')}}
```

mgpinf:
Can you add the keybindings in `docs/examples/default-copy-mode-key-table.markdown` similar to what you added in the file `wezterm-gui/src/overlay/copy.rs`?

docs/config/lua/keyassignment/CopyMode/MoveForwardNonBlankWord.md
```diff
@@ -0,0 +1,19 @@
...
+local wezterm = require 'wezterm'
+local act = wezterm.action
+
+return {
+  key_tables = {
+    copy_mode = {
+      { key = 'W', mods = 'NONE', action = act.CopyMode 'MoveForwardNonBlankWord' },
+    },
+  },
+}
```
mgpinf: Can you run the command `./ci/build-docs.sh`? It should format the Lua code in the documentation.

joeledwardson:
is it possible to get this merged? happy to make any changes if need be to help

PR #6007, opened 2024-08-21
feat: add a way to spawn populated LineEditor

### LINKED ISSUE #6002: Add option to pre-populate user prompts with custom text
I would love an option to specify "default" text for `wezterm.action.PromptInputLine`, for example when using it to rename the active tab. For this case, that could be the tab name or a shared prefix for manually named tabs (e.g. `M: `)

Add another parameter `default` to function `wezterm.action.PromptInputLine` that would define a text string to place in prompt for user to edit, rather than starting with an empty input.

Hooking up a SendText callback with the desired text, but it is not immediately obvious how to do it after spawning the prompt.

**Additional context**
A callback parameter to conditionally style the text while it is being entered would be good, too, but it is a lot of overhead for little use-case.

ekorchmar:
Closes #6002

Adds a new constructor for `LineEditor`, that allows to specify text to pre-populate the input.
Also removes a call to line.clear() in a method to initiate editing, as it is currently only ever called after editor struct had been newly spawned.

### DIFF
--- a/config/src/keyassignment.rs
+++ b/config/src/keyassignment.rs
@@ -459,7 +459,7 @@ pub struct PromptInputLine {
     pub action: Box<KeyAssignment>,
     /// Optional label to pre-fill the input line with
     #[dynamic(default)]
-    pub with_content: String,
+    pub initial_value: Option<String>,
     /// Descriptive text to show ahead of prompt
     #[dynamic(default)]
     pub description: String,
--- a/config/src/keyassignment.rs
+++ b/config/src/keyassignment.rs
@@ -457,6 +457,9 @@ pub struct QuickSelectArguments {
 #[derive(Debug, Clone, PartialEq, FromDynamic, ToDynamic)]
 pub struct PromptInputLine {
     pub action: Box<KeyAssignment>,
+    /// Optional label to pre-fill the input line with
+    #[dynamic(default)]
+    pub with_content: String,
     /// Descriptive text to show ahead of prompt
     #[dynamic(default)]
     pub description: String,
--- a/docs/config/lua/keyassignment/PromptInputLine.md
+++ b/docs/config/lua/keyassignment/PromptInputLine.md
@@ -20,7 +20,7 @@ upon the input.
   anything, or CTRL-C to cancel the input.
 * `prompt` - the text to show as the prompt. You may embed escape sequences
   and/or use [wezterm.format](../wezterm/format.md).  Defaults to: `"> "`. {{since('nightly', inline=True)}}
-* `with_content` - optional.  If provided, the initial content of the input
+* `initial_value` - optional.  If provided, the initial content of the input
   field will be set to this value.  The user may edit it prior to submitting
   the input.
 
@@ -37,7 +37,7 @@ config.keys = {
     mods = 'CTRL|SHIFT',
     action = act.PromptInputLine {
       description = 'Enter new name for tab',
-      with_content = 'Tab: ',
+      initial_value = 'My Tab Name',
       action = wezterm.action_callback(function(window, pane, line)
         -- line will be `nil` if they hit escape without entering anything
         -- An empty string if they just hit enter
--- a/docs/config/lua/keyassignment/PromptInputLine.md
+++ b/docs/config/lua/keyassignment/PromptInputLine.md
@@ -8,7 +8,7 @@ from the user.
 When the user enters the line, emits an event that allows you to act
 upon the input.
 
-`PromptInputLine` accepts three fields:
+`PromptInputLine` accepts four fields:
 
 * `description` - the text to show at the top of the display area. You may
   embed escape sequences and/or use [wezterm.format](../wezterm/format.md).
@@ -20,6 +20,9 @@ upon the input.
   anything, or CTRL-C to cancel the input.
 * `prompt` - the text to show as the prompt. You may embed escape sequences
   and/or use [wezterm.format](../wezterm/format.md).  Defaults to: `"> "`. {{since('nightly', inline=True)}}
+* `with_content` - optional.  If provided, the initial content of the input
+  field will be set to this value.  The user may edit it prior to submitting
+  the input.
 
 ## Example of interactively renaming the current tab
 
@@ -34,6 +37,7 @@ config.keys = {
     mods = 'CTRL|SHIFT',
     action = act.PromptInputLine {
       description = 'Enter new name for tab',
+      with_content = 'Tab: ',
       action = wezterm.action_callback(function(window, pane, line)
         -- line will be `nil` if they hit escape without entering anything
         -- An empty string if they just hit enter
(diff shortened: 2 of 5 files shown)


config/src/keyassignment.rs
```diff
@@ -457,6 +457,9 @@ pub struct QuickSelectArguments {
 #[derive(Debug, Clone, PartialEq, FromDynamic, ToDynamic)]
 pub struct PromptInputLine {
     pub action: Box<KeyAssignment>,
+    /// Optional label to pre-fill the input line with
+    #[dynamic(default)]
+    pub with_content: String,
```
wez, line 462: I'd prefer a name like this, and I think we should make this optional to work better with another suggestion I'm going to make

```suggestion
    pub initial_value: Option<String>,
```

docs/config/lua/keyassignment/PromptInputLine.md
```diff
@@ -32,6 +35,7 @@ config.keys = {
     mods = 'CTRL|SHIFT',
     action = act.PromptInputLine {
       description = 'Enter new name for tab',
+      with_content = 'Tab: ',
```
wez, line 38: I think that the colon in the value make this look a bit like a prompt rather than an initial value, and since we now support setting a custom prompt, I think it might be clearer as an example to change this.

```suggestion
      initial_value = 'My Tab Name',
```

termwiz/src/lineedit/mod.rs
```diff
@@ -163,6 +163,12 @@ impl<'term> LineEditor<'term> {
         }
     }
 
+    pub fn with_text(terminal: &'term mut dyn Terminal, default: &str) -> Self {
+        let mut editor = Self::new(terminal);
+        editor.line.insert_text(default);
+        return editor;
+    }
+
```
wez, line 171: I prefer to avoid changing the public interface for this, as there is already a method that can be used to update the contents of the editor

```suggestion
```

wezterm-gui/src/overlay/prompt.rs
```diff
@@ -66,7 +66,7 @@ pub fn show_line_prompt_overlay(
     term.render(&[Change::Text(text)])?;
 
     let mut host = PromptHost::new();
-    let mut editor = LineEditor::new(&mut term);
+    let mut editor = LineEditor::with_text(&mut term, &args.with_content);
```
wez, line 69: I think we can achieve this this way instead:

```suggestion
    let mut editor = LineEditor::new(&mut term);
    if let Some(value) = &args.initial_value {
      editor.set_line_and_cursor(value, value.len());
   }
```

wez (changes requested):
Thanks for this! Please rebase this PR as there is a partial conflict with #6054 

I have a couple of suggestions, but this otherwise looks good!

config/src/keyassignment.rs
ekorchmar: Moved all changes to a single commit. Thank you!

ekorchmar:
Rebased onto new master and implemented requested changes.

docs/config/lua/keyassignment/PromptInputLine.md
```diff
@@ -20,6 +20,9 @@ upon the input.
   anything, or CTRL-C to cancel the input.
 * `prompt` - the text to show as the prompt. You may embed escape sequences
   and/or use [wezterm.format](../wezterm/format.md).  Defaults to: `"> "`. {{since('nightly', inline=True)}}
+* `initial_value` - optional.  If provided, the initial content of the input
+  field will be set to this value.  The user may edit it prior to submitting
+  the input.
```
wez, line 25: Let's tag this with the version it was introduced with:

```suggestion
  the input. {{since('nightly', inline=True)}}
```

termwiz/src/lineedit/mod.rs
```diff
@@ -811,7 +811,6 @@ impl<'term> LineEditor<'term> {
     }
 
     fn read_line_impl(&mut self, host: &mut dyn LineEditorHost) -> Result<Option<String>> {
-        self.line.clear();
```
wez, line 814: ah, I didn't spot this initially. I think this is a big semantic change that breaks behavior. You can probably see this in the debug overlay (ctrl-shift-L) if you type in `foo` and hit enter.  Does it cause foo to show up in the next line? 

What I think should be done here is to adjust this method:

```rust
fn read_line_impl(&mut self,
    host: &mut dyn LineEditorHost,
    initial_value: Option<&str>
) -> Result<Option<String>> {
    self.line.clear();
    if let Some(value) = initial_value {
        self.line.set_line_and_cursor(value, value.len());
    }
```

Then, refactor:

```rust
pub fn read_line(&mut self, host: &mut dyn LineEditorHost) -> Result<Option<String>>
```

 into

```rust
pub fn read_line_with_optional_initial_value(&mut self, host: &mut dyn LineEditorHost, initial_value: Option<&str>) -> Result<Option<String>>
```

 that calls the above function:

```rust
...
let res = self.read_line_impl(host, initial_value);
...
```

Then finally add back the original method, but make it call the refactored one:

```rust
pub fn read_line(&mut self, host: &mut dyn LineEditorHost) -> Result<Option<String>> {
   self.read_line_with_optional_initial_value(host, None)
}
```

Note that the public methods on this struct need to have doc comments because they are part of the public API that is published to https://docs.rs/termwiz/latest/termwiz/lineedit/struct.LineEditor.html

wezterm-gui/src/overlay/prompt.rs
```diff
@@ -68,6 +68,9 @@ pub fn show_line_prompt_overlay(
     let mut host = PromptHost::new();
     let mut editor = LineEditor::new(&mut term);
     editor.set_prompt(&args.prompt);
+    if let Some(value) = &args.initial_value {
+        editor.set_line_and_cursor(value, value.len());
+    }
     let line = editor.read_line(&mut host)?;
```
wez, line 74: with the above changes, this would become:

```suggestion
    let line = editor.read_line_with_optional_initial_value(&mut host, args.initial_value.as_deref())?;
```

termwiz/src/lineedit/mod.rs
ekorchmar: > I think this is a big semantic change that breaks behavior. You can probably see this in the debug overlay (ctrl-shift-L) if you type in foo and hit enter. Does it cause foo to show up in the next line?

From what I can test, it does not appear as if removal of this line causes old contents to appear anywhere; I tested various combinations of launching REPL, custom editor, pressing enter, cancelling input with Esc and Ctrl+C and other things.

If I understand correctly, in every instance where a `LineEditor` struct is spawned, it's `read_line` method is only ever called once, with no loops. As such, the "clear" function does not actually do anything: it only ever sees an already empty, freshly created line.

wez: The termwiz crate is a public export and used by others, so just looking at the usage within wezterm is not sufficient to determine that nothing is impacted by this change.

I went ahead and merged this with my suggested changes.

Thank you!

ekorchmar: This makes sense. Thank you for reviewing this

Issue #5985, opened 2024-08-18
Bell event triggers too much


gramedek:
### What Operating System(s) are you seeing this problem on?

Linux X11

### Which Wayland compositor or X11 Window manager(s) are you using?

X11 with AwesomeWM as window manager

### WezTerm version

20240812.215703.30345b36

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

Whenever a BEL character is being printed (`'\a'` / `'\007'` / `'\x07'`), the bells rings multiple times, I noticed it rings the exact same amount of times as the count of opened wezterm windows.

### To Reproduce

- open wezterm
- run `printf "\a"`

### Configuration

no config

### Expected Behavior

The bell only rings once.

### Logs

```log
Debug Overlay
wezterm version: 20240812-215703-30345b36 x86_64-unknown-linux-gnu
Window Environment: X11 awesome
Lua Version: Lua 5.4
OpenGL: Mesa Intel(R) Graphics (ADL GT2) 4.6 (Compatibility Profile) Mesa 24.1.5-arch1.1
Enter lua statements or expressions and hit Enter.
Press ESC or CTRL-D to exit
03:07:50.080 ERROR wezterm_mux_server_impl::local > writing pdu data buffer: Broken pipe (os error 32)
03:07:50.114 INFO logging > lua: the bell was rung in pane 42 in window 27!
03:07:50.114 INFO logging > lua: the bell was rung in pane 42 in window 29!
03:07:50.115 INFO logging > lua: the bell was rung in pane 42 in window 28!
03:07:50.115 INFO logging > lua: the bell was rung in pane 42 in window 0!
03:07:50.116 INFO logging > lua: the bell was rung in pane 42 in window 25!
03:07:50.117 INFO logging > lua: the bell was rung in pane 42 in window 26!
03:07:50.117 INFO logging > lua: the bell was rung in pane 42 in window 30!
```

### Anything else?

I added the following to log the bell occurrences, but the problem is the same whether or not I have a config file:
```lua
wezterm.on('bell', function(window, pane)
	_, _ = window, pane;
	wezterm.log_info(
		'the bell was rung in pane '
		.. pane:pane_id()
		.. ' in window '
		.. window:window_id()
		.. '!'
	);
end);
```

The problem isn't the audible bell, but the actual bell event, because both trigger the same amount of times.

[...]

PR #6012, opened 2024-08-22
Restrict bell events to the proper window.

loops:
As per the comment in mod.rs, bell events are sent to all windows; not just the window containing the pane which generated the event.

To prevent each bell ringing multiple times, the window event handler must check if it has the pane, and ignore the bell event if it doesn't.

This fixes bug #5985

### DIFF
--- a/wezterm-gui/src/termwindow/mod.rs
+++ b/wezterm-gui/src/termwindow/mod.rs
@@ -1223,21 +1223,23 @@ impl TermWindow {
                     alert: Alert::Bell,
                     pane_id,
                 } => {
-                    if self.window_contains_pane(pane_id) {
-                        match self.config.audible_bell {
-                            AudibleBell::SystemBeep => {
-                                Connection::get().expect("on main thread").beep();
-                            }
-                            AudibleBell::Disabled => {}
-                        }
-
-                        log::trace!("Ding! (this is the bell) in pane {}", pane_id);
-                        self.emit_window_event("bell", Some(pane_id));
-
-                        let mut per_pane = self.pane_state(pane_id);
-                        per_pane.bell_start.replace(Instant::now());
-                        window.invalidate();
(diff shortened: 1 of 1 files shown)


wezterm-gui/src/termwindow/mod.rs
```diff
@@ -1223,19 +1223,21 @@ impl TermWindow {
                     alert: Alert::Bell,
                     pane_id,
                 } => {
-                    match self.config.audible_bell {
-                        AudibleBell::SystemBeep => {
-                            Connection::get().expect("on main thread").beep();
+                    if self.window_contains_pane(pane_id) {
```
wez, line 1226: I prefer early-return style to avoid excessive nesting; that will also minimize the diff here in this PR!

```suggestion
                    if !self.window_contains_pane(pane_id) {
                       return Ok(());
                    }
```

wezterm-gui/src/termwindow/mod.rs
```diff
@@ -1223,19 +1223,21 @@ impl TermWindow {
...
+                        log::trace!("Ding! (this is the bell) in pane {}", pane_id);
+                        self.emit_window_event("bell", Some(pane_id));
 
-                    let mut per_pane = self.pane_state(pane_id);
-                    per_pane.bell_start.replace(Instant::now());
-                    window.invalidate();
+                        let mut per_pane = self.pane_state(pane_id);
+                        per_pane.bell_start.replace(Instant::now());
+                        window.invalidate();
+                    }
```
wez, line 1240:
```suggestion
```

wez (commented):
Thanks for this!
Please amend to use early return style, otherwise looks good!

loops:
Made the change, and the diff looks more presentable :-)    I added, a second commit with the changes, but I could squash them  instead if preferred.

wez:
Thanks!

PR #6013, opened 2024-08-22
Update appearance.md docs to clarify TOML is used

craigmac:
Clarify that these files are in TOML format, and add a link to TOML.


wez:
Thanks!

PR #6029, opened 2024-08-25
Update pane index when moving panes in same tab

### LINKED ISSUE #6028: split-pane --move-pane-id can kill panes

macOS

AeroSpace

### WezTerm version

20240203-110809-5046fc22 + main

### Describe the bug

Example:
```
+-----+
|  A  |
+-----+
|  B  |
+-----+
```
Command: `split-pane --pane-id $B_PANE_ID --move-pane-id $A_PANE_ID`

In `Domain::split_pane` (in `mux/src/domain.rs`), with a `SplitSource::MovePane`, the splitting pane (pane B) is first translated into a pane index (1) for its tab. Then, pane A is removed from its tab. However, this changes pane B's index (now 0). Subsequently, when we call `tab.split_and_insert(b_old_index, ...`, the old index 1 is incorrect; in this example it is in fact invalid as there is only a single pane, and pane A never gets reattached to the tab.

### To Reproduce

In a new tab, split the pane vertically. In the bottom pane, run:
`wezterm cli split-pane --move-pane-id $(wezterm cli get-pane-direction up)`

### Configuration

no config

### Expected Behavior

Top pane should be moved into the right side of a horizontal split of the previously bottom pane.

### Logs

_No response_

### Anything else?

_No response_

scauligi:
Fixes #6028.
Not sure if this is the best solution, but it works.

### DIFF
--- a/mux/src/domain.rs
+++ b/mux/src/domain.rs
@@ -84,7 +84,7 @@ pub trait Domain: Downcast + Send + Sync {
             None => anyhow::bail!("Invalid tab id {}", tab),
         };
 
-        let pane_index = match tab
+        let mut pane_index = match tab
             .iter_panes_ignoring_zoom()
             .iter()
             .find(|p| p.pane.pane_id() == pane_id)
@@ -118,6 +118,18 @@ pub trait Domain: Downcast + Send + Sync {
                 let pane = src_tab.remove_pane(src_pane_id).ok_or_else(|| {
                     anyhow::anyhow!("pane {} not found in its containing tab!?", src_pane_id)
                 })?;
+
+                // May need to update pane_index if src_pane was also in the same tab
+                if src_tab.tab_id() == tab.tab_id() {
+                    pane_index = match tab
+                        .iter_panes_ignoring_zoom()
+                        .iter()
+                        .find(|p| p.pane.pane_id() == pane_id)
+                    {
+                        Some(p) => p.index,
+                        None => anyhow::bail!("invalid pane id {}", pane_id),
+                    };
+                }
 
                 if src_tab.is_dead() {
                     mux.remove_tab(src_tab.tab_id());
--- a/mux/src/domain.rs
+++ b/mux/src/domain.rs
@@ -84,7 +84,7 @@ pub trait Domain: Downcast + Send + Sync {
             None => anyhow::bail!("Invalid tab id {}", tab),
         };
 
-        let mut pane_index = match tab
+        let pane_index = match tab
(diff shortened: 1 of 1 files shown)


mux/src/domain.rs
```diff
@@ -119,6 +119,18 @@ pub trait Domain: Downcast + Send + Sync {
...
+                    pane_index = match tab
+                        .iter_panes_ignoring_zoom()
+                        .iter()
+                        .find(|p| p.pane.pane_id() == pane_id)
+                    {
+                        Some(p) => p.index,
+                        None => anyhow::bail!("invalid pane id {}", pane_id),
+                    };
+                }
+
```
wez, line 133: Let's make this unconditional but move it to just before the `tab.split_and_insert` call on line 142:

```suggestion
                   // May need to update pane_index if the logic above mutated the tab contents
                   let pane_index = match tab
                        .iter_panes_ignoring_zoom()
                        .iter()
                        .find(|p| p.pane.pane_id() == pane_id)
                    {
                        Some(p) => p.index,
                        None => anyhow::bail!("invalid pane id {}", pane_id),
                    };
```

mux/src/domain.rs
```diff
@@ -84,7 +84,7 @@ pub trait Domain: Downcast + Send + Sync {
             None => anyhow::bail!("Invalid tab id {}", tab),
         };
 
-        let pane_index = match tab
+        let mut pane_index = match tab
```
wez, line 87: with the changes suggested below, we can keep this immutable, which makes the code overall easier to reason about:

```suggestion
        let pane_index = match tab
```

wez (changes requested):
Thanks for this!

I have a suggestion that I think makes the flow clearer overall, and which will be more robust against other changes that might mutate the tab contents in the future

scauligi:
I think this change is in line with what you're suggesting? There's a minor complication in that the pre-removal `pane_index` is needed to calculate `split_size`, but the post-removal `pane_index` is what we need for the final `tab.split_and_insert` call.

wez:
Yeah, this is what I was thinking. The complication you mentioned just sounds like how this works now, or did you mean that phrasing the code this way introduces another issue?

scauligi:
Whoops! Lost track of this and forgot to reply.
The "complication" was in reference to the fact that the new pane index is needed for the `tab.split_and_insert` call, so recalculating it inside the nested scope wouldn't work since we need the new index in the outer scope (which my latest change should be taking care of).

wez:
Thank you!

PR #6051, opened 2024-08-28
nix: output WezTerm.app on Mac

rickyelopez:
currently only the binaries are output. Build and output the .app dir so that users can launch the app as-normal on Mac

Added code is taken from [nixos](https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/applications/terminal-emulators/wezterm/default.nix#L116-L123) where it seems to be working fine. Confirmed working on my system.


rickyelopez:
@wez could you please take a look at this small change when you get a chance? Thanks!

wez:
Thanks!

PR #6054, opened 2024-08-28
Custom prompt for PromptInputLine

mgpinf:
Hello @wez 

This change introduces an additional optional parameter to set the prompt in the PromptInputLine. If not passed, the default value of "> " is used.

I'm pretty new to rust, and not sure what test cases to add/modify for these.

Please suggest improvements if any

### DIFF
--- a/docs/config/lua/keyassignment/PromptInputLine.md
+++ b/docs/config/lua/keyassignment/PromptInputLine.md
@@ -19,7 +19,7 @@ upon the input.
   user entered. `line` may be `nil` if they hit Escape without entering
   anything, or CTRL-C to cancel the input.
 * `prompt` - the text to show as the prompt. You may embed escape sequences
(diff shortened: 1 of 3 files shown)


docs/config/lua/keyassignment/PromptInputLine.md
```diff
@@ -18,6 +18,8 @@ upon the input.
   objects from the current pane and window, and `line` is the text that the
   user entered. `line` may be `nil` if they hit Escape without entering
   anything, or CTRL-C to cancel the input.
+* `prompt` - the text to show as the prompt. You may embed escape sequences
+  and/or use [wezterm.format](../wezterm/format.md).  Defaults to: `"> "`
```
wez, line 22: let's annotate this with the upcoming version:

```suggestion
  and/or use [wezterm.format](../wezterm/format.md).  Defaults to: `"> "`. {{since('nightly', inline=True)}}
```

wez (approved):
Looks good, thanks!

Issue #5142, opened 2024-03-10
resizing in mux domains has issues


crides:
### What Operating System(s) are you seeing this problem on?

Linux X11, macOS

### WezTerm version

wezterm 20240102-011725-2f7ca41a

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

No, and I'll explain why below

### Describe the bug

1. Resizing makes the bar glitchy
2. Resizing can change the window size
3. Resizing can take a long time, where the server resizes and redraws each pane one step at a time, even if the dragging itself is very quick (doesn't happen for resizing with keybinds, which is expected)

### To Reproduce

1. attach to mux ((slow) ssh is better, but unix works fine for at least the first 2)
2. split (it's easier to see the problems when split vertically)
3. drag the bar

### Configuration

no config

### Logs

```
01:05:37.087 ERROR wezterm_term::terminalstate::iterm > my pixel dimensions are wacky! 0x0
01:05:37.149 ERROR mux::connui > while running ConnectionUI loop: recv_timeout: channel is empty and disconnected
01:05:37.560 ERROR wezterm_term::terminalstate::iterm > my pixel dimensions are wacky! 0x0
01:05:37.633 ERROR mux::connui > while running ConnectionUI loop: recv_timeout: channel is empty and disconnected
01:05:38.008 ERROR wezterm_term::terminalstate::iterm > my pixel dimensions are wacky! 0x0
01:05:38.074 ERROR mux::connui > while running ConnectionUI loop: recv_timeout: channel is empty and disconnected
01:05:38.445 ERROR wezterm_term::terminalstate::iterm > my pixel dimensions are wacky! 0x0
01:05:38.514 ERROR mux::connui > while running ConnectionUI loop: recv_timeout: channel is empty and disconnected
```

### Anything else?

mux with unix

https://github.com/wez/wezterm/assets/15631555/6544b25b-b26b-49a9-b80a-bb408bf70d51

mux with ssh. Slightly easier to see the effects, but I wanted to show issue 3 which didn't show up. Issue 3 is clearly visible on my work computer.

https://github.com/wez/wezterm/assets/15631555/ace74c15-6692-4d5d-a128-749db0c4359b

yshui:
I have the same issue.

The problem is much more pronounced if there is a lot of text in the pane. And sometimes the panes will keep dancing around even after I let go of them, and will only stop after I have detached then re-attached the domain.

yshui:
[...]

PR #6062, opened 2024-08-30
don't resize pane with delta

yshui:
say the pane with 100 cols wide, we got 3 mouse events, one move the split to 99, the next one move it to 98, then to 90

we would first call resize_split_by with -1, -1, then with -8. the first resize is completed just as we call resize_split_by(-8), this triggers a resync, and updates the split position to 99. because of that resize_split_by(-8) will try to resize it to 91, instead of the desire 90.

in practice, there are a lot more mouse events and the timing is more unpredictable, the end result can be very chaotic.

partially fix wez#5142

(here is a clip illustrating the problem: https://fosstodon.org/@yshui/113013856546255206)

### DIFF
--- a/mux/src/tab.rs
+++ b/mux/src/tab.rs
@@ -614,7 +614,7 @@ impl Tab {
     /// Called when running in the mux server after an individual pane
     /// has been resized.
     /// Because the split manipulation happened on the GUI we "lost"
-    /// the information that would have allowed us to call resize_split_by()
+    /// the information that would have allowed us to call resize_split_to()
     /// and instead need to back-infer the split size information.
     /// We rely on the client to have resized (or be in the process
     /// of resizing) affected panes consistently with its own Tab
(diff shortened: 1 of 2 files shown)


yshui:
there are more issues that need to be fixed to resolve #5142 

there is a feedback loop: (client) resize -> (server) resize -> (client) resync -> (client) resize.
there is an issue with display size vs real size.

and there is another issue that cause the server to become unresponsive after the client sent it a large number of messages. this one i haven't started looking into yet.

AlD:
WezTerm is unusable for me without this patch, even if it's just partial. I'd recommend merging it.

yshui:
I don't think wezterm is actively developed anymore these days. Author has said before he doesn't have enough bandwidth. I'd recommend using something else, I am no longer using wezterm. There are many good alternatives.

PR #6063, opened 2024-08-30
Handle non-sRGB displays on macOS

rianmcguire:
Fixes https://github.com/wez/wezterm/issues/5824

Explicitly set the window's color space to sRGB, as the OpenGL output is always in sRGB.

If you don't set a color space on the NSWindow, macOS appears use the current display's color space by default. This causes colors to render incorrectly on non-sRGB displays (like the built-in LCD on some Macs).

I've tested this change on both the built-in LCD and an external sRGB monitor, and the colors now render the same as Safari:


wez:
Thanks!

PR #6082, opened 2024-09-03
fix: link to launcher menu

silasbue:
The link was pointing to the top of the page instead of the correct section.


wez:
Thanks!

PR #6084, opened 2024-09-04
Avoid potential panic, when starting in Wayland.

loops:
As per the documentation for the OutputState info method:

/// This may be none if the output has been destroyed or the
/// compositor has not sent information about the output yet.

Unwrapping its return has lead to crashes at startup, which to
the user can appear as WezTerm refusing to start.


wez:
Thanks!

PR #6103, opened 2024-09-07
Add support for ProxyUseFdPass to ssh

loops:
As per https://github.com/wez/wezterm/issues/6093

Utilizes a small, 3rd party crate (passfd) which allows the proxy
command to return a connected FD, instead of copying data.
This support is not implemented on Microsoft, and the crate
was marked as a dependency on unix-like systems only.

Some, associated code cleanup and reorg changes are included
in a trailing commit, and can be safely dropped if not wanted.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3657,6 +3657,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "passfd"
+version = "0.1.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b332c50e4d07c0011fff51ea305374408319908908bc1dbed7a0ffaaf63a8151"
+dependencies = [
+ "libc",
+]
+
+[[package]]
 name = "paste"
 version = "1.0.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -6479,6 +6488,7 @@ dependencies = [
  "libssh-rs",
  "log",
  "once_cell",
+ "passfd",
  "portable-pty",
  "predicates",
  "regex",
--- a/wezterm-ssh/Cargo.toml
+++ b/wezterm-ssh/Cargo.toml
@@ -40,6 +40,9 @@ wezterm-uds = { path = "../wezterm-uds" }
 # Not used directly, but is used to centralize the openssl vendor feature selection
 async_ossl = { path = "../async_ossl" }
 
+[target.'cfg(unix)'.dependencies]
+passfd = "0.1.6"
+
 [dev-dependencies]
 assert_fs = "1.0.4"
 clap = {version="4.0", features=["derive"]}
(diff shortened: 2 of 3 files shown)


wezterm-ssh/src/config.rs
```diff
@@ -5,6 +5,20 @@ use std::path::{Path, PathBuf};
...
+
+impl ConfigMapOps for ConfigMap {
+    fn is_yes(&self, name: &'static str) -> bool {
+        Some("yes") == self.get(name).map(|s| s.as_str())
+    }
+    fn as_str(&self, name: &'static str) -> Option<&str> {
+        self.get(name).map(|s| s.as_str())
+    }
+}
+
```
wez, line 21: Heh, this is (un?)surprisingly the most contentious part of this PR.  I have opinions; I think they are minor but there are several!

First, I think the use of a trait here is a walking a line of laziness; I say that without judgement, as economy of line-of-code count is generally desirable. I think that it makes things a bit repetitive here, which feels like more effort overall. It's not so bad that I hate it, but it doesn't quite sit with me.

What I propose instead is to make `ConfigMap` a newtype wrapper that derefs to the underlying map, so it "feels like" inheritance. Yes, it is more lines of code, but less repetition:

```rust
pub struct ConfigMap(BTreeMap<String, String>);

impl std::ops::Deref for ConfigMap {
   type Target = BTreeMap<String, String>;
   fn deref(&self) -> &Self::Target {
      &self.0
   }
}

impl std::ops::DerefMut for ConfigMap {
   fn deref_mut(&mut self) -> &mut Self::Target {
     &mut self.0
   }
}
```

Then these helper methods can be added as methods more directly; I suggest slightly different naming and implementation, something like this:

```rust
impl ConfigMap {
   pub fn get_str(&self, key: &str) -> Option<&str> {
      self.get(key).as_deref()
   }

   pub fn is(&self, key: &str, candidate: &str) -> bool {
      self.get_str(key) == Some(candidate)
   }

   pub fn is_yes(&self, key: &str) -> bool {
       self.is(key, "yes")
   }
}
```

wez (commented):
Thanks for this!

I have comments about the cleanup portion of this, but the rest seems good to me!

wezterm-ssh/src/config.rs
loops: Okay,  I gave it a try.  Had to add a ConfigMap::new, and then was still left with errors
for callers not being able to use it as an iterator.  Making it an iterator is beyond my
knowledge at this point.  Sorry.

But this really isn't an integral part of the feature request; i'd suggest we just go
with the first 3 commits, and drop this one altogether for now.  Can always circle
back later.

wez: re: iterator, there are some iterator traits but I think declaring them is probably not warranted for this. You should be able to explicitly call `.iter()` on the newtype to get at the iterator. 

I'm happy to take the first three commits for now.

loops: Okay, dropped that part of the series, and pushed.

wez:
Thanks!

cpuguy83:
Thanks for working on this!

Issue #3266, opened 2023-03-16
iterm2 inline image protocol: the screen is scrolled whenever an image (with`doNotMoveCursor=0`) reaches the last line


AnonymouX47:
### What Operating System(s) are you seeing this problem on?

Linux X11

### Which Wayland compositor or X11 Window manager(s) are you using?

KWin

### WezTerm version

20230316-093132-88e53a5f

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

With `doNotMoveCursor=0`, the screen is scrolled if an image reaches the last line, even though the cursor has neither reached the last column nor advanced to the next line. This only occurs if the image reaches the last line of the screen.

This also occurs on the alternate screen/buffer and **messes up TUIs**, which is my major concern.

See the screencast attached below.

### To Reproduce

Emit an iterm2 inline image escape sequence to the terminal **without a trailing EOL (_newline_, "\n")**.

Sample sequence (all white image, 1x1 cells)
```
\x1b]1337;File=size=86;width=1;height=1;preserveAspectRatio=0;inline=1:iVBORw0KGgoAAAANSUhEUgAAAAoAAAAWCAIAAAB2RJoKAAAAHUlEQVR4XmP8//8/A27AhC6ACkalsYJRaayAMmkALcYDKZtUpxIAAAAASUVORK5CYII=\x1b\\
```

If in a shell or REPL, ensure you put the program to sleep for some seconds immediately after emitting the sequence **before an EOL is emitted**... without the sleep, this behaviour will not be evident.

Observe the behaviour when the image reaches the last screen line versus when it doesn't.

### Configuration

no config

### Expected Behavior

The screen should not be scrolled if the image's height doesn't require it i.e when the image ends exactly on last line of the screen.

If the image requires more lines than what is left on the screen, the screen shouldn't be scrolled any more than the image's height requires i.e no extra line.

### Anything else?

This doesn't occur with `doNotMoveCursor=1` but:
- there are concrete reasons why this argument can't work with my use case
- politely, this isn't an excuse for `doNotMoveCursor=0` to be broken

Screencast:

https://user-images.githubusercontent.com/61663146/225728821-bce6ada3-38b1-4055-9138-44feb840ef1b.mp4

Note that the cursor never advances to the next line until after the sleep nor does it reach the last column.

wez:
[...]

PR #6114, opened 2024-09-10
fix non-iTerm2 cursor position after sixel image

### LINKED ISSUE #6110: inconsistent vertical cursor position after sixel display

Linux Wayland

Gnome Shell 46.4-1

### WezTerm version

20240812-215703-30345b36

### Describe the bug

After displaying a sixel image, the vertical cursor is often not in the expected position, and the image is subsequently (partially) overwritten by the prompt. This also happen when displaying multiple sixel images one after the other: each one overwrites the bottom (or the whole) of the next one. 

### To Reproduce

I'm using images of the [Spleen bitmap fonts](https://github.com/fcambus/spleen) for demonstration purposes ([8×16](https://camo.githubusercontent.com/073f58b1089321519ee7bb61607785cc8ae2c14dc4b7e1d22f1dbe88dd5ab132/68747470733a2f2f7777772e63616d6275732e6e65742f66696c65732f73706c65656e2f73706c65656e2d387831362e706e67), [12×24](https://camo.githubusercontent.com/63d9aaad34cbd8396fc58392bf4dedbdb83731ae5e1aae8198f46ebba7977fdf/68747470733a2f2f7777772e63616d6275732e6e65742f66696c65732f73706c65656e2f73706c65656e2d31327832342e706e67), [16×32](https://camo.githubusercontent.com/a02160741c91230c38edd751c27ea9599b2d3c901ca61783847bc189663f080d/68747470733a2f2f7777772e63616d6275732e6e65742f66696c65732f73706c65656e2f73706c65656e2d31367833322e706e67)):

Using ImageMagick to display the image 3 times:

---

**WezTerm:**

**8×16:**
image

**12×24:**
image

**16×32:**
image

In the worst case, we only see one of the 3 repeats, and most of it is immediately overwritten by the prompt. For the 16×32 font, we can see each image overwrites the bottom of the previous one.

---

In contrast, we don't get that with e.g. mlTerm:

**mlTerm:**

**8×16:**
image

**12×24:**
image

**16×32:**
image

### Configuration

no config

### Expected Behavior

The vertical cursor should always be left at the expected position, _after_ the last line of the sixel image.

### Logs

```
Debug Overlay
wezterm version: 20240812-215703-30345b36 x86_64-unknown-linux-gnu
Window Environment: Wayland
Lua Version: Lua 5.4
OpenGL: AMD Radeon RX 590 Series (radeonsi, polaris10, LLVM 18.1.8, DRM 3.57, 6.10.8-arch1-1) 4.6 (Compatibility Profile) Mesa 24.2.2-arch1.1
Press ESC or CTRL-D to exit
```

### Anything else?

_No response_

jdtournier:
Fixes #6110.

It seems the problem was introduced by a115fc0d to fix issue #3266. The fix was intended to address issues with cursor positioning when displaying iTerm images, but prevents the final newline from being emitted in other modes. 

This commit fixes the issue I had in #6110 for sixel display, producing behaviour that matches mlTerm and minTTY. That said, I'm not sure whether the additional test performed is the most appropriate - please advise either way.

### DIFF
--- a/term/src/terminalstate/image.rs
+++ b/term/src/terminalstate/image.rs
@@ -217,7 +217,7 @@ impl TerminalState {
                 xpos += x_delta;
             }
             ypos += y_delta;
-            if !params.do_not_move_cursor && y < height_in_cells - 1 {
+            if !params.do_not_move_cursor && (params.style != ImageAttachStyle::Iterm || y < height_in_cells - 1) {
                 self.new_line(false);
             }
         }


term/src/terminalstate/image.rs
```diff
@@ -217,7 +217,7 @@ impl TerminalState {
                 xpos += x_delta;
             }
             ypos += y_delta;
-            if !params.do_not_move_cursor && y < height_in_cells - 1 {
+            if !params.do_not_move_cursor && (params.style != ImageAttachStyle::Iterm || y < height_in_cells - 1) {
```
wez, line 220: I feel like the logic here is hard to grok. Does this make it clearer and reflect the correct intent?

```suggestion
            let new_line = match params.style {
               ImageAttachStyle::Iterm => !params.do_not_move_cursor && y < height_in_cells - 1,
               _ => !params.do_not_move_cursor
            };
            if new_line {
               self.new_line(false);
            }
```

wez (commented):
Thanks for this!

I'm not sure that I understand the error condition, specifically, why the bounds check should only be applicable to iterm style and not the others?

term/src/terminalstate/image.rs
jdtournier: As a disclaimer: this is my first time looking at Rust code (I tend to focus on C++). I'll defer to you as to whether this makes the intent clearer!

jdtournier:
I must also admit, I'm not sure why this would only apply to iterm images. The issue in #3266 was that the newline made the screen scroll one line beyond the image displayed (when at the bottom of the screen). With this change, this happens with sixel images too (and this matches the behaviour of the other terminals I've tested). 

But for some reason, with the current code, the subsequent prompt appears below the iterm image, as it should (at least that's what I can see from #3266), but not for sixel images: the prompt is displayed at the current line, on top of the last line of the sixel image. 

Is it possible that an additional newline is inserted for iterm images at some later stage, when the prompt is about to be displayed...?

jdtournier:
> But for some reason, with the current code, the subsequent prompt appears below the iterm image, as it should (at least that's what I can see from #3266)

I should add that the statement above is completely untested! I've never used this type of image, happy to test if there's an easy example I can get my hands on?

wez:
I think I'd feel happier removing that bounds check in all cases, assuming that is valid.  In other words: that original commit may have had some extraneous stuff, but it's been long enough that I don't recall the details, and my current mode right now is "light triage" of PRs, so I don't have bandwidth to go deeper on this right now.

If you have some cycles and are willing, could you take a look at the original issue to see if it would be resolved without that condition? It looks like there were two parts to the offending commit and maybe it would be fine without that part?

jdtournier:
Sure, I'll take a look into it. I've been meaning to delve into Rust, I guess this is as good as opportunity as any!

jdtournier:
Closing the PR as there doesn't seem to be anything to fix - see association discussion on #6110.

PR #6122, opened 2024-09-11
shell-integration: be resilient to nounset

antoineco:
In a shell started with the `nounset` option set (e.g. `zsh -u`), WezTerm's shell integration fails with noisy errors such as `.wezterm.sh:18: BASH_VERSION: parameter not set`.

This can be trivially avoided by using `${parameter-}` substitutions, which is supported in all the shells supported by WezTerm. (In fact, substitutions were already used in a few places.)


wez:
Thanks!

Issue #5482, opened 2024-05-27
wezterm crash when copy in x11


HIDE-r:
### What Operating System(s) are you seeing this problem on?

Linux Wayland

### Which Wayland compositor or X11 Window manager(s) are you using?

hyprland

### WezTerm version

wezterm 20240203-110809-5046fc22

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

crash when use cursor select text, or use copy mode 

```
❯ wezterm --config enable_wayland=false --skip-config start -- bash
01:49:27.850  ERROR  wezterm_gui > running message loop: process_queued_xcb: process_xcb_event_ime: ChangeProperty {
    mode: Replace,
    window: Window {
        res_id: 2097322,
    },
    property: Atom {
        res_id: 246,
    },
    type: Atom {
        res_id: 294,
    },
    data: [
        99,
        111,
        108,
        108,
        105,
        110,
        64,
        76,
        101,
        110,
        111,
        118,
        111,
        45,
        82,
        57,
        48,
        48,
        48,
        75,
    ],
}: X(Window(ValueError { response_type: 0, error_code: 3, sequence: 1032, bad_value: 2097322, minor_opcode: 0, major_opcode: 18, pad: 1 }), Some("x::ChangeProperty")); terminating
```

### Configuration

no config also happens

### Expected Behavior

no crash

ystyle:
same issues

GertBurger:
Just to add, this is likely to affect at least all Arch users as I think the incompatibility was introduced with a recent-ish hyprland update.

The crash also kills all open wezterm terminals which is quite disruptive.

Has anyone found a workaround yet?

GertBurger:
As a workaround I have disabled the default mouse bindings using `config.disable_default_mouse_bindings = true`

but I guess one can also try to disable the problematic bindings individually:
```
config.mouse_bindings = {
    {
        event = { Down = { streak = 1, button = 'Left' } },
        action = wezterm.action.DisableDefaultAssignment
    },
    {
        event = { Down = { streak = 2, button = 'Left' } },
        action = wezterm.action.DisableDefaultAssignment
    },
    {
        event = { Down = { streak = 3, button = 'Left' } },
        action = wezterm.action.DisableDefaultAssignment
    },
    {
        event = { Down = { streak = 4, button = 'Left' } },
[...]

PR #6135, opened 2024-09-14
Don't exit-program on X11 selection-event protocol errors 

loops:
There are situations where transient xcb errors can be generated in regard to copy/paste selection.  One example was reported in connection with "wl-clip-persist" in issue #6128.  And another in issue #5482.

But there's no reason for us to commit harakiri in response, so just quietly ignore such errors.

### DIFF
--- a/window/src/os/x11/window.rs
+++ b/window/src/os/x11/window.rs
@@ -795,13 +795,21 @@ impl XWindowInner {
                 conn.child_to_parent_id.borrow_mut().remove(&self.child_id);
             }
             Event::X(xcb::x::Event::SelectionClear(e)) => {
-                self.selection_clear(e)?;
+                if let Err(err) = self.selection_clear(e) {
+                    log::error!("Error handling SelectionClear: {err:#}");
+                }
             }
             Event::X(xcb::x::Event::SelectionRequest(e)) => {
-                self.selection_request(e)?;
+                if let Err(err) = self.selection_request(e) {
+                    // Don't propagate this, as it is not worth exiting the program over it.
+                    // <https://github.com/wez/wezterm/pull/6135>
+                    log::error!("Error handling SelectionRequest: {err:#}");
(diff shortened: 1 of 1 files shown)


wez:
Thanks for this, but I don't see evidence that wezterm is actually panicking or crashing here or in the linked issues, as the call stack behind the error propagation here doesn't panic, it just logs the error and shares it as a notification.

Is the goal really just to silence errors instead, or is there actually a panic?

loops:
Hi wez,

Hope you're doing well!

You're right it doesn't actually segfault.  But the code reads like it will escape the "forever" loop, as the error is propagated up the call chain via the "anyhow" type.   I was unable to reproduce the problem locally, so based my description on the results reported in the issues.

I am happy to make a more appropriate commit message, that changes "panic" to "exit program".

wez:
Ah, I see.  OK, the way I look at this sort of error is, what is the highest level at which we can/should catch it?
I think right now this PR is a bit naughty in that it is suppressing errors and continuing to respond to the incoming request when it didn't really complete its work, which propagates undefined state across an IPC boundary, and that feels a bit bad.

What I propose instead is to unwind those changes and instead catch and log over here:

https://github.com/wez/wezterm/blob/dd53cee8d89e2a0cfaab701f08ae931eef79d8c8/window/src/os/x11/window.rs#L800-L802

```rust
 Event::X(xcb::x::Event::SelectionRequest(e)) => { 
     if let Err(err) = self.selection_request(e) {
        // Don't propagate this, as it is not worth exiting the program over it.
        // <https://github.com/wez/wezterm/pull/6135>
        log::error!("error handling SelectionRequest: {err:#}");
     } 
 } 
```

It seems like similar treatment should be applied to the other selection related events in that file?

What do you think?

loops:
Makes a lot more sense.     Pushed.

PR #6146, opened 2024-09-15
shell-integration: Fix syntax error due to missing double quote

chenxiaolong:
This fixes a regression introduced in #5999.


wez:
Thanks!

PR #6148, opened 2024-09-16
Remove `--cwd` from `wezterm.desktop`

rindeal:
Whatever the `--cwd` was trying to fix in https://github.com/wez/wezterm/pull/2271, forcing `$HOME` to be the default dir upon everyone just is not cutting it.

Reverts: https://github.com/wez/wezterm/commit/4e2b5cd2cdcfd7ccce37d19e6742fdae8db2881b
Fixes: https://github.com/wez/wezterm/issues/2933


wez:
Thanks for diving in!

However, I won't merge this, because it is seems very much like an angry reaction to things not working the way you want, and willfully ignores the problems that regressing this will cause for other users, and the increase in support burden that this will cause for me if it were to be merged as-is.

If you really want to remove this, you will need to put in the work to understand the original issues across all platforms and come up with a proper solution.

PR #6156, opened 2024-09-18
Add CI for Fedora Linux 41, drop Fedora Linux 38 as it's EOL

Guara92:


wez:
Thanks for this, however, the build appears to be failing on fedora 41, seemingly due to an openssl installation issue.
That may be a transient thing, or it may be an indication that `get-deps` needs to be updated for fedora 41.
I don't have bandwidth to explicitly test on fedora 41 at this time, and I won't merge this while it is broken, because I already get too many github notifications and don't want any more!

Guara92:
I'll try to have a look at it!

Guara92:
I think I found the root cause, the problem is inside `libssh-rs`'s `vendored` feature, it can be fixed applying [this patch](https://src.fedoraproject.org/rpms/libssh/blob/1e405a954baf1265974db264e7c2fd28fec84525/f/libssh-0.10.6-no-engine.patch) but I don't know the best way to do it in a FFI crate, I also tried to upgrade the vendored libssh to 0.11.1, that contains the fixes [[1](https://gitlab.com/libssh/libssh-mirror/-/commit/2fe9ed17647a8312e18002cb76c0919a60a4fec8),[2](https://gitlab.com/libssh/libssh-mirror/-/commit/5a2654c8370bba171007358440c1f39de72f8440)], but other problems I think not related to F41 arise.
@wez Do you have any suggestion to move forward?

wez:
I've updated libssh; can you rebase this PR and let's see what happens with the build?
Thanks!

wez:
Thank you!

Issue #6093, opened 2024-09-05
wezterm ssh does not support ProxyUseFdPass


cpuguy83:
**Is your feature request related to a problem? Please describe.**

I use a custom proxy command to ssh into my machines to handle things like make sure the VPN is connected, make sure the machine is started, etc.
I always had some issues with the setup until I found `ProxyUseFdPass` in ssh.

`ProxyUseFdPass` tells ssh that the command is going to pass back a file descriptor to SSH to use rather than keeping the proxy command alive.
In this case stdout of the proxy process is wired up to a unix socket, which is used to pass back the file descriptor for the ssh connection.

Unfortunately the ssh implementation wezterm uses does not support this and even ignores the ssh config option erroring out with a message (after a long timeout it complains that the proxy process exited) that took me a bit to put the two things together.

**Describe the solution you'd like**

It would be great if wezterm ssh could support this option.

**Describe alternatives you've considered**

Only thing I can do here is go back to the more error prone case I had before using `ProxyUseFdPass`... or just not use wezterm's ssh implementation.

**Additional context**

wez:
This is implemented in current nightly builds

cpuguy83:
Thank you!

Issue #6166, opened 2024-09-20
Single line scrolling fails in mux-server, if application uses alternate screen mode


loops:
### What Operating System(s) are you seeing this problem on?

Linux Wayland

### Which Wayland compositor or X11 Window manager(s) are you using?

Sway

### WezTerm version

20240915-082426-9bf30b8d

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

When connected to a Unix or Ssh mux-server, attempting to scroll the screen, when an application has initialized the alternate-screen mode, only the bottom two lines appear to scroll, rather than the full screen.

### To Reproduce

The most common place to see this would be in "less" (as long as the -X option is not used):

Make sure -X not used:

     $ unset LESS

Now generate lines, and press ENTER multiple times in less to scroll screen:

      $ seq 1 1000 | less

Another way to see the is to use this script (which works fine unless going through mux):

     $ tput smcup ; tput clear cup $(tput lines) ; for i in $(seq 1 20) ; do echo $i ; sleep 0.1; done

As an aside, the above script works fine if you remove the sleep.

### Configuration

no config

PR #6168, opened 2024-09-20
ssh doc addition, and remove compiler warning 

loops:
Two small separate changes.

Address a compiler warning, and make a small doc change for changes from  #6093

### DIFF
--- a/docs/ssh.md
+++ b/docs/ssh.md
@@ -55,6 +55,10 @@ All other options are parsed but have no effect.  Notably, neither `Match` or
 
 `Include` is now supported.
 
+{{since('nightly')}}
+
+`ProxyUseFDpass` is now supported. (But not on Microsoft Windows).
+
 ### CLI Overrides
 
 `wezterm ssh` CLI allows overriding config settings via the command line.  This
--- a/config/src/lua.rs
+++ b/config/src/lua.rs
@@ -276,7 +276,7 @@ end
         "#,
         )
         .set_name("=searcher")
-        .eval()
+        .eval::<()>()
         .context("replace package.searchers")?;
 
         wezterm_mod.set(


docs/ssh.md
```diff
@@ -55,6 +55,10 @@ All other options are parsed but have no effect.  Notably, neither `Match` or
 
 `Include` is now supported.
 
+{{since('20240914-091937-5212cd4c:')}}
```
wez, line 58: I use nightly for this sort of thing; after I cut a release, I will replace these since-nightly markers with the version number of the newly tagged release.

```suggestion
{{since('nightly')}}
```

term/src/screen.rs
```diff
@@ -660,7 +660,10 @@ impl Screen {
         // changed by the scroll operation.  For normal newline at the bottom
         // of the screen based scrolling, the StableRowIndex does not change,
         // so we use the scroll region bounds to gate the invalidation.
-        if scroll_region.start != 0 || scroll_region.end as usize != self.physical_rows {
+        if scroll_region.start != 0
+            || scroll_region.end as usize != self.physical_rows
+            || !self.allow_scrollback
```
wez, line 665: I'm not sure about this change; it implies that the seqno's of the viewport are not being correctly invalidated by the main logic below for the common case. In particular, I'm concerned that if this is needed, it should probably be unconditional rather than scoped to the !allow_scrollback case.

The reason this is conditional is to avoid an additional O(screen-height) fixed overhead for every scroll operation.

Please translate your repro scenario from #6166 into a unit test in
https://github.com/wez/wezterm/blob/main/term/src/test/mod.rs that triggers the problem; I'd like to see what is happening to the seqno's of the lines being scrolled without attempting to fix it, then we can figure out what the right fix is.

wez (commented):
Thanks for tackling this! I think the scrolling issue is probably more complex, so I'd suggest splitting that out from the other two fixes so that we can get those merged in the meantime

docs/ssh.md
loops: Okay, I'll remember that for the future too.

wez:
Thanks!

Issue #6158, opened 2024-09-18
current_working_dir.file_path returns partial result for paths with '#' in the name


Syntaxheld:
### What Operating System(s) are you seeing this problem on?

macOS

### WezTerm version

wezterm 20240915-153243-2d0c5cdd

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

When the current working directory name contains a hash ('#'), the result of `tab.active_pane.current_working_dir.file_path` ends before that hash.

### To Reproduce

- use supplied config
- `mkdir 'folder with #tag'`
-  `cd 'folder with #tag'`
- look at debug overlay

The `current_working_dir` result will include the whole name, but the `current_working_dir.file_path` representation will end before the `#`.

### Configuration

```lua
wezterm.on('format-window-title', function(tab, pane, tabs, panes, config)
	local cwd = tab.active_pane.current_working_dir
	if not cwd then
		return "W?"
	end

	wezterm.log_info(string.format("%s %s", "CWD:", tab.active_pane.current_working_dir))
	wezterm.log_info(string.format("%s %s", " FP:", tab.active_pane.current_working_dir.file_path))
	
    return cwd.file_path or "W??"
end)
```

### Expected Behavior

`file_path` should include the hash and the rest of the path name.

### Logs

```
00:32:14.521 INFO logging > lua: CWD: file:///Users/guest/Desktop/path%20#%20slupu
00:32:14.521 INFO logging > lua:  FP: /Users/guest/Desktop/path
```

loops:
Hi Syntaxheld,

The URL parser treats anything after '#' as a fragment.  And the same thing happens with everything after '?' being treated as a query.

This could be fixed.  But, since the current working directory should always return a nil file_path, what are you trying to do by using  .file_path in the first place?   What is the situation where this bug is a problem for you?

Syntaxheld:
@loops I stumbled over this as part of a workaround for https://github.com/wez/wezterm/issues/6154 ... I wanted the window title to show the working directory of the current tab (/ active pane).

This bug is not a problem for me, since by now I am using user variables to do what I need.

> [...] current working directory should always return a nil file_path
[...]

PR #6171, opened 2024-09-21
Handle # and ? characters in directory path

loops:
When referencing the current-working-directory, before it is set by an OSC 7 escape sequence, we ask the OS for the correct path.  This path was then being parsed as a URL; where a "#" or "?" character would be
interpreted as the start of a fragment or query
component of a URL -- which is a mistake.

So this change parses the returned directory as such, where those characters will be treated as a normal character in the path.

Nothing is changed for the OSC 7 escape sequence case. In that case, the application must percent-encode the path before sending, so that those characters are not misinterpreted.

As per issue #6158 reported by Syntaxheld

### DIFF
--- a/mux/src/localpane.rs
+++ b/mux/src/localpane.rs
@@ -1045,17 +1045,14 @@ impl LocalPane {
         {
             let leader = self.get_leader(policy);
             if let Some(path) = &leader.current_working_dir {
-                return Url::parse(&format!("file://localhost{}", path.display())).ok();
+                return Url::from_directory_path(path).ok();
             }
             return None;
         }
 
         #[cfg(windows)]
         if let Some(fg) = self.divine_foreground_process(policy) {
-            // Since windows paths typically start with something like C:\,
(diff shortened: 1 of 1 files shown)


mux/src/localpane.rs
```diff
@@ -1055,7 +1055,7 @@ impl LocalPane {
             // Since windows paths typically start with something like C:\,
             // we cannot simply stick `localhost` on the front; we have to
             // omit the hostname otherwise the url parser is unhappy.
```
wez, line 1057: this comment is not longer in sync with the implementation. I feel reasonably confident that we can remove this comment and not have to directly test this change on windows, because we are already using `from_directory_path` in `wezterm set-working-directory`.

So, let's remove this comment!

wez (commented):
Thanks!

mux/src/localpane.rs
loops: Done!

wez:
Thanks!

PR #6185, opened 2024-09-23
adds HideOtherApplications (for macOS)

tgfisher:
Adds the ability to [hide all other apps](https://developer.apple.com/documentation/appkit/nsapplication/1428746-hideotherapplications), `hide_other_apps` while keeping wezterm in focus on mac OS (`OPT+CMD+h`).

Adds a `wezterm.action`, "HideOtherApplications".

Adds to documentation.

_Did my best to follow the `CONTRIBUTING.md`, but I'm a little green. (1) Couldn't get docs built due to "bad credentials" from `docs.github.com/rest` after playing around with authentication. (2) Looked for tests for related/similar functionality, `hide` and `hide_application`, and didn't see any. Happy to circle back on both things if anyone can provide some guidance._


wez:
Thanks for this! Please rebase the PR so that I can trigger the CI (it's a little wonky since moving to the wezterm org)

window/src/lib.rs
```diff
@@ -264,6 +264,9 @@ pub trait WindowOps {
         Ok(())
     }
 
+    /// Hide other app windows
+    fn hide_other_apps(&self);
```
wez, line 268: This should have a default NOP implementation so that the build doesn't break for all the other systems

```suggestion
    fn hide_other_apps(&self) {}
```

window/src/lib.rs
```diff
@@ -264,6 +264,9 @@ pub trait WindowOps {
         Ok(())
     }
 
+    /// Hide other app windows
```
wez, line 267: I don't think we need to add a method to the window for a function that belongs on the connection; remove this?

window/src/os/macos/window.rs
```diff
@@ -749,6 +749,13 @@ impl WindowOps for Window {
         });
     }
 
+    fn hide_other_apps(&self) {
```
wez, line 752: I think the implementation here in this file is not needed; just use the connection object's method

wezterm-gui/src/termwindow/mod.rs
```diff
@@ -2716,6 +2716,11 @@ impl TermWindow {
                     w.hide();
                 }
             }
+            HideOtherApplications => {
+                if let Some(w) = window.as_ref() {
+                    w.hide_other_apps();
+                }
```
wez, line 2722: Should probably do this like the `HideApplication` action down on line 2751 (please move this alongside it)

```suggestion
                let con = Connection::get().expect("call on gui thread");
                con.hide_other_applications();
```

wez (changes requested):

gloaysa:
Hi @wez I see this one is quite old, but I'm very much interested in this feature. Should I take it over and open a new PR after applying your suggestions?

tgfisher:
My apologies, after @wez force-pushed I thought it was out of my hands. I will apply the suggestions today.

wezterm-gui/src/termwindow/mod.rs
tgfisher: Thanks, your suggested approach makes more sense.

window/src/lib.rs
tgfisher: Yep!

window/src/os/macos/window.rs
tgfisher: Agreed, thanks!

tgfisher (commented):
Thanks @wez, totally agree with your comments.

craftzdog:
Hi @wez, could you merge this?

salazr:
hey @wez! new to wezterm, just need this added to it :)

Cg9291:
any update on the status of this change?

Issue #4102, opened 2023-08-06
Screen redraw on macOS when using (Neo)vim


omar25h:
### What Operating System(s) are you seeing this problem on?

macOS

### Which Wayland compositor or X11 Window manager(s) are you using?

yabai

### WezTerm version

20230806-091840-2a7167a0

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

It seems that there is an issue with the redrawing of the screen whenever scrolling in neovim.

https://github.com/wez/wezterm/assets/5558211/6c5646cb-1330-4470-aef6-f891627da1c6

### To Reproduce

* In a pane, open vim or neovim with a file that has at least half the number of lines shown on the screen
* Split the windows horizontally with `:sp`
* In the upper window, go to the top line in the buffer and start scrolling down. Notice how the screen is redrawn in the bottom window.

* **Note**: Hitting CTRL+L refreshes the full screen and the bottom window is redrawn as expected

[wezterm-recording-REMpIT.cast.txt](https://github.com/wez/wezterm/files/12270165/wezterm-recording-REMpIT.cast.txt)

### Configuration

```lua
local wezterm = require 'wezterm'
local act = wezterm.action

local function font_with_fallback(name, params)
  local names = { name, 'Noto Color Emoji', 'JetBrains Mono' }
  return wezterm.font_with_fallback(names, params)
end

local function get_appearance()
  if wezterm.gui then return wezterm.gui.get_appearance() end
  return 'Dark'
end

local function scheme_for_appearance(appearance)
  if appearance:find 'Dark' then return 'GruvboxDark' end
  return 'Gruvbox (Gogh)'
end

local function colors_for_appearance(appearance)
  if appearance == nil then return {} end
  if appearance:find 'Dark' then
    return {
      foreground = '#ebdbb2',
      background = '#282828',

      cursor_bg = '#bdae93',
      cursor_fg = '#665c54',

      selection_fg = '#ebdbb2',
      selection_bg = '#d65d0e',

      scrollbar_thumb = '#222222',

      split = '#444444',

      ansi = {
        '#3c3836',
        '#cc241d',
        '#98971a',
        '#d79921',
        '#458588',
        '#b16286',
        '#689d6a',
        '#a89984',
      },
      brights = {
        '#928374',
        '#fb4934',
        '#b8bb26',
        '#fabd2f',
        '#83a598',
        '#d3869b',
[...]

Issue #4607, opened 2023-11-22
Wezterm Mux Server causes weird line artifacts in neovim


miversen33:
### What Operating System(s) are you seeing this problem on?

Linux X11

### Which Wayland compositor or X11 Window manager(s) are you using?

So this is a weird one. I don't know when this happened, I have no idea if this is a neovim specific issue. I just notice the issue as an end user. And the issue is that while using neovim, I will often see lines being blank, in the wrong place, shifting around as I move the cursor throughout a file, etc.

I was able to verify this is a Wezterm issue and not a neovim issue by using neovim in several different terminals with the same neovim configuration. I have tested this in the following terminals
- Alacritty
- Konsole
- Wezterm (no mux server)

In all cases, neovim performs as expected. However if I use a unix domain, suddenly I see weird artifacts when navigating a file. Below is a video of this
https://github.com/wez/wezterm/assets/2640668/ac9ea99a-b05b-443b-9f6a-41e7069a48a8

And here is a video showing me using wezterm with the same file _without_ the mux server.
https://github.com/wez/wezterm/assets/2640668/f297228f-1986-4ab6-8668-a600b5dbeb1f

I have no idea what is going on. Help?

### WezTerm version

20231120-164150-fde92672

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

There are line artifacts happening when using neovim within the wezterm mux server

### To Reproduce

To make life as easy as possible, below is a bash script that can be used to get you into a similar state as what I was seeing (hopefully)

```bash
#!/usr/bin/bash

TEST_DIR=/tmp/wezterm-test
EDIT_FILE=https://raw.githubusercontent.com/wez/wezterm/main/mux/src/tab.rs
EDIT_FILE_NAME=tab.rs
NEOVIM_VERSION=nightly # can also be stable

mkdir -p $TEST_DIR
cat > $TEST_DIR/wezterm.lua << EOF
local wezterm = require("wezterm")
return {
    unix_domains = {
        { name = "wezterm-test-domain" }
    },
}
EOF

cat > $TEST_DIR/neovim-config.lua << EOF
-- Minimal configuration
-- mini.lua
-- Use with the --clean -u flags. EG nvim --clean -u mini.lua
-- This config will create a temp directory and will blow away that temp directory
-- everytime this configuration is loaded. Great for simulating a new installation
-- of a plugin

-- Setting some basic vim options
[...]

PR #6186, opened 2024-09-24
Fix alternate-screen scrolling

loops:
Hi Wez,

Did the test case you requested, but it didn't demonstrate the problem.  So
after staring at it a bit more, found this one-liner.  Hope it's okay.

### DIFF
--- a/term/src/screen.rs
+++ b/term/src/screen.rs
@@ -649,6 +649,8 @@ impl Screen {
     ) {
         let phys_scroll = self.phys_range(scroll_region);
         let num_rows = num_rows.min(phys_scroll.end - phys_scroll.start);
+        let scrollback_ok = scroll_region.start == 0 && self.allow_scrollback;
+        let insert_at_end = scroll_region.end as usize == self.physical_rows;
 
         debug!(
             "scroll_up {:?} num_rows={} phys_scroll={:?}",
@@ -660,7 +662,7 @@ impl Screen {
         // changed by the scroll operation.  For normal newline at the bottom
         // of the screen based scrolling, the StableRowIndex does not change,
         // so we use the scroll region bounds to gate the invalidation.
-        if scroll_region.start != 0 || scroll_region.end as usize != self.physical_rows {
+        if !scrollback_ok {
             for y in phys_scroll.clone() {
                 self.line_mut(y).update_last_change_seqno(seqno);
             }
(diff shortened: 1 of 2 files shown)


wez:
Seems like this is reverting https://github.com/wez/wezterm/pull/6099 which only just got merged.
@tbung what do you think about this and #6166 ?

loops:
Yeah, i'm able to reproduce something like the bug in #6099 after applying this PR.

So, i'm not sure what the correct answer is to fix that problem, while not breaking alt-mode mux
scrolling.  This PR made sense to me, since it seems logical that having no scrollback, should be
treated the same as having a deleted scrollback.  But will leave it to all of you who actually know
and understand the codebase.

tbung:
I'm busy until next week, then I'll take a look at what the difference between `less` and `vim` is here and devise another test case so we can come up with a solution to both issues.

loops:
Hi tbung,

Looking for a test case, I found that the problem from #6099 isn't limited to
alternate screen mode, and reappears if you disable the alternate screen
capability.  With Vim it's: 

      vim -c 'set t_ti= t_te='

But for Neovim, you have to compile a custom terminal entry that removes smcup:

     tic - <<< 'noalt|fake term, rmcup@, smcup@, use=xterm-256color,'
     TERM=noalt  nvim

In either case, the split window problem, where the status bar floats up
the screen, shows up again.

The problem is that whenever an extra line is inserted in the upper scroll
region, it invalidates the StableRowIndex for all the lines below the scroll
region, but they aren't being marked as dirty.

loops:
Hey @tbung,

  I've had a couple kicks at the can already, but couldn't help trying again.   Pushed an update that resolves the issues I was seeing.  Let me know what you think.  Hopefully it saves you a bit of time.

tbung:
Works beautifully, thanks for figuring it out! Also, really nice that you added tests for both bugs now. This should now finally mean that #4607 is solved.

loops:
Hi @tbung

That's great.   I did just make a small push, to move the invalidation logic to the end of the function, which makes more logical sense, and fix a small comment in the tests.  But it's otherwise unchanged.

We'll just see what Wez says now, and if any other cleanups or anything else is needed.

kaddkaka:
Awesome that this is getting fixed 👍

Is the check "owner approval" referring to @loops in this case?

tjex:
Can report that this resolves #6415 

I wonder if it also fixes #4102 ?

wez:
Finally got around to looking at this; thank you @loops, and everyone that has tested it; it looks good, and I really appreciate the additional test coverage for this.

vohoanglong0107:
This also seems to fix https://github.com/wez/wezterm/issues/4102. Tested on MacOS 15.1.1 arm64 with SSH multiplexing

Issue #6015, opened 2024-08-22
Builds in copr are broken


kappa:
### What Operating System(s) are you seeing this problem on?

Linux Wayland

### WezTerm version

n/a

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

buils in https://copr.fedorainfracloud.org/coprs/wezfurlong/wezterm-nightly/builds/ are showing as "failed" for many days

### Configuration

n/a

wez:
Seem specific to fedora 41 and rawhide, the other platforms are building fine.
I have no bandwidth to debug this right now.

rawhide:

```
warning: libssh-rs-sys@0.2.4: vendored/src/pki_crypto.c:36:10: fatal error: openssl/engine.h: No such file or directory
warning: libssh-rs-sys@0.2.4:    36 | #include <openssl/engine.h>
warning: libssh-rs-sys@0.2.4:       |          ^~~~~~~~~~~~~~~~~~
warning: libssh-rs-sys@0.2.4: compilation terminated.
```

fedora 41:

```
>>> Downloading successful, but checksum doesn't match. Calculated: 2030a42da5af
>>> Downloading successful, but checksum doesn't match. Calculated: 2030a42da5af
>>> Downloading successful, but checksum doesn't match. Calculated: 2030a42da5af
>>> Librepo error: Cannot download repomd.xml: Cannot download repodata/repomd.xFailed to download metadata (metalink: "https://mirrors.fedoraproject.org/metalink?repo=fedora-41&arch=aarch64") for repository "fedora"
 Librepo error: Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried
```

seems like some environmental issue in copr or with fedora 41.

SlouchyButton:
Hi, for rawhide it is very possibly related to this change https://fedoraproject.org/wiki/Changes/OpensslDeprecateEngine

OpenSSL engine in fedora was removed, since the affected file seems to be just a dependency, I don't think it would be easy to solve on `wezterm` side.

_This does impact F41 as well_, looking at last copr build logs for F41, the copr issue was resolved so now it crashes on OpenSSL engine not being present as well.

aliaksandr-trush:
Fedora build broken\disabled now? Last successful build 11 days ago.

wez:
[...]

PR #6200, opened 2024-09-27
Alter copr spec file for Fedora 41 and Rawhide

loops:
SlouchyButton mentioned in #6015 that a Fedora Openssl packaging change was likely causing a problem for the copr build.

The deprecated engine header has been broken out into its own package.  Add a dependency on it for Fedora 41 and Rawhide.

### DIFF
--- a/ci/deploy.sh
+++ b/ci/deploy.sh
@@ -171,6 +171,9 @@ BuildRequires: Mesa-libEGL-devel
 %else
 BuildRequires: mesa-libEGL-devel
 %endif
+%if 0%{?fedora} >= 41
+BuildRequires: openssl-devel-engine
+%endif
 Source0: wezterm-${TAR_NAME}.tar.gz
 
 %global debug_package %{nil}


SlouchyButton:
Just a note here, this might stop working in future Fedora versions as the use is deprecated. Ideally, the OpenSSL engine should not be used, but since `wezterm` is not using it directly afaik, there is a chance that newer versions of the dependency that does use it will fix it before it becomes a problem.

Otherwise, this looks good, slight problem is that when I tried to mockbuild the new spec file it seems that `time` dependency fails to build, possibly due to newer rust.
```
error[E0282]: type annotations needed for `Box<_>`
  --> /builddir/.cargo/registry/src/index.crates.io-6f17d22bba15001f/time-0.3.31/src/format_description/parse/mod.rs:83:9
   |
83 |     let items = format_items
   |         ^^^^^
...
86 |     Ok(items.into())
   |              ---- type must be known at this point
   |
   = note: this is an inference error on crate `time` caused by an API change in Rust 1.80.0; update `time` to version `>=0.3.35` by calling `cargo update`
```

This shouldn't really block this PR as it is unrelated, but will also need further investigation to fix.

rphillips:
@wez This does fix the issue and build on Fedora 41 beta. The github CI run looks like a flake. Can we get this merged?

SlouchyButton:
> Or if it affects some fedora versions only, a condition is needed

F41+ - since F41, the OpenSSL engine use is deprecated.

wez:
Thanks!

PR #6229, opened 2024-10-04
Fixed mail-to regex allowing email addresses to have dots in

### LINKED ISSUE #5080: Implicit mailto links should match dots before the @-sign

At my $DAYJOB we have firstname.lastname@company.com as email addresses. The default regex for finding mail links don't include dots before the @ sign, so only the lastname@company.com part gets included in the link.

Change the default regexes to include dots before the at sign. Perhaps

    \b\w+(\.\w+)*@[\w-]+(\.[\w-]+)+\b

I can of course change this myself in my local config but I figure that it would be good to change the default option as well.

Regexes are an insufficient tool to correctly parse all possible email addresses but this is a pretty common format for email addresses and it would make sense to have them included.

Looking at the regex, perhaps hyphens should be included as well?

lilHermit:
Fixes #5080


loops:
Hi lilHermit,

This change appears to also allow the hyphen character to precede the @ symbol. However, whenever you want to match an explicit hyphen inside a character-class, the usual practice is to make it the final character of the set.  Otherwise, it appears that you're actually trying to match a range of characters, as in  [F-T] (all the characters F through T).

Indeed, if you put the proposed regex into the online regex tool at https://rustexp.lpil.uk/ it complains about a boundary error.

It may work as-is, but it's probably worth altering, to move the "-" to the final character after the "\."   and also mentioning in the commit comment the addition of the hyphen as a valid character before the "@".

lilHermit:
@loops Good shout, I have now rebased

PR #6239, opened 2024-10-06
feat: implement opt-in OSC52 clipboard querying

### LINKED ISSUE #2050: Support OSC 52 clipboard querying (opt-in)

Wezterm supports writing to the clipboard via OSC 52 sequence, but does not implement reading the clipbiard. Since this could be a risky feature to enable by default, it should be opt-in, which is what kitty and xterm do.

A config setting should enable OSC 52 clipboard reading. Then applications can request the clipboard contents. In my case I have bash integration that connects readline copy bindings to write to the system clipboard, and I would like to also connect readline paste commands the same way.

I have had this readline integration working using xclip, but the OSC approach means it would also work in situations where X forwarding is not enabled. (I don't know if the wezterm ssh domain + multiplexer allows X forwarding and updating the DISPLAY variable inside the multiplexer when reattaching).

39555:
Closes: #2050 

Hi Wez! This is my first time contribution to the project. I've implemented the missing osc52 clipboard reading command. The feature is disabled by default and can be enabled via the config option `config.enable_osc52_clipboard_reading = true` .  I've also added the documentation with security notes and some tests related to osc52.

The clipboard content is received asynchronously through a channel, which is passed as the second parameter in the `Clipboard::get_contents` function. For the `TerminalState` it is the `ThreadedWriter` directly, for the mux client it is a tx-rx channel that than sends the pdu responce with the clipboard data back to the server and than to the server's `ThreadedWriter`. 

The waiting of the clipboard data on the mux server side is a bit clunky [wezterm-mux-server-impl/src/dispatch.rs](https://github.com/wez/wezterm/compare/main...39555:osc52-read?expand=1#diff-5f7131e27c54368222955f6e8059ba1ed6d1fae984ad36eb425be27d7c4102f8)
because there is no promise handling mechanism similar to what the client has so I left a mutable promise variable before the message loop and handled the response with `if Pdu::QueryClipboardResponce(..) =` before passing a pdu to the actual server's  message handler. I would like to hear any idea how to improve it.

I am open to any suggestions for naming, fixes and improvements.

### DIFF
--- a/docs/config/lua/config/enable_osc52_clipboard_reading.md
+++ b/docs/config/lua/config/enable_osc52_clipboard_reading.md
@@ -1,0 +1,30 @@
+---
+tags:
+  - osc
+  - clipboard
+---
+
+# `enable_osc52_clipboard_reading = false`
+
+{{since('nightly')}}
+
+When set to `true`, the terminal will allow access to the system clipboard by
+terminal applications via `OSC 52` [escape sequence](../../../shell-integration.md#osc-52-clipboard-paste).
+
+The default for this option is `false`.
+
+Note that it is not recommended to enable this option due to serious security
+implications.
+
+### Security Concerns
+
+Clipboards are often used to store sensitive information, and granting any
+terminal application (especially from remote machines) access to it poses a
+security risk. A malicious server could spam the terminal with OSC 52 paste
+sequences to monitor whatever you have on the clipboard, which may occasionally
+contain sensitive data.
+
+Setting clipboard data that contains escape sequences or malicious commands and
+reading it back could allow an attacker to inject harmful characters into the
+input stream. Although, the risk is somewhat mitigated as the pasted text is
+encoded in BASE64.
--- a/docs/shell-integration.md
+++ b/docs/shell-integration.md
@@ -3,6 +3,7 @@ wezterm supports integrating with the shell through the following means:
 * `OSC 7` Escape sequences to advise the terminal of the working directory
 * `OSC 133` Escape sequence to define Input, Output and Prompt zones
 * `OSC 1337` Escape sequences to set user vars for tracking additional shell state
+* `OSC 52` Escape sequences for writing to and reading from the clipboard
 
 `OSC` is escape sequence jargon for *Operating System Command*.
 
@@ -180,3 +181,33 @@ return config
 
 Now, rather than just running `cmd.exe` on its own, this will cause `cmd.exe`
 to self-inject the clink line editor.
+
+## OSC 52 Clipboard Copy
+
+The data can be copied to the `System Clipboard` or `Primary Selection` with a
+command like this:
+
+```bash
+printf '\033]52;c;%s\033\\' $(base64 <<< "hello world")
+```
+
+- `c` copies to the system clipboard
+- `p` copies to the primary selection buffer
+
+The second parameter provides the selection data, which is a string encoded in base64 (RFC-4648).
+
+## OSC 52 Clipboard Paste
+
+The data can be pasted from the `System Clipboard` or `Primary Selection` with a
+command like this:
+
+```bash
+printf "\033]7;c;?\033\\"
+```
+
+- `c` pastes to the system clipboard
+- `p` pastes to the primary selection buffer
+
+Note that this feature poses a potential security risk and is disabled by
+default. It requires enabling via [a configuration option](config/lua/config/enable_osc52_clipboard_reading.md).
+You should carefully consider whether you're willing to accept the associated risks before enabling it.
--- a/docs/shell-integration.md
+++ b/docs/shell-integration.md
@@ -205,8 +205,8 @@ command like this:
 printf "\033]7;c;?\033\\"
 ```
 
-- `c` pastes to the system clipboard
-- `p` pastes to the primary selection buffer
+- `c` pastes from the system clipboard
+- `p` pastes from the primary selection buffer
 
 Note that this feature poses a potential security risk and is disabled by
 default. It requires enabling via [a configuration option](config/lua/config/enable_osc52_clipboard_reading.md).
--- a/docs/shell-integration.md
+++ b/docs/shell-integration.md
@@ -202,7 +202,7 @@ The data can be pasted from the `System Clipboard` or `Primary Selection` with a
 command like this:
 
 ```bash
-printf "\033]7;c;?\033\\"
+printf "\033]52;c;?\033\\"
 ```
 
 - `c` pastes from the system clipboard
--- a/codec/src/lib.rs
+++ b/codec/src/lib.rs
@@ -461,6 +461,8 @@ pdu! {
     SendPaste: 13,
     Resize: 14,
     SetClipboard: 20,
+    QueryClipboard: 21,
+    QueryClipboardResponse: 63,
     GetLines: 22,
     GetLinesResponse: 23,
     GetPaneRenderChanges: 24,
@@ -516,6 +518,7 @@ impl Pdu {
             | Self::SendPaste(_)
             | Self::Resize(_)
             | Self::SetClipboard(_)
+            | Self::QueryClipboard(_)
             | Self::SetPaneZoomed(_)
             | Self::SpawnV2(_) => true,
             _ => false,
@@ -594,6 +597,7 @@ impl Pdu {
             | Pdu::SetPalette(SetPalette { pane_id, .. })
             | Pdu::NotifyAlert(NotifyAlert { pane_id, .. })
             | Pdu::SetClipboard(SetClipboard { pane_id, .. })
+            | Pdu::QueryClipboard(QueryClipboard { pane_id, .. })
             | Pdu::PaneFocused(PaneFocused { pane_id })
             | Pdu::PaneRemoved(PaneRemoved { pane_id }) => Some(*pane_id),
             _ => None,
@@ -767,6 +771,18 @@ pub struct SetClipboard {
     pub pane_id: PaneId,
     pub clipboard: Option<String>,
     pub selection: ClipboardSelection,
+}
+
+#[derive(Deserialize, Serialize, PartialEq, Debug)]
+pub struct QueryClipboard {
+    pub pane_id: PaneId,
+    pub selection: ClipboardSelection,
+}
+
+#[derive(Deserialize, Serialize, PartialEq, Debug)]
+pub struct QueryClipboardResponse {
+    pub pane_id: PaneId,
+    pub content: Option<String>,
 }
 
 #[derive(Deserialize, Serialize, PartialEq, Debug)]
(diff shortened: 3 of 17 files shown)


loops:
Hi 39555,

This works nicely!  Especially nice that you made it work with the mux-server as well.  (It might be worth
noting in the docs that the config option  also needs to be enabled on the remote machine, for the mux-server).

Great work.

39555:
"@loops, what do you mean by the config on the remote server? I thought WezTerm only reads the config from the local machine. Could you write a small guide?

39555:
I’m experiencing a strange bug on windows whether in the gui or when `wezterm connect unix` or connecting via SSH. The OSC sequence isn’t being parsed at all. However, everything works perfectly on WSL (via default_domain), macOS, and when using SSH from WSL to macOS or from Windows to macOS via `wezterm connect <ssh_domain>` or `wezterm ssh <addr>`."

39555:
I've spend some time digging in.  Unfortunately. This feature can't work on windows due to ConPTY. They implicitly filter this osc without any user notification.... Both Alacritty and Rio terminals implement this feature and it doesn't work on windows either. 

https://github.com/alacritty/alacritty/issues/7962

Jendker:
Thank you for this! This feature was essential and previously missing in wezterm.

While using your code, I noticed something that might be useful. When I use a terminal without OSC52 query support (like wezterm without your PR) and attempt to paste with OSC 52 in neovim, I see the message:
```
Waiting for OSC 52 response from the terminal. Press Ctrl-C to interrupt...
```
This informs the user that OSC 52 paste is unavailable.

However, with your PR in wezterm, if I don’t set `config.enable_osc52_clipboard_reading = true`, the query fails silently and retrieves the last copied item in neovim. I'm not sure if this is expected behavior, but it could be confusing for users who might overlook the opt-in setting.

melMass:
Thanks a lot! Is there a blocker for getting a review/merge?

39555:
> Thanks a lot! Is there a blocker for getting a review/merge?

Wez currently has no bandwidth to work on Wezterm. Maybe someone else could review it and suggest improvements before Wez is ready

docs/config/lua/config/enable_osc52_clipboard_reading.md
```diff
@@ -0,0 +1,30 @@
...
+
+### Security Concerns
+
+Clipboards are often used to store sensitive information, and granting any
+terminal application (especially from remote machines) access to it poses a
+security risk. A malicious server could spam the terminal with OSC 52 paste
+sequences to monitor whatever you have on the clipboard, which may occasionally
+contain sensitive data.
+
+Setting clipboard data that contains escape sequences or malicious commands and
```
pablospe, line 27: One question, since I don't understand. Is this still valid even using ssh (encrypted tunnel)? How can an attacker inject harmful characters without already knowing how to access to the encrypted tunnel?

docs/shell-integration.md
```diff
@@ -180,3 +181,33 @@ return config
...
+
+The data can be pasted from the `System Clipboard` or `Primary Selection` with a
+command like this:
+
+```bash
+printf "\033]7;c;?\033\\"
+```
+
+- `c` pastes to the system clipboard
+- `p` pastes to the primary selection buffer
```
Lenbok, line 209:
```suggestion
- `c` pastes from the system clipboard
- `p` pastes from the primary selection buffer
```

docs/shell-integration.md
```diff
@@ -180,3 +181,33 @@ return config
...
+
+The second parameter provides the selection data, which is a string encoded in base64 (RFC-4648).
+
+## OSC 52 Clipboard Paste
+
+The data can be pasted from the `System Clipboard` or `Primary Selection` with a
+command like this:
+
+```bash
+printf "\033]7;c;?\033\\"
```
Lenbok, line 205: I'm trying to test this out and the above didn't do anything. Surely it's supposed to be:
```suggestion
printf "\033]52;c;?\033\\"
```
which results in wezterm sending the base64ed clipboard to the screen. How would I capture/de-base64 it?

Lenbok: BTW, my goal is to write a bash function that will output the clipboard so I can bind that to a key to insert via readline. Essentially I want something like this:
```bash
function _shell_paste_p() {
    printf "\033]52;p;?\033\\"
}
_rl_paste_p() {
    CLIP=$(_shell_paste_p)
    COUNT=$(echo -n "$CLIP" | wc -c)
    READLINE_LINE="${READLINE_LINE:0:$READLINE_POINT}${CLIP}${READLINE_LINE:$READLINE_POINT}"
    READLINE_POINT=$(($READLINE_POINT + $COUNT))
}
bind -m emacs -x '"\C-y": _rl_paste_p'
```
The "cut line" equivalent is this (which works fine):
```bash
function _shell_copy() {
    printf "\033]52;c;$(base64 | tr -d '\r\n')\a"
}
_rl_kill() {
    echo -n "${READLINE_LINE:$READLINE_POINT}" | _shell_copy
    READLINE_LINE="${READLINE_LINE:0:$READLINE_POINT}"
}
bind -m emacs -x '"\C-k": _rl_kill'
```

39555: Thanks for fixing the typos! Typically, decoding is handled by editors. For example, in Neovim https://github.com/neovim/neovim/blob/2024c10d74cea760ff9e4ac8208cc4c9c176f4d3/runtime/lua/vim/ui/clipboard/osc52.lua#L22. There's also a command-line utility for this: https://github.com/theimpostor/osc. You need to read the tty device directly to obtain the response https://github.com/theimpostor/osc/blob/8a313c60d67353669509e9461d22b403df72b5e7/main.go#L273. I’m not sure how to do this in Bash myself, so if you manage to decode the response successfully, please share your solution!

Lenbok: Thanks for pointing me at the command line `osc` tool. I've got my bash integration working now with:
```
function _shell_paste_c() {
    osc paste
}
_rl_paste_c() {
    CLIP=$(_shell_paste_c)
    COUNT=$(echo -n "$CLIP" | wc -c)
    READLINE_LINE="${READLINE_LINE:0:$READLINE_POINT}${CLIP}${READLINE_LINE:$READLINE_POINT}"
    READLINE_POINT=$(($READLINE_POINT + $COUNT))
}
bind -m emacs -x '"\C-y": _rl_paste_c'
```
(I gave my earlier functions a "_p" suffix as I was expecting to use the primary selection -- the osc tool doesn't seem to let you choose primary vs clipboard, so I'm using the clipboard instead (hence I named the functions with "_c" suffix). It's all working fine, thanks!

Kreijstal:
> I've spend some time digging in. Unfortunately. This feature can't work on windows due to ConPTY. They implicitly filter this osc without any user notification.... Both Alacritty and Rio terminals implement this feature and it doesn't work on windows either.
> 
> [alacritty/alacritty#7962](https://github.com/alacritty/alacritty/issues/7962)

how come this works on mintty tho?

39555:
@Kreijstal Just discovered your issue at https://github.com/mintty/mintty/issues/1264 Thanks! Does `mintty`  use `conpty`? Because neither Allacritty nor Rio can paste from the clipboard on Windows. We need to dig inside the https://github.com/microsoft/terminal/tree/aa5459df4a33c731b25be9f47fe481d8c73db14d/src/winconpty

Kreijstal:
> @Kreijstal Just discovered your issue at [mintty/mintty#1264](https://github.com/mintty/mintty/issues/1264) Thanks! Does `mintty` use `conpty`? Because neither Allacritty nor Rio can paste from the clipboard on Windows. We need to dig inside the https://github.com/microsoft/terminal/tree/aa5459df4a33c731b25be9f47fe481d8c73db14d/src/winconpty

I guess windows terminal vendors their won conpty? crazy

39555:
I researched this a little bit. `winconpty` starts the `OpenConsole.exe` process. `OpenConsole.exe` is built from the `src/host` folder https://github.com/microsoft/terminal/blob/aa5459df4a33c731b25be9f47fe481d8c73db14d/doc/ORGANIZATION.md?plain=1#L39C1-L39C58 than some magic happens inside that scary state machine https://github.com/microsoft/terminal/blob/aa5459df4a33c731b25be9f47fe481d8c73db14d/src/terminal/parser/stateMachine.cpp and seems like the sequence gets lost here but im not sure https://github.com/microsoft/terminal/blob/0961a77a5aec003e9f79d388f8e7bffb9a550cdb/src/terminal/parser/OutputStateMachineEngine.cpp#L803

adoal:
Maybe there can be a third option `ask` for reading clipboard besides true and false, just like kitty's `clipboard_control`.

Lenbok:
Is lack of support for windows a blocker for merging this PR?

docs/config/lua/config/enable_osc52_clipboard_reading.md
wez: A plausible scenario here might be:
* Remote system has been compromised with a rootkit
* Attacker replaces something that gets run when an ssh session is incoming
* You copy and paste your password in order to login to that remote system
* your clipboard is extracted and exfiltrated to some remote site for analysis by the hacker
* your ssh session continues to login, and you are unaware that the password you had in your clipboard is now compromised

wez:
Just wanted to note that I've seen this PR, but it's pretty big and I haven't had a chance to really review it yet.

re: conpty discussion above, I wanted to note that I updated to the most recent build of that for wezterm yesterday.  It has some improvements around OSC handling that might help here.

Alby407:
I really hope this change will be merged soon, it's the only thing missing right now.

Lenbok:
@39555 Would you please be able to update this PR to resolve conflicts?

Lenbok:
FYI, I have an branch with this PR updated to recent wezterm at https://github.com/Lenbok/wezterm/tree/osc52-read

hueindahaus:
What is the status on this? Has this been abandoned?

Lenbok:
The original author of this PR seems to have lost interest. I'm currently using the updated version I linked above. It's really up to @wez or some other maintainer to review/update/merge AFAIK.

AnveshJarabani-WellSky:
yooo merge this!!!

ldelossa:
Hello all. 

This functionality is crucial to my workflow. Therefore I took the latest patch and used AI to setup an automated "patch and release" flow, with AI assisted merge conflict resolution. Right now the patched builds are available from brew. 

See: https://github.com/ldelossa/wezterm-osc52

The patched builds follow wezterm HEAD.

Kreijstal:
> Hello all.
> 
> This functionality is crucial to my workflow. Therefore I took the latest patch and used AI to setup an automated "patch and release" flow, with AI assisted merge conflict resolution. Right now the patched builds are available from brew.
> 
> See: https://github.com/ldelossa/wezterm-osc52
> 
> The patched builds follow wezterm HEAD.

@ldelossa now enable issues

Issue #5806, opened 2024-07-13
get_text_from_semantic_zone eats last line


yerlaser:
### What Operating System(s) are you seeing this problem on?

Windows

### WezTerm version

20240203-110809-5046fc22

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

I have enabed shell integration to send the appropriate OSC ANSI sequences and bound the following callback to a keyboard combination:
```lua
function(win, pane)
  local zones = pane:get_semantic_zones('Output')
  win:copy_to_clipboard(pane:get_text_from_semantic_zone(zones[#zones]), 'Clipboard')
end
```

This works in that it copies the output of the last command.
However, the last line of the output gets lost.
I know that the shell send correct OSC sequences because this mouse binding selects exactly the last command output:
```lua
action = wezterm.action.SelectTextAtMouseCursor 'SemanticZone',
```
This was taken from the documentation.

### To Reproduce

Bind below function to a keyboard shortcut.

### Configuration

```lua
function(win, pane)
  local zones = pane:get_semantic_zones('Output')
  win:copy_to_clipboard(pane:get_text_from_semantic_zone(zones[#zones]), 'Clipboard')
end
```

### Expected Behavior

Should copy the full output of the last command.

mgpinf:
There is a similar issue https://github.com/wez/wezterm/issues/5346 to which a workaround is mentioned in the comment https://github.com/wez/wezterm/issues/5346#issuecomment-2121003351. There is a PR which fixes the issue, but for the time being, the workaround does the job.

Issue #5346, opened 2024-04-27
Last line of semantic zone output not being returned in pane:get_text_from_semantic_zone(zone)


mgpinf:
### What Operating System(s) are you seeing this problem on?

macOS

### WezTerm version

20240405-180910-cce0706b

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

Last line of semantic zone output not being returned in `pane:get_text_from_semantic_zone(zone)`. Rest of the content is present.
Enabled shell integration and tried in both zsh and fish.

Tried running the following command:
<img width="435" alt="image" src="https://github.com/wez/wezterm/assets/56510798/ae4cf378-c393-41ab-a8d6-0eba61cdfeaa">

Obtained output (from logs):
```
01:01:57.041  INFO   logging > lua: total 0
drwxr-xr-x  2 unbxd  staff    64B Apr 28 00:40 another_dir
drwxr-xr-x  2 unbxd  staff    64B Apr 28 00:41 another_dir_2
drwxr-xr-x  2 unbxd  staff    64B Apr 28 00:38 dir1
-rw-r--r--  1 unbxd  staff     0B Apr 28 00:38 file1

```

### To Reproduce

Enable shell integration in zsh/fish.
Add below config and try out

### Configuration

```lua
{
  key = 'v',
  action = wezterm.action_callback(function(window, pane)
    local output_semantic_zones = pane:get_semantic_zones("Output")
    if #output_semantic_zones > 0 then
      local output_content
      for i = 1, #output_semantic_zones, 1 do
        output_content = pane:get_text_from_semantic_zone(output_semantic_zones[i])
        wezterm.log_info(output_content)
      end
    end
  end)
}
```

### Expected Behavior

Expected output should include last line as well.

### Logs

```
01:01:57.041  INFO   logging > lua: total 0
drwxr-xr-x  2 unbxd  staff    64B Apr 28 00:40 another_dir
drwxr-xr-x  2 unbxd  staff    64B Apr 28 00:41 another_dir_2
drwxr-xr-x  2 unbxd  staff    64B Apr 28 00:38 dir1
-rw-r--r--  1 unbxd  staff     0B Apr 28 00:38 file1

```

piechologist:
As a workaround, I've been using the following function ever since:
```lua
local function my_fixed_get_text_from_semantic_zone(pane, zone)
[...]

PR #6248, opened 2024-10-08
add last_row in get_text_from_semantic_region

mgpinf:
This commit adds the last row into account while computing the `pane:get_text_from_semantic_region(zone)` method.

This fixes the following:
- #5806
- #5346 

I have tested the code, and the last line of semantic zone is now part of the result.

Just want to confirm that this does not conflict with anything


wez:
Ah, this makes sense: the end coordinates are inclusive. Thank you!

PR #6271, opened 2024-10-13
docs: Put APT key into /etc/apt/keyrings

fumiyas:
sources.list(5) says:

> The recommended locations for keyrings are
> /usr/share/keyrings for keyrings managed by packages, and
> /etc/apt/keyrings for keyrings managed by the system operator.


wez:
Thanks!

Issue #5360, opened 2024-05-01
Wezterm creates crazy large window on Wayland in Fedora 40


nbenitez:
### What Operating System(s) are you seeing this problem on?

Linux Wayland

### Which Wayland compositor or X11 Window manager(s) are you using?

Mutter: default in Fedora 40

### WezTerm version

wezterm 20240429_102818_0184e88e  (Using Copr repo)

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

I just installed Fedora 40, installed Wezterm from Copr repo so it must be latest version, and when launching Wezterm it displays a very large window (goes further off-screen) and within it the terminal window only occupies a quarter of it, please check out the screenshots I attached below.

My display configuration, which wezterm correctly reports in `wezterm.gui.screens(),` (attached below) is, an external HiDPI monitor 3840x2160 at scale 2 connected to my laptop, the laptop screen is always off (set as inactive in Gnome display settings), so it's correct that it does not appear in  `wezterm.gui.screens()` output.

### To Reproduce

Just launch Wezterm from desktop file or from a terminal i.e. `wezterm start --cwd .` on a Fedora 40 Wayland session.

### Configuration

-- I think the Bug is not affected by a specific configuration, but nonetheless I'm pasting mine below:
config.window_decorations = "RESIZE"
config.enable_wayland = true;
config.enable_tab_bar = false
config.scrollback_lines = 10000 -- default is 3500
config.hide_mouse_cursor_when_typing = false
config.pane_focus_follows_mouse = true
config.check_for_updates = false
config.font_size = 10.0
config.adjust_window_size_when_changing_font_size = false
-- no ligatures
config.harfbuzz_features = {"calt=0", "clig=0", "liga=0"}

### Expected Behavior

The terminal window (i.e. black terminals as shown in the screenshot) should occupy all the window not just a quarter of it, and the window itself should have a normal size, not a width that goes far far right off-screen.

I think my setup it's pretty normal on Linux i.e. latest Fedora in a laptop connected to an external monitor.

### Logs

Debug Overlay
wezterm version: 20240429_102818_0184e88e x86_64-unknown-linux-gnu
Window Environment: Wayland
Lua Version: Lua 5.4
OpenGL: Mesa Intel(R) HD Graphics 530 (SKL GT2) 4.6 (Compatibility Profile) Mesa 24.0.5
[...]

Issue #6262, opened 2024-10-10
Wayland xdg_surface::configure event sends incorrect window size


aliaksandr-trush:
### What Operating System(s) are you seeing this problem on?

Linux Wayland

### Which Wayland compositor or X11 Window manager(s) are you using?

Mutter

### WezTerm version

wezterm 20241007_103714_ed430415

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

When WezTerm window Maximized\FullScreen\Tiled and it is loosing the focus, window receiving incorrect window size in the configure event (initial windows size), instead of correct one.
In logs you can see a few configure events received by WezTerm:
1. Window creation: new_size(none, none)
2. Window creation in windowed mode: new_size(Some(1260), Some(900))
3. TIle to the right WindowConfigure { new_size: (Some(960), Some(1048))
4. Loose focus: WindowConfigure { new_size: (Some(960), Some(900)), 

### To Reproduce

1. Run WezTerm
2. Make it FullScreen/Maximized/Tiled
3. Switch workspace or Focus another application

### Configuration

local config = wezterm.config_builder()
config.initial_cols = 140
config.initial_rows = 50
config.enable_tab_bar = false
config.window_decorations = "NONE"
config.window_padding = { left = 0, right = 0, top = 0, bottom = 0 }

### Expected Behavior

Window size should be defined correctly.

### Logs

 » WAYLAND_DEBUG=1 WEZTERM_LOG=info ./target/debug/wezterm start                                                                                                    aliaksandr@t14s-fedora
[2652321.299]  -> wl_display#1.get_registry(new id wl_registry#2)
[2652321.353]  -> wl_display#1.sync(new id wl_callback#3)
[2652321.600] {Display Queue} wl_display#1.delete_id(3)
[2652321.646] wl_registry#2.global(1, "wl_compositor", 6)
[2652321.800] wl_registry#2.global(2, "wl_drm", 2)
[2652321.820] wl_registry#2.global(3, "wl_shm", 2)
[2652321.836] wl_registry#2.global(4, "wl_output", 4)
[2652321.858] wl_registry#2.global(5, "zxdg_output_manager_v1", 3)
[2652321.884] wl_registry#2.global(6, "wl_data_device_manager", 3)
[2652321.928] wl_registry#2.global(7, "zwp_primary_selection_device_manager_v1", 1)
[...]

PR #6281, opened 2024-10-15
use surface height/width for client side decoration frame resizes

dani162:
Fixes the problem with too big window frames when client side decorations are used on Wayland in combination with fractional scaling. See #5360 
However, there is still a problem when resizing the window, but I have not been able to find out why.

### DIFF
--- a/window/src/os/wayland/window.rs
+++ b/window/src/os/wayland/window.rs
@@ -845,11 +845,9 @@ impl WaylandWindowInner {
 
                 log::trace!("Resizing frame");
                 if !self.window_frame.is_hidden() {
-                    // Clamp the size to at least one pixel.
-                    let width =
-                        NonZeroU32::new(pixel_width as u32).unwrap_or(NonZeroU32::new(1).unwrap());
-                    let height =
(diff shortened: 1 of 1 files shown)


dani162:
Currently, while the window is resized in the Y axis, the window grows in the X direction and vice versa. 
Or if you tab out of the window and tab in, the window grows slightly in the X and Y direction each time.

ShadowRZ:
@dani162 Probably due to #6262 I guess?

wez:
I merged a couple of wayland changes recently, and I'm not sure how this interacts with them. Could you rebase and see if things are still working as expected?
Thanks!

dani162:
The changes work as expected, but the Problem mentioned above while resizing still remains
Before:
image
After:
image

wez:
I'm not sure that I follow: does this PR cause that remaining issue?  If not, then I think that we can merge this and track that other issue separately.

dani162:
The pr does not create the issue and can be merged in my opinion.

wez:
Thanks!  Does this fully resolve https://github.com/wezterm/wezterm/issues/5360 ?

dani162:
Yes i think so

PR #6286, opened 2024-10-16
update flake.lock to fix build failure

Neurarian:
Updating the flake inputs fixes the current upstream build failure due to a bug in [rust-overlay](https://github.com/oxalica/rust-overlay) introduced by a nixpkgs change in the rust build helper:

https://github.com/oxalica/rust-overlay/issues/191
https://github.com/oxalica/rust-overlay/pull/192 

- [x] wezterm builds successfully with updated flake inputs.

### DIFF
--- a/nix/flake.lock
+++ b/nix/flake.lock
@@ -5,11 +5,11 @@
         "systems": "systems"
       },
       "locked": {
-        "lastModified": 1710146030,
-        "narHash": "sha256-SZ5L6eA7HJ/nmkzGG7/ISclqe6oZdOZTNoesiInkXPQ=",
+        "lastModified": 1726560853,
+        "narHash": "sha256-X6rJYSESBVr3hBoH0WbKE5KvhPU5bloyZ2L4K60/fPQ=",
         "owner": "numtide",
         "repo": "flake-utils",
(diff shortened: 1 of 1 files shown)


cguentherTUChemnitz:
I don't understand, why this is closed, but not merged.

Should this be the fix for: `error: attribute 'targetPlatforms' missing` caused by https://github.com/oxalica/rust-overlay/issues/191 ?

cguentherTUChemnitz:
Ok i was able to "workaround" the missing rust-overlay update problem, by overriding the rust-overlay input to latest master for defining the flake import of wezterm as follows

```
    wezterm = {
      url = "github:wez/wezterm?dir=nix";
      inputs = {
        nixpkgs.follows = "nixpkgs";
        rust-overlay = {
          url = "github:oxalica/rust-overlay/master";
          inputs.nixpkgs.follows = "nixpkgs";
        };
      };
    };
```

Neurarian:
Hi, I closed this again because as you stated, it's quite easy to circumvent this. Furthermore, it also builds when not following nixpkgs for the wezterm input. I'd just like to avoid having multiple nixpkgs instances on my system, so your workaround is much better. I figured a simple flake update may not be worth the pull request and that I'd just wait for the bot to update the inputs. If this should be properly fixed asap, I could also reopen it of course.

Issue #6289, opened 2024-10-17
Add solution for East Asian Width Problem


hamano:
**Is your feature request related to a problem? Please describe.**

We would like to express our gratitude from the Far East for implementing the `treat_east_asian_ambiguous_width_as_wide` option in wezterm.
However, this option is a adhoc workaround and is inadequate.
- fullwidth for box drawing breaks the TUI screen.
- fullwidth for block elements breaks the progress bar.
- Inconsistent circled numbers width ⓪①..⑳㉑.
- Traditional Japanese charactor width does not match the Unicode recommended width.

**Describe the solution you'd like**

Regarding the EAW issue, better solutions:

1. Using wcswidth()
2. Set cellwidth for code points.

Since wezterm is a cross-platform application, using `wcswidth()` might not be appropriate.
I propose adding a cellwidth configuration option to wezterm.
This is a feature provided by mlterm, Emacs, and Vim, and it offers a practical solution to the ambiguous Unicode standard.
With this option, it is possible to unify character widths across terminal, shell, and text editors.

Based on vim's `setcellwidths()`, I propose the following syntax for the settings

```
config.cellwidths = {
  {0x2460, 0x2473, 2}, -- ①..⑳
  {0x24EA, 0x24EA, 2}, -- ⓪
  {0x2668, 0x2668, 2}, -- ♨
  {0xF113, 0xF113, 2}, -- 
}
```

However, it seems that keyword omission is not preferred in wezterm

```
config.cellwidths = {
  {first = 0x2460, last = 0x2473, width = 2}, -- ①..⑳
  {first = 0x24EA, last = 0x24EA, width = 2}, -- ⓪
  {first = 0x2668, last = 0x2668, width = 2}, -- ♨
  {first = 0xF113, last = 0xF113, width = 2}, -- 
}
```

I will handle the actual work and submit a pull request.
Additionally, I understand that there is a difference in perspective regarding the EAW issue between Latin users and CJK users.
If additional explanations are needed regarding the difficulties faced by the CJK users, I am more than willing to provide them.

hamano:
ref: https://github.com/wez/wezterm/issues/1888

NathanCummings:
Is the issue I am having in #6228 related to this do you think? Particularly look at my most recent comment where running `wezterm ls-fonts --list-system` shows many asian characters not being displayed properly.

hamano:
@NathanCummings #6228 is not EAW issue.
[...]

PR #6290, opened 2024-10-17
add cellwidths option #6289

hamano:
https://github.com/wez/wezterm/issues/6289

### DIFF
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -22,10 +22,10 @@ use crate::unix::UnixDomain;
 use crate::wsl::WslDomain;
 use crate::{
     default_config_with_overrides_applied, default_one_point_oh, default_one_point_oh_f64,
-    default_true, default_win32_acrylic_accent_color, GpuInfo, IntegratedTitleButtonColor,
-    KeyMapPreference, LoadedConfig, MouseEventTriggerMods, RgbaColor, SerialDomain, SystemBackdrop,
-    WebGpuPowerPreference, CONFIG_DIRS, CONFIG_FILE_OVERRIDE, CONFIG_OVERRIDES, CONFIG_SKIP,
-    HOME_DIR,
+    default_true, default_win32_acrylic_accent_color, CellWidth, GpuInfo,
+    IntegratedTitleButtonColor, KeyMapPreference, LoadedConfig, MouseEventTriggerMods, RgbaColor,
+    SerialDomain, SystemBackdrop, WebGpuPowerPreference, CONFIG_DIRS, CONFIG_FILE_OVERRIDE,
+    CONFIG_OVERRIDES, CONFIG_SKIP, HOME_DIR,
 };
 use anyhow::Context;
 use luahelper::impl_lua_conversion_dynamic;
@@ -835,6 +835,9 @@ pub struct Config {
 
     #[dynamic(default)]
     pub treat_east_asian_ambiguous_width_as_wide: bool,
+
+    #[dynamic(default)]
+    pub cell_widths: Option<Vec<CellWidth>>,
 
     #[dynamic(default = "default_true")]
     pub allow_download_protocols: bool,
--- a/config/src/terminal.rs
+++ b/config/src/terminal.rs
@@ -102,10 +102,7 @@ impl wezterm_term::TerminalConfiguration for TermConfig {
 
     fn unicode_version(&self) -> UnicodeVersion {
         let config = self.configuration();
-        UnicodeVersion {
-            version: config.unicode_version,
-            ambiguous_are_wide: config.treat_east_asian_ambiguous_width_as_wide,
-        }
+        config.unicode_version()
     }
 
     fn debug_key_events(&self) -> bool {
--- a/term/src/terminalstate/performer.rs
+++ b/term/src/terminalstate/performer.rs
@@ -134,7 +134,7 @@ impl<'a> Performer<'a> {
         for g in Graphemes::new(text) {
             let g = self.remap_grapheme(g);
 
-            let mut print_width = grapheme_column_width(g, Some(self.unicode_version));
+            let mut print_width = grapheme_column_width(g, Some(&self.unicode_version));
             if print_width == 0 {
                 // We got a zero-width grapheme.
 
@@ -836,7 +836,7 @@ impl<'a> Performer<'a> {
                     self.unicode_version.version = n;
                 }
                 ITermProprietary::UnicodeVersion(ITermUnicodeVersionOp::Push(label)) => {
-                    let vers = self.unicode_version;
+                    let vers = self.unicode_version.clone();
                     self.unicode_version_stack
                         .push(UnicodeVersionStackEntry { vers, label });
                 }
--- a/termwiz/src/cell.rs
+++ b/termwiz/src/cell.rs
@@ -8,6 +8,7 @@ use crate::widechar_width::WcWidth;
 use finl_unicode::grapheme_clusters::Graphemes;
 #[cfg(feature = "use_serde")]
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use std::collections::HashMap;
 use std::hash::{Hash, Hasher};
 use std::mem;
 use std::sync::Arc;
@@ -652,7 +653,7 @@ impl TeenyString {
     pub fn from_str(
         s: &str,
         width: Option<usize>,
-        unicode_version: Option<UnicodeVersion>,
+        unicode_version: Option<&UnicodeVersion>,
     ) -> Self {
         // De-fang the input text such that it has no special meaning
         // to a terminal.  All control and movement characters are rewritten
@@ -848,7 +849,7 @@ impl Cell {
     pub fn new_grapheme(
         text: &str,
         attrs: CellAttributes,
-        unicode_version: Option<UnicodeVersion>,
+        unicode_version: Option<&UnicodeVersion>,
     ) -> Self {
         let storage = TeenyString::from_str(text, None, unicode_version);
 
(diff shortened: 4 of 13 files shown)


term/src/terminalstate/performer.rs
```diff
@@ -132,7 +132,7 @@ impl<'a> Performer<'a> {
         for g in Graphemes::new(text) {
             let g = self.remap_grapheme(g);
 
-            let print_width = grapheme_column_width(g, Some(self.unicode_version));
+            let print_width = grapheme_column_width(g, Some(self.unicode_version.clone()));
```
wez, line 135: re: performance anxiety, this is a very hot loop, especially for folks doing `cat bigfile`.

I think we should be passing `Some(&self.unicode_version)` here to avoid any additional overhead.

There will be more API fan-out, but I think it is worth it to keep the performance reasonable.

The termwiz Cargo.toml version number would also need to be bumped because it would be a breaking API change, even though it is a minor one.

termwiz/src/cell.rs
```diff
@@ -881,17 +882,40 @@ impl Cell {
     }
 }
 
-#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+#[derive(Clone, Debug, Eq, PartialEq, FromDynamic, ToDynamic)]
+pub struct CellWidth {
+    pub first: u32,
+    pub last: u32,
+    pub width: u8,
+}
```
wez, line 890: From a layering perspective, I think this struct should move to the `config` crate, and that the `setcellwidths` function should move along with it.  I would probably also move `setcellwidths` to be an associated function of `CellWidth`

```rust
impl CellWidth {
   fn compile_to_map(widths: Option<Vec<Self>>) -> Option<Arc<HashMap<u32,u8>>> {
     ...
   }
}
```

termwiz/src/cell.rs
```diff
@@ -881,17 +882,40 @@ impl Cell {
...
+        let mut map: HashMap<u32, u8> = HashMap::new();
+        for cellwidth in cellwidths {
+            for i in cellwidth.first..cellwidth.last+1 {
+                map.insert(i, cellwidth.width);
+            }
+        }
+        return Some(map);
+    } else {
+        return None;
+    }
```
wez, line 903: FWIW, you can simply this, something like:

```suggestion
    let cellwidths = cellwidths.as_ref()?;
    let mut map = HashMap::new();
    for cellwidth in cellwidths {
        for i in cellwidth.first..=cellwidth.last {
            map.insert(i, cellwidth.width);
        }
    }
    Some(map)
```

wezterm-gui/src/main.rs
```diff
@@ -886,6 +887,7 @@ pub fn run_ls_fonts(config: config::ConfigHandle, cmd: &LsFontsCommand) -> anyho
     let unicode_version = UnicodeVersion {
         version: config.unicode_version,
         ambiguous_are_wide: config.treat_east_asian_ambiguous_width_as_wide,
+        cellwidths: setcellwidths(config.cellwidths.clone()),
```
wez, line 890: I think it's worth adding a helper method to the `Config` class that encapsulates this:

in `config/src/config.rs`:

```rust
impl Config {
   pub fn unicode_verson(&self) -> UnicodeVersion {
      ...
   }
}
```

then the logic in this file is simply:

```rust
let unicode_version = config.unicode_version();
```

config/src/terminal.rs
```diff
@@ -105,6 +106,7 @@ impl wezterm_term::TerminalConfiguration for TermConfig {
         UnicodeVersion {
             version: config.unicode_version,
             ambiguous_are_wide: config.treat_east_asian_ambiguous_width_as_wide,
+            cellwidths: setcellwidths(config.cellwidths.clone()),
```
wez, line 109: per my comment in `main.rs`, I think this method could simplify down to:

```rust
 fn unicode_version(&self) -> UnicodeVersion {
      let config = self.configuration();
      config.unicode_version()
}
```

termwiz/src/surface/line/line.rs
```diff
@@ -138,7 +138,7 @@ impl Line {
         let mut cells = Vec::new();
 
         for sub in Graphemes::new(s) {
-            let cell = Cell::new_grapheme(sub, attrs.clone(), unicode_version);
+            let cell = Cell::new_grapheme(sub, attrs.clone(), unicode_version.clone());
```
wez, line 141: This is also a hot code path, should probably pass `&unicode_version` here as well

termwiz/src/cell.rs
```diff
@@ -928,7 +963,7 @@ pub const LATEST_UNICODE_VERSION: UnicodeVersion = UnicodeVersion {
 /// and sums up the length.
 pub fn unicode_column_width(s: &str, version: Option<UnicodeVersion>) -> usize {
     Graphemes::new(s)
-        .map(|g| grapheme_column_width(g, version))
+        .map(|g| grapheme_column_width(g, version.clone()))
```
wez, line 966: let's pass the version by reference here too

wez (changes requested):
Thanks for this!
It makes sense to me, and the approach seems reasonable and easy to follow.
I have a couple of stylistic suggestions, but the main point of feedback is that we need to adjust this a bit to avoid excessive copying of the map in the hot code paths.

wezterm-gui/src/main.rs
hamano: added unicode_version() in struct ConfigHandle

hamano:
Thank you for reviewing.
I have fixed the issues you pointed out and rebased them into a single commit.

wez:
Hi! There are some issues with the tests, could you fix those and re-submit?

config/src/config.rs
```diff
@@ -824,6 +825,9 @@ pub struct Config {
     #[dynamic(default)]
     pub treat_east_asian_ambiguous_width_as_wide: bool,
 
+    #[dynamic(default)]
+    pub cellwidths: Option<Vec<CellWidth>>,
```
wez, line 829: Let's use an underscore here to separate the two words

```suggestion
    pub cell_widths: Option<Vec<CellWidth>>,
```

termwiz/src/cell.rs
```diff
@@ -967,8 +981,8 @@ pub fn unicode_column_width(s: &str, version: Option<UnicodeVersion>) -> usize {
 /// The terminal emulator can then pass the unicode version through to
 /// the Cell that is used to hold a grapheme, and that per-Cell version
 /// can then be used to calculate width.
-pub fn grapheme_column_width(s: &str, version: Option<UnicodeVersion>) -> usize {
-    let version = version.as_ref().unwrap_or(&LATEST_UNICODE_VERSION);
+pub fn grapheme_column_width(s: &str, version: Option<&UnicodeVersion>) -> usize {
+    let version = version.as_ref().unwrap_or(&&LATEST_UNICODE_VERSION);
```
wez, line 985:
```suggestion
    let version = version.as_deref().unwrap_or(&LATEST_UNICODE_VERSION);
```

wez (commented):
What we need now is a page for the docs; please see https://github.com/wezterm/wezterm/blob/main/CONTRIBUTING.md#contributing-documentation for information on how to run the doc build.

For this PR, you should add a file alongside `docs/config/lua/config/treat_east_asian_ambiguous_width_as_wide.md` for the new option: `docs/config/lua/config/cell_widths.md`.

For the `since()` version part, use the string `nightly` as the version.

Please include an example of using the new option in that page.

Thanks!

ProfessorMOB:
This might sound like a insane idea, but what if we exposed the harfbuzz API directly through lua in some way?

If we exposed the harfbuzz API directly through Lua, then we could give the user an absurd amount of control over what their characters should look like. I don't think I'll be able to flesh this idea any further do to my own lack of expertise, but I did want to throw it out there. 

Nonetheless, I'm onboard with hamano's approach.

hamano:
@wez Could you please review the doc 0b3d1e1

wez (approved):

hamano:
rebased

hamano:
@wez  Every time you edit Cargo.lock, a conflict occurs. Should I remove the Cargo.lock changes from the PR?

ProfessorMOB:
The option works now, but characters like ﷽ still aren't displayed properly, its now taking two spaces instead of one. 

Another problem I ran into was that it was VERY buggy. Here's a video of what I mean. 

https://github.com/user-attachments/assets/de433efc-849c-46b4-a9fc-5ce608ad3a2a

wez:
if your shell, or other application that you are using, doesn't agree with the terminal on the width, then this is exactly the sort of behavior that you can expect. This is mentioned in the documentation.
https://wezterm.org/config/lua/config/cell_widths.html

ProfessorMOB:
I pointed out two problems in a way that was very unclear, and I apologize. You explained the one of the problems I ran into (where its bugging out and I can't move beyond the character), but there's another problem I'm facing. I had set the cell width to be 5 cells, but it's only displaying the character in the space of 2 cells. Is that also because of the same problem (cellwidths of shell and terminal don't agree)?

And also, how would one force the shell to agree with the cellwidth specified? I read the docs its because of the glibc locale, would I have to change the locale for me to fix that?

hamano:
The cell_widths option is intended only for widths of 1 or 2.
Even if support for widths of 3 or more were supported, it wouldn't be useful since both Vim and the glibc locale only handle widths of 1 or 2. Supporting terminals alone would not be practical.

For customizing the locale, we use this repository.
https://github.com/hamano/locale-eaw
Apologies that the README is in Japanese.
Add the following to config.ini:
```
[CUSTOM]
U+FDFD = 2
```

Then run `./gen.py`, and `UTF-8-CUSTOM.gz` will be generated.
Linux users can use this to customize character widths in Bash and Zsh.

hamano:
As for macOS, I’ve heard that modifying the locale is difficult due to security protections.
I’d like to know more as well, so please let me know if you find a way to do it.

ProfessorMOB:
Personally, no need to apologize to me about the Readme being in Japanese, I just wish technology wasn't so anglicised. 

As for MacOS, I found this (https://apple.stackexchange.com/questions/384388/how-to-add-a-new-locale-to-macos-catalina) from stackoverflow, I don't have experience with Linux locale or locale in general. I found this with a quick google search. Dunno if it applies to later versions of MacOS though.

PR #6292, opened 2024-10-18
Add support for OSC 22 control sequence to change mouse shape

reykjalin:
## Description

> [!NOTE]
> Opening this as a draft PR for now because there are some kinks I need to work through and I'm not sure how. I'll describe the problems in a review on this PR. I'm also not at all sure if the approach here is correct.

* Adds handling for the [OSC 22](https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h4-Operating-System-Commands:OSC-Ps;Pt-ST:Ps-=-2-2.1018) control sequence.
* Follows [the example set by Kitty](https://github.com/kovidgoyal/kitty/blob/2efa394dea2ec72aa60d6f94bc9ac241c3cc0037/gen/cursors.py#L15-L57) to support [CSS cursor](https://developer.mozilla.org/en-US/docs/Web/CSS/cursor) values for the OSC 22 parameter.
* Currently only supports `pointer`, `text`, `row-resize`, `ns-resize`, `col-resize`, and `ew-resize`, based on the currently available `MouseCursor` options in the codebase. Any other value defaults to `MouseCursor::Arrow`.

### Videos

Running [comlink](https://github.com/rockorager/comlink/tree/main) (terminal IRC client) before and after:

| Before (9ddca7bde92090792dbcdc65c1e9897c362196d7) | After |
|:---:|:---:|
| <video src="https://github.com/user-attachments/assets/58af48cc-c5a0-4ed4-a53b-ff6b18f2b7e8"> | <video src="https://github.com/user-attachments/assets/bb3237d1-1361-44ea-9cc6-a3f0ed5f7829"> |

## Testing

This requires a terminal program that makes use of OSC 22 codes to change the mouse cursor. The only one I know of and actively use is comlink, but I'm sure there are other TUIs out there that do use the codes, I'm just not aware of them.

Run a terminal program that uses OSC 22 codes and check if they work.

I'll try to make a terminal program to demonstrate at least some of them as I work on this PR. It should make testing easier.

### DIFF
--- a/mux/src/pane.rs
+++ b/mux/src/pane.rs
@@ -553,6 +553,10 @@ mod test {
     }
 
     impl Pane for FakePane {
+        fn get_mouse_cursor_shape(&self) -> MouseCursor {
+            unimplemented!()
+        }
+
         fn pane_id(&self) -> PaneId {
             unimplemented!()
         }
--- a/mux/src/pane.rs
+++ b/mux/src/pane.rs
@@ -16,6 +16,7 @@ use termwiz::surface::{Line, SequenceNo};
 use url::Url;
 use wezterm_dynamic::Value;
 use wezterm_term::color::ColorPalette;
+pub use wezterm_term::MouseCursor;
 use wezterm_term::{
     Clipboard, DownloadHandler, KeyCode, KeyModifiers, MouseEvent, SemanticZone, StableRowIndex,
     TerminalConfiguration, TerminalSize,
@@ -230,6 +231,8 @@ pub trait Pane: Downcast + Send + Sync {
 
     /// Returns render related dimensions
     fn get_dimensions(&self) -> RenderableDimensions;
+
+    fn get_mouse_cursor_shape(&self) -> MouseCursor;
 
     fn get_title(&self) -> String;
     fn send_paste(&self, text: &str) -> anyhow::Result<()>;
--- a/term/src/terminal.rs
+++ b/term/src/terminal.rs
@@ -51,6 +51,8 @@ pub enum Alert {
     TabTitleChanged(Option<String>),
     /// When the color palette has been updated
     PaletteChanged,
+    /// When the mouse cursor shape has been updated
+    MouseCursorShapeChanged,
     /// A UserVar has changed value
     SetUserVar {
         name: String,
--- a/term/src/terminalstate/mod.rs
+++ b/term/src/terminalstate/mod.rs
@@ -62,6 +62,15 @@ pub(crate) enum MouseEncoding {
     SgrPixels,
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum MouseCursor {
+    Arrow,
+    Hand,
+    Text,
+    SizeUpDown,
+    SizeLeftRight,
+}
+
 impl TabStop {
     fn new(screen_width: usize, tab_width: usize) -> Self {
         let mut tabs = Vec::with_capacity(screen_width);
@@ -342,6 +351,9 @@ pub struct TerminalState {
 
     palette: Option<ColorPalette>,
 
+    /// The mouse cursor shape (OSC 22)
+    mouse_cursor_shape: String,
+
     pixel_width: usize,
     pixel_height: usize,
     dpi: u32,
@@ -547,6 +559,7 @@ impl TerminalState {
             any_event_mouse: false,
             button_event_mouse: false,
             mouse_tracking: false,
+            mouse_cursor_shape: String::from(""),
             last_mouse_move: None,
             cursor_visible: true,
             g0_charset: CharSet::Ascii,
@@ -663,6 +676,26 @@ impl TerminalState {
             .as_ref()
             .cloned()
             .unwrap_or_else(|| self.config.color_palette())
+    }
+
+    /// Returns a copy of the current mouse cursor shape.
+    pub fn get_mouse_cursor_shape(&self) -> MouseCursor {
+        // FIXME: The log here is clearing a cache somewhere causing the mouse shape
+        //        to be updated correctly. Not sure what I need to change.
+        log::info!("Mouse cursor shape: {}", self.mouse_cursor_shape);
+        if self.mouse_cursor_shape == "pointer" {
+            MouseCursor::Hand
+        } else if self.mouse_cursor_shape == "text" {
+            MouseCursor::Text
+        } else if self.mouse_cursor_shape == "row-resize" || self.mouse_cursor_shape == "ns-resize"
+        {
+            MouseCursor::SizeUpDown
+        } else if self.mouse_cursor_shape == "col-resize" || self.mouse_cursor_shape == "ew-resize"
+        {
+            MouseCursor::SizeLeftRight
+        } else {
+            MouseCursor::Arrow
+        }
     }
 
     /// Called in response to dynamic color scheme escape sequences.
@@ -924,6 +957,12 @@ impl TerminalState {
         self.make_all_lines_dirty();
         if let Some(handler) = self.alert_handler.as_mut() {
             handler.alert(Alert::PaletteChanged);
+        }
+    }
+
+    fn mouse_cursor_shape_did_change(&mut self) {
+        if let Some(handler) = self.alert_handler.as_mut() {
+            handler.alert(Alert::MouseCursorShapeChanged);
         }
     }
 
--- a/term/src/terminalstate/performer.rs
+++ b/term/src/terminalstate/performer.rs
@@ -678,6 +678,7 @@ impl<'a> Performer<'a> {
                 self.focus_tracking = false;
                 self.mouse_tracking = false;
                 self.mouse_encoding = MouseEncoding::X10;
+                self.mouse_cursor_shape = String::from("");
                 self.keyboard_encoding = KeyboardEncoding::Xterm;
                 self.sixel_scrolls_right = false;
                 self.any_event_mouse = false;
@@ -704,6 +705,7 @@ impl<'a> Performer<'a> {
                 self.erase_in_display(EraseInDisplay::EraseScrollback);
                 self.erase_in_display(EraseInDisplay::EraseDisplay);
                 self.palette_did_change();
+                self.mouse_cursor_shape_did_change();
             }
 
             _ => {
@@ -718,6 +720,10 @@ impl<'a> Performer<'a> {
         self.pop_tmux_title_state();
         self.flush_print();
         match osc {
+            OperatingSystemCommand::SetMouseCursorShape(mouse_shape) => {
+                self.mouse_cursor_shape = mouse_shape;
+                self.mouse_cursor_shape_did_change();
+            }
             OperatingSystemCommand::SetIconNameSun(title)
             | OperatingSystemCommand::SetIconName(title) => {
                 if title.is_empty() {
--- a/wezterm-client/src/pane/clientpane.rs
+++ b/wezterm-client/src/pane/clientpane.rs
@@ -27,7 +27,7 @@ use url::Url;
 use wezterm_dynamic::Value;
 use wezterm_term::color::ColorPalette;
 use wezterm_term::{
-    Alert, Clipboard, KeyCode, KeyModifiers, Line, MouseEvent, StableRowIndex,
+    Alert, Clipboard, KeyCode, KeyModifiers, Line, MouseCursor, MouseEvent, StableRowIndex,
     TerminalConfiguration, TerminalSize,
 };
 
@@ -245,6 +245,10 @@ impl ClientPane {
 
 #[async_trait(?Send)]
 impl Pane for ClientPane {
+    fn get_mouse_cursor_shape(&self) -> MouseCursor {
+        MouseCursor::Arrow
+    }
+
     fn pane_id(&self) -> PaneId {
         self.local_pane_id
     }
--- a/wezterm-gui/src/overlay/copy.rs
+++ b/wezterm-gui/src/overlay/copy.rs
@@ -27,8 +27,8 @@ use unicode_segmentation::*;
 use url::Url;
 use wezterm_term::color::ColorPalette;
 use wezterm_term::{
-    unicode_column_width, Clipboard, KeyCode, KeyModifiers, Line, MouseEvent, SemanticType,
-    StableRowIndex, TerminalSize,
+    unicode_column_width, Clipboard, KeyCode, KeyModifiers, Line, MouseCursor, MouseEvent,
+    SemanticType, StableRowIndex, TerminalSize,
 };
 use window::{KeyCode as WKeyCode, Modifiers, WindowOps};
 
@@ -1101,6 +1101,10 @@ impl CopyRenderable {
 }
 
 impl Pane for CopyOverlay {
+    fn get_mouse_cursor_shape(&self) -> MouseCursor {
+        MouseCursor::Arrow
+    }
+
     fn pane_id(&self) -> PaneId {
         self.delegate.pane_id()
     }
--- a/wezterm-gui/src/overlay/quickselect.rs
+++ b/wezterm-gui/src/overlay/quickselect.rs
@@ -19,7 +19,7 @@ use termwiz::surface::{SequenceNo, SEQ_ZERO};
 use url::Url;
 use wezterm_term::color::ColorPalette;
 use wezterm_term::{
-    Clipboard, KeyCode, KeyModifiers, Line, MouseEvent, StableRowIndex, TerminalSize,
+    Clipboard, KeyCode, KeyModifiers, Line, MouseCursor, MouseEvent, StableRowIndex, TerminalSize,
 };
 use window::WindowOps;
 
@@ -317,6 +317,10 @@ impl QuickSelectOverlay {
 }
 
 impl Pane for QuickSelectOverlay {
+    fn get_mouse_cursor_shape(&self) -> MouseCursor {
+        MouseCursor::Arrow
+    }
+
     fn pane_id(&self) -> PaneId {
         self.delegate.pane_id()
     }
--- a/wezterm-gui/src/termwindow/mod.rs
+++ b/wezterm-gui/src/termwindow/mod.rs
@@ -1210,6 +1210,13 @@ impl TermWindow {
                     self.update_title();
                 }
                 MuxNotification::Alert {
+                    alert: Alert::MouseCursorShapeChanged,
+                    pane_id,
+                } => {
+                    // FIXME: Is there a cache that needs to be invalidated like for PaletteChanged?
+                    self.mux_pane_output_event(pane_id);
+                }
+                MuxNotification::Alert {
                     alert: Alert::PaletteChanged,
                     pane_id,
                 } => {
@@ -1515,6 +1522,12 @@ impl TermWindow {
             | MuxNotification::WindowWorkspaceChanged(_) => return true,
             MuxNotification::Alert {
                 alert: Alert::PaletteChanged { .. },
+                ..
+            } => {
+                // fall through
+            }
+            MuxNotification::Alert {
+                alert: Alert::MouseCursorShapeChanged { .. },
                 ..
             } => {
                 // fall through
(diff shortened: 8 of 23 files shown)


term/src/terminalstate/mod.rs
```diff
@@ -62,6 +62,15 @@ pub(crate) enum MouseEncoding {
...
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum MouseCursor {
+    Arrow,
+    Hand,
+    Text,
+    SizeUpDown,
+    SizeLeftRight,
+}
```
reykjalin, line 72: I moved this struct because I figured the terminal state getter was a good place to convert the string from the control sequence to a type understood by the current codebase (see `get_mouse_cursor_shape` in this file).

Not sure if this is the right place though, especially considering that means it needs to be referenced much higher up the stack, and I needed to add a new `wezterm-term` reference to the `window/Cargo.toml` file.

term/src/terminalstate/mod.rs
```diff
@@ -665,6 +678,26 @@ impl TerminalState {
             .unwrap_or_else(|| self.config.color_palette())
     }
 
+    /// Returns a copy of the current mouse cursor shape.
+    pub fn get_mouse_cursor_shape(&self) -> MouseCursor {
+        // FIXME: The log here is clearing a cache somewhere causing the mouse shape
+        //        to be updated correctly. Not sure what I need to change.
+        log::info!("Mouse cursor shape: {}", self.mouse_cursor_shape);
```
reykjalin, line 685: I noticed this sort of by accident. I was trying to figure out why the cursor shape wouldn't change after closing an app that supports OSC 22 codes:

https://github.com/user-attachments/assets/8cb5310b-6bb5-462a-9fe4-c57283dc787d

I wanted to see if the cursor shape wasn't being updated, but it turns out it is! And logging the value clears some cache I'm not aware of that causes the cursor to be drawn correctly up the call stack.

I don't know what I'm missing, so I'd appreciate some direction on how this could be addressed 🙏

wezterm-client/src/pane/clientpane.rs
```diff
@@ -245,6 +245,10 @@ impl ClientPane {
 
 #[async_trait(?Send)]
 impl Pane for ClientPane {
+    fn get_mouse_cursor_shape(&self) -> MouseCursor {
+        MouseCursor::Arrow
+    }
```
reykjalin, line 250: Just a default implementation for now, needs a proper implementation.

wezterm-gui/src/overlay/copy.rs
```diff
@@ -1101,6 +1101,10 @@ impl CopyRenderable {
 }
 
 impl Pane for CopyOverlay {
+    fn get_mouse_cursor_shape(&self) -> MouseCursor {
+        MouseCursor::Arrow
+    }
```
reykjalin, line 1106: Just a default implementation for now, needs a proper implementation.

wezterm-gui/src/overlay/quickselect.rs
```diff
@@ -317,6 +317,10 @@ impl QuickSelectOverlay {
 }
 
 impl Pane for QuickSelectOverlay {
+    fn get_mouse_cursor_shape(&self) -> MouseCursor {
+        MouseCursor::Arrow
+    }
```
reykjalin, line 322: Just a default implementation for now, needs a proper implementation.

wezterm-gui/src/termwindow/mod.rs
```diff
@@ -1519,6 +1526,12 @@ impl TermWindow {
             } => {
                 // fall through
             }
+            MuxNotification::Alert {
+                alert: Alert::MouseCursorShapeChanged { .. },
+                ..
+            } => {
+                // fall through
+            }
```
reykjalin, line 1534: I copied what's done for `Alert::PaletteChanged` here, not sure if that's correct.

wezterm-gui/src/termwindow/mouseevent.rs
```diff
@@ -843,6 +843,10 @@ impl super::TermWindow {
             MouseCursor::Text
         }));
 
+        if let Some(pane) = self.get_active_pane_or_overlay() {
+            context.set_cursor(Some(pane.get_mouse_cursor_shape()));
+        }
```
reykjalin, line 848: Not sure if this is the right way to get the correct pane to get the mouse cursor shape.

I think I also need to correctly include the hyperlink hover code in here, but figured I'd keep it simple for now.

As an additional note, I only added handling for OSC 22 in the `mouse_event_terminal` handler since it's the only one that seemed relevant. The OSC 22 control sequence should be emitted by a running terminal app and the other events seem related to native WezTerm events?

Do any of the other event handlers need to take into account the OSC 22 control sequence?

wezterm-gui/src/termwindow/mod.rs
```diff
@@ -1209,6 +1209,13 @@ impl TermWindow {
                 } => {
                     self.update_title();
                 }
+                MuxNotification::Alert {
+                    alert: Alert::MouseCursorShapeChanged,
+                    pane_id,
+                } => {
+                    // FIXME: Is there a cache that needs to be invalidated like for PaletteChanged?
```
wez, line 1216: There isn't an explicit cache, but you will need to cause the logic in the normal mouse event processing flow to trigger and assess what the cursor should be here, based on its position.

I don't have a great suggestion for this off the top of my head. Most likely will need to add a helper function to figure out if the mouse cursor is over the terminal area and if so, get the shape from the pane identified by `pane_id` and apply it to the window.

wezterm-gui/src/termwindow/mouseevent.rs
wez: GH won't let me suggest into the text above your change, so I'm sketching it out here. What I think you want is something like this:

```rust
        context.set_cursor(Some(if self.current_highlight.is_some() {
            // When hovering over a hyperlink, show an appropriate
            // mouse cursor to give the cue that it is clickable
            MouseCursor::Hand
        } else if outside_window {
            MouseCursor::Arrow
        } else if let Some(shape) = pane.get_mouse_cursor_shape() {
             // Apply the mapping here. This mapping logic should
             // be factored out into a helper function that you call here
              match shape.as_str() {
               "text" => MouseCursor::Text,
               ..
               _ => MouseCursor::Text,
            } 
        } else if pane.is_mouse_grabbed() {
            MouseCursor::Arrow
        } else {
            MouseCursor::Text
        }));
```

window/Cargo.toml
```diff
@@ -46,6 +46,7 @@ wezterm-bidi = { path = "../bidi" }
 wezterm-color-types = { path = "../color-types" }
 wezterm-font = { path = "../wezterm-font" }
 wezterm-input-types = { path = "../wezterm-input-types" }
+wezterm-term = { path = "../term", features=["use_serde"] }
```
wez, line 49: as suggested elsewhere, I think the mapping can be done in the wezterm-gui layer and that means that you can undo these changes to the window crate; in theory the window crate can be used to create non-terminal programs, so it is desirable to avoid pulling in a terminal emulator core from wezterm-term for that

wez (commented):
I'm surprised that you got to this so soon!
A few suggestions below!

wezterm-gui/src/termwindow/mouseevent.rs
reykjalin: Addressed in https://github.com/wez/wezterm/pull/6292/commits/b67cf84b159515d2789c946019763bb3dca61ed6 and https://github.com/wez/wezterm/pull/6292/commits/683d1348a35a558dc781faa135a4de71f8cea49c

mux/src/pane.rs
```diff
@@ -231,6 +232,8 @@ pub trait Pane: Downcast + Send + Sync {
     /// Returns render related dimensions
     fn get_dimensions(&self) -> RenderableDimensions;
 
+    fn get_mouse_cursor_shape(&self) -> MouseCursor;
```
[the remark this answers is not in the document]
reykjalin: Done in https://github.com/wez/wezterm/pull/6292/commits/b67cf84b159515d2789c946019763bb3dca61ed6

term/src/terminalstate/mod.rs
```diff
@@ -342,6 +351,9 @@ pub struct TerminalState {
 
     palette: Option<ColorPalette>,
 
+    /// The mouse cursor shape (OSC 22)
+    mouse_cursor_shape: String,
```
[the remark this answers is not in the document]
reykjalin: Done in https://github.com/wez/wezterm/pull/6292/commits/b67cf84b159515d2789c946019763bb3dca61ed6

term/src/terminalstate/mod.rs
```diff
@@ -665,6 +678,26 @@ impl TerminalState {
             .unwrap_or_else(|| self.config.color_palette())
     }
 
+    /// Returns a copy of the current mouse cursor shape.
+    pub fn get_mouse_cursor_shape(&self) -> MouseCursor {
```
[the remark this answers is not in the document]
reykjalin: Thanks for the detailed explanation! I didn't have any of that context :)

Done in https://github.com/wez/wezterm/pull/6292/commits/b67cf84b159515d2789c946019763bb3dca61ed6!

term/src/terminalstate/mod.rs
```diff
@@ -665,6 +678,26 @@ impl TerminalState {
...
+        } else if self.mouse_cursor_shape == "row-resize" || self.mouse_cursor_shape == "ns-resize"
+        {
+            MouseCursor::SizeUpDown
+        } else if self.mouse_cursor_shape == "col-resize" || self.mouse_cursor_shape == "ew-resize"
+        {
+            MouseCursor::SizeLeftRight
+        } else {
+            MouseCursor::Arrow
+        }
+    }
```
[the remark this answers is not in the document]
reykjalin: Makes sense! Moved to the `wezterm-gui` crate as part of https://github.com/wez/wezterm/pull/6292/commits/b67cf84b159515d2789c946019763bb3dca61ed6 and https://github.com/wez/wezterm/pull/6292/commits/683d1348a35a558dc781faa135a4de71f8cea49c

term/src/terminalstate/performer.rs
```diff
@@ -678,6 +678,7 @@ impl<'a> Performer<'a> {
                 self.focus_tracking = false;
                 self.mouse_tracking = false;
                 self.mouse_encoding = MouseEncoding::X10;
+                self.mouse_cursor_shape = String::from("");
```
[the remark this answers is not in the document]
reykjalin: Done in https://github.com/wez/wezterm/pull/6292/commits/b67cf84b159515d2789c946019763bb3dca61ed6

window/Cargo.toml
reykjalin: Yes, fixed in https://github.com/wez/wezterm/pull/6292/commits/b67cf84b159515d2789c946019763bb3dca61ed6 and https://github.com/wez/wezterm/pull/6292/commits/683d1348a35a558dc781faa135a4de71f8cea49c!

reykjalin:
> I'm surprised that you got to this so soon!

You and me both 😅 

I think I've addressed most of the feedback, but I'm still seeing the issue where it's like the mouse shape is cached. I'll dig into that some more next week :)

wezterm-gui/src/termwindow/mod.rs
```diff
@@ -3539,6 +3540,16 @@ impl TermWindow {
...
+
+    fn parse_mouse_cursor_shape(&self, mouse_cursor_shape: &str) -> MouseCursor {
+        match mouse_cursor_shape {
+            "pointer" => MouseCursor::Hand,
+            "text" => MouseCursor::Text,
+            "row-resize" | "ns-resize" => MouseCursor::SizeUpDown,
+            "col-resize" | "ew-resize" => MouseCursor::SizeLeftRight,
+            _ => MouseCursor::Arrow,
+        }
+    }
```
wez, line 3552: Since this doesn't do anything with self, it can just be a "free" function

```suggestion

}

pub fn parse_mouse_cursor_shape(mouse_cursor_shape: &str) -> MouseCursor {
        match mouse_cursor_shape {
            "pointer" => MouseCursor::Hand,
            "text" => MouseCursor::Text,
            "row-resize" | "ns-resize" => MouseCursor::SizeUpDown,
            "col-resize" | "ew-resize" => MouseCursor::SizeLeftRight,
            _ => MouseCursor::Arrow,
        }
}
```

wezterm-gui/src/termwindow/mouseevent.rs
```diff
@@ -839,14 +839,12 @@ impl super::TermWindow {
             MouseCursor::Hand
         } else if pane.is_mouse_grabbed() || outside_window {
```
wez, line 840: FWIW, I think you you will want to allow the shape to override in the grabbed-mouse case. That is the most likely situation where this escape sequence will be used.

reykjalin: Ah right, I missed that in the initial recommendation, thank you for catching that!
Fixed in 11cf8b68ba43689f11dd27a26101b9bee552e609

wezterm-gui/src/termwindow/mod.rs
reykjalin: I wasn't sure what the preferred approach here was, so thank you for this suggestion! Done in 0cd76fc302db43bc14396130f2fe3b2aec82e500

reykjalin:
I think I figured out the reason the mouse cursor shape gets "stuck" on whatever was last set through the OSC 22 sequence. It's because the TUI doesn't set the OSC22 code back to the default, or doesn't clear it I guess? So once the TUI is closed whichever pointer was active is still stored in the `mouse_cursor_shape` state, so it keeps getting applied over everything else.

I'm not sure if that's "correct" behavior or not. I also don't know what the right time to clear the OSC 22 setting would be, if that even is the correct approach here. The xterm docs don't seem to specify that, at least not in relation to what the control sequence does.

I'll keep looking into this over the next couple of days.

wezterm-gui/src/termwindow/mouseevent.rs
```diff
@@ -837,7 +838,17 @@ impl super::TermWindow {
...
+            MouseCursor::Arrow
+        } else if let Some(shape) = pane.get_mouse_cursor_shape() {
+            if pane.is_mouse_grabbed() {
+                parse_mouse_cursor_shape(shape.as_str())
+            } else {
+                // If the mouse is no longer grabbed by a TUI, reset the cursor shape.
+                pane.clear_mouse_cursor_shape();
+                MouseCursor::Arrow
+            }
+        } else if pane.is_mouse_grabbed() {
```
reykjalin, line 851: As it turns out, this fixed the issue of the mouse cursor shape being wrong. I have absolutely no idea if this is the right way to approach it though.

But basically; if the mouse is grabbed by whichever application is running, we defer to whatever mouse shape is set by any OSC 22 escape sequences.

If we find that a shape has been set by an OSC 22 escape sequence _but the mouse isn't grabbed_ we clear it and just use the default pointer instead.

Some thoughts:

1. I'm realizing that using `mouse_cursor_shape` for the names of this might be confusing? Should we specify that this is a shape set through an OSC 22 escape sequence somehow?
    * It's more confusing in the GUI layer, less so in the terminal state itself.
    * But it may be useful to use something like `osc_22_mouse_cursor_shape` instead of just `mouse_cursor_shape`?
2. I'm not sure this is the best place to clear the mouse cursor shape. Maybe there's an event sent when the mouse is grabbed an released that would be more appropriate?
3. As stated earlier; I have no clue if this is the right way to manage OSC 22 control sequences, but at least it behaves the way I expect.

| Before this change | After this change |
|:---:|:---:|
| <video src="https://github.com/user-attachments/assets/690078e0-4dab-4919-84f2-5018499e1b90"> | <video src="https://github.com/user-attachments/assets/95b03ab1-0c23-411d-826c-a70b1f981675"> | 

Happy to get some guidance on any improvements here!

reykjalin:
Hi @wez 👋  Just wanted to ask if there's anything I can do to make moving this along easier?

I think the changes here are at least functionally good, so I'd love to help move this along if there's anything I can do to help with that :)

mux/src/pane.rs
```diff
@@ -550,6 +553,14 @@ mod test {
     }
 
     impl Pane for FakePane {
+        fn get_mouse_cursor_shape(&self) -> Option<String> {
+            unimplemented!()
```
wez, line 557:
```suggestion
            None
```

term/src/terminal.rs
```diff
@@ -51,6 +51,8 @@ pub enum Alert {
     TabTitleChanged(Option<String>),
     /// When the color palette has been updated
     PaletteChanged,
+    /// When the mouse cursor shape has been updated
+    MouseCursorShapeChanged,
```
wez, line 55: Let's pass the name along with this, like with the title changed event above.
That will make it easier to do a half-baked mux client implementation (see 44866cc137e336d8c00b23f5e0d7d1f03983b591)

```suggestion
    MouseCursorShapeChanged(Option<String>)
```

term/src/terminalstate/mod.rs
```diff
@@ -665,6 +669,16 @@ impl TerminalState {
...
+    /// Returns a copy of the current mouse cursor shape.
+    pub fn get_mouse_cursor_shape(&self) -> Option<String> {
+        self.mouse_cursor_shape.clone()
+    }
+
+    /// Clears the mouse cursor shape.
+    pub fn clear_mosue_cursor_shape(&mut self) {
+        self.mouse_cursor_shape = None;
+    }
+
```
wez, line 681: as mentioned elsewhere, I think this should be removed from the interface

```suggestion
```

wezterm-gui/src/termwindow/mouseevent.rs
wez: I think we should implicitly clear the cursor shape when the mouse is un-grabbed, then the original gating logic:

```rust
else if pane.is_mouse_grabbed() || outside_window {
   MouseCursor::Arrow
```

can be used there to very clearly show that we're not going to use the cursor shape in those conditions.

Each of these locations are places where the effective result is `is_mouse_grabbed` might change to false:

https://github.com/wezterm/wezterm/blob/c06cc87d9c50b6f45a03a0c9bfe24381cbb194c5/term/src/terminalstate/mod.rs#L1764-L1767

https://github.com/wezterm/wezterm/blob/c06cc87d9c50b6f45a03a0c9bfe24381cbb194c5/term/src/terminalstate/mod.rs#L1799-L1802

https://github.com/wezterm/wezterm/blob/c06cc87d9c50b6f45a03a0c9bfe24381cbb194c5/term/src/terminalstate/mod.rs#L1783-L1788

There should probably be a little helper function called in those three cases that does something like:

```rust
if !self.is_mouse_grabbed() && self.mouse_cursor_shape.is_some() {
   self.mouse_cursor_shape.take();
   mouse_cursor_shape_did_change
}
```

with all that done, the logic here in this file should simplify a lot and be very clear and easy to follow.

FWIW, I think `mouse_cursor_shape` is OK to refer to this stuff.

term/src/terminal.rs
wez: Note that by changing the `Alert` enum, you're changing the binary signature of types on which the mux protocol depends. You'll need to bump the CODEC_VERSION:

https://github.com/wezterm/wezterm/blob/c06cc87d9c50b6f45a03a0c9bfe24381cbb194c5/codec/src/lib.rs#L441-L444

wezterm-client/src/pane/clientpane.rs
```diff
@@ -245,6 +245,12 @@ impl ClientPane {
 
 #[async_trait(?Send)]
 impl Pane for ClientPane {
+    fn get_mouse_cursor_shape(&self) -> Option<String> {
```
wez, line 248: Take a look at the changes over in https://github.com/wezterm/wezterm/commit/44866cc137e336d8c00b23f5e0d7d1f03983b591#diff-b8703182c50d1b02b59221d8680fa9e6ec9eeb8520beb8e34d85f933fca7fac5R195 for how to do an imperfect but probably good enough in practice implementation of this that will work over the mux protocol

reykjalin:
Thanks for the review Wez! Just want to let you know this is on my radar, but I'm going to be busy until the end of March so I might not be able to address the review until then.

I'll follow up as soon as I'm able to! :)

PR #6316, opened 2024-10-24
Update log_error.md to include using ShowDebugOverlay to view logs

haphamdev:
I have struggled to find how to view the logs from Wezterm.log_info (warn, error).
I propose to make a minor update in the docs.


wez:
Thank you!

PR #6320, opened 2024-10-24
Select active tab choice as initial choice in ShowTabNavigator

mgpinf:
This introduces the following changes
- In `ShowTabNavigator`, select the choice corresponding to the active tab as the initial choice


mgpinf:
In my opinion, it would be useful to have active tab choice selected as the initial choice in ShowTabNavigator because if we disable tab bar and use ShowTabNavigator to navigate between tabs, it would be indicate the current tab we are on.

wez:
Thank you!

PR #6323, opened 2024-10-25
Update leader_is_active.md

haphamdev:
I think it should be `window:leader_is_active()` instead of `wezterm.leader_is_active()`


bew (approved):

wez:
Thank you!

Issue #5545, opened 2024-06-11
[FEATURE] Request to align tabs left or right on the tab bar


NishantJoshi00:
**Is your feature request related to a problem? Please describe.**
I was trying to setup my terminal to have a similar look to that of my old terminal. While configuring I wasn't able to find any config to align the tabs to the right side

**Describe the solution you'd like**
I had 2 solutions in mind:
1. Either we can have this configurable directly as a config
2. Or we can have this configurable in `format-tab-title`

**Additional context**
I would be interested in contributing if the opportunity exists

PR #6324, opened 2024-10-25
add support to align tabs to the right

NishantJoshi00:
A PR for addressing this issue.  #5545


NishantJoshi00:
This is far from a complete PR, but I just wanted to run this by the maintainer before adding configuration flag, handling some of the concerns with regards to how this will behave with most of other things when constructing the `tab-bar` `Line`.

PR #6349, opened 2024-11-02
Introduce sequencing for 'pane focus' related events

dzderic:
Switching between panes (especially over domains where large latencies are involved) can cause a loop where we get stuck repeatedly switching back and forth between two panes. This happens because handlers of `PaneFocused` events can also produce these events, causing us to get stuck in a loop if we let the GUI process stale `PaneFocused` events.

The solution is to introduce a 'serial' (AKA 'sequence number') for GUI-initiated pane focus changes. This will be echo'd back by the server, which allows the GUI to avoid processing the response if it has sent a new request in the mean time. CLI-initiated pane focus changes (eg. via `wezterm cli activate-pane`) will continue to be unconditionally handled by the client.

Inspired by #4737 and https://github.com/wez/wezterm/pull/4737#issuecomment-1876173247

Verified `activate-pane-direction`, `activate-pane` and `activate-tab` still work correctly

closes: #4390
closes: #4693

### DIFF
--- a/codec/src/lib.rs
+++ b/codec/src/lib.rs
@@ -16,6 +16,7 @@ use config::keyassignment::{PaneDirection, ScrollbackEraseMode};
 use mux::client::{ClientId, ClientInfo};
 use mux::pane::PaneId;
 use mux::renderable::{RenderableDimensions, StableCursorPosition};
+use mux::serial::InputSerial;
 use mux::tab::{PaneNode, SerdeUrl, SplitRequest, TabId};
 use mux::window::WindowId;
 use portable_pty::CommandBuilder;
@@ -24,7 +25,6 @@ use serde::{Deserialize, Serialize};
 use smol::io::AsyncWriteExt;
 use smol::prelude::*;
 use std::collections::HashMap;
-use std::convert::TryInto;
 use std::io::Cursor;
 use std::ops::Range;
 use std::path::PathBuf;
@@ -594,7 +594,7 @@ impl Pdu {
             | Pdu::SetPalette(SetPalette { pane_id, .. })
             | Pdu::NotifyAlert(NotifyAlert { pane_id, .. })
             | Pdu::SetClipboard(SetClipboard { pane_id, .. })
-            | Pdu::PaneFocused(PaneFocused { pane_id })
+            | Pdu::PaneFocused(PaneFocused { pane_id, .. })
             | Pdu::PaneRemoved(PaneRemoved { pane_id }) => Some(*pane_id),
             _ => None,
         }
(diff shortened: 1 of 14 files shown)


mux/src/lib.rs
```diff
@@ -111,6 +116,7 @@ pub struct Mux {
     num_panes_by_workspace: RwLock<HashMap<String, usize>>,
     main_thread_id: std::thread::ThreadId,
     agent: Option<AgentProxy>,
+    pane_focus_serial: RwLock<InputSerial>,
```
dzderic, line 119: Is there a better place to put this?

huylv:
Hi @dzderic, could you test if it fixes the problem when using wezterm via SSH domains as I described in [my comment](https://github.com/wez/wezterm/issues/4390#issuecomment-2586806627) in #4390? Many thanks!

codec/src/lib.rs
```diff
@@ -822,6 +788,7 @@ pub struct WindowTitleChanged {
 #[derive(Deserialize, Serialize, PartialEq, Debug)]
 pub struct PaneFocused {
     pub pane_id: PaneId,
+    pub pane_focus_serial: Option<InputSerial>,
```
wez, line 791: the various changes in this file require that we bump the codec version:

https://github.com/wezterm/wezterm/blob/364a158b6c615ee0d5cdfcd0ffa83fe73714ecdf/codec/src/lib.rs#L441-L444

mux/src/lib.rs
wez: hmm, maybe?

I think the biggest question on my mind is why this is only incremented by the local pane's implementation of `focus_changed` rather than centrally in the mux in `record_focus_for_client`, which is what is responsible for calling that.

What's on my mind here is that while we do have a mux client pane implementation that maybe you want to exclude from this, we also have a number of other pane implementations that run only locally.

mux/src/localpane.rs
```diff
@@ -478,6 +478,11 @@ impl Pane for LocalPane {
     }
 
     fn focus_changed(&self, focused: bool) {
+        if focused {
+            let mux = Mux::get();
+            mux.increment_pane_focus_serial();
```
wez, line 483: why does this happen only here and not for any other Pane impl?
There should probably be some comments here in the code to explain it!

wez (commented):
Thanks for this!
I like the approach, but I unsure about the rules around when the pane focus serial should and should not be incremented.
It would be great to capture those, because it would make it easier to understand the logic you intend, whether there are gaps in it, and easier for others to avoid breaking it in the future!

PR #6368, opened 2024-11-07
Update user-var-changed.md

343max:
By default base64 inserts a line break at some column which will truncate the user var. `-w 0` disables line wrapping

### DIFF
--- a/docs/config/lua/window-events/user-var-changed.md
+++ b/docs/config/lua/window-events/user-var-changed.md
@@ -8,10 +8,15 @@ used to set a user var.
 You can use something like the following from your shell:
 
 ```bash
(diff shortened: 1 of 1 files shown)


wez:
Thanks for this. That option doesn't appear to be available on the mac; sounds like maybe a GNU extension.
That makes things a bit of a PITA here; it's simpler to pretend that `-w -0` doesn't exist, but nicer to talk about it here.
But that in turn means explaining some more in the docs about both cases.
What would you like to do about it?

343max:
Good point, I wasn't aware. I would really like to make users aware of this gotcha because it isn't obvious and might create bugs. So I added a note so users are aware of the potential issue and know they have to make sure it works for them. Does this work for you?

wez:
Looks good, thank you!

PR #6370, opened 2024-11-07
Add Fallbacks for Setting WEZTERM_HOST in Non-Systemd and Minimal Environments

Kreijstal:
This PR addresses issues with setting the WEZTERM_HOST environment variable in environments lacking certain hostname utilities, specifically hostname and hostnamectl, which may not be available in minimal or non-Systemd setups.


wez:
Thank you!

Issue #4061, opened 2023-07-27
Key repeat gets stuck after pressing two keys in fast succession (again)


wookayin:
### What Operating System(s) are you seeing this problem on?

macOS

### Which Wayland compositor or X11 Window manager(s) are you using?

macOS. No window manager used.

### WezTerm version

wezterm 20230712-072601-f4abf8fd

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

Same in the nightly wezterm 20230726-161042.

I think the same bug as in #2452 happens again on my environments. The key repeat is stuck when pressing two keys (for a sufficient long time to repeat) in quick succession results in the second key printing one character.

e.g.

Actual:
```
aaaaaaaaaaaaaaabaaaaaaaaaab
```

### To Reproduce

Same as in #2452; Go into a terminal. Press the key A on your keyboard and hold it. Release it and quickly press the letter B on the keyboard and hold it. b is only printed once, even though it should be continuously be printed.

### Configuration

No config needed

### Expected Behavior

Expected:
```
aaaaaaaaaaaaaaabbbbbbbbbbbbbbbbbaaaaaaaaaaaabbbbbbbbbbbb
```

### Anything else?

![demo](https://github.com/wez/wezterm/assets/1009873/91be05a0-da73-4080-ac36-b52eb6ad062a)

wez:
Please include the debug key events logs from https://wezfurlong.org/wezterm/troubleshooting.html#debugging-keyboard-related-issues

wookayin:
<details>
   <summary>$ wezterm show-keys --lua</summary>
```
local wezterm = require 'wezterm'
local act = wezterm.action

return {
  keys = {
    { key = 'Tab', mods = 'CTRL', action = act.ActivateTabRelative(1) },
    { key = 'Tab', mods = 'SHIFT|CTRL', action = act.ActivateTabRelative(-1) },
    { key = 'Enter', mods = 'ALT', action = act.ToggleFullScreen },
    { key = '!', mods = 'CTRL', action = act.ActivateTab(0) },
    { key = '!', mods = 'SHIFT|CTRL', action = act.ActivateTab(0) },
    { key = '\"', mods = 'ALT|CTRL', action = act.SplitVertical{ domain =  'CurrentPaneDomain' } },
[...]

PR #6388, opened 2024-11-11
fix: don't clear ime_last_event when a key is UP

psyclaudeZ:
# Issue

When certain keys [1] are pressed and held before a previously held key is released, the key up event from the previously held key will prevent the new key from repeating. More details in #4061.

# Approach

The immediate reason of this is that the key up event clears `ime_last_event` which subsequently clears `ime_state` hence breaks the logic guarded by this comment:
```
// ... if the current event
// is a repeat, and the IME previously generated
// `Acted`, we will assume that we're safe to replay
// that last action.
```

Since, IMO, IME events are generally associated with key being pressed down, I'm adding a simple check in `key_common` to prevent the event from being cleared by a key up event.

Technically there're other places that clear `ime_last_event`, but I'm limiting my change to the immediate blast radius of the issue for simplicity.

# Test Plan

1. `cargo test --all`
2. Verified the said issue wouldn't happen anymore by holding `k` -> holding `j` -> releasing `k`.

***

[1] - b, f, j, m, p, q, v, x to be exact. The accent menu won't be loaded for these characters on macOS anyway.

### DIFF
--- a/window/src/os/macos/window.rs
+++ b/window/src/os/macos/window.rs
@@ -2560,7 +2560,7 @@ impl WindowView {
                             // but didn't call one of our callbacks.
                             // In theory, we should stop here, but the IME
                             // mysteriously swallows key repeats for certain
-                            // keys (eg: `f`) but not others.
+                            // keys (i.e. b, f, j, m, p, q, v, x) but not others.
                             // To compensate for that, if the current event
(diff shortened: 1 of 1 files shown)


wez:
This sounds reasonable, thank you!

PR #6391, opened 2024-11-11
fix: better handling of hyperlinks with parentheses

psyclaudeZ:
# Problem

I noticed there're a few cases where hyperlinks aren't handled correctly (see added unit tests). The root cause is that `)` is a confusing character in the context of URL and we should probably add more checks while doing the regex matching.

Also the [example explicit link](https://github.com/psyclaudeZ/wezterm/blob/main/docs/hyperlinks.md#explicit-hyperlinks) isn't clickable. 

# Approach

Instead of handle the exhaustive list of cases of `)` (which is inefficient and complicated), I decided to focus on the case where there's a terminating `)`. Particularly, I split the original regex into two: 1) match a URL with an explicit terminating `)` 2) mach a URL  with other "safe" terminating characters.

I also added `_` to the safe terminating character list.

Also added unit tests.

# Test Plan

Unit test.

`cargo build` and ran through a few examples.

I didn't explicitly attempt to fix the explicit hyperlink example. But after this regex change, `printf '\e]8;;http://example.com\e\\This is a link\e]8;;\e\\\n'` will actually generate a clickable link.


wez:
Thank you Bowei, this looks good!
Hope you are well!

psyclaudeZ:
No problem! Thank you Wez, same to you!

PR #6392, opened 2024-11-11
chore: resolve nit from linter

psyclaudeZ:
Just picked up Rust, but I don't think this will be a breaking change?

```
qwarning: variable does not need to be mutable
   --> window/src/os/macos/window.rs:681:13
    |
681 |         let mut handle =
    |             ----^^^^^^
    |             |
    |             help: remove this `mut`
    |
    = note: `#[warn(unused_mut)]` on by default


```


wez:
Thanks!

PR #6405, opened 2024-11-17
QuickSelectArgs: optionally bypass custom action on paste

nhurlock:
QuickSelect will paste the selection when the capitalized selection form is used. This is fine on its own when dealing with the default actions (lower=copy upper=copy+paste), but can cause odd behavior for QuickSelectArgs when a custom action is defined.

The example used in the docs for QuickSelectArgs (opening a URL) I believe is one of those cases where we'd want to perform a paste but NOT perform the action when a capitalized selection is used. Without the option, the user would experience both a paste of the URL as well as their browser opening the link.

In this PR I'm proposing the addition of a `paste_performs_action` override option for QuickSelectArgs which will allow users to disable this behavior. For the URL open case, when disabled, capital selection would only paste and not open the URL. I've defaulted the option to `true` to persist the current behavior.

### DIFF
--- a/config/src/keyassignment.rs
+++ b/config/src/keyassignment.rs
@@ -446,7 +446,7 @@ pub struct QuickSelectArguments {
     pub patterns: Vec<String>,
     #[dynamic(default)]
     pub action: Option<Box<KeyAssignment>>,
-    /// Call `action` after paste is performed (capital selection)
+    /// Skip triggering `action` after paste is performed (capital selection)
     #[dynamic(default)]
     pub skip_action_on_paste: bool,
     /// Label to use in place of "copy" when `action` is set
--- a/config/src/keyassignment.rs
+++ b/config/src/keyassignment.rs
@@ -446,10 +446,13 @@ pub struct QuickSelectArguments {
     pub patterns: Vec<String>,
     #[dynamic(default)]
     pub action: Option<Box<KeyAssignment>>,
+    /// Call `action` after paste is performed (capital selection)
+    #[dynamic(default)]
+    pub skip_action_on_paste: bool,
     /// Label to use in place of "copy" when `action` is set
     #[dynamic(default)]
     pub label: String,
-    /// How man lines before and how many lines after the viewport to
+    /// How many lines before and how many lines after the viewport to
     /// search to produce the quickselect results
     pub scope_lines: Option<usize>,
 }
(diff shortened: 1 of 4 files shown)


config/src/keyassignment.rs
```diff
@@ -446,10 +446,13 @@ pub struct QuickSelectArguments {
     pub patterns: Vec<String>,
     #[dynamic(default)]
     pub action: Option<Box<KeyAssignment>>,
+    /// Call `action` after paste is performed (capital selection)
+    #[dynamic(default = "default_true")]
+    pub paste_performs_action: bool,
```
wez, line 451: naming is hard... I would be inclined to swap the sense of this and name it `skip_action_on_paste` (and default `false`), which I think is slightly easier (for my native English brain) to parse because of the word order.

What do you think?

wez (commented):
Looks good, thank you!
My only question is on naming!

config/src/keyassignment.rs
nhurlock: works for me!
i do think having true-to-enable for a flag like this is better 👍

wez:
We're currently having some issues with CI; could you rebase your pr onto main and push it again? That way we'll at least have on task that is able to build and run the tests for this!

config/src/keyassignment.rs
```diff
@@ -446,10 +446,13 @@ pub struct QuickSelectArguments {
     pub patterns: Vec<String>,
     #[dynamic(default)]
     pub action: Option<Box<KeyAssignment>>,
+    /// Call `action` after paste is performed (capital selection)
```
wez, line 449:
```suggestion
    /// Skip triggering `action` after paste is performed (capital selection)
```

docs/config/lua/keyassignment/QuickSelectArgs.md
```diff
@@ -31,6 +31,7 @@ The `QuickSelectArgs` struct allows for the following fields:
 * `patterns` - if present, completely overrides the normal set of patterns and uses only the patterns specified
 * `alphabet` - if present, this alphabet is used instead of [quick_select_alphabet](../config/quick_select_alphabet.md)
 * `action` - if present, this key assignment action is performed as if by [window:perform_action](../window/perform_action.md) when an item is selected.  The normal clipboard action is NOT performed in this case.
+* `skip_action_on_paste` - overrides whether `action` is performed after an item is selected using a capital value (when paste occurs).
```
wez, line 34:
```suggestion
* `skip_action_on_paste` - overrides whether `action` is performed after an item is selected using a capital value (when paste occurs). {{since('nightly', inline=True)}}
```

PR #6411, opened 2024-11-19
Use serial2 crate instead of serial

jeevithakannan2:
The serial crate is not maintained has heavy dependencies. An alternative to serial is serial2 an actively maintained crate.


pty/src/serial.rs
```diff
@@ -22,71 +20,40 @@ use std::path::PathBuf;
...
+            baud: 9600,
         }
     }
 
-    pub fn set_baud_rate(&mut self, baud: BaudRate) {
+    pub fn set_baud_rate(&mut self, baud: u32) {
         self.baud = baud;
     }
-
-    pub fn set_char_size(&mut self, char_size: CharSize) {
```
wez, line 52: removing these methods is a breaking API change in the pty crate, and they are important things to be able to change.

Please restore these methods and arrange for them to configure the port via https://docs.rs/serial2/latest/serial2/struct.Settings.html

pty/src/serial.rs
```diff
@@ -149,7 +116,7 @@ impl Child for SerialChild {
         loop {
             std::thread::sleep(Duration::from_secs(5));
 
-            let mut port = self.port.lock().unwrap();
+            let port = self.port.lock().unwrap();
```
wez, line 119: Since serial2::SerialPort is Send + Sync, I think we can and should remove the Mutex that is used here so that we're not adding an additional bottleneck on top of whatever internal synchronization it is already doing.

wez (commented):
Thanks for this!
It's currently breaking the API for existing consumers by removing some important configuration options, and we need to reconcile that.

If we cannot retain the current interface then we need to have a strategy and suggestion in place for users when they upgrade, and importantly, we need to bump the pty crate version to `0.9.0` to safely indicate that the API is not backwards compatible.

pty/src/serial.rs
jeevithakannan2: Yeah you're right

jeevithakannan2:
> Thanks for this! It's currently breaking the API for existing consumers by removing some important configuration options, and we need to reconcile that.

Added API methods back now it should not the break the APIs

wez:
Thank you!

PR #6440, opened 2024-11-27
Disable libpng VSX_OPT flag for PowerPC 64

tom-miller1:
Fix a linker error in wezterm-gui due to a missing libpng feature on modern PowerPC 64-bit platforms.

Tested build on RHEL8.8 running on IBM POWER9 platform.
image


wez:
Thank you!

Issue #195, opened 2020-05-25
More styling options


wez:
Tidy up the window styling options so that padding and border sizes are configurable, along with border coloring.  (Today it's minimum effort for the window)

Allow specifying those options for tabs in the tab bar too.

cc: @dfrankland

bew:
Would be nice also to be able to hide the tab bar when there is only one tab open

wez:
> Would be nice also to be able to hide the tab bar when there is only one tab open.

You can do that today; I thought this was in the docs but it's only in the changelog:

http://wezfurlong.org/wezterm/changelog.html?highlight=hide_tab#20200406-151651-5b700e4

```lua
return {
   hide_tab_bar_if_only_one_tab = true,
}
```

PR #6445, opened 2024-11-29
adds movement for start of word, regardless of stop characters

rwese:
regarding #195 https://github.com/wez/wezterm/issues/1954#issuecomment-1568451852

```
Sample test/with-characters that cause stopps
                    -(cursor)
       -(after hitting MoveBackwardWordStart)

otherwise would need 2 MoveBackwardWord actions.
```

### DIFF
--- a/config/src/keyassignment.rs
+++ b/config/src/keyassignment.rs
@@ -672,6 +672,7 @@ pub enum CopyModeAssignment {
     MoveToStartOfNextLine,
     MoveToSelectionOtherEnd,
     MoveToSelectionOtherEndHoriz,
+    MoveBackwardWordStart,
     MoveBackwardWord,
     MoveForwardWord,
     MoveForwardWordEnd,
--- a/wezterm-gui/src/overlay/copy.rs
+++ b/wezterm-gui/src/overlay/copy.rs
@@ -822,6 +822,54 @@ impl CopyRenderable {
         }
     }
 
+    fn move_backward_one_word_start(&mut self) {
+        let y = if self.cursor.x == 0 && self.cursor.y > 0 {
+            self.cursor.x = usize::max_value();
+            self.cursor.y.saturating_sub(1)
+        } else {
+            self.cursor.y
+        };
+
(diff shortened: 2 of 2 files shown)


rwese:
it's missing documentation, I just wanted to get some feedback if this is the proper way to implement the new movement.

bew:
About naming, I think `MoveBackwardWordStart` & `MoveBackwardWord` are not different enough (I thought they were the same at the beginning).
Maybe rename to `MoveBackwardWORD` (but that breaks camelcase-ing..), or `MoveBackwardBigWord` ?

rwese:
Naming is always hard, I thought it would fit well with "moveForwardEnd" > "BackwardStart".

In terms of code, I took the easy way, not wanting to have to deal with breaking changes and refactoring. But I can give it a refactor, need to look into the test-cases.

config/src/keyassignment.rs
```diff
@@ -672,6 +672,7 @@ pub enum CopyModeAssignment {
     MoveToStartOfNextLine,
     MoveToSelectionOtherEnd,
     MoveToSelectionOtherEndHoriz,
+    MoveBackwardWordStart,
```
wez, line 675: TBH, I think this name is probably about as good as it's going to get; it is consistent with `MoveForwardWordEnd` and it does describe what it does reasonably well.

wezterm-gui/src/overlay/copy.rs
```diff
@@ -822,6 +822,54 @@ impl CopyRenderable {
         }
     }
 
+    fn move_backward_one_word_start(&mut self) {
```
wez, line 825: It's hard to tell at a glance just how similar this is to the `move_backward_one_word` function. If adding a parameter to control whether it should be in "start mode" or not makes for a smaller diff and a more readable function, that would get my vote!

wez (changes requested):
Thanks for this!

Please also add a page to the docs for the new action; look at `docs/config/lua/keyassignment/CopyMode/MoveForwardWordEnd.md` for inspiration!

Issue #3975, opened 2023-07-13
macOS: drag-n-drop folder into the app icon


conga0k:
**Describe the solution you'd like**
In other terminal emulators, such as kitty and Terminal.app, it is possible to open a new session by dragging a folder onto the application icon. I would like to see this feature implemented in Wezterm too.

wez:
Potentially the same/similar implementation details as:
* https://github.com/wez/wezterm/issues/3944

PR #6456, opened 2024-12-04
macOS: drag-n-drop folder into the app icon

catlan:
This implements #3975: macOS: drag-n-drop folder into the app icon

This requires in the _Info.plist_ to expand the  _CFBundleDocumentTypes_ array (copied the parts of the definition from _Terminal.app/Contents/Info.plist_). With this in place, the _application:openFile:_ gets called. I copied the existing OpenCommandScript code and adapted it.

This enables to open folders by drag-n-drop into the app icon, most commonly when in the Dock. But it also allows other applications to open wezterm at a certain directory.

Example Script:
```bash
# opens that directory in WezTerm (Path)
open -a ~/Downloads/WezTerm-macos-20240203-110809-5046fc22/WezTerm.app ~/Downloads/WezTerm-macos-20240203-110809-5046fc22

# opens that directory in WezTerm (Bundle ID)
open -b com.github.wez.wezterm ~/Downloads/WezTerm-macos-20240203-110809-5046fc22
```

Example Objective-C:
```objc
NSWorkspace *workspace = NSWorkspace.sharedWorkspace;
NSURL *appURL = [workspace URLForApplicationWithBundleIdentifier:@"com.github.wez.wezterm"];
if (appURL) {
	NSWorkspaceOpenConfiguration *configuration = [NSWorkspaceOpenConfiguration configuration];
	[workspace openURLs:@[url] withApplicationAtURL:appURL configuration:configuration completionHandler:^(NSRunningApplication *application, NSError *error) {
		if (error) {
			completionHandler(NO, error);
		} else {
			completionHandler(YES, nil);
		}
	}];
}
```

Todo:
* When WezTerm is launched with a directory to open, only open one window with the directory as cwd. Currently the default window opens and a window with the directory as cwd. I don't have a rust dev env, so I can't debug this.

### DIFF
--- a/window/src/os/macos/app.rs
+++ b/window/src/os/macos/app.rs
@@ -121,9 +121,14 @@ extern "C" fn application_open_file(
     let launched: BOOL = unsafe { *this.get_ivar("launched") };
     if launched == YES {
         let file_name = unsafe { nsstring_to_str(file_name) }.to_string();
+        let path = std::path::Path::new(&file_name);
         if let Some(conn) = Connection::get() {
             log::debug!("application_open_file {file_name}");
-            conn.dispatch_app_event(ApplicationEvent::OpenCommandScript(file_name));
+            if path.is_dir() {
+                conn.dispatch_app_event(ApplicationEvent::OpenDirectory(file_name));
+            } else {
+                conn.dispatch_app_event(ApplicationEvent::OpenCommandScript(file_name));
+            }
         }
     }
 }
--- a/window/src/os/macos/app.rs
+++ b/window/src/os/macos/app.rs
@@ -118,7 +118,7 @@ extern "C" fn application_open_file(
     _sel: Sel,
     _app: *mut Object,
     file_name: *mut Object,
-) {
+) -> BOOL {
     let file_name = unsafe { nsstring_to_str(file_name) }.to_string();
     let path = std::path::Path::new(&file_name);
     if let Some(conn) = Connection::get() {
@@ -128,7 +128,9 @@ extern "C" fn application_open_file(
         } else {
             conn.dispatch_app_event(ApplicationEvent::OpenCommandScript(file_name));
         }
+        return YES;
     }
+    NO
 }
 
 extern "C" fn application_dock_menu(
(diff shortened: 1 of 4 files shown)


window/src/os/macos/app.rs
```diff
@@ -68,27 +69,27 @@ extern "C" fn application_will_finish_launching(
     log::debug!("application_will_finish_launching");
 }
 
-extern "C" fn application_did_finish_launching(this: &mut Object, _sel: Sel, _notif: *mut Object) {
+extern "C" fn application_did_finish_launching(_self: &mut Object, _sel: Sel, _notif: *mut Object) {
     log::debug!("application_did_finish_launching");
-    unsafe {
-        (*this).set_ivar("launched", YES);
```
wez, line 74: FWIW, I suspect that removing this launched flag is probably the thing that's causing your multiple window issue

wez (changes requested):
Thanks for this!

re: not having a rust dev env, it's pretty low friction to install rust and build wezterm locally, and I recommend that you do that to iron out the kinks with this PR.

The launched flag in the ivars was important to this macos activation stuff, so I suspect that your TODO is around that.

I think this PR will regress the other application launch situations, so I really would like you to get to the bottom of that before we can merge this.

catlan:
> re: not having a rust dev env, it's pretty low friction to install rust and build wezterm locally, and I recommend that you do that to iron out the kinks with this PR.

What I meant here was: I couldn’t figure out how to do an incremental debug build. I had to do a clean release build; otherwise, only the binaries would get built, but not the macOS app bundle. And this got tiring pretty quickly

> The launched flag in the ivars was important to this macos activation stuff, so I suspect that your TODO is around that.

No, the flag was just a result of not understanding the macOS application lifecycle. If you read 67814d7, I mention that it wasn’t used in applicationWillFinishLaunching: or applicationDidFinishLaunching:. And in application_open_untitled_file:, where it was used, I implemented a check based on the number of open windows with e0b301f.

> I think this PR will regress the other application launch situations, so I really would like you to get to the bottom of that before we can merge this.

I tested all launch situations, and it is only that opening the application by dropping a folder on the icon that will the default window and a window with the directory as cwd.

Issue #6475, opened 2024-12-09
Mouse multiple button click requires pixel precision


jbiosca78:
### What Operating System(s) are you seeing this problem on?

Linux X11

### Which Wayland compositor or X11 Window manager(s) are you using?

Enlightenment 0.26

### WezTerm version

20241209-140259-6c171dd0

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

In #910, the solution taked was to check the click position for calculate the streak. It was ok because ClickPosition was character only.
But in #1805, ClickPosition added x_pixel_offset and y_pixel_offset, so now the multiple click must be in the same pixel.
Sometimes when pressing the first click, the mouse moves a little, making the double or more clicks to fail.
I have high sensibility mouse and high dpi display (4K), so many times my double clicks fails.

### To Reproduce

Do a double click moving a bit the mouse. The event (selection in my example) will fail.
It doesn't fail in other terminals.

### Configuration

no config

### Expected Behavior

Multiple click would not need to be in same pixel, it must work at least if the clicks are inside the same character, or if the click positions are very near.

PR #6476, opened 2024-12-09
fix mouse button streak requiring pixel precision

jbiosca78:
The fix made for #1805 introduced a problem where the multiple clicks must be done in the same pixel because the add of pixel offset in ClickPosition.

Multiple clicks in same character but with a little pixel displacement should be taken as multiple clicks as was before #1805.

This pull updates term/src/input.rs function LastMouseClick.add to check streak only if position.column and position.row matches, instead of checking full position with pixel offsets, fixing #6475.

### DIFF
--- a/term/src/input.rs
+++ b/term/src/input.rs
@@ -80,7 +80,8 @@ impl LastMouseClick {
     pub fn add(&self, button: MouseButton, position: ClickPosition) -> Self {
         let now = Instant::now();
         let streak = if button == self.button
-            && position == self.position
+            && position.column == self.position.column
+            && position.row == self.position.row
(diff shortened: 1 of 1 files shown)


jbiosca78:
This fix reduces so much my double click fails, but I still get some fails when I go fast and I double click in a word but the mouse moves from one character to another.

Other terminals like alacritty, terminology or xterm don't require that double click be in the same character. I prefer that behaviour makes me more fast with mouse. Maybe we should make the position check optional. The only pitfall is that if we miss the word to double click and click outside for 'unselect', it would do a triple click if clicking fast, but for me is still desirable because is much less common.

wez:
Thanks!

PR #6478, opened 2024-12-09
added bash example to portable_pty crate

boredcoder411:
I just added this example because I thought it wasn't really clear and I saw some discussions about how to open a shell without it hanging.


wez:
Thanks!

PR #6482, opened 2024-12-11
docs: replace %HOME% with %USERPROFILE% for Windows

ficcialfaint:
`%HOME%` is not a thing on Windows, `%USERPROFILE%` is.
image

Apologies, I messed up the commit name a bit by doing so:
image


wez:
Thanks!

PR #6493, opened 2024-12-17
Make rustflags in cargo config an array

dmyyy:
If you have top-level cargo settings that set rustflags in some parent directory you can't build wezterm from source because a string and an array can't be merged. This fixes that, and better matches the [reference](https://doc.rust-lang.org/cargo/reference/config.html) where rustflags is an array.


wez:
Thanks!

PR #6511, opened 2024-12-24
Fixed cursor blinking on transparent backgrounds

### LINKED ISSUE #3886: Blinking cursor is broken on transparent background

Linux X11

Tried on Mutter and Awesome.

### WezTerm version

20230408-112425-69ae8472 and 20230619-104421-c1f49509 (nightly)

### Describe the bug

The cursor isn't actually blinking, it's just switching colors from `cursor_bg` to `background` and it becomes visible when you set a transparent background or a background image. It's easier to notice on block styled cursors but it's the case with all cursors.

### To Reproduce

Just set a cursor style to any blinking one and either set your `window_background_image` to some image or `window_background_opacity` to anything smaller than 1. You just need a background that can contrast from whatever's setup as your `background` color.

### Configuration

```
window_background_opacity = 0.8,
audible_bell = 'Disabled',
default_cursor_style = 'BlinkingBlock',
cursor_blink_ease_in = 'Constant',
cursor_blink_ease_out = 'Constant',
cursor_blink_rate = 500,
```

### Expected Behavior

The cursor should switch from being visible to being invisible, as in, render the cell as it if the cursor wasn't there instead of just coloring the cursor `background`.

If this is intentional, there should at least be an option to make it work like this.

### Logs

_No response_

### Anything else?

_No response_

dangerous-frog:
Fixes #3886

Hi, 
It seems the mask was completely overriding the alpha values which means that any opacity stuff was getting thrown out - in case of bug above, the proper color with alpha 0 was being passed but then still rendered. Now it should take the alpha of the color into account.

### DIFF
--- a/wezterm-gui/src/glyph-frag.glsl
+++ b/wezterm-gui/src/glyph-frag.glsl
@@ -148,7 +148,7 @@ void main() {
     // and we need to tint with the fg_color
     color = fg_color;
     if (!subpixel_aa) {
-      color.a = colorMask.a;
+      color.a *= colorMask.a;
     }
     color = apply_hsv(color, foreground_text_hsb);
   }


lithium-hypercube:
This modification only partially works on my system; when I clear the terminal the cursor blinks as intended on a transparent background, but if I type in text and then erase the text the cursor starts switching between cursor_bg and cursor_fg again.

dangerous-frog:
Thanks for noticing! I couldn't replicate the exact issue on my side, but I did find the background color coming back after clearing the terminal (happening right after clearing, regardless of anything typed). I've fixed it and I think it has a high chance of fixing it for you. Would you be able to check again? If not could you share your config or more details so I can look into it more?

wez:
Thanks for this, but I'm concerned that this will impact other rendering cases; the shader handles more than just the cursor, and it's not clear to me that the change to the cursor bg color is correct for the non-transparent background case.
I think this needs more testing in the various situations before we can decide if the implementation is 100% good.

lithium-hypercube:
Thank you for the new patch! I can confirm that the cursor now blinks into transparency, but (as wez commented) this happens in all cases, including when the window is not transparent. A configurable cursor_bg opacity value might be helpful.

Issue #549, opened 2021-03-16
Ability to drag and drop windows


alexherbo2:
It could be nice to have similar features to [Konsole] for managing the terminal windows (tabs, panes) with the mouse.

[Konsole]: https://apps.kde.org/en/konsole

- [Konsole and splits]

[Konsole and splits]: https://tcanabrava.github.io/2019/06/27/konsole-splits/

The only other place where I have seen a good mouse support for tiles like that was in [Sway].

[Sway]: https://swaywm.org

carneeki:
Would be awesome to drag tabs (and maybe panes?) between windows as part of this - similar to how browsers behave.

wez:
Related to https://github.com/wez/wezterm/issues/1253

bennypowers:
Same, except not-necessarily by mouse, for example emulating tmux' `rotate-window` and `swap-pane` commands

wez:
In the nightly there are a few new key assignments:

* [RotatePanes](https://wezfurlong.org/wezterm/config/lua/keyassignment/RotatePanes.html)
* [PaneSelect](https://wezfurlong.org/wezterm/config/lua/keyassignment/PaneSelect.html)

And from the CLI:

* https://wezfurlong.org/wezterm/cli/cli/move-pane-to-new-tab.html
* https://wezfurlong.org/wezterm/cli/cli/split-pane.html - `wezterm cli split-pane --move-pane-id ID` can move a pane into a newly created split

There's no mouse equivalent functionality as of yet.

dstromberg:
I too would love to see the ability to drag tabs to rearrange them in wezterm.

tgross35:
Would it be possible to to add default bindings for the reorder tab left/right options? Something with `[`/`]` would be great to keep it similar to the change tabs action. Maybe `alt+[`?

wez:
@tgross35 there are already default assignments. You are welcome to add whatever assignments you like via your config file.

```console
$ wezterm -n show-keys | grep MoveTab
        SHIFT | CTRL         PageUp             ->   MoveTabRelative(-1)
        SHIFT | CTRL         PageDown           ->   MoveTabRelative(1)
```

ibash:
I use kitty right now but this would be the reason I'd switch to wez. I used iterm2 before and dragging was natural. Tbh I still have to lookup the kitty shortcut to reorder panes each and every time I want to.

VVVHawk:
+1, the way `terminator` handles it is perfect.

[...]

PR #6527, opened 2024-12-30
feat: reorder tabs via left mouse drag

Jefffrey:
Relates to #549

Be able to reorder tabs within a window via left mouse drag. Doesn't have animation dragging (like on Konsole), so it looks snappy for now.

[Screencast_20241230_204103.webm](https://github.com/user-attachments/assets/c612f8b6-f03d-4e2a-95fa-f2ca6b5ca39f)

### DIFF
--- a/docs/config/lua/keyassignment/MoveTabRelative.md
+++ b/docs/config/lua/keyassignment/MoveTabRelative.md
@@ -4,6 +4,8 @@ Move the current tab relative to its peers.  The argument specifies an
 offset. eg: `-1` moves the tab to the left of the current tab, while `1` moves
 the tab to the right.
 
+It is also possible to reorder tabs via left mouse drag on the tab bar entry.
+
 ```lua
 local act = wezterm.action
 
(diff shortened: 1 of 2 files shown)


lihaohong6:
I tested this commit on Fedora KDE and Windows. This feature works on both platforms. 

I tried on macOS as well, but couldn't get wezterm to compile from source. Maybe it works for someone else who can try it?

9999years:
I've tested this on macOS, it works well!

The tabs aren't draggable with `config.window_decorations = "INTEGRATED_BUTTONS"`, but that's because of #7116 and not a deficiency with this PR.

jayjohnston:
Tested this daily for a while — the core reordering works well. Found and
fixed one bug that made it look broken (the tab bar never repaints during
the drag, so a successful reorder is invisible until something else forces
a redraw), and added a floating drag preview so the dragged tab visibly
tracks the cursor instead of same-titled tabs just flickering active/inactive
as you cross them.

Opened #7926 with both fixes on top of this branch — happy to have it
squashed in here instead if that's easier for you, just wanted to share the
diff and a demo:

PR #6533, opened 2025-01-01
Improve Wayland scroll behavior

### LINKED ISSUE #3142: Scrolling too fast in Wayland

gokselk:
This PR enhances the scroll input handling in Wayland by introducing source-specific scroll adjustments:

- For mouse wheels:
  - Uses discrete values when available
  - Falls back to normalizing continuous values using standard 15-degree wheel clicks
- For touchpads and other input sources:
  - Applies a 1/10 scale factor to smooth out scrolling behavior

This change helps address the oversensitive scrolling reported in #3142, particularly making touchpad scrolling more consistent with X11 behavior and other terminal emulators.

Tested on KDE Plasma 6 Wayland with 3-line scroll increments, showing improved consistency with system-wide scrolling behavior.

### DIFF
--- a/window/src/os/wayland/pointer.rs
+++ b/window/src/os/wayland/pointer.rs
@@ -8,7 +8,7 @@ use smithay_client_toolkit::seat::pointer::{
     PointerData, PointerDataExt, PointerEvent, PointerEventKind, PointerHandler,
 };
 use wayland_client::backend::ObjectId;
-use wayland_client::protocol::wl_pointer::{ButtonState, WlPointer};
+use wayland_client::protocol::wl_pointer::{AxisSource, ButtonState, WlPointer};
 use wayland_client::protocol::wl_seat::WlSeat;
 use wayland_client::{Connection, Proxy, QueueHandle};
 use wezterm_input_types::MousePress;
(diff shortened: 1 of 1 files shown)


gokselk:
Identified inconsistent behavior across different DPI settings. This PR is currently a draft while I investigate potential solutions.

window/src/os/wayland/pointer.rs
```diff
@@ -157,12 +157,29 @@ impl PendingMouse {
...
+
+                // Handle scroll based on the input source
+                let (scroll_x, scroll_y) = match source {
+                    // For mouse wheels, use discrete values when available
+                    Some(AxisSource::Wheel) => {
+                        if vertical.discrete != 0 {
+                            (horizontal.discrete as f64, vertical.discrete as f64)
+                        } else {
+                            // Standard wheel click is 15 degrees
+                            (horizontal.absolute / 15.0, vertical.absolute / 15.0)
```
wez, line 174: I'm thinking that adding a couple of config options would be a good idea, at least until we can dial things in to be great by default.

Over in this other PR:
* https://github.com/wezterm/wezterm/pull/5104
a general scroll factor was added; I don't really want to do that, but that PR does show how you can add a config option and plumb it through.

I'm thinking that perhaps it would be good to add:

* wayland_scroll_wheel_factor 
* wayland_scroll_finger_factor
* wayland_scroll_continuous_factor
* wayland_scroll_tilt_factor
* wayland_scroll_unknown_source_factor

eg: one option for each of the possible axis source enum values, and one for the unknown source case.

I'm not 100% sure how to reconcile the discrete vs. absolute cases for the wheel.  Do we permute these options so that we have eg: `wayland_scroll_wheel_discrete_factor` and `wayland_scroll_wheel_absolute_factor` ?
Those could default to `1` and `15` to match the logic above, but that would allow the user to fine tune this stuff if necessary.

gokselk: I took another look at this a few weeks ago but ran into issues with getting the correct fractional scaling ratio using the smithay-client-toolkit. It might be a good idea to first focus on getting integer scaling right, or possibly provide an option to override scaling through the config file. I’ll revisit it later to continue investigating.

bew:
Hello @gokselk, are you willing to continue working on this?

NOTE: I've been using this patch in my Wezterm install (via Nix) since ~2025-11, but I'm still seeing quite speedy scrolling using a touchpad 🤔

PR #6540, opened 2025-01-02
nix/flake.lock: Update; nix/flake.nix: backport patches from NixOS/nixpkgs

thiagokokada:
Fix compilation in `aarch64-darwin` with latest `nixpkgs` changes. Also general clean-ups and adding a formatter to make the code more consistent.

I am also committed to be an unnoficial maintainer of this since this affects my developer environment. I can't commit to be a maintainer forever, but I should keep this as much up-to-date as possible until it is useful for me.


thiagokokada:
I also build for `x86_64-darwin` and seems to be working fine.

bew (approved):
Change looks good 👍

nix/zlib-fdopen.patch
```diff
@@ -0,0 +1,23 @@
+diff --git a/deps/freetype/zlib/zutil.h b/deps/freetype/zlib/zutil.h
```
wez, line 1: is this patch really necessary? I *really* don't like the precedent of dumping a lot of stuff into wezterm's repo just for nix--I'm already on the fence about retaining the lock and nix files in here as it is.

Can you find a way to achieve whatever this is doing, without the patch?

wez (commented):
Thanks for diving in!

nix/zlib-fdopen.patch
thiagokokada: If we want to pin `zlib` `1.2.11`, yes.

The other option would be to bump `zlib` to `1.3.1`. I am not sure if this will bring other issues though, but at least this issue should be solved.

wez: 1.3.1 sounds good to me; I've pushed that dep update.  Please rebase this PR and we'll see what happens in CI!

thiagokokada: BTW, if you can bump `libpng` we can also remove the other patch probably.

wez: just pushed a bump for libpng -> 1.6.44

thiagokokada: Done, thanks @wez.

wez:
Thank you!

PR #6548, opened 2025-01-03
wayland: prevent setting cursor when window is not focused

jmbaur:
Fixes https://github.com/wez/wezterm/issues/5760

### DIFF
--- a/window/src/os/wayland/window.rs
+++ b/window/src/os/wayland/window.rs
@@ -932,6 +932,10 @@ impl WaylandWindowInner {
     }
 
     fn set_cursor(&mut self, cursor: Option<MouseCursor>) {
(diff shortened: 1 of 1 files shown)


holmanb (approved):
I can reliably reproduce this issue. I just applied this change to the latest upstream commit (6c443bee9a21b6d37e06227) and tested it. The error log disappears with this change: it fixes my reproducer.

I'm not familiar with the code enough to give a proper code review, but the fix appears logical based on the [symptoms that I observe](https://github.com/wez/wezterm/issues/5760#issuecomment-2598746671).

Thanks @jmbaur!

blg666 (approved):

blg666 (approved):

wez:
Thank you!

PR #6558, opened 2025-01-06
feat(macos): add support for titlebar background matching theme background on macos

Jay-Madden:
This PR adds support for changing the background color of the titlebar on macos.
Adding a new macos window decoration option `MACOS_USE_BACKGROUND_COLOR_AS_TITLEBAR_COLOR`

With the following wezterm config
```lua
config.window_decorations = "TITLE|RESIZE|MACOS_USE_BACKGROUND_COLOR_AS_TITLEBAR_COLOR"
```
the titlebar is rendered like this 

image

It also works for the default theme, being rendered like this 
image

### Tasks

- [x] Active titlebar background color

Ref: https://github.com/wez/wezterm/issues/6557


window/Cargo.toml
```diff
@@ -45,6 +45,7 @@ url = "2"
 wezterm-bidi = { path = "../bidi" }
 wezterm-color-types = { path = "../color-types" }
 wezterm-font = { path = "../wezterm-font" }
+ wezterm-term = { path = "../term" }
```
Jay-Madden, line 48:
```suggestion
wezterm-term = { path = "../term" }
```

blg666 (approved):

wezterm-input-types/src/lib.rs
```diff
@@ -1976,6 +1976,7 @@ bitflags! {
         const MACOS_FORCE_DISABLE_SHADOW = 4;
         const MACOS_FORCE_ENABLE_SHADOW = 4|8;
         const INTEGRATED_BUTTONS = 16;
+        const MACOS_RESPECT_THEME_BACKGROUND = 32;
```
wez, line 1979: naming wise, we don't use the term `theme` to refer to color schemes or color preferences, so this is an inconsistency.

I'd prefer it for the name to spell out what it does, which to me sounds like `MACOS_USE_BACKGROUND_COLOR_AS_TITLEBAR_COLOR` or something like that, as that is what this PR appears to be doing

window/src/os/macos/window.rs
```diff
@@ -1109,6 +1110,27 @@ impl WindowInner {
         }
     }
 
+    fn update_titlebar_background(&self, color: RgbaColor) {
+        unsafe {
+            let titlebar_view_container = get_titlebar_view_container(&self.window).unwrap();
```
wez, line 1115: I really don't like bare `.unwrap()` calls, as each one represents a place where the program can either die or end up broken if the expectations aren't met.  I'd prefer for this to return without doing anything instead of panicking the program

window/src/os/macos/window.rs
```diff
@@ -1132,12 +1154,23 @@ impl WindowInner {
...
                 .setStyleMask_(NSWindowStyleMask::NSBorderlessWindowMask);
+
             apply_decorations_to_window(
                 &self.window,
                 self.config.window_decorations,
                 self.config.integrated_title_button_style,
             );
 
+            // Set the titlebar background to the theme color falling back to the default if there
+            // is no specified theme
```
wez, line 1165: s/theme/color scheme/

wez (changes requested):
Thanks for this!
I have a few comments that are mostly minor.

One thing that isn't is that I think it would be great to try to do the view traversal stuff using the objc2 crate; it should cut down on a bunch of the ownership management and make the associated code a bit smaller and easier to reason about.

That might be a big ask right now because the rest of the file is using the legacy crate and I don't know how easily the two will coexist until the rest of the file is cut over.

window/src/os/macos/window.rs
```diff
@@ -1694,6 +1799,26 @@ impl Inner {
 
 const VIEW_CLS_NAME: &str = "WezTermWindowView";
 const WINDOW_CLS_NAME: &str = "WezTermWindow";
+const TITLEBAR_VIEW_NAME: &str = "NSTitlebarContainerView";
+
+// core-graphics does not currently expose a binding by default for CGColor srgb conversions
+// pull request to upstream this here: https://github.com/servo/core-foundation-rs/pull/716
+// IMPORTANT: This binding was introduced in catalina, attempting to use it on older systems will fail
+fn srgb(red: CGFloat, green: CGFloat, blue: CGFloat, alpha: CGFloat) -> CGColor {
```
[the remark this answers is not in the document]
Jay-Madden: I named it just `srgb` to allow for easily switching to the servo binding I upstreamed here https://github.com/servo/core-foundation-rs/pull/716 whenever they get around to cutting that release. 

That being said, as you found, objc2 already has this [binding](https://docs.rs/objc2-core-graphics/latest/objc2_core_graphics/fn.CGColorCreateSRGB.html) so all of this binding code has been removed and replaced with that single call from the crate.

window/src/os/macos/window.rs
```diff
@@ -1694,6 +1799,26 @@ impl Inner {
...
+}
+
+extern "C" {
+    fn CGColorCreateSRGB(
+        red: CGFloat,
+        green: CGFloat,
+        blue: CGFloat,
+        alpha: CGFloat,
+    ) -> core_graphics::sys::CGColorRef;
+}
```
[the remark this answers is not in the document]
Jay-Madden: Great find, this does indeed allow for us to not need the binding directly.

window/src/os/macos/window.rs
Jay-Madden: Switch this to an `if let` and added a log line on failure.

wezterm-input-types/src/lib.rs
Jay-Madden: done

window/src/os/macos/window.rs
```diff
@@ -1132,12 +1154,23 @@ impl WindowInner {
...
+
             apply_decorations_to_window(
                 &self.window,
                 self.config.window_decorations,
                 self.config.integrated_title_button_style,
             );
 
+            // Set the titlebar background to the theme color falling back to the default if there
+            // is no specified theme
+            if self.config.window_decorations.contains(WindowDecorations::MACOS_RESPECT_THEME_BACKGROUND) {
```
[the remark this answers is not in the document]
Jay-Madden: Consolidated this logic into `update_titlebar_background` and called it in both `show` and `config_did_change`

window/Cargo.toml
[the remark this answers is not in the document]
Jay-Madden: Its needed because we pull the default color value to fall back to from `wezterm-term`

```rust
        let color = self.config.resolved_palette.background.unwrap_or(
            wezterm_term::color::ColorPalette::default().background.into()
        );
```

I could not find the default values in the config object passed into the window. If there is a way to grab the default color values without referencing `ColorPallette` the dep can be removed

Jay-Madden:
@wez Apologies for the delay, i unfortunately got rather sick right about the time you left a review 

I have addressed the majority of the comments

1. Handle the failure to get the titlebar container gracefully, log and do nothing
2. Add the `objc2-core-graphics` crate and directly use their `CGColorCreateSRGB` binding instead of creating our own. 

Now, re using the objc2 crate to do the view traversal work. I looked into it for a while and as best i can tell theres no non horrifically hacky way to cast an `objc` crate `id` or `StrongPtr` object into the related `objc2` crates `Retained<T>` object. 

While T is likely the same underlying objective c memory. There is no way that i could find to construct a `Retained<T>` from the raw pointer created by 
```rust
            let window: id = msg_send![get_window_class(), alloc];
```

Is the desire to move entirely to objc2 for all macos window rendering before this is merged? 

I am willing to do work to move the macos window rendering to `objc2` crate, its just a separate change and will take a bit of time.

window/src/os/macos/window.rs
```diff
@@ -1401,6 +1436,76 @@ fn decoration_to_mask(
     }
 }
 
+unsafe fn get_view_class_name(id: id) -> Option<String> {
```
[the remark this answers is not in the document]
Jay-Madden: Addressed this in the top level comment

wez:
> Is the desire to move entirely to objc2 for all macos window rendering before this is merged?
> 
> I am willing to do work to move the macos window rendering to `objc2` crate, its just a separate change and will take a bit of time.

There's desire to migrate, but it's a non-blocking desire. If you have bandwidth to undertake migrating to objc2 that's great and very much welcomed, but it can be done in a follow-up PR

window/src/os/macos/window.rs
```diff
@@ -1111,6 +1112,40 @@ impl WindowInner {
...
 
+    fn update_titlebar_background(&self) {
+        if !self.config.window_decorations.contains(WindowDecorations::MACOS_USE_BACKGROUND_COLOR_AS_TITLEBAR_COLOR) {
+            return
+        }
+
+        // Set the titlebar background to the theme color falling back to the default if there
+        // is no specified color scheme
+        let color = self.config.resolved_palette.background.unwrap_or(
+            wezterm_term::color::ColorPalette::default().background.into()
```
wez, line 1123: ah, I see.  It's a bit undesirable to add a dep to wezterm_term because the window crate is intended to ostensibly be used for more than just wezterm's terminal stuff.  However, since window already depends on config and that is very much wezterm centric, that argument is growing weaker these days.

A more practical argument against doing this is that creating the full palette object is relatively speaking a lot of work to compute the full color cube and do some other stuff with heap allocated arrays.

I think you can just default to using Black directly in the fallback case to avoid that work and adding that dep.

I also think that you can obtain that just by defaulting the RgbaColor like this, although it's possible that it might be transparent black, in which case you'll need to explicitly make an opaque black RgbaColor

```suggestion
        let color = self.config.resolved_palette.background.unwrap_or_default()
```

Jay-Madden: The default is indeed transparent, likely due to the alpha of 0 
image

I just created the color black manually like this 

```
        let color = self.config.resolved_palette.background.unwrap_or(
            RgbaColor::from(SrgbaTuple(0.,0.,0.,255.))
        );
```

Jay-Madden: `wezterm-term` dependency has been removed in [4edf878](https://github.com/wezterm/wezterm/pull/6558/commits/4edf878a5c6e225f5ecd990fa72ed66e4ba7b65f)

Jay-Madden:
> If you have bandwidth to undertake migrating to objc2 that's great and very much welcomed, but it can be done in a follow-up PR

From the few hours this morning i spent poking through both libraries code I think the api surface area of the two libraries is more or less compatible. The conversion is mostly just grunt work. I'll look a bit deeper and open a subsequent PR if i don't encounter any major blockers.

wez:
Thank you!

I manually pulled your fork, started a new branch, cherry-picked your commits, unpicking and recreating the work that got blended into the merge commit, then fixed up the fmt errors and pushed a squashed version of that.

Please avoid using merge commits in future PRs; rebasing is the one true way!

Jay-Madden:
Oh yeesh, noted, ya this branch was not at all setup to be rebased lol, that was a gnarly and hacky merge commit too.

pbgc:
@Jay-Madden @wez this was great! Unfortunately when MacOS is using a light system theme and the background color of the terminal is dark ... we get a dark title on top of a dark background:
Captura de ecrã 2025-04-14, às 13 48 08

In this case the title should be light (or at least have a configuration settings to choose the color) .. like ghostty does very well:
Captura de ecrã 2025-04-14, às 13 51 43

PR #6584, opened 2025-01-19
feat: enforce minimum contrast ratio for reverse video cursor

jameshurst:
This PR ensures the reverse video cursor meets a minimum contrast ratio, otherwise the default fg/bg is used instead.

The minimum contrast ratio of `2.5` was chosen from the kitty implementation linked in https://github.com/wez/wezterm/discussions/2861.

The `SrgbaTuple::contrast_ratio` implementation also appeared to be incorrect since it was comparing lightness instead of luminance.

This should resolve https://github.com/wez/wezterm/discussions/2861.

### DIFF
--- a/docs/config/lua/config/reverse_video_cursor_min_contrast.md
+++ b/docs/config/lua/config/reverse_video_cursor_min_contrast.md
@@ -5,6 +5,8 @@ tags:
 ---
 # `reverse_video_cursor_min_contrast = 2.5`
 
+{{since('nightly')}}
+
 The minimum contrast ratio required to use the reverse video cursor.
 
 When the contrast ratio between the reverse video cursor foreground and
--- a/docs/config/lua/config/reverse_video_cursor_min_contrast.md
+++ b/docs/config/lua/config/reverse_video_cursor_min_contrast.md
@@ -1,0 +1,12 @@
+---
+tags:
+  - appearance
+  - text_cursor
+---
+# `reverse_video_cursor_min_contrast = 2.5`
+
+The minimum contrast ratio required to use the reverse video cursor.
+
+When the contrast ratio between the reverse video cursor foreground and
+background is below this threshold then the default cursor foreground and
+background will be used instead.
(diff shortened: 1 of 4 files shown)


blg666 (approved):

wezterm-gui/src/termwindow/render/mod.rs
```diff
@@ -484,6 +484,8 @@ impl crate::TermWindow {
     }
 
     pub fn compute_cell_fg_bg(&self, params: ComputeCellFgBgParams) -> ComputeCellFgBgResult {
+        const CURSOR_MIN_CONTRAST: f32 = 2.5;
```
wez, line 487: I think we should make this an option because I just know that someone is going to complain about this otherwise, regardless of how well selected the default value might be!

I'd suggest adding an option named something like `cursor_minimum_contrast: f32` over here:

https://github.com/wezterm/wezterm/blob/b251509a2f9c488f5379a8457258eb0f07053fce/config/src/config.rs#L612-L613

You'll also need to add a little helper function to return the 2.5 default value and assign that using `#[dynamic(default = "default_contrast_ratio")]` as an attribute on that option.

then I think you should be good to examine `self.config.cursor_minimum_contrast` in this function instead of the constant.

wez (changes requested):
Thanks for this!

Let's add a config option to control that contrast ratio.  There will also need to be an entry in the docs for that page, you can add one to https://github.com/wezterm/wezterm/tree/main/docs/config/lua/config

Thanks again!

jameshurst:
I've added the option `reverse_video_cursor_min_contrast` to the config and created a docs page for it as well. I opted for that name instead of `cursor_minimum_contrast` to make it clear that the option only applies when the reverse video cursor is used.

docs/config/lua/config/reverse_video_cursor_min_contrast.md
```diff
@@ -0,0 +1,12 @@
+---
+tags:
+  - appearance
+  - text_cursor
+---
+# `reverse_video_cursor_min_contrast = 2.5`
+
```
wez, line 7:
```suggestion

{{since('nightly')}}

```

wez:
Thank you!

PR #6587, opened 2025-01-20
feat(macOS): add support to disable rounded corners on macOS

### LINKED ISSUE #2182: disable rounded corners macos
Hi, I've seen discussion about making the macos build preserve the rounded corners when the title bar is disabled here: #1034, I'm assuming the squared corners were originally considered a bug?

Is it possible to restore the squared corners in macos via a setting? If not, I think it would be a good idea to extend the window decorations setting to include a corners option.

amadeusdotpng:
This PR adds support to disable rounded corners on macOS with a new window decoration option `MACOS_FORCE_SQUARE_CORNERS`.
```lua
config.window_decorations = 'MACOS_FORCE_SQUARE_CORNERS'
```
Before:
image

After:
image

It is recommended to use `RESIZE` with this option since it does not make the window resizable by itself.

It does not work with `TITLE` and `INTEGRATED_BUTTONS`.

resolves #2182

### DIFF
--- a/wezterm-input-types/src/lib.rs
+++ b/wezterm-input-types/src/lib.rs
@@ -1975,7 +1975,8 @@ bitflags! {
         // so that we effective have Option<bool>
         const MACOS_FORCE_DISABLE_SHADOW = 4;
         const MACOS_FORCE_ENABLE_SHADOW = 4|8;
-        const INTEGRATED_BUTTONS = 16;
+        const MACOS_FORCE_SQUARE_CORNERS = 16;
+        const INTEGRATED_BUTTONS = 32;
     }
 }
 
@@ -1995,6 +1996,8 @@ impl Into<String> for &WindowDecorations {
             s.push("MACOS_FORCE_ENABLE_SHADOW");
         } else if self.contains(WindowDecorations::MACOS_FORCE_DISABLE_SHADOW) {
             s.push("MACOS_FORCE_DISABLE_SHADOW");
+        } else if self.contains(WindowDecorations::MACOS_FORCE_SQUARE_CORNERS) {
+            s.push("MACOS_FORCE_SQUARE_CORNERS");
         }
         if s.is_empty() {
             "NONE".to_string()
@@ -2020,6 +2023,8 @@ impl TryFrom<String> for WindowDecorations {
                 flags |= Self::MACOS_FORCE_DISABLE_SHADOW;
             } else if ele == "MACOS_FORCE_ENABLE_SHADOW" {
                 flags |= Self::MACOS_FORCE_ENABLE_SHADOW;
+            } else if ele == "MACOS_FORCE_SQUARE_CORNERS" {
+                flags |= Self::MACOS_FORCE_SQUARE_CORNERS;
             } else if ele == "INTEGRATED_BUTTONS" {
                 flags |= Self::INTEGRATED_BUTTONS;
             } else {
--- a/wezterm-input-types/src/lib.rs
+++ b/wezterm-input-types/src/lib.rs
@@ -1975,8 +1975,8 @@ bitflags! {
         // so that we effective have Option<bool>
         const MACOS_FORCE_DISABLE_SHADOW = 4;
         const MACOS_FORCE_ENABLE_SHADOW = 4|8;
-        const MACOS_FORCE_SQUARE_CORNERS = 16;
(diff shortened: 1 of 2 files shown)


blg666 (approved):

blg666 (approved):

wezterm-input-types/src/lib.rs
```diff
@@ -1975,7 +1975,8 @@ bitflags! {
         // so that we effective have Option<bool>
         const MACOS_FORCE_DISABLE_SHADOW = 4;
         const MACOS_FORCE_ENABLE_SHADOW = 4|8;
-        const INTEGRATED_BUTTONS = 16;
+        const MACOS_FORCE_SQUARE_CORNERS = 16;
+        const INTEGRATED_BUTTONS = 32;
```
wez, line 1979: prefer to minimize the diff unless there is good reason to re-number the enum

```suggestion
        const INTEGRATED_BUTTONS = 16;
        const MACOS_FORCE_SQUARE_CORNERS = 32;
```

wez (commented):
Thanks for this!

alberti42:
# Bug report: crash in debug, wrong scale on launch, no drag/resize

## Summary

* **Debug-only crash** when `MACOS_FORCE_SQUARE_CORNERS` removes the titlebar: we call `setHidden:` on a `nil` standard window button.
* **Scaling wrong on launch** if the app **starts** with `MACOS_FORCE_SQUARE_CORNERS` (content appears ~2× larger, clipped). If you start with `RESIZE` and then **toggle** to square corners at runtime, scaling is OK.
* **No drag/resize** when square corners are active: the window can’t be resized or moved (even ⌃⌘-drag doesn’t work).

## Environment

* macOS: Sequoia 15.6.1 (24G90)
* Hardware: Apple Silicon M2
* WezTerm: nightly (built from source); also verified using the original commit fcae187b29dcfc6d51372be814b0b74621c37fdd
* Build modes: **debug** triggers crash; **release** doesn’t crash but shows scaling+interaction issues

## Repro

### A) Debug crash (nil button)

* Start with config enabling `MACOS_FORCE_SQUARE_CORNERS` (and no `TITLE`).
* Run a **debug** build.
* Crash with backtrace at `window/src/os/macos/window.rs` around the loop calling `standardWindowButton_` then `setHidden:`.

**Cause:** When the style mask drops `NSTitledWindowMask`, `standardWindowButton_(…)` returns `nil`. Rust `msg_send!` to `setHidden:` on `nil` panics in debug (Objective-C would silently ignore).

### B) Wrong scale on fresh launch

* Launch with `MACOS_FORCE_SQUARE_CORNERS` active from the start.
* The terminal content renders ~2× bigger and is clipped by the window bounds.

**Observation:** If you launch with `RESIZE` (normal titled window) and **then** switch to square corners, scale is correct. So the mismatch happens at initial style/layer setup time.

**Likely cause:** Drawable layer `contentsScale` stays at **1.0** while `convertRectToBacking` reports **2.0** (Retina). Rendering targets 2× pixels into a 1× drawable ⇒ everything is large/clipped.

### C) No dragging / resizing with square corners

* With `MACOS_FORCE_SQUARE_CORNERS`, the window can’t be moved or resized (even ⌃⌘-drag).
* This is consistent with removing the standard titlebar and not providing an alternative drag/resize path.

## Logs / snippet

Crash location (representative):

```
rustCopy codefor titlebar_button in &[
    appkit::NSWindowButton::NSWindowMiniaturizeButton,
    appkit::NSWindowButton::NSWindowCloseButton,
    appkit::NSWindowButton::NSWindowZoomButton,
] {
    let button = window.standardWindowButton_(*titlebar_button);
    // `button` is nil when NSTitledWindowMask is not present
    let _: () = msg_send![button, setHidden: if hide_buttons { YES } else { NO }];
}
```

alberti42:
For your info, I opened a PR #7193 which significantly refactor the code about window decorations. In my view, if you can fix the issues I mentioned about MACOS_FORCE_SQUARE_CORNERS, it would be great if you could adopt / test the code refactoring I introduced.

The original purpose of PR #7193 was to introduce a new decoration `MACOS_DISABLE_TITLEBAR_DRAG` to allow the user to interact with the first line when no title was shown. Unfortunately, this new decoration still has some limitations that I described in the PR. However, the code refactoring is, in my view, solid and contributes to improving wezterm eventually.

Issue #2347, opened 2022-08-02
missing license file in many crates published from this repo


decathorpe:
One of the requirements of the terms of the MIT license is that redistributed sources contain a copy of the license text. This is currently not satisfied for many crates that are published from this workspace, for example, the vtparse crate.

Adding a symbolic link from `LICENSE.md -> ../LICENSE.md` inside the crates that are missing a license file should have `cargo package` and `cargo publish` pick up those files (unless you run `cargo publish` from Windows, I think). Thanks!

wez:
I'd rather fixup the manifests to reference `../LICENSE.md` to avoid polluting the filesystem in the repo

davide125:
Note that the top-level LICENSE has a bunch of extra provisions that probably don't apply to the sub crates.

decathorpe:
> I'd rather fixup the manifests to reference `../LICENSE.md` to avoid polluting the filesystem in the repo

As far as I know, this doesn't actually work, and will result in published crates that don't contain the referenced file, but only a reference to something that's not in the crate.

bew:
Can be closed I think since #6601 is merged

wez:
I don't think everything has been covered, so keeping this open seems reasonable

PR #6601, opened 2025-01-24
add a few missing license files

martinvonz:
I didn't symlink to the license file in the root because it mentions license that don't seem to apply to the crates I added license files to here.

I used the year each directory was created as the first copyright year.

I didn't add license files to all crates because I don't need them for what I'm working on right now, and I want to avoid having to figure out which licenses apply to which crates.

#2347


wez:
Thanks!

PR #6606, opened 2025-01-27
Add description and fuzzy_description arguments to ShowLauncherArgs

mgpinf:
Add `description`, `fuzzy_description` optional arguments in `ShowLauncherArgs`.
Similar to `description`, `fuzzy_description` arguments present in `InputSelector`.
If `description` argument is not provided, default value of `"Select an item and press Enter=launch Esc=cancel  /=filter"` is displayed
If `fuzzy_description` argument is not provided, default value of `"Fuzzy matching: "` is displayed

### DIFF
--- a/config/src/keyassignment.rs
+++ b/config/src/keyassignment.rs
@@ -17,6 +17,8 @@ use wezterm_term::SemanticType;
 pub struct LauncherActionArgs {
     pub flags: LauncherFlags,
     pub title: Option<String>,
+    pub help_text: Option<String>,
+    pub fuzzy_help_text: Option<String>,
 }
 
 bitflags::bitflags! {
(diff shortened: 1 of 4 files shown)


config/src/keyassignment.rs
```diff
@@ -17,6 +17,8 @@ use wezterm_term::SemanticType;
 pub struct LauncherActionArgs {
     pub flags: LauncherFlags,
     pub title: Option<String>,
+    pub description: Option<String>,
```
wez, line 20: Naming is the hardest part of this.  I think I regret using `title` for the tab title.
I think description is a bit too vague.

Rather than description, I think that maybe something like `help_text` is a bit more indicative of what this option does.

And then `fuzzy_help_text` for the fuzzy mode?

wez (changes requested):
Thanks for this! I think the code overall looks fine; one minor change to passing the params, and then a question on naming is really the only thing I want to flag and tweak.

PR #6609, opened 2025-01-28
fix: display whitespace before reset of attributes in InputSelector

mgpinf:
Display white space before reset of background and setting attributes to default in `InputSelector`. This is in order for whitespace after the option to be highlighted in the same background color as the currently selected option.
The above mentioned points are currently reflected in `ShowLauncherArgs` but not in `InputSelector`.

### DIFF
--- a/wezterm-gui/src/overlay/selector.rs
+++ b/wezterm-gui/src/overlay/selector.rs
@@ -155,11 +155,11 @@ impl SelectorState {
                 line.resize(max_width, termwiz::surface::SEQ_ZERO);
             }
             changes.append(&mut line.changes(&attr));
+            changes.push(Change::Text(" \r\n".to_string()));
             if entry_idx == self.active_idx {
                 changes.push(AttributeChange::Reverse(false).into());
             }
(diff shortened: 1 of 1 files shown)


mgpinf:
Current behavior in InputSelector:
Screenshot 2025-01-28 at 09 43 55

Current behavior in ShowLauncherArgs:
Screenshot 2025-01-28 at 09 44 50

wezterm-gui/src/overlay/selector.rs
```diff
@@ -155,11 +155,11 @@ impl SelectorState {
                 line.resize(max_width, termwiz::surface::SEQ_ZERO);
             }
             changes.append(&mut line.changes(&attr));
+            changes.push(Change::Text(" \r\n".to_string()));
```
wez, line 158: I think this should just be the space here, then emit the CRLF after the attributes have changed (where the `" \r\n"` was originally).

It's a subtle difference, but I think moving to the next line before the others are applied might behave slightly wonky in some edge cases.

```suggestion
            changes.push(Change::Text(" ".to_string()));
```

wez:
Thank you!

PR #6610, opened 2025-01-28
wezterm-ssh: make structs/fields public

iamazy:
I'm using `wezterm-ssh`, but some of its fields and structs are inaccessible from outside. I want to make their visibility public

### DIFF
--- a/wezterm-ssh/src/pty.rs
+++ b/wezterm-ssh/src/pty.rs
@@ -17,18 +17,18 @@ pub(crate) struct NewPty {
 }
 
 #[derive(Debug)]
-pub(crate) struct ResizePty {
+pub struct ResizePty {
     pub channel: ChannelId,
     pub size: PtySize,
 }
 
 #[derive(Debug)]
 pub struct SshPty {
-    pub(crate) channel: ChannelId,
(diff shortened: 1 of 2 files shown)


wez:
I'm not sure about this; it's making something of an API commitment, and it is revealing all of the internal implementation details for this struct.

Can you share more about what you want to do with those fields? Maybe there's a more supportable interface that could be provided instead?

iamazy:
Hi, @wez, Thanks for replying. I am building an application that needs to implement some `pty` interface for ssh (see [ssh-pty](https://github.com/iamazy/nxshell/blob/main/crates/egui-term/src/ssh/mod.rs)). However, the fields in `SshPty` in `wezterm-ssh` are not visible to the outside, so I hope to make them public

wez:
I understand that you want to make them public, but it is not a supportable change: it will either lock me into a specific set of implementation details, or will break you when I need to make a change in the future. If you can elaborate on specifically what you need from the interface, we can consider how to support that.

iamazy:
> If you can elaborate on specifically what you need from the interface, we can consider how to support that.

I just want to obtain the mutable and immutable references of the reader/writer in `SshPty`

PR #6626, opened 2025-02-02
feat: add cli flag to specify output file in `wezterm record`

### LINKED ISSUE #6625: Specify output file in `wezterm record`
When using `wezterm record`, there is no way to specify the output file.

Add a new flag to the cli `-o` that accepts a path and write the asciicast to that file when that flag is present, otherwise, keep writing to a random filename in the temporary directory.

I can always look in the temp directory and look for the newest file with the `wezterm-recording` but I think this is a better solution an easy to implement.

**Additional context**
If there are no issues with this feature, I'm willing to implement it myself, shouldn't be too hard.

Tyarel8:
Adds a new cli flag `-o` to `wezterm record` to specify the file where the asciicast will be saved.

Resolves #6625


bew:
Tested, works correctly 👍

Issue #1124, opened 2021-09-08
Center window contents


valpackett:
**Is your feature request related to a problem? Please describe.**

Unlike #291 I just want to be able to center the content in the window to avoid ugly asymmetrical padding on the bottom and right sides of the window. Or even align it to be horizontally centered but vertically always sticking to the bottom edge.

(BTW even if/when there would be snap-to-character-grid resizing, there would still be fullscreen/maximized mode…)

**Describe the solution you'd like**

It is sort of possible to achieve with scripting…

```lua
wezterm.on("window-resized", function(window, pane)
  local window_dims = window:get_dimensions();
  local overrides = window:get_config_overrides() or {}
  local new_padding = {
    left = math.floor((window_dims.pixel_width % 19) / 2),
    right = 0,
    top = math.floor((window_dims.pixel_height % 43) / 2),
    bottom = 0
  };
  overrides.window_padding = new_padding
  window:set_config_overrides(overrides)
end);
```

…but only with hardcoded cell size. So, I guess there should be a Lua call for getting character grid cell dimensions!

**Describe alternatives you've considered**

Or make this just a built-in mode (flag) to avoid having stuff in configs.

karimlevallois:
This is interesting. How would a use the above in my 3 pane setup (below) to add padding to my top right (terminal) split? Currently it's a bit far to the left and looks a bit out of place against my centred Neovim main pane and auto-sized Bottom right split

```
-- Startup using 3 panes, lvim, wezterm and btm
wezterm.on("gui-startup", function(cmd)
	local tab, pane, window = mux.spawn_window(cmd or {
		args = { "/opt/homebrew/bin/fish", "-c", "/Users/Karim/.local/bin/lvim" },
	})
	window:gui_window():toggle_fullscreen()
	-- Create a right side pane
	local right_pane = pane:split { direction = "Right", size = 0.4 }
	-- Split right pane into two, with new pane on bottom
	local bottom_pane = right_pane:split { direction = "Bottom", size = 0.9, args = { "/opt/homebrew/bin/btm" } }
	-- Activate primary left pane
	pane:activate()
end)
```

bartdorsey:
I really would like an option to "center" the terminal inside whatever padding you had set. Right now all the excess padding goes to the right and bottom.  The idea way for this to work is you set your padding and then any extra padding is divided equally between opposite sides.

wez:
[...]

PR #6629, opened 2025-02-04
Add config option to adjust window alignment

Shiphan:
Hi @wez, first time here.

This PR add a config option `window_alignment`, it can be used like so:
```lua
config.window_alignment = {
  horizontal = "Center",
  vertical = "Center",
}
```
one small concern is that the new option works fine with window padding, scroll bar and tab bar, but I don't know if there is something that i didn't tested.

related issue: #1124

### DIFF
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -521,6 +521,9 @@ pub struct Config {
     #[dynamic(default)]
     pub window_padding: WindowPadding,
 
+    #[dynamic(default)]
+    pub window_content_alignment: WindowContentAlignment,
+
     /// Specifies the path to a background image attachment file.
     /// The file can be any image format that the rust `image`
     /// crate is able to identify and load.
@@ -1895,6 +1898,28 @@ impl Default for WindowPadding {
             bottom: default_half_cell(),
         }
     }
+}
+
+#[derive(FromDynamic, ToDynamic, Clone, Copy, Debug, Default)]
+pub struct WindowContentAlignment {
+    pub horizontal: HorizontalWindowContentAlignment,
+    pub vertical: VerticalWindowContentAlignment,
+}
+
+#[derive(Debug, FromDynamic, ToDynamic, Clone, Copy, PartialEq, Eq, Default)]
+pub enum HorizontalWindowContentAlignment {
+    #[default]
+    Left,
+    Center,
+    Right,
+}
+
+#[derive(Debug, FromDynamic, ToDynamic, Clone, Copy, PartialEq, Eq, Default)]
+pub enum VerticalWindowContentAlignment {
+    #[default]
+    Top,
+    Center,
+    Bottom,
 }
 
 #[derive(FromDynamic, ToDynamic, Clone, Copy, Debug, PartialEq, Eq)]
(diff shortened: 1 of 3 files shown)


config/src/config.rs
```diff
@@ -1897,6 +1900,28 @@ impl Default for WindowPadding {
...
 }
 
+#[derive(FromDynamic, ToDynamic, Clone, Copy, Debug, Default)]
+pub struct WindowAlignment {
+    pub horizontal: HorizontalWindowAlingment,
+    pub vertical: VerticalWindowAlingment,
+}
+
+#[derive(Debug, FromDynamic, ToDynamic, Clone, Copy, PartialEq, Eq, Default)]
+pub enum HorizontalWindowAlingment {
```
bew, line 1910: Note that you have a typo on Alignment, same for the enum for vertical below

bew (commented):
A few comments ;)

Shiphan:
I had fix the typo and add a bit more docs, but I am still thinking about the naming
it seems in docs the "content" is been called "terminal cells" in other part of document, so maybe `cells_alignment`? but I also think `window_content_alignment` is more intuitive than this one...

wezterm-gui/src/termwindow/render/mod.rs
```diff
@@ -352,9 +355,43 @@ impl crate::TermWindow {
...
+                h_context.pixel_cell
+            } else {
+                padding_right.evaluate_as_pixels(h_context)
+            };
+        let vertical_gap = self.dimensions.pixel_height as f32
+            - self.terminal_size.pixel_height as f32
+            - padding_top
+            - padding_bottom
+            - if self.show_tab_bar {
+                self.tab_bar_pixel_height().unwrap()
```
wez, line 379: it's always desirable to avoid bare `.unwrap()`, as unwrap can panic and break or terminate the program

```suggestion
                self.tab_bar_pixel_height().unwrap_or(0.0)
```

config/src/config.rs
```diff
@@ -521,6 +521,9 @@ pub struct Config {
     #[dynamic(default)]
     pub window_padding: WindowPadding,
 
+    #[dynamic(default)]
+    pub window_alignment: WindowAlignment,
```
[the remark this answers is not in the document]
wez: Yeah, `window_content_alignment` feels less likely to be misunderstood by users. I'm open to other suggestions here too if you have a flash of insight, but I do think that `window_alignment` implies that it controls how the window itself might align to the desktop--that was what I thought this PR was about when I first saw the title.

docs/config/lua/config/window_alignment.md
```diff
@@ -0,0 +1,28 @@
+---
+tags:
+  - appearance
+---
+# `window_alignment`
+
+Controls the alignment of the terminal cells inside the window when window size is not a multiple of terminal cell size.
+In other words, this option controls where the additional gap will be.
```
wez, line 8: This sentence makes sense to you and me, but I think it needs more exposition for other users. Right now it talks about a gap without explaining what gap or where.

It should probably explain about how we have to paint in the excess pixels in that case, and that results in a gap.  Then I think it makes it clearer to say that this option allows the user to control where that gap is.

wez (commented):
Thanks for this!

wez:
Please note that the CI is currently a bit wonky due to the move to a github org; it should get resolved in a couple of days.

I think that if you rebase your base PR you will at least get some more reasonable results from the cirrus CI

wez:
Thank you!

PR #6641, opened 2025-02-08
docs: Try fix mkdocs rendering of nested lists

bew:
Nested lists rendering with mkdocs is basically completely broken at the moment.

This happens in these pages:
- https://wezterm.org/config/lua/wezterm/font.html
- https://wezterm.org/config/lua/keyassignment/ActivateKeyTable.html
  (bulletpoint under `timeout_milliseconds` should be indented)

I remember from last time I used mkdocs (~2years ago) that we had to add spaces around (nested) lists to 'unbreak' the nested list rendering.

But at the same time https://github.com/squidfunk/mkdocs-material/issues/6509 seems to mention we need to have 4 indent spaces instead of 2.

I can't run the script to build the docs at the moment to verify this 😬.
Maybe someone else can check this out 👀


wez:
I tweaked it a bit and pushed this as cb593f6af5485e0de29ccc19ef27fc1416a85eef
Thanks!

bew:
Spaces it is then 👍

PR #6648, opened 2025-02-10
Reset attributes before emitting CRLF in the Launcher Menu

mgpinf:
Below changes are made:
- Emit CRLF after attributes have been reversed
- Reset Attributes before emitting CRLF in order to handle cases where formatted string is passed against `label` field in `SpawnCommand` object entries in `config.launch_menu`

### DIFF
--- a/wezterm-gui/src/overlay/launcher.rs
+++ b/wezterm-gui/src/overlay/launcher.rs
@@ -410,6 +410,7 @@ impl LauncherState {
             if entry_idx == self.active_idx {
                 changes.push(AttributeChange::Reverse(false).into());
             }
+            changes.push(Change::AllAttributes(CellAttributes::default()));
             changes.push(Change::Text("\r\n".to_string()));
         }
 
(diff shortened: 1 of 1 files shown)


bew:
Hello, could you please explain the problem, why you're proposing these changes?

mgpinf:
@bew 

For the below commit:
https://github.com/wezterm/wezterm/pull/6648/commits/f067c44ac6b23605769bdd2cae37883577496a9f

I had pushed the same change for InputSelector based on the suggestion I received:
https://github.com/wezterm/wezterm/pull/6609#discussion_r1947835818

For the below commit:
https://github.com/wezterm/wezterm/pull/6648/commits/b882fcfa75224ef566b5c95f01c1c698296ec89b

For most cases we don't require this, but since launch_menu items are configurable and it contains a list of SpawnCommand objects, and SpawnCommand has a an optional field to show in the launcher called 'label', this change handles cases where user passes a value formatted
using wezterm.format against 'label'.
As in the attributes reset is required in order to display the lines that follow it in default colors.

wez:
Thank you!

Issue #2831, opened 2022-12-04
Gnome: Allow double click on title bar to maximize the window


jankatins:
**Is your feature request related to a problem? Please describe.**

I'm used to double clicking the title bar of a gnome app (and also a mac one, but right now I'm an a GNOME laptop) to maximize a window. wezterm doesn't allow me that.

I think I remember not having that problem a while ago?

My gnome is configured to do this:

```
λ  gsettings get org.gnome.desktop.wm.preferences action-double-click-titlebar
'toggle-maximize'
```

**Describe the solution you'd like**

I would like to have a double click in the title bar maximize the wezterm window.

**Describe alternatives you've considered**

I could click the maximize button, but it's a much smaller target than the whole title bar. I can also use a custom key combo (WIN+UP in my case)

wez:
Can you share more about the environment in which you are running?

If you are running in wayland mode with a compositor that doesn't support server side decorations (eg: Mutter doesn't), then you are out of luck. This kind of titlebar action is handled by the window manager/compositor.  For Wayland with client-side decorations, there is no standard protocol for sharing this kind of window environment policy with the application, unless the application is a GTK program and happens to cheat and encode all the same logic on the client side.

I recommend `enable_wayland=false` to cause wezterm to run under XWayland instead, which will effectively enable server side decorations and should make this work again.

jankatins:
Gnome reports that it is running Wayland and I run whatever debian gives me when I install gnome. So I'm probably out of luck, I guess?

```
λ  wmctrl -m
Name: GNOME Shell
```

I noticed that Chrome also is a bit sluggish: if my perception is not misleading me, I have to tripple click... gitkraken (an electron app) works on double click. No idea if that is actually running in a gtk window under the hood?

valpackett:
> happens to cheat and encode all the same logic on the client side

That's not cheating, that's kinda what you're supposed to do. Alacritty does implement that for example.

Miuler:
With the configuration `enable_wayland=false` the redrawing in resize window is slow.

PR #6657, opened 2025-02-11
Improve dragging and double-clicking for maximize on the tab bar

Fidelxyz:
- Fix the bug where double-clicking on the tab bar would not maximize the window if extra `window_decorations` flag other than `INTEGRATED_BUTTONS | RESIZE` were set (e.g. `INTEGRATED_BUTTONS | RESIZE | MACOS_FORCE_ENABLE_SHADOW`)
- Allow dragging on the tab bar to restore a maximized window, consistent with OS behavior in handling dragging windows

https://github.com/user-attachments/assets/615b5581-4c61-4541-8ab0-5f8ac05e7a10

### DIFF
--- a/wezterm-gui/src/termwindow/mouseevent.rs
+++ b/wezterm-gui/src/termwindow/mouseevent.rs
@@ -472,9 +472,9 @@ impl super::TermWindow {
                         .window_state
                         .intersects(WindowState::MAXIMIZED | WindowState::FULL_SCREEN);
                     if let Some(ref window) = self.window {
-                        if self.config.window_decorations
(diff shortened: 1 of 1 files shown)


wez:
Thanks for this!

@jankatins: would you care to try this to see if it resolves https://github.com/wezterm/wezterm/issues/2831 for you?

@tmccombs: I would appreciate you taking a quick peek just to make sure this won't collide with your PRs; it looks distinct from the code you're changing, but I'd like to make sure that it's compatible so that you're not surprised by it if it does happen to impact it!

tmccombs:
This shouldn't collide with my changes.

PR #6661, opened 2025-02-11
glyph: Fix glyph '┽', was rendering as '┥'

bew:
image

### DIFF
--- a/wezterm-gui/src/customglyph.rs
+++ b/wezterm-gui/src/customglyph.rs
@@ -1776,7 +1776,7 @@ impl BlockKey {
                 },
                 Poly {
                     path: &[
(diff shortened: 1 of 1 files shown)


wez:
Thank you!

wez:
heh, did a double take when reviewing the changelog, because of the bug:
image

bew:
> heh, did a double take when reviewing the changelog, because of the bug: image

I'm sorry I don't understand the goal of that message+screenshot, are you saying it doesn't actually work for you?

wez:
it's fine: I was running a build without your change and had to double check that I'd written the correct text for the changelog!

bew:
Ahah I had the same double-take while writing the commit message 😅

PR #6677, opened 2025-02-14
Add alphabet to launcher args

mgpinf:
This PR adds the below changes:
- Similar to `InputSelector`, the labels for the Launcher are generated based on the provided argument or the default value
- Add configuration option `launcher_alphabet` for Launcher alphabet.
  - This is in order to be able to configure the alphabet for actions corresponding to KeyAssignment enums `ShowLauncher`, `ShowTabNavigator`
  - If `launcher_alphabet` configuration option not set, the value defaults to `"1234567890abcdefghilmnopqrstuvwxyz"`
- Optional argument `alphabet` is added to `ShowLauncherArgs`
- The `alphabet` string value defaults to the same value as the configuration option `launcher_alphabet` 
- Push `(max_label_len+3)` whitespaces if always_fuzzy false in InputSelector
  - This is to handle edge case in `InputSelector` when `max_label_len` is 2 and the labels shift one place to the left when entering fuzzy mode and back one place to the right when exiting fuzzy mode
- Push `(max_label_len+3)` whitespaces if `FUZZY` not present in `args` argument of the Launcher. This is to handle the edge case in the Launcher mentioned in the previous point

### DIFF
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -414,6 +414,9 @@ pub struct Config {
     pub disable_default_key_bindings: bool,
     pub leader: Option<LeaderKey>,
 
+    #[dynamic(default = "default_num_alphabet")]
+    pub launcher_alphabet: String,
+
     #[dynamic(default)]
     pub disable_default_quick_select_patterns: bool,
     #[dynamic(default)]
@@ -1802,6 +1805,10 @@ fn default_status_update_interval() -> u64 {
 
 fn default_alternate_buffer_wheel_scroll_speed() -> u8 {
     3
+}
+
+fn default_num_alphabet() -> String {
+    "1234567890abcdefghilmnopqrstuvwxyz".to_string()
 }
 
 fn default_alphabet() -> String {
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -1808,6 +1808,7 @@ fn default_alternate_buffer_wheel_scroll_speed() -> u8 {
 }
 
 fn default_num_alphabet() -> String {
+    // Note: vi motion keys are intentionally excluded from this alphabet
     "1234567890abcdefghilmnopqrstuvwxyz".to_string()
 }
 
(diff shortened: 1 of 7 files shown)


mgpinf:
Have added the `InputSelector` commit in this PR because it is a related to the alphabet and a small change

config/src/config.rs
```diff
@@ -1804,6 +1807,10 @@ fn default_alternate_buffer_wheel_scroll_speed() -> u8 {
     3
 }
 
+fn default_num_alphabet() -> String {
+    "1234567890abcdefghilmnopqrstuvwxyz".to_string()
```
wez, line 1811:
```suggestion
    // Note: vi motion keys are intentionally excluded from this alphabet
    "1234567890abcdefghilmnopqrstuvwxyz".to_string()
```

docs/config/lua/config/launcher_alphabet.md
```diff
@@ -0,0 +1,13 @@
+---
+tags:
+  - launcher
+---
+# `launcher_alphabet`
+
+{{since('nightly')}}
+
+Specify a string of unique characters. The characters in the string are used
+to calculate one or two click shortcuts that can be used to quickly choose from
```
wez, line 10:
```suggestion
to calculate one or two key press shortcuts that can be used to quickly choose from
```

wezterm-gui/src/overlay/launcher.rs
```diff
@@ -5,6 +5,7 @@
 //! be rendered as a popup/context menu if the system supports it; at the
 //! time of writing our window layer doesn't provide an API for context
 //! menus.
+use super::quickselect;
```
wez, line 8: I prefer to use `crate::` relative paths for this sort of import; it makes it a little easier to refactor in the future

wez (commented):
Doh, I just saw that the review I wrote for this was pending all this time!
A few minor comments here, but otherwise this looks good!

wez:
Thank you!

PR #6682, opened 2025-02-15
Add configuration option for InputSelector label background and foreground color

mgpinf:
Add below configuration option in config.colors:
- `input_selector_label_bg`
- `input_selector_label_fg`

If above mentioned options are not set in the configuration file then:
- `input_selector_label_bg` defaults to default background color
- `input_selector_label_fg` defaults to default foreground color


wez (commented):
Looks good; just a minor comment!

wez:
Thank you!

PR #6683, opened 2025-02-16
Option to disable attributes for non matching text in quick select mode

### LINKED ISSUE #4022: Disable terminal colors temporarily in quick select mode
It is a bit too noisy with colors on when trying to select something. Kitty by default turns colors all off in quick select.

Of course it is possible to write my own function to turn off colors, do quick select, and turn it back on, but it would much easier if there is an option to just turn off colors in quick select mode.

mgpinf:
Add a configuration option `disable_attributes_for_non_matching_text_in_quick_select` that disables all attributes for non-matching text in Quick Select Mode.

The default value for `disable_attributes_for_non_matching_text_in_quick_select` is `false` thereby keeping the existing behavior as default.

This resolves #4022


mgpinf:
Hi Wez. I have few doubts

- Could you suggest a better name for the configuration flag? I feel this name is too long, but at the same time don't know of a shorter name that clarifies the intention properly.
- Is it ok to add bold attribute to the quick_select label text and quick_select match text? This would further emphasize it.

bew:
Hello, if you're still working on the PR, can you mark it as a 'draft' PR, and only set it as 'ready for review' when you think it's finished?

wezterm-gui/src/overlay/quickselect.rs
```diff
@@ -575,47 +579,72 @@ impl Pane for QuickSelectOverlay {
...
+                                    .set_intensity(Intensity::Normal)
+                                    .set_underline(Underline::None)
+                                    .set_italic(false)
+                                    .set_blink(Blink::None)
+                                    .set_reverse(false)
+                                    .set_strikethrough(false)
+                                    .set_invisible(false)
+                                    .set_foreground(ColorAttribute::Default)
+                                    .set_background(ColorAttribute::Default)
+                                    .set_hyperlink(None);
```
mgpinf, line 595: Is there a better way to reset all attributes? Or does this work?

wezterm-gui/src/overlay/quickselect.rs
```diff
@@ -575,47 +579,72 @@ impl Pane for QuickSelectOverlay {
...
-                                            colors
-                                                .quick_select_match_fg
-                                                .unwrap_or(AnsiColor::Green.into()),
-                                        )
-                                        .set_reverse(false);
-                                }
+                    } else {
+                        if disable_attr_for_non_matching_text {
+                            for cell in line.cells_mut_for_attr_changes_only().iter_mut() {
+                                cell.attrs_mut()
```
mgpinf, line 585: Could you help with a better way to reset attributes for a cell? Or is this fine?

wezterm-gui/src/overlay/quickselect.rs
```diff
@@ -575,47 +579,72 @@ impl Pane for QuickSelectOverlay {
...
+                                    .set_intensity(Intensity::Normal)
+                                    .set_underline(Underline::None)
+                                    .set_blink(Blink::None)
+                                    .set_italic(false)
```
mgpinf, line 589: Is there a better way to clear attributes for a cell? Or is this fine?

mgpinf: I have added a public method in `termwiz/src/cell.rs` to clear all attributes from a cell.

mgpinf:
@bew This PR is ready for review now

wez:
> * Is it ok to add bold attribute to the quick_select label text and quick_select match text? This would further emphasize it.

Can you share screenshots contrasting the before/after, and for the after, how it looks both when this option is enabled and disabled?

Either way, I think that's a separate thesis and probably should be a separate PR from this one: I did find it hard to review because the diff was much larger than I expected due to indentation changing.

wezterm-gui/src/overlay/quickselect.rs
```diff
@@ -616,6 +623,8 @@ impl Pane for QuickSelectOverlay {
                             }
                         }
                         line.clear_appdata();
+                    } else if disable_attr {
+                        line.clear_appdata();
```
wez, line 627: should this go together with the attribute cleaing on line 559? Feels like it should, but not sure if you put it here because it only needs to go here

wezterm-gui/src/overlay/quickselect.rs
```diff
@@ -642,6 +652,11 @@ impl Pane for QuickSelectOverlay {
             let stable_idx = idx as StableRowIndex + top;
             renderer.dirty_results.remove(stable_idx);
             if stable_idx == search_row {
+                if disable_attr {
+                    line.cells_mut_for_attr_changes_only()
```
wez, line 656: hmm, should this go here? we always rewrite this particular line with the search instructions. Should this move to line 652 instead, so it happens at the top of this loop?

mgpinf:
> Can you share screenshots contrasting the before/after, and for the after, how it looks both when this option is enabled and disabled?

Screenshots:

Git log without QuickSelect overlay:
Screenshot from 2025-02-23 23-39-29


Git log with QuickSelect overlay when `quick_select_remove_styling` option is disabled:
Screenshot from 2025-02-23 23-39-44


Git log with QuickSelect overlay when `quick_select_remove_styling` option is enabled:
Screenshot from 2025-02-23 23-40-15

wezterm-gui/src/overlay/quickselect.rs
mgpinf: I have pushed the changes to this. I have a doubt, when is the `get_lines` method called? I see that the `with_lines_mut` method is called during Quick Select mode

wez:
re: screenshots, thanks for sharing! I was mostly thinking about the question of whether we should adjust the intensity or not; when you put that into a PR, that's the thing I'd like to compare with the screenshots here to help decide.

wez:
Thank you!

PR #6689, opened 2025-02-17
Adapt get-deps for NetBSD, and add NetBSD installation instructions.

0-wiz-0:
There is one remaining (minor) issue: NetBSD's sort doesn't support `--version-sort`, so when running `get-deps` when that runs `ci/check-rust-version.sh`, you'll get:
```
sort: unknown option -- -
usage: sort [-bdfHilmnrSsu] [-k kstart[,kend]] [-o output] [-R char] [-T dir]
             [-t char] [file ...]
   or: sort -C|-c [-bdfilnru] [-k kstart[,kend]] [-o output] [-R char]
             [-t char] [file]
Installed rustc version '1.82.0' is less than required '1.71.0'

Check if your OS provides newer version of Rust, if not
use rustup to manage installed versions of Rust
https://www.rust-lang.org/en-US/install.html

See https://wezterm.org/install/source.html for complete
installation instructions for wezterm
```

The link to the package (in the installation instructions for NetBSD) will start working in a couple hours.

I packaged git head as of 20250213-133349-ee0c04e7, since the last release was too old.

### DIFF
--- a/docs/features.md
+++ b/docs/features.md
@@ -5,7 +5,7 @@ hide:
 
 ## Available Features
 
-* Runs on Linux, macOS, Windows 10 and FreeBSD
+* Runs on Linux, macOS, Windows 10, FreeBSD and NetBSD
 * [Multiplex terminal panes, tabs and windows on local and remote hosts, with native mouse and scrollback](multiplexing.md)
 * <a href="https://github.com/tonsky/FiraCode#fira-code-monospaced-font-with-programming-ligatures">Ligatures</a>, Color Emoji and font fallback, with true color and [dynamic color schemes](config/appearance.md).
 * [Hyperlinks](hyperlinks.md)
--- a/docs/index.md
+++ b/docs/index.md
@@ -11,7 +11,7 @@ hide:
 
 ## Features
 
-* Runs on Linux, macOS, Windows 10 and FreeBSD
+* Runs on Linux, macOS, Windows 10, FreeBSD and NetBSD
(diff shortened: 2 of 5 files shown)


0-wiz-0:
Oh, and I was wondering if `sync-color-schemes` and `generate-bidi` should be installed or not?

wez:
> Oh, and I was wondering if `sync-color-schemes` and `generate-bidi` should be installed or not?

See https://github.com/wezterm/wezterm/blob/main/README-DISTRO-MAINTAINER.md for info about that sort of thing

0-wiz-0:
Thanks for the pointer - I had found that file, but it doesn't mention the two programs at all, so I was wondering.
(The package doesn't install them for now.)

0-wiz-0:
Btw, [NetBSD CI](https://github.com/vmactions/netbsd-vm) are available on github if you want to set one up.

wez:
> Btw, [NetBSD CI](https://github.com/vmactions/netbsd-vm) are available on github if you want to set one up.

Oh, interesting!  I'll take a PR that runs the build and tests if you want to put it together :)

I'm not sure about having the CI here produce installable binaries/packages for NetBSD (like we do for most other systems), because I don't understand the full scope of that setup right now, and I'm honestly pretty bandwidth constrained.  If it would be useful for CI to do that, and you're willing to help set that up, I'd take PR(s) for that too :)

wez:
Thank you!

PR #6696, opened 2025-02-19
Add font configuration option for ActivateCommandPalette, CharSelect, PaneSelect

mgpinf:
Add configuration option for font for the below entities:
- `ActivateCommandPalette`
- `CharSelect`
- `PaneSelect`

For the above entities, if the font is not set in configuration, it defaults to the font same font as `config.window_frame.font`

### DIFF
--- a/docs/config/lua/config/char_select_font.md
+++ b/docs/config/lua/config/char_select_font.md
@@ -1,0 +1,19 @@
+---
+tags:
+  - appearance
+  - char_select
+  - font
+---
+# `char_select_font`
+
+{{since('nightly')}}
+
+Configures the font to use for character selection. The `char_select_font`
+setting can specify a set of fallbacks and other options, and is described
+in more detail in the [Fonts](../../fonts.md) section.
+
+If not specified, the font is same as the font in `window_frame.font`
+
+You will typically use [wezterm.font](../wezterm/font.md) or
+[wezterm.font_with_fallback](../wezterm/font_with_fallback.md) to specify the
+font.
--- a/docs/config/lua/config/char_select_font.md
+++ b/docs/config/lua/config/char_select_font.md
@@ -17,3 +17,9 @@ If not specified, the font is same as the font in `window_frame.font`
 You will typically use [wezterm.font](../wezterm/font.md) or
 [wezterm.font_with_fallback](../wezterm/font_with_fallback.md) to specify the
 font.
+
+To specify `char_select_font`:
+
+```lua
+config.char_select_font = wezterm.font 'Roboto'
+```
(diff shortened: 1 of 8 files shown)


docs/config/lua/config/char_select_font.md
```diff
@@ -0,0 +1,19 @@
...
+
+Configures the font to use for character selection. The `char_select_font`
+setting can specify a set of fallbacks and other options, and is described
+in more detail in the [Fonts](../../fonts.md) section.
+
+If not specified, the font is same as the font in `window_frame.font`
+
+You will typically use [wezterm.font](../wezterm/font.md) or
+[wezterm.font_with_fallback](../wezterm/font_with_fallback.md) to specify the
+font.
```
wez, line 19: Can you show an actual example here? I think that makes the docs feel more accessible.
I typically will just show how to replicate the default config.

Something like this?

```
    ```lua
    config.char_select_font = wezterm.font 'Roboto'
    ```
```

same for the other doc pages

wezterm-font/src/lib.rs
```diff
@@ -655,8 +678,12 @@ impl FontConfigInner {
             return Ok(Rc::clone(entry));
         }
 
-        let loaded =
-            self.make_title_font_impl(myself, Some(config.command_palette_font_size), false)?;
+        let loaded = self.make_entity_font_impl(
+            myself,
+            Some(config.command_palette_font_size),
```
wez, line 683: since you have `Entity`, you can resolve this value in your impl as well

wezterm-font/src/lib.rs
```diff
@@ -584,18 +591,33 @@ impl FontConfigInner {
...
+        entity: Entity,
     ) -> anyhow::Result<Rc<LoadedFont>> {
         let config = self.config.borrow();
         let (sys_font, sys_size) = self.compute_title_font(&config, make_bold);
 
         let font_size = pref_size.unwrap_or(sys_size);
 
-        let text_style = config.window_frame.font.as_ref().unwrap_or(&sys_font);
+        let text_style = match entity {
+            Entity::Title => config.window_frame.font.as_ref().unwrap_or(&sys_font),
```
wez, line 607: I think you can eliminate the `pref_size` param and make this more succinct:

```rust
        let (pref_size, text_style) = match entity {
            Entity::Title => (config.window_frame.font_size, &config.window_frame.font),
            ...
         };

         let font_size = pref_size.unwrap_or(sys_size);
         let text_style = text_style.unwrap_or(&sys_font);
```

mgpinf:
As of now I have made `char_select_font`, `command_palette_font`, `pane_select_font` default value to be `config.window_frame` font in order to avoid breaking changes. Should I keep it that way or default to system font rather than relying on `config.window_frame.font`?

wez:
What you have now sounds good to me; preserving the current behavior on upgrade is desirable

wez:
Thank you!

PR #6700, opened 2025-02-20
shift check bit when comparing to pattern

digitallyserviced:
There seems to have been a regression when rendering sextant characters when `custom_block_glyphs=true` in the new "dynamic" rendering of bit/marched glyphs like sextant/octants.

Octants has the proper bit shifting when comparing, however this did not seem to get utilized for sextants.

Left is regressed, right is fixed
Screenshot_2025-02-19-22-17-25_1920x1080


wez:
Thank you!

PR #6706, opened 2025-02-21
flake.nix: fix build on darwin

zaninime:
The `openssl` dependency is also needed on Darwin.


bew (approved):
Tested on macos, works well and indeed fixes the build 👍

wez:
Thank you!

PR #6707, opened 2025-02-21
Add Confirmation to Key assignment

mgpinf:
A confirmation menu similar to close menu is displayed.
If `yes` is selected, a custom event is emitted which causes the provided callback to be called.
If `no` is selected, no further action takes place.
The callback accepts parameters `window, pane`

### DIFF
--- a/docs/config/lua/keyassignment/Confirmation.md
+++ b/docs/config/lua/keyassignment/Confirmation.md
@@ -1,0 +1,43 @@
+# `Confirmation`
+
+{{since('nightly')}}
+
+Activates an overlay to display a confirmation menu
+
+When the user accepts a line, emits an event that allows you to act
+upon the input.
+
+`Confirmation` accepts the following fields:
+
+* `message` - the text to show for confirmation. You may embed
+  escape sequences and/or use [wezterm.format](../wezterm/format.md).
+  Defaults to: `"🛑 Really continue?"`.
+
+## Example of choosing a program with user confirmation
+
+```lua
+local wezterm = require 'wezterm'
+local act = wezterm.action
+local config = wezterm.config_builder()
+
+config.keys = {
+  {
+    key = 'E',
+    mods = 'CTRL|SHIFT',
+    action = act.Confirmation {
+      message = "Do you want to run htop in a new window?",
+      action = wezterm.action_callback(function(window, pane)
+        window:perform_action(act.SpawnCommandInNewWindow { args = { 'htop' } }, pane)
+      end),
+    },
+  },
+}
+
+return config
+```
+
+
+
+
+See also [InputSelector](InputSelector.md).
+See also [PromptInputLine](PromptInputLine.md).
--- a/docs/config/lua/keyassignment/Confirmation.md
+++ b/docs/config/lua/keyassignment/Confirmation.md
@@ -40,13 +40,16 @@ config.keys = {
     key = 'E',
     mods = 'CTRL|SHIFT',
     action = act.Confirmation {
-      message = "Do you want to run htop in a new window?",
+      message = 'Do you want to run htop in a new window?',
       action = wezterm.action_callback(function(window, pane)
-        window:perform_action(act.SpawnCommandInNewWindow { args = { 'htop' } }, pane)
+        window:perform_action(
+          act.SpawnCommandInNewWindow { args = { 'htop' } },
+          pane
+        )
       end),
-      cancel = wezterm.action_callback(function(window,pane)
-        wezterm.log_error("user declined")
-      end)
+      cancel = wezterm.action_callback(function(window, pane)
+        wezterm.log_error 'user declined'
+      end),
     },
   },
 }
--- a/docs/config/lua/keyassignment/Confirmation.md
+++ b/docs/config/lua/keyassignment/Confirmation.md
@@ -18,7 +18,7 @@ upon the input.
   objects from the current pane and window. This callback is called when the
   user selects `Yes`.
 * `cancel` - event callback registered via `wezterm.action_callback`.  The
-  callback's function signature is `(window, pane, id, label)` where `window` and
+  callback's function signature is `(window, pane)` where `window` and
   `pane` are the [Window](../window/index.md) and [Pane](../pane/index.md).
   This is an optional argument. If present, this callback is called when the
   user selects `No` or closes the confirmation menu.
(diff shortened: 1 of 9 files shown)


mgpinf:
Hi Wez. Which of the below options do you think is better?
- Perform action irrespective of yes, no with an extra argument in the callback called `confirm` which is true/false.
- Only call the callback if confirmation is true (current implementation).

wez:
> Hi Wez. Which of the below options do you think is better?
> 
> * Perform action irrespective of yes, no with an extra argument in the callback called `confirm` which is true/false.
> * Only call the callback if confirmation is true (current implementation).

Third option, is to allow an optional cancellation action that is triggered only if they cancel/decline; that would keep the parameter list a bit simpler overall:

```lua
config.keys = {
  {
    key = 'E',
    mods = 'CTRL|SHIFT',
    action = act.Confirmation {
      message = "Do you want to run htop in a new window?",
      action = wezterm.action_callback(function(window, pane)
        window:perform_action(act.SpawnCommandInNewWindow { args = { 'htop' } }, pane)
      end),
      cancel = wezterm.action_callback(function(window, pane)
        print("user declined")
      end),
    },
  },
}
```

wezterm-gui/src/overlay/confirm.rs
```diff
@@ -0,0 +1,216 @@
...
+    use termwiz::surface::TextureCoordinate;
+
+        let image_data = Arc::new(ImageData::with_raw_data(ICON_DATA.to_vec()));
+        let decoded_image = image::load_from_memory(ICON_DATA)?;
+        let logo_width_cells = 3usize;
+        let logo_width = logo_width_cells * size.xpixel;
+        let logo_scale = logo_width as f32 / decoded_image.width() as f32;
+        let logo_height = decoded_image.height() as f32 * logo_scale;
+        let logo_height_cells = (logo_height / size.ypixel as f32).ceil() as usize;
+        */
```
wez, line 35: I know this is moving from existing code, but I think it's fine to just delete this dead commented code; it won't serve any purpose over here

wezterm-gui/src/overlay/confirm.rs
```diff
@@ -0,0 +1,216 @@
...
+            Change::CursorVisibility(CursorVisibility::Hidden),
+            /*
+            Change::Image(termwiz::surface::change::Image {
+                width: logo_width_cells,
+                height: logo_height_cells,
+                top_left: TextureCoordinate::new_f32(0., 0.),
+                bottom_right: TextureCoordinate::new_f32(1., 1.),
+                image: Arc::clone(&image_data),
+            }),
+            */
```
wez, line 78:
```suggestion
```

docs/config/lua/keyassignment/Confirmation.md
```diff
@@ -0,0 +1,56 @@
...
+When the user accepts a line, emits an event that allows you to act
+upon the input.
+
+`Confirmation` accepts the following fields:
+
+* `message` - the text to show for confirmation. You may embed
+  escape sequences and/or use [wezterm.format](../wezterm/format.md).
+  Defaults to: `"🛑 Really continue?"`.
+* `action` - event callback registered via `wezterm.action_callback`.  The
+  callback's function signature is `(window, pane, id, label)` where `window` and
```
wez, line 16:
```suggestion
  callback's function signature is `(window, pane)` where `window` and
```

docs/config/lua/keyassignment/Confirmation.md
```diff
@@ -0,0 +1,56 @@
...
+  `pane` are the [Window](../window/index.md) and [Pane](../pane/index.md)
+  objects from the current pane and window. This callback is called when the
+  user selects `Yes`.
+* `cancel` - event callback registered via `wezterm.action_callback`.  The
+  callback's function signature is `(window, pane, id, label)` where `window` and
```
wez, line 21:
```suggestion
  callback's function signature is `(window, pane)` where `window` and
```

docs/config/lua/keyassignment/Confirmation.md
```diff
@@ -0,0 +1,56 @@
...
+config.keys = {
+  {
+    key = 'E',
+    mods = 'CTRL|SHIFT',
+    action = act.Confirmation {
+      message = "Do you want to run htop in a new window?",
+      action = wezterm.action_callback(function(window, pane)
+        window:perform_action(act.SpawnCommandInNewWindow { args = { 'htop' } }, pane)
+      end),
+      cancel = wezterm.action_callback(function(window,pane)
```
wez, line 42: if you run `./ci/build-docs.sh` it should format the lua code in the example

docs/config/lua/keyassignment/Confirmation.md
```diff
@@ -0,0 +1,56 @@
...
+}
+
+return config
+```
+
+
+
+
+See also [InputSelector](InputSelector.md).
+See also [PromptInputLine](PromptInputLine.md).
```
wez, line 56: I'd suggest turning this into a bulleted list:
```suggestion
See also:
   * [InputSelector](InputSelector.md).
   * [PromptInputLine](PromptInputLine.md).
```

docs/config/lua/keyassignment/InputSelector.md
```diff
@@ -232,4 +232,5 @@ return config
 
 
 See also [PromptInputLine](PromptInputLine.md).
+See also [Confirmation](Confirmation.md).
```
wez, line 235: make this a list as well

docs/config/lua/keyassignment/PromptInputLine.md
```diff
@@ -96,3 +96,4 @@ return config
 ```
 
 See also [InputSelector](InputSelector.md).
+See also [Confirmation](Confirmation.md).
```
wez, line 99: list here also

docs/config/lua/keyassignment/Confirmation.md
```diff
@@ -0,0 +1,56 @@
+# `Confirmation`
```
wez, line 1: Let's add a `prompt` tag to this page, and also to the the other prompting doc pages; that way the docs will group them together and it become less important to maintain the `See Also` links at the bottom.

If you can think of a better tag than `prompt`, please suggest/use it :)

```suggestion
---
tags:
  - prompt
---

# `Confirmation`
```

mgpinf: I think `prompt` is good.

wez:
Thank you!

PR #6709, opened 2025-02-21
Update termwiz Changelog with git-cliff

joshka:
This is not perfect, but maybe useful. Obvious downside is that the manually added breaking changes note is omitted.

In https://github.com/ratatui/ratatui/pull/1682, I noticed that there wasn't a full changelog for termwiz, which would have been useful to link to when updating the ratatui termwiz backend. I went through the termwiz history and tagged the various versions that I was able to determine with some accuracy. There's a couple (0.72, and 0.15.0) that I wasn't sure of, but they're old releases and the impact is that these commits are just rolled into the next release.

If you want to replicate locally, here's the relevant tags / commits:

```
91ea7cd576dc785e8b4c7f105ace1ef70036085f refs/tags/termwiz-0.1.0
3c6552b975ae22008a98baa0ff34086b7a6a0594 refs/tags/termwiz-0.2.0
00c1000caa97ec84a6b15cc7043275e0657dcea5 refs/tags/termwiz-0.3.0
e9780d6a342a7905ccd02c99a5c846e3ed3a8ccd refs/tags/termwiz-0.3.1
d18383416ee23b8c1e8525de5ed973b699f09c21 refs/tags/termwiz-0.4.0
662227a7257ea30e6f6355490e0529f4440fc17c refs/tags/termwiz-0.5.0
51545ab0992608d48142bbb2e05f6dcef8b200b3 refs/tags/termwiz-0.6.0
3dea5a59fce5a671b127eab1ea9fda062d320c1b refs/tags/termwiz-0.7.0
fb6d0bbc7cd10ff5845cb6504c8c45e51ab88383 refs/tags/termwiz-0.7.1
4858194ef4d6a093156c8eaed7453f43be32cae1 refs/tags/termwiz-0.8.0
1ad6f00696ea7d4588ee7eeb93f4189f155871fa refs/tags/termwiz-0.9.0
86a0f0939c522851587338cf0f12dc4c1d97acad refs/tags/termwiz-0.12.0
b0f5e673c6e1033d64304b1546205474c9850ac2 refs/tags/termwiz-0.13.0
c9c6d869f4c6411014efbab82e0db538d789e2b5 refs/tags/termwiz-0.16.0
673ca3ad077956e345b2bcdcdcdb09079dceb1f7 refs/tags/termwiz-0.17.0
33866dad7065c1a2e622a356639e8d337c40241f refs/tags/termwiz-0.17.1
c91d6c396c010f7adfde022488a57f2f0b3002fd refs/tags/termwiz-0.18.0
a6fc9323e20906d6b7836697b5ba6b4be08b2d1d refs/tags/termwiz-0.19.0
3509e210cea534fcf374b31f3d9162190af5c5e4 refs/tags/termwiz-0.20.0
1d4d6df045054704da87c91d17f281604b62a691 refs/tags/termwiz-0.22.0
958804c8ac16b750954b3bcaaf847e7498cd2205 refs/tags/termwiz-0.23.0
```

The git cliff confg file is the default file from git-cliff --init with no modifications. The changelog is generated using:

```shell
cd termwiz
# the --tag-pattern arg ignores several other tags like`t-parse-*` that are applied to commits that changed the termwiz folder.
git cliff --tag-pattern 'termwiz-*' -o CHANGELOG.md
```


joshka:
@orhun wrote in https://github.com/ratatui/ratatui/pull/1682#issuecomment-2675603758

> Yup, that's not very bad - similar to what I got. I also recommend setting tag_pattern to termwiz-* in the config file.

orhun (approved):

wez:
Thank you!

EvanCarroll:
I was referenced this commit from a convo on git cliff. If this isn't the right place, ignore this message (or ask and i'll delete this).

I'm confused at the current workflow and what is desired here.

```
git cliff --tag-pattern 'termwiz-*' -o CHANGELOG.md
```

Will generate the changelog **after** the version is already tagged. What good is that? Why would you want the changelog _not_ in the tagged version.

Wouldn't it be better to always run


```
git cliff --unreleased --tag-pattern 'termwiz-*' -o CHANGELOG.md
```

and then to commit that in a chore chore commit..

```
git add CHANGELOG.md -m 'chore: git-cliff'
```

and then to tag that chore commit...

```
git tag $(git cliff --unreleased --tag-pattern 'termwiz-*' --bumped-version)
```

Now your commit the tag points to has the changelog generated by git cliff..

I'm asking because I want to make `git cliff` better and work on the documentation. I've written multiple blog entries on git cliff,

https://substack.evancarroll.com/s/technology

And, I don't understand why `git-cliff` would ever be invoked without `--unreleased` as it is here.

wez:
@EvanCarroll from my perspective, I don't care too much about those details: the changelog doesn't show up on docs.rs or crates.io, so it doesn't matter a great deal whether the changelog update is part of the tag or not. It's a supplemental documentation artifact rather than a code artifact.

What I want to optimize for is not having to remember little details or having to run down a wide range of deps, so having a script like the https://github.com/wezterm/wezterm/blob/main/termwiz/update-changelog.sh that I ended up adding (which uses the docker flavor of git-cliff) that I run after tagging sounds fine to me.

If there is a desire to make the tagging atomic wrt. the release/publish process, then I'd be open to having an alternative `tag-release.sh` script that handles that, but I have limited desire to sit down and sweat the details of such a script myself: I just don't have the bandwidth for it.

If you want to propose and pursue changes around this, I'd suggest opening up a PR!

EvanCarroll:
Well, with a positive attitude like that and a simple request. I'd be happy to take a look at it for you. Because it's super easy. Looking at it now. 

This is what I expected though, it's more of a you took the default option because the defaults make it easier to do the bad. Should be a _very_ simple update.

joshka:
Probably the easiest way to handle this more generically is to use [release-plz](https://release-plz.dev/) (or your choice of other release workflow management). There's a small amount of initial investment of time in learning it and configuring it so it works in your workflow, but it pays off in not having to run a script for managing releases. Releasing by merging a PR is pretty neat IMO.

> And, I don't understand why git-cliff would ever be invoked without --unreleased as it is here.

My motivation in this PR was to record what has happened already. For that purpose, the unreleased flag is not relevant. Most projects don't keep a running changelog of unreleased changes as part of their process.

With the updated changelog, I can point Ratatui's changelog at the right place which provides users the ability to see what's changed downstream.

The process when you're doing release workflows is a bit different as you've called out above as the release tag does not already exist while creating a release.

> the changelog doesn't show up on docs.rs or crates.io, so it doesn't matter a great deal whether the changelog update is part of the tag or not. It's a supplemental documentation artifact rather than a code artifact.

As an aside - I think a changelog might be a good thing to actually show up on crates.io and docs.rs. I always check what changed when updating deps in libraries and apps I maintain. When there's no changelog, or it's hard to find the info this can be a pain. Changelog + Dependabot = falling into a pit of success.

PR #6723, opened 2025-02-25
Add DisplayText to Key Assignment

mgpinf:
This displays the provided text in an overlay.
This is useful when want to display anything to read like stdout or stderr of a process.

### DIFF
--- a/config/src/keyassignment.rs
+++ b/config/src/keyassignment.rs
@@ -532,6 +532,11 @@ fn default_message() -> String {
 }
 
 #[derive(Debug, Clone, PartialEq, FromDynamic, ToDynamic)]
+pub struct DisplayText {
+    pub text: String,
+}
+
+#[derive(Debug, Clone, PartialEq, FromDynamic, ToDynamic)]
 pub enum KeyAssignment {
     SpawnTab(SpawnTabDomain),
     SpawnWindow,
@@ -646,6 +651,7 @@ pub enum KeyAssignment {
     PromptInputLine(PromptInputLine),
     InputSelector(InputSelector),
     Confirmation(Confirmation),
+    DisplayText(DisplayText),
 }
 impl_lua_conversion_dynamic!(KeyAssignment);
 
--- a/docs/config/lua/keyassignment/DisplayText.md
+++ b/docs/config/lua/keyassignment/DisplayText.md
@@ -1,0 +1,47 @@
+# `DisplayText`
+
+{{since('nightly')}}
+
+Activates an overlay to display the provided text.
+
+`DisplayText` accepts the following fields:
+
+* `text` - The content to display in the overlay .
+
+### Key Assignments
+
+The default key assignments in the InputSelector are as follows:
+
+| Action  |  Key Assignment |
+|---------|-------------------|
+| Quit     | <kbd>Ctrl</kbd> + <kbd>G</kbd> |
+|          | <kbd>Ctrl</kbd> + <kbd>C</kbd> |
+|          | <kbd>Ctrl</kbd> + <kbd>D</kbd> |
+|          | <kbd>Ctrl</kbd> + <kbd>&#91;</kbd> |
+|          | <kbd>Escape</kbd> |
+
+## Example of displaying the status of wezterm run_child_process
+
+```lua
+local wezterm = require 'wezterm'
+local act = wezterm.action
+local config = wezterm.config_builder()
+
+config.keys = {
+  {
+    key = 'E',
+    mods = 'CTRL|SHIFT',
+    action = wezterm.action_callback(function(window, pane)
+      local success, stdout, stderr = wezterm.run_child_process {
+        'docker',
+        'container',
+        'ls',
+      }
+      local display_text = success and stdout or stderr
+      window:perform_action(act.DisplayText { text = display_text }, pane)
+    end),
+  },
+}
+
+return config
+```
--- a/docs/config/lua/keyassignment/DisplayText.md
+++ b/docs/config/lua/keyassignment/DisplayText.md
@@ -11,7 +11,7 @@ Activates an overlay to display the provided text.
 
 ### Key Assignments
 
-The default key assignments in the InputSelector are as follows:
+The default key assignments in DisplayText are as follows:
 
 | Action  |  Key Assignment |
 |---------|-------------------|
--- a/docs/config/lua/keyassignment/DisplayText.md
+++ b/docs/config/lua/keyassignment/DisplayText.md
@@ -6,7 +6,8 @@ Activates an overlay to display the provided text.
 
 `DisplayText` accepts the following fields:
 
-* `text` - The content to display in the overlay .
+* `text` - the content to display in the overlay. You may embed
+  escape sequences and/or use [wezterm.format](../wezterm/format.md)
 
 ### Key Assignments
 
--- a/docs/config/lua/keyassignment/DisplayText.md
+++ b/docs/config/lua/keyassignment/DisplayText.md
@@ -4,6 +4,14 @@
 
 Activates an overlay to display the provided text.
 
+We can use the below keybindings to activate the respective
+overlay on top of the `DisplayText` overlay:
+
+* [ActivateCopyMode](./ActivateCopyMode.md)
+* [Search](./Search.md)
+* [QuickSelect](./QuickSelect.md),
+* [QuickSelectArgs](./QuickSelectArgs.md)
+
 `DisplayText` accepts the following fields:
 
 * `text` - the content to display in the overlay. You may embed
--- a/mux/src/termwiztermtab.rs
+++ b/mux/src/termwiztermtab.rs
@@ -6,7 +6,7 @@
 use crate::domain::{alloc_domain_id, Domain, DomainId, DomainState};
 use crate::pane::{
     alloc_pane_id, CachePolicy, CloseReason, ForEachPaneLogicalLine, LogicalLine, Pane, PaneId,
-    WithPaneLines,
+    Pattern, SearchResult, WithPaneLines,
 };
 use crate::renderable::*;
 use crate::tab::Tab;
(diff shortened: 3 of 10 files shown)


wez (commented):
I appreciate the simplicity of this, vs. the floating pane feature from #5576.

However, I wonder about the limitations here; I can see folks using this as a means for popping a pane show some output but then perhaps being frustrated if they can't then use eg: QuickSelect to copy and paste from it because it is an overlay and we only support a single overlay rather than an overlay stack.

I also suspect that there will be almost immediately someone wanting more key assignments; they might want to be able to use CopyMode's assignments to navigate the output for example.

What do you think about this?

Maybe we should have an overlay stack, so that you could DisplayText and then QuickSelect or CopyMode over the top? Do you feel like taking a crack at implementing that?

The biggest concern I have about the idea of an overlay stack it is that there might be code that makes assumptions about there only being a single overlay, and then also doing something reasonable around handling eg: CopyMode being active then trying to QuickSelect. Some overlays may not stack over each other very well; I don't know if we need to actively handle that or just tell people to use common sense and avoid doing that.

mgpinf:
I have a doubt, isn't creating an overlay pane for `DisplayText` similar to `CloseCurrentPane` going to create a new `TermWizTerminalPane` with a new pane ID (id2), and we assign overlay (id2) to current pane (id1) by calling `pane_state(id1).overlay.replace(pane (id2), key_table_state)`. Quickselect and CopyMode overlays are created in the same pane id, so when Quickselect is triggered, Overlay pane id for the pane_state of the pane would be the same as id of the pane on which QuickSelect was triggered. So if we trigger QuickSelect on The `TermwizTerminalPane` overlay, should it not be enough?

mgpinf:
e.g.
pane 1 would have pane 2 (TermWizTerminalPane) as overlay
pane 2 would have QuickSelect overlay

mgpinf:
Hi Wez.

Assuming a pane with id `p1`

`QuickSelect` and `CopyMode` are pane overlays that are on the current pane without a new pane getting added to the mux.
In this case, `pane_state(p1).overlay` is `Some(OverlayState {pane_p1, KeyTableState::default()})`

`CloseCurrentPane` is a pane overlay that is on the current pane with a new pane with id `p2` getting added to the mux.
In this case, `pane_state(p1).overlay` is `Some(OverlayState {pane_p2, KeyTableState::default()})`

I am thinking of implementing `DisplayText` as a pane overlay similar to the `CloseCurrentPane`.
This is because `QuickSelect`, `CopyMode` are pane overlays and would not work on `DisplayText` pane if `DisplayText` was a tab overlay

Assuming that we have a `LocalPane` with id `p1`, if we `DisplayText`, we obtain an overlay that is a `TermWizTerminalPane`  with id `p2`
If we want to create a `QuickSelect` overlay on pane `p2`, we need to fetch the pane `p2` by calling 'get_active_pane_or_overlay()'

Based on above points, the pane structure would be:

1. QuickSelect (QuickSelectOverlay) on LocalPane
```
LocalPane (p1)
QuickSelect (p1) (overlay of p1)
```

2. DisplayText (TermWizTerminalPane) on LocalPane
```
LocalPane (p1)
DisplayText (p2) (overlay of p1)
```

3. QuickSelect (QuickSelectOverlay) on DisplayText (TermWizTerminalPane) on LocalPane
```
LocalPane (p1)
DisplayText (p2) (overlay of p1)
QuickSelect (p2) (overlay of p2)
```

The part of the region of the pane to be rendered is fetched using `get_active_pane_or_overlay`
Cases 1 and 2 would work as expected with the earlier implementation of `get_active_pane_or_overlay`
But case 3 would fail, because even though QuickSelect overlay is assigned to DisplayText overlay, the `get_active_pane_or_overlay` used to return only the first overlay if it was active.

I have updated `get_active_pane_or_overlay` in order to return the topmost overlay in a pane region

This is kind of stacked overlay of maximum size 2 which is only applicable if the first overlay is a `TermWizTerminalPane` with the second overlay being `QuickSelect`

Earlier the pane to assign overlay for `QuickSelect` and `QuickSelect` was fetched using `get_active_pane_no_overlay`
I have updated it to use `get_active_pane_or_overlay` to fetch the pane to assign overlay for, in order for it to work on `DisplayText` overlay

I have copied the search implementation of `Pane` trait in `LocalPane` to `TermWizTerminalPane` in order for results to be fetched in `QuickSelect` when called on a pane containing `DisplayText` overlay

The pane to assign overlay to for `DisplayText` is fetched using `get_active_pane_no_overlay` to avoid unnecessary stacking of overlays.

For `CopyMode`
Since the current pane id might be a `TermWizTerminalPane` overlay,
resolving pane id using `mux::Mux::get().resolve_pane_id(pane.pane_id())`
will throw error because it internally calls `iter_panes_ignoring_zoom`
method which only returns non-overlay panes. So, to resolve this, we
first check if the current pane is the active pane, else we fetch
overlay pane by calling pane_state method on active pane and check if
the current pane is the overlay pane if the overlay is present in the
pane_state. If current pane is not the active pane, and the overlay pane
is present and current pane is not the overlay pane or the overlay is
blank, we throw an error

mgpinf:
I realised that the `TermWizTerminalPane` overlay does not get resized along with the panes. 

For example while moving the split, the content on the Termwiz pane overlay is still on adjacent pane (the pane in which the moving the split is increasing it's width/height. Also, when toggling the zoom of an active pane, I am experiencing the same issue. These are few of the cases I have tested on.

This is happening because in case of the resizing the panes by moving the split, the panes are resized in the mux part of the code, while the information about the overlay for a pane can be obtained in the GUI layer. So, I tried explicitly calling resize the overlays in the GUI layer after the call to resize pane in mux layer, which works, but I feel it is kind of a bandaid fix.

Also, actions like `SplitPane` call `spawn_command_impl` which in turn calls an asynchronous function `spawn_command_internal` which is inside a closure that takes ownership of values from the environment. To fetch the overlay for a pane, we need a `TermWindow` object which we can only pass as reference to `spawn_command_impl` because `TermWindow` currently does not implement `Clone`, `Copy` traits. Add the mentioned traits would require changes to fields as well.

As of now, the implementation of `QuickSelect` and `CopyMode` is working as expected on `TermWizTerminalOverlay`.

I shall probably revert implementing `DisplayText` as a tab level overlay and add `QuickSelect` and `CopyMode` to work with tab overlays.

What do you think?

mgpinf:
Hi Wez.

`QuickSelect` or `CopyMode` overlays have same ID as the pane on which it is assigned to.
`DisplayText`, `CloseCurrentPane`, `PromptInputLine`, `InputSelector`, `Confirmation` overlays result in new pane (`TermWizTerminalPane`) getting added to the mux.

Based on above points, the pane structure would be:

1. QuickSelect (QuickSelectOverlay) on LocalPane
```
LocalPane (p1)
QuickSelect (p1) (overlay of p1)
```

2. DisplayText (TermWizTerminalPane) on LocalPane
```
LocalPane (p1)
DisplayText (p2) (overlay of p1)
```

3. QuickSelect (QuickSelectOverlay) on DisplayText (TermWizTerminalPane) on LocalPane
```
LocalPane (p1)
DisplayText (p2) (overlay of p1)
QuickSelect (p2) (overlay of p2)
```

The part of the region of the pane to be rendered is fetched using `get_active_pane_or_overlay`
Cases 1 and 2 would work as expected with the earlier implementation of `get_active_pane_or_overlay`
But case 3 would fail, because even though QuickSelect overlay is assigned to DisplayText overlay, the `get_active_pane_or_overlay` used to return only the first overlay if it was active.

I have updated `get_active_pane_or_overlay` in order to return the topmost overlay in a pane region

This is kind of stacked overlay of maximum size 2 which is only applicable if the first overlay is a `TermWizTerminalPane` with the second overlay being `QuickSelect`

Earlier the pane to assign overlay for `QuickSelect` and `QuickSelectArgs` was fetched using `get_active_pane_no_overlay`
I have updated it to use `get_active_pane_or_overlay` to fetch the pane to assign overlay for, in order for it to work on `DisplayText` overlay

I have copied the search implementation of `Pane` trait in `LocalPane` to `TermWizTerminalPane` in order for results to be fetched in `QuickSelect` when called on a pane containing `DisplayText` overlay

For `CopyMode`
Since the current pane id might be a TermWizTerminal overlay, resolving pane id using `mux::Mux::get().resolve_pane_id(pane.pane_id())` will throw error because it internally calls `iter_panes_ignoring_zoom` method which only returns non-overlay panes.
To resolve this:
* Fetch active pane for the current tab
* If current pane is not active pane:
    * Fetch overlay pane by calling tab_state method on active tab
    * If overlay pane is not present, throw error
    * If current pane is not the overlay pane, throw error

manish-gowd-unbxd:
I have made `TabState` contain 2 optional overlays. This is in order to not clear the `DisplayText` overlay in case any other
overlay is activated. `DisplayText` overlay is activated at the bottom layer (layer 0), and the other overlays are activated at the top layer (layer 1). A higher layer overlay can be activated irrespective of status of the lower layer overlay. A lower layer overlay can be activated only when the higher layer overlay is inactive

mgpinf:
I have made a tab to contain 2 overlays. This is in order to not clear the `DisplayText` overlay in case any other overlay is activated. `DisplayText` overlay is activated at the bottom layer (layer 0), and the other overlays are activated at the top layer (layer 1). Layer 1 overlay can be activated irrespective of whether layer 0 overlay is active. Layer 0 overlay can be activated only when layer 1 overlay is not active

mgpinf:
I have updated the `TabState` to have 3 optional overlays in order to provide better precedence. An overlay at a higher layer can be activated irrespective of whether an overlay is activated at a lower layer. An overlay at a lower layer cannot be activated if an overlay is present at a higher layer. An overlay at a layer replaces any existing overlay at that layer provided that the layer is the topmost layer for which an overlay is present.

Overlay for layer 0:
* `show_display_text_overlay`

Overlays for layer 1:
* `show_line_prompt_overlay`
* `selector`
* `show_confirmation_overlay`
* `show_debug_overlay`
* `launcher`

Overlays for layer 2:
* `confirm_close_window`
* `confirm_close_tab`
* `confirm_quit_program`

What do you think about these? Should I move `show_debug_overlay` to a separate layer? I think the `PromptInputLine`, `InputSelector`, `Confirmation` belong to the same layer because these 3 can be composed together. While `show_debug_overlay` and `launcher` based overlays are separate entities

mgpinf:
I have added unnecessary complex logic to this PR. I shall close this PR and raise a PR to activate `CopyMode` on `TermwizTerminal` overlays at tab level

PR #6734, opened 2025-02-27
Add `Clone` derive to `termwiz::surface::Surface`

tombh:
I think this is my first PR here, so just wanted give a big thank you for all the wonderful code in this repo! ❤️

I don't know if this change is a small thing, or of course I wouldn't be surprised if it were a big thing.

Just a little context as to why it would be useful for me. I'm currently accessing some `Surface` instances between threads and so using locks. But in certain cases it would just be much simpler to share copies of surfaces between threads instead.

I'd also be interested if there were some other way of making quick duplicates?

### DIFF
--- a/termwiz/src/surface/mod.rs
+++ b/termwiz/src/surface/mod.rs
@@ -101,7 +101,7 @@ pub const SEQ_ZERO: SequenceNo = 0;
 /// difference between the updated screen and apply those changes to
 /// the render target, and then use `get_changes` to render those without
 /// repainting the world on each update.
-#[derive(Default)]
+#[derive(Default, Clone)]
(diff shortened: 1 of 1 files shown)


wez:
I think `Clone` is probably fine here. I don't know that you can do a quicker-than-clone copy of this state.  The main concern is really whether there are a lot changes buffered up that might bloat the surface, so you might want to consider , in you consuming code, calling `Surface::flush_changes_older_than` with some appropriate sequence value prior to copying if that might be a concern.

tombh:
Thank you! It's working great ❤️

And that's a great point about `flush_changes_older_than`, thanks.

PR #6741, opened 2025-02-28
Fix outdated tap

VinnyPutty:


docs/install/linux.md
```diff
@@ -270,7 +270,7 @@ hide:
     can install wezterm from our tap:
 
     ```console
-    $ brew tap wezterm/wezterm-linuxbrew
+    $ brew tap wez/wezterm-linuxbrew
```
wez, line 273: I don't think this is outdated so much as "ahead of its time". I just moved the brew related repos over to the newer wezterm org, so I think that the current `wezterm/wezterm-linuxbrew` is correct again now.

Could you verify that?

Issue #6749, opened 2025-03-03
Wezterm hangs when tmux control mode session ssh connection times out


iDarshan:
### What Operating System(s) are you seeing this problem on?

macOS

### Which Wayland compositor or X11 Window manager(s) are you using?

NA

### WezTerm version

20250224-215103-f06be357

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

This is in context of #6602. I have built that PR locally and tested tmux control mode. In that this issue of wezterm hang is found. @joexue provided the patch on that PR, which seems to have fixed the issue, but @joexue suggests that's not the right fix and requested to raise an issue for it. Here is the patch which fixes the hang issue:
```diff
diff --git a/mux/src/lib.rs b/mux/src/lib.rs
index f46e5ee..f27b338 100644
--- a/mux/src/lib.rs
+++ b/mux/src/lib.rs
@@ -807,9 +807,9 @@ impl Mux {
         if let Some(pane) = self.panes.write().remove(&pane_id).clone() {
             log::debug!("killing pane {}", pane_id);
             pane.kill();
-            self.recompute_pane_count();
             self.notify(MuxNotification::PaneRemoved(pane_id));
         }
+        self.recompute_pane_count();
     }

     fn remove_tab_internal(&self, tab_id: TabId) -> Option<Arc<Tab>> {
```

### To Reproduce

Build PR #6602 and run tmux control mode over ssh with remote server. When ssh connection times out, wezterm will hang.

### Configuration

no config

### Expected Behavior

Wezterm should not hang when ssh times out to remote server where tmux is running in control mode. It should safely detach tmux session.

joexue:
This is not a just MacOS issue.
Both gdb and parking_lot::deadlock::check_deadlock point to same culprit **_domain_was_detached_**. Below are back trace from parking_lot::deaklock when the weztern hangs.
Basically, the domain_was_detached spwan another thread when it calls remove_pane_if, then calls remove_pane. while in its own thread, it calls remove_pane_internal and they lock windows and panes of mux in reversed order, hence cause dead lock.

```
11:58:40.634  DEBUG  mux::tmux                   > 1 deadlocks detected
11:58:40.634  DEBUG  mux::tmux                   > Deadlock #0
11:58:40.634  DEBUG  mux::tmux                   > Thread Id 140273256580800
[...]

PR #6750, opened 2025-03-04
fix: dead lock in domain_was_detached

joexue:
refs: #6749


wez:
Looks good, thank you!

Issue #6763, opened 2025-03-07
tmux -CC error on %config-error


wez:
### What Operating System(s) are you seeing this problem on?

Linux Wayland

### Which Wayland compositor or X11 Window manager(s) are you using?

N/A

### WezTerm version

60c3f756e49da4cb361c6f800531566b6709b687

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

I rarely use tmux, but I was testing this with wezterm just now; I ran `wezterm -n` to start with no config, then in that session ran `tmux -CC`.  No other windows tabs or panes appear, but in the error log I see:

```console
; ./target/debug/wezterm -n
10:05:14.093  ERROR  termwiz::tmux_cc > Unrecognized tmux cc line:  --> 1:1
  |
1 | %config-error /home/wez/wez-personal/dotfiles/tmux.conf:12: invalid option: pane-border-fg
  | ^---
  |
  = expected begin, end, error, client_session_changed, client_detached, output, exit, sessions_changed, pane_mode_changed, window_add, window_close, window_pane_changed, window_renamed, session_changed, session_renamed, session_window_changed, or layout_change
10:05:14.096  INFO   mux::localpane   > tmux -CC mode requested
```

### To Reproduce

Have an invalid `tmux.conf`. The line in the error message looks like this: `set -g pane-border-fg black`

### Configuration

Default; `wezterm -n`, but with a tmux conf with outdated options inside

### Expected Behavior

We should be able to continue past the config errors and show the tmux tabs from the session

wez:
@joexue FYI, I think we could do with handling this configuration error case!

joexue:
Will check it

wez:
Fixed by #6773

PR #6773, opened 2025-03-08
fix: tmux -CC error on %config-error

joexue:
refs: #6763

In the PR, the %config-error is handled

we also add all events we missed according to latest Tmux 3.6 version although most of them we are not interested but we need to parse them in case they come by accident.

Looks messy, but not all are real changes, because I move the rules in alphabet order to easily organize the code later.

### DIFF
--- a/mux/src/tmux.rs
+++ b/mux/src/tmux.rs
@@ -114,7 +114,7 @@ impl TmuxDomainState {
                 // Tmux specific events
                 Event::ConfigError { error } => {
                     // tmux config file error, not our fault, just log it and go
-                    log::info!("The tmux configuration error: {error}");
+                    log::warn!("tmux configuration error: {error}");
                 }
                 Event::Exit { reason: _ } => {
                     *self.state.lock() = State::Exit;
--- a/mux/src/tmux.rs
+++ b/mux/src/tmux.rs
@@ -112,6 +112,10 @@ impl TmuxDomainState {
                 },
 
                 // Tmux specific events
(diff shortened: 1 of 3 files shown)


wez:
I'd recommend (and prefer!) that the re-organizing part of this be a separate commit that doesn't otherwise change any functionality. That has the nice property of being easier to review and easier to `git bisect`.

On top of that commit you can add the bits for %config-error.

That way I can more easily view just the refactor commit first, see that it has no functional change, then view the %config-error parts second and see what they are doing.

Thanks!

joexue:
> I'd recommend (and prefer!) that the re-organizing part of this be a separate commit that doesn't otherwise change any functionality. That has the nice property of being easier to review and easier to `git bisect`.
> 
> On top of that commit you can add the bits for %config-error.
> 
> That way I can more easily view just the refactor commit first, see that it has no functional change, then view the %config-error parts second and see what they are doing.
> 
> Thanks!

sure, let's do that

mux/src/tmux.rs
```diff
@@ -112,6 +112,10 @@ impl TmuxDomainState {
                 },
 
                 // Tmux specific events
+                Event::ConfigError { error } => {
+                    // tmux config file error, not our fault, just log it and go
+                    log::info!("The tmux configuration error: {error}");
```
wez, line 117:
```suggestion
                    log::warn!("tmux configuration error: {error}");
```

wez:
Thank you!

Issue #6769, opened 2025-03-08
tmux -CC mode not parsing output correctly (no problem in normal tmux mode)


bfstree:
https://github.com/user-attachments/assets/35e20687-6c3b-4b9a-b82d-be864ff9977f

`lazygit` is not working properly in tmux CC mode. I am receiving the following errors:

```
18:11:17.356  INFO   mux::localpane > tmux -CC mode requested
18:11:17.357  INFO   mux::tmux      > Tmux create window id 1
18:11:17.357  INFO   mux::tmux      > tmux session changed:0
18:11:17.417  INFO   mux::tmux_commands > new pane synced, id: 0
18:11:39.375  ERROR  termwiz::tmux_cc   > Failed to parse line from tmux: incomplete utf-8 byte sequence from index 1043
18:11:39.376  ERROR  termwiz::tmux_cc   > Failed to parse line from tmux: invalid utf-8 sequence of 1 bytes from index 11
18:11:40.221  ERROR  termwiz::tmux_cc   > Failed to parse line from tmux: incomplete utf-8 byte sequence from index 1255
18:11:40.221  ERROR  termwiz::tmux_cc   > Failed to parse line from tmux: invalid utf-8 sequence of 1 bytes from index 11
18:11:43.500  ERROR  termwiz::tmux_cc   > Failed to parse line from tmux: incomplete utf-8 byte sequence from index 1457
18:11:43.500  ERROR  termwiz::tmux_cc   > Failed to parse line from tmux: invalid utf-8 sequence of 1 bytes from index 11
18:11:50.287  ERROR  termwiz::tmux_cc   > Failed to parse line from tmux: incomplete utf-8 byte sequence from index 1523
18:11:50.288  ERROR  termwiz::tmux_cc   > Failed to parse line from tmux: invalid utf-8 sequence of 1 bytes from index 11
```

This is on macOS 15.3, tmux 3.5a, lazygit 0.46, TERM=tmux-256color
This does not happen in normal tmux mode.

Thank you for implementing this feature!

_Originally posted by @bfstree in https://github.com/wezterm/wezterm/issues/6602#issuecomment-2672930677_

joexue:
The video you gave is expected, right?

digitallyserviced:
> The video you gave is expected, right?

The video shows artifacts that are not cleared properly near the bottom of the screen.

digitallyserviced:
@bfstree 

[...]

PR #6779, opened 2025-03-10
fix: tmux -CC mode not parsing output correctly

joexue:
refs: #6769

    The tmux output may split into multiple parts, if this splitting happens
    in the middle of multiple bytes UTF8 characters, then the issue raised.
    That the first part will have incomplete UTF8 character while the second
    part will have a invalid UTF8 character.

    In this case, we have to treat the output text as bytes rather than
    UTF8 string.

    This fix unvis the output text and send to pane as bytes.

Force pushed to prevent previews invalid commits polluting code history

### DIFF
--- a/termwiz/src/tmux_cc/mod.rs
+++ b/termwiz/src/tmux_cc/mod.rs
@@ -922,7 +922,7 @@ impl Parser {
                 if self.begun.is_some() {
                     log::error!(
                         "expected %end or %error before %begin ({})",
-                        String::from_utf8_lossy(&self.buffer).as_ref()
+                        String::from_utf8_lossy(&self.buffer)
                     );
                 }
                 self.begun.replace(Guarded {
@@ -937,9 +937,7 @@ impl Parser {
             Ok(event) => Some(event),
             Err(err) => {
                 log::error!("Unrecognized tmux cc line: {}", err);
-                return Err(anyhow::anyhow!(String::from_utf8_lossy(&self.buffer)
-                    .as_ref()
-                    .to_owned()));
+                anyhow::bail!("{}", String::from_utf8_lossy(&self.buffer));
             }
         };
 
@@ -966,7 +964,7 @@ mod tests {
                 number: 321,
                 flags: 0,
             },
-            parse_line("%begin 12345 321 0".to_owned().as_bytes()).unwrap()
+            parse_line(b"%begin 12345 321 0").unwrap()
         );
 
         assert_eq!(
@@ -975,7 +973,7 @@ mod tests {
                 number: 321,
                 flags: 0,
             },
-            parse_line("%end 12345 321 0".to_owned().as_bytes()).unwrap()
+            parse_line(b"%end 12345 321 0").unwrap()
         );
     }
 
--- a/termwiz/src/tmux_cc/mod.rs
+++ b/termwiz/src/tmux_cc/mod.rs
@@ -60,7 +60,7 @@ pub enum Event {
     },
     ExtendedOutput {
         pane: TmuxPaneId,
-        text: String,
+        text: Vec<u8>,
     },
     Exit {
         reason: Option<String>,
@@ -76,7 +76,7 @@ pub enum Event {
     },
     Output {
         pane: TmuxPaneId,
-        text: String,
+        text: Vec<u8>,
     },
     PaneModeChanged {
         pane: TmuxPaneId,
@@ -215,8 +215,10 @@ fn parse_guard(mut pairs: Pairs<Rule>) -> anyhow::Result<(i64, u64, i64)> {
     Ok((timestamp, number, flags))
 }
 
-fn parse_line(line: &str) -> anyhow::Result<Event> {
-    let mut pairs = parser::TmuxParser::parse(Rule::line_entire, line)?;
+fn parse_line(line: &[u8]) -> anyhow::Result<Event> {
+    let binding = String::from_utf8_lossy(line);
+    let parsed_line = binding.as_ref();
+    let mut pairs = parser::TmuxParser::parse(Rule::line_entire, parsed_line)?;
     let pair = pairs.next().ok_or_else(|| anyhow::anyhow!("no pairs!?"))?;
     match pair.as_rule() {
         // Tmux generic rules
@@ -296,12 +298,10 @@ fn parse_line(line: &str) -> anyhow::Result<Event> {
         Rule::extended_output => {
             let mut pairs = pair.into_inner();
             let pane = parse_pane_id(pairs.next().ok_or_else(|| anyhow!("missing pane id"))?)?;
-            let text = unvis(
-                pairs
-                    .next()
-                    .ok_or_else(|| anyhow!("missing text"))?
-                    .as_str(),
-            )?;
+            let pair = pairs.next().ok_or_else(|| anyhow!("missing text"))?;
+
+            let (_, pos) = pair.line_col();
+            let text = unvis_bytes(&line[pos - 1..])?;
             Ok(Event::ExtendedOutput { pane, text })
         }
         Rule::exit => {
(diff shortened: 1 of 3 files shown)


joexue:
@bfstree
I tested by using lazygit, but I'm not familiar with it, just run as your video and did not find the issue you met.
please help to test it.

termwiz/src/tmux_cc/mod.rs
```diff
@@ -793,6 +795,49 @@ pub fn parse_layout(layout: &str) -> anyhow::Result<Vec<WindowLayout>> {
     Ok(result)
 }
 
+// https://github.com/tmux/tmux/wiki/Control-mode
+// The tmux encode all characters < 32 as Oct '\xxx'
+// This function decode those encoded characters to raw bytes
+fn decode_tmux_output(bytes: &[u8]) -> Vec<u8> {
```
wez, line 801: how does this differ from the `unvis` function that we have already? `unvis` returns a `String`, so that's an obvious difference, but I'm wondering if this is logically the same function that should be applied in the same places?

termwiz/src/tmux_cc/mod.rs
```diff
@@ -76,7 +77,8 @@ pub enum Event {
     },
     Output {
         pane: TmuxPaneId,
-        text: String,
+        bytes: Vec<u8>,
+        bytes_pos: usize,
```
wez, line 81: it's not clear to me what the purpose is for these `bytes_pos` fields; I think that it needs a better name and probably also some comments to explain it

termwiz/src/tmux_cc/mod.rs
```diff
@@ -902,42 +947,61 @@ impl Parser {
...
+                    timestamp,
+                    number,
+                    flags,
+                    error: false,
+                    output: String::new(),
+                });
+                None
+            }
+            Ok(event) => {
+                if let Event::Output {
```
wez, line 973: this should be a `match` block instead of an if-else-if-else

termwiz/src/tmux_cc/mod.rs
```diff
@@ -902,42 +947,61 @@ impl Parser {
...
+                    pane,
+                    bytes: _,
+                    bytes_pos,
+                } = event
+                {
+                    Some(Event::Output {
+                        pane,
+                        bytes: decode_tmux_output(&self.buffer[bytes_pos..]),
```
wez, line 981: ah, I think I see what's happening here, bytes_pos is trying to walk back to the start of the byte sequence that we potentially "lost" when doing the lossy string transformation.

That doesn't feel great.  As an alternative, perhaps `parse_line` should operate on bytes instead of a string?

joexue: I’m afraid we can’t. Pest says it only accepts Unicode. If we do that, we need to escape non-UTF8 firstly then convert it back, kind of same solution.

joexue: Another solution is just use the UTF8 from from_utf8_lossy but that doesn’t respect the apps

wez: keep in mind that wezterm only supports UTF-8 anyway; the vtparse crate is responsible for decoding that data.
It seems reasonable to me to perform an equivalent lossy/eliding transform on the bytes we get from decoding tmux's output, at the tmux_cc layer.

joexue: So, just use the lossy data? That woks too when I tested it.

bfstree:
@joexue The commit I checked out 08b7a698d92de4fc7951f79eaa93ac5e3d5c255b is now drawing `lazygit` correctly. Thanks for fixing this issue.

ghost:
Compared to the previous commit 08b7a698d92de4fc7951f79eaa93ac5e3d5c255b,
the latest commit f502537 is replacing valid cells to � in `lazygit`.

Attached is the screenshot:

And the video:

These rendering issues did not happen in 08b7a69. Also, the issues do not happen in normal tmux mode.

joexue:
> @joexue Compared to the previous commit [08b7a69](https://github.com/wezterm/wezterm/commit/08b7a698d92de4fc7951f79eaa93ac5e3d5c255b), the latest commit [f502537](https://github.com/wezterm/wezterm/commit/f50253740fa96eb7f58e3f1f1bb865767933c61e) is replacing valid cells to � in `lazygit`.
> 
> Attached is the screenshot:
> 
> render1
> And the video:
> 
>  render1-video.mp4 
> 
> Also, as you can see at the end of the video, "us" string is also appended at the current selection, so I think some parsing is not correct. However, there are no error logs showing though. These rendering issues did not happen in [08b7a69](https://github.com/wezterm/wezterm/commit/08b7a698d92de4fc7951f79eaa93ac5e3d5c255b). Also, the issues do not happen in normal tmux mode.

Hmm, that is expected, we change to a simple solution , this solution just replace the non-UTF8 to the ‘?’ that you saw. As @wez pointed, the wezterm supports only UTF8. But the previous version did not see that ‘?’ , how wezterm handled ascii 216(I debuted the lazygit and found the non-UTF8), maybe @wez can answer the question.

joexue:
Ok, I think I know what happened. If The output too long, then it is split to multiple parts, and if the split happens in the middle of multiple bytes UTF8, then the issue raised. That why we have (within fix) two errors, one is incomplete UTF8 character, that is first split part, one is invalid UTF8 character, that is second part. 
I can recover those characters from error log that verified my thought.
This also explains why previous commit works. Because we send raw bytes to pane, and pane treat it as stream, so the split UTF8 will eventually concat together.

joexue:
@bfstree , I did test the last commit by using lazygit. But I would like you to verify it if you have chance. Thanks.

bfstree:
@joexue Rendering is working correctly with your new commit 2bc7d97c. Thanks for the prompt fix!

termwiz/src/tmux_cc/mod.rs
```diff
@@ -961,7 +966,7 @@ mod tests {
                 number: 321,
                 flags: 0,
             },
-            parse_line("%begin 12345 321 0").unwrap()
+            parse_line("%begin 12345 321 0".to_owned().as_bytes()).unwrap()
```
wez, line 969: I think you can write this more simply like this, and use the same technique for the other cases below:

```suggestion
            parse_line(b"%begin 12345 321 0").unwrap()
```

termwiz/src/tmux_cc/mod.rs
```diff
@@ -902,42 +909,40 @@ impl Parser {
...
-                        None
-                    }
-                    Ok(event) => Some(event),
-                    Err(err) => {
-                        log::error!("Unrecognized tmux cc line: {}", err);
-                        return Err(anyhow::anyhow!(line.to_owned()));
-                    }
+                    log::error!(
+                        "expected %end or %error before %begin ({})",
+                        String::from_utf8_lossy(&self.buffer).as_ref()
```
wez, line 925: I think this can simplify like this?

```suggestion
                        String::from_utf8_lossy(&self.buffer)
```

termwiz/src/tmux_cc/mod.rs
```diff
@@ -902,42 +909,40 @@ impl Parser {
...
+                None
             }
+            Ok(event) => Some(event),
             Err(err) => {
-                log::error!("Failed to parse line from tmux: {}", err);
-                None
+                log::error!("Unrecognized tmux cc line: {}", err);
+                return Err(anyhow::anyhow!(String::from_utf8_lossy(&self.buffer)
+                    .as_ref()
+                    .to_owned()));
```
wez, line 942: Simpler this way?

```suggestion
                anyhow::bail!("{}", String::from_utf8_lossy(&self.buffer));
```

wez (commented):
Thanks for this!
A couple of minor suggestions, but this looks good!

wez:
Thank you!

Issue #6776, opened 2025-03-09
Starship not work on a Linux machine which install wezterm via ssh


lifei:
### What Operating System(s) are you seeing this problem on?

Windows

### WezTerm version

20240203-110809-5046fc22

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

my linux server wezterm version is:
```
wezterm --version
wezterm 20240203-110809-5046fc22
```

I dig into this issue a little. I found that starship use the following line to init the bash.

```
elif [[ -n "${bash_preexec_imported:-}" || -n "${__bp_imported:-}" || -n "${preexec_functions-}" || -n "${precmd_functions-}" ]]; then
  # bash-preexec needs a single function--wrap the args into a closure and pass
  starship_preexec_all() { starship_preexec "$_"; }
  preexec_functions+=(starship_preexec_all)
  precmd_functions+=(starship_precmd)
```

Then I dig into bash_preexec and found it's enabled by `/etc/profile.d/wezterm.sh`.

more detail:
1. it works when i using wezterm connect sshmux:xxx.
2. it does work when i use ssh xxx.

so i think the file `/etc/profile.d/wezterm.sh` should be check whether bash-preexec is enabled.

### Configuration

none

dangooddd:
Same problem. Starship stops working on all terminals after installing wezterm

lifei:
> Same problem. Starship stops working on all terminals after installing wezterm

by running `__bp_install `  or `__bp_install_after_session_init` will help you.

wez:
Please try the nightly version of wezterm

dangooddd:
Problem appears on nightly for me. Wezterm installed from copr repo (Fedora 41)

digitallyserviced:
@wez @dangooddd @lifei 

This may be related to this, as it would seem the same file is causing problems...

#6755

digitallyserviced:
@lifei @dangooddd 

I have been trying to replicate this and have been unable to. Granted I am testing the same integration script versions, my setup may not be completely the same.

Can anyone run `bash --verbose --login` and see if it works proper? and also can they provide the output? Ensure no secrets in env vars are leaked...

How and in what file are you running the starship init? `.bashrc`, `.bash_profile` another file? And what commands?

dangooddd:
[...]

Issue #6755, opened 2025-03-06
Starship bash prompt is not displayed only for the first prompt after installing Wezterm


dmi-dmi-dmi:
### What Operating System(s) are you seeing this problem on?

Linux X11

### Which Wayland compositor or X11 Window manager(s) are you using?

xfwm4

### WezTerm version

20250301-123420-7d0bff06

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

When a new wezterm window, pane, or tab is created, the default bash prompt is shown instead of the [starship](https://starship.rs/) bash prompt. Once I enter and execute any command starship prompt appears.

This bug is similar to https://github.com/wezterm/wezterm/issues/2870 but it is not the same since other terminal emulators/login shell itself started to expose the same behaviour after wezterm installation. If I remove it (more specifically, if I remove `/etc/profile.d/wezterm.sh` file) everything is back to normal. 

### Configuration

no config

digitallyserviced:
@dmi-dmi-dmi 

Can you share your bashrc at least specifically your starship invocation. 

Also I am unsure about bash minutiae compared to zsh but I can still help you troubleshoot.

When sitting at your bash prompt what does `source /etc/profile.d/wezterm.sh` and see if it has any errors.

Does wezterm install place it there ? Or did you? I don't use a package version so I don't know.

While profile.d is picked up by most shells and includes them, the order in which they get included can mess things up.

I assume you are using the .bashrc to load starship.

I would actually suggest you load/source the wezterm.sh after starship.

The shell integration uses some hacky ways to polyfill preexec and precmd hooks on bash whereas other shells support them natively. Because profile.d will be prioritized during shell init the hacky nature could fuck up the full eval of your bashrc.

digitallyserviced:
@dmi-dmi-dmi 

Check #6780 and my comment https://github.com/wezterm/wezterm/issues/6776#issuecomment-2710664373

mattvv:
Looking forward to this landing

PR #6780, opened 2025-03-10
check that ps1 has not changed before restoring it

digitallyserviced:
#6755 
#6776 

When running semantic preexec we restore the ps1 prompt to a saved version that can possibly have changed during the initial session setup when custom prompts such as starship modify it.

Resolves the first prompt not rendering properly when non-login shells are started.  

~~**edit** accidentally restored an older version wezterm.sh when testing and making changes so converted to draft while resolving this~~


wez:
Thank you!

Issue #6764, opened 2025-03-07
tmux -CC mouse reporting issue


wez:
### What Operating System(s) are you seeing this problem on?

Linux X11

### Which Wayland compositor or X11 Window manager(s) are you using?

Wayland, but N/A

### WezTerm version

20250307-100237-60c3f756

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

Starting `tmux -CC a` then opening `vim`, then in vim, `:split` and `:se mouse=a`, I would expect to be able to drag the split with the mouse.

This doesn't happen unless `setw -g mode-mouse on` is enabled in the `tmux.conf`

### Configuration

`wezterm -n`

### Expected Behavior

I would expect the mouse reporting to work automatically.

Should we implicitly enable mouse mode when we attach `tmux -CC`?
If not, we should probably have a little guide section in the docs that have recommended config for using `tmux -CC` that covers this.

wez:
cc: @joexue what do you think about this?

joexue:
Will check it.

PR #6782, opened 2025-03-10
fix: tmux -CC mouse reporting issue

joexue:
refs: #6764

The app (such as vim "set mouse=a" command) may enable terminal mouse reporting capability. When tmux control session was detached and attach again, we lost the setting.

This change give the ability to fetch the tmux terminal mouse reporting status and restore to the local terminal when attaching.

mouse_sgr and mouse_utf8 are not supported by old versions of tmux, so just leave them along for now.

### DIFF
--- a/mux/src/tmux_commands.rs
+++ b/mux/src/tmux_commands.rs
@@ -12,7 +12,7 @@ use std::collections::HashSet;
 use std::fmt::{Debug, Write};
 use std::io::Write as _;
 use std::sync::Arc;
-use termwiz::escape::csi::{Cursor, CSI};
+use termwiz::escape::csi::{Cursor, DecPrivateMode, DecPrivateModeCode, Mode, CSI};
 use termwiz::escape::{Action, OneBased};
 use termwiz::tmux_cc::*;
 use wezterm_term::TerminalSize;
@@ -35,6 +35,8 @@ pub(crate) struct PaneItem {
     pane_left: u64,
     pane_top: u64,
     pane_active: bool,
+    pane_mouse_any: bool,
+    pane_mouse_buttton: bool,
 }
 
 #[derive(Debug)]
@@ -166,6 +168,18 @@ impl TmuxDomainState {
         ))]);
     }
 
+    fn enable_mouse_reporting_any(&self, pane: &Arc<dyn Pane>) {
+        pane.perform_actions(vec![Action::CSI(CSI::Mode(Mode::SetDecPrivateMode(
+            DecPrivateMode::Code(DecPrivateModeCode::AnyEventMouse),
+        )))]);
+    }
+
+    fn enable_mouse_reporting_button(&self, pane: &Arc<dyn Pane>) {
+        pane.perform_actions(vec![Action::CSI(CSI::Mode(Mode::SetDecPrivateMode(
+            DecPrivateMode::Code(DecPrivateModeCode::ButtonEventMouse),
+        )))]);
+    }
+
     fn create_pane(&self, pane: &PaneItem) -> anyhow::Result<Arc<dyn Pane>> {
         let local_pane_id = alloc_pane_id();
         let active_lock = Arc::new((Mutex::new(false), Condvar::new()));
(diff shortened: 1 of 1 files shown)


wez:
I think there's a misunderstanding here; what I was referring to in #6764 was the tmux mouse mode configuration, not the wezterm pane mouse reporting mode. I thought that the solution would be to tell tmux to enable its own mouse mode, which will in turn cause tmux to request mouse reporting from wezterm, if necessary.  I think the solution here in this PR might have some bad side effects by sending mouse reporting in when it isn't expected.

joexue:
> I think there's a misunderstanding here; what I was referring to in #6764 was the tmux mouse mode configuration, not the wezterm pane mouse reporting mode. I thought that the solution would be to tell tmux to enable its own mouse mode, which will in turn cause tmux to request mouse reporting from wezterm, if necessary. I think the solution here in this PR might have some bad side effects by sending mouse reporting in when it isn't expected.

From my test, nothing to do with tmux, as long as we set mouse=a in vim, then mouse works. The only thing is, this state will go when we detach then reattach. That why this change restore the mouse report state from tmux status when we reattach a session.

not sure if I understand you correctly

sdrik:
I spent some time reproducing the behaviour here, and I think there's a terminology split in the thread that explains the disagreement, plus one concrete bug worth fixing. Sharing what I found.

**Two different "mouse modes" are being conflated**

- *tmux's own mouse option* (`set -g mouse on`; the old `mode-mouse on` from #6764) drives tmux's own mouse features — dragging pane borders, selecting panes, wheel→copy-mode. This is what @wez was referring to.
- *The application's mouse reporting* — the DECSET `1000`/`1002`/`1003` that e.g. vim's `:set mouse=a` turns on — is per-pane state that tmux tracks and exposes through `#{mouse_*_flag}`. This is what this PR restores.

In `-CC`, wezterm mirrors each tmux pane as a local pane and injects mouse reports straight into the app via `send-keys -H`; tmux's own `mouse` option isn't in that path. So @joexue is right that the app-level reporting is the relevant thing, and that it's silently dropped on reattach (the mirrored pane is a fresh terminal and `capture-pane` carries no mode-setting sequences). Restoring it from tmux is the correct approach.

**The concrete issue: `mouse_any_flag` is the wrong variable**

`#{mouse_any_flag}` is not "any-event mouse (1003)" — it's a *summary* bit set for **any** mouse mode. From tmux (`format.c`, `tmux.h`):

- `mouse_any_flag`  = `mode & ALL_MOUSE_MODES` (= `STANDARD|BUTTON|ALL`, i.e. 1000 ∨ 1002 ∨ 1003)
- `mouse_all_flag`  = `mode & MODE_MOUSE_ALL` (**1003 only**)
- `mouse_button_flag` = 1002, `mouse_standard_flag` = 1000

Reproduced on tmux 3.6a (app enables each mode, then `list-panes`):

| app enables | any | all | button | standard |
|---|:--:|:--:|:--:|:--:|
| `1000h` | 1 | 0 | 0 | 1 |
| `1002h` | 1 | 0 | 1 | 0 |
| `1003h` | 1 | 1 | 0 | 0 |

Because the PR does `if mouse_any_flag { enable AnyEventMouse (1003) }`, **every** mouse app gets `1003` (any-event) enabled on reattach — including `vim :set mouse=a`, which only requests `1002`. That makes wezterm report motion even with no button held (input the app never subscribed to), and since the code only ever enables (never disables), `any_event_mouse` stays on even after the app later turns the mouse off. I think this is the "reporting when it isn't expected" side effect @wez was worried about.

**Suggested fix** — map each flag 1:1:

```
if mouse_standard_flag { SetDecPrivateMode(MouseTracking)    }  // 1000
if mouse_button_flag   { SetDecPrivateMode(ButtonEventMouse) }  // 1002
if mouse_all_flag      { SetDecPrivateMode(AnyEventMouse)    }  // 1003  (was mouse_any_flag)
```

One smaller note: the *encoding* isn't restored either, so an app that negotiated SGR (`1006`) comes back in X10 after reattach. That matters beyond compatibility — X10 encodes each coordinate as a single byte (`32 + 1 + coord`), so it can't represent positions past ~column/row 223, and wide panes need SGR. `#{mouse_sgr_flag}`/`#{mouse_utf8_flag}` would cover it, but they're tmux 3.0+ only (and a missing format variable expands to empty, which would break the current `parse()`), so that part should be version-gated.

PR #6796, opened 2025-03-14
Add configuration option for Launcher label background and foreground color

mgpinf:
Add below configuration option in config.colors:
* `launcher_label_bg`
* `launcher_label_fg`

If above mentioned options are not set in the configuration file then:
* `launcher_label_bg` defaults to default background color
* `launcher_label_fg` defaults to default foreground color


wez:
Thanks!

PR #6800, opened 2025-03-15
docs: hyperlinks: integration example

aorith:
Add a recipe to show an example on how to customize which action will be taken in the event of clicking an hyperlink. (https://github.com/wezterm/wezterm/discussions/6797)


docs/recipes/hyperlinks.md
```diff
@@ -0,0 +1,73 @@
...
+This recipe is an example on how to leverage hyperlinks directly in the terminal.
+Hyperlinks are generated by terminal applications by using the [OSC-8 escape sequences](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda).
+
+## Requirements
+
+By default, `ls` doesn't generate hyperlinks, here is an example of aliases to enable them in some common tools:
+
+```sh
+alias ls='ls --hyperlink --color=auto'
+alias ll='ls --hyperlink -Ahl --group-directories-first --color=auto'
```
bew, line 12: I think this page should focus on the hyperlink options in the aliases, other options like listing mode, color auto or grouping directories are a bit out of scope 🤔. 
Following this, I think the entire `ll` alias could be removed here.

What do you/others think?

docs/recipes/hyperlinks.md
```diff
@@ -0,0 +1,73 @@
...
+    action = act.OpenLinkAtMouseCursor,
+  },
+}
+
+return config
+```
+
+## Drawback
+
+This setup requires that your command prompt be empty before clicking a hyperlink; otherwise, the sent command may not execute as expected.
```
bew, line 73: Also note that this setup only works locally, not when we're connected on a remote machine

docs/recipes/hyperlinks.md
```diff
@@ -0,0 +1,73 @@
...
+      end
+    end
+  end
+
+  -- without a return value, we allow default actions
+end)
+
+config.mouse_bindings = {
+  {
+    -- Trigger open selection (default is without a modifier)
```
bew, line 61: Note that you should also disable Down event to avoid unexpected program behavior in some cases.
This is documented here:
https://wezterm.org/config/mouse.html#gotcha-on-binding-an-up-event-only

docs/recipes/hyperlinks.md
aorith: Totally true, I left the colors on `ls` tho

docs/recipes/hyperlinks.md
aorith: You're right, without that config and tmux copy-mode is activated afterwards

docs/recipes/hyperlinks.md
```diff
@@ -0,0 +1,78 @@
...
+end)
+
+config.mouse_bindings = {
+    -- Trigger open selection (default is without a modifier)
+    {
+      event = { Up = { streak = 1, button = "Left" } },
+      mods = "SUPER",
+      action = act.OpenLinkAtMouseCursor,
+    },
+    -- Disable the 'Down' event of CTRL-Click to avoid weird program behaviors
```
bew, line 65: Note: The comment mentions CTRL, but the keybind uses SUPER..
What do you think about using CTRL instead for the keybind?

docs/recipes/hyperlinks.md
```diff
@@ -0,0 +1,78 @@
...
+      action = act.Nop,
+    },
+}
+
+return config
+```
+
+## Drawbacks
+
+This setup works only locally (not over SSH) and requires an empty command prompt before clicking a hyperlink; otherwise, the command may not execute correctly.
```
bew, line 78: It also requires that you are at a shell prompt!
If you're in neovim for example it'll give very unexpected behavior.

I'd suggest making a list of items for the limitations, it's easier to skim through visually without having to read a paragraph

docs/recipes/hyperlinks.md
```diff
@@ -0,0 +1,78 @@
...
+
+This recipe is an example on how to leverage hyperlinks directly in the terminal.
+Hyperlinks are generated by terminal applications by using the [OSC-8 escape sequences](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda).
+
+## Requirements
+
+By default, `ls` doesn't generate hyperlinks, here is an example of aliases to enable them in some common tools:
+
+```sh
+alias ls='ls --hyperlink --color=auto'
```
digitallyserviced, line 11: while this may beyond the initial scope of this recipe, it may be worthwhile to note that core `ls` is not the only `ls` tool which provides hyperlink support and could provide an alternative to the coreutils on MacOS if desired.

this is actually what I use and afaik maintains argument compatibility as well as additional theming and other features.

[eza-community/eza](https://github.com/eza-community/eza?tab=readme-ov-file)

docs/recipes/hyperlinks.md
```diff
@@ -0,0 +1,78 @@
+# Use hyperlinks directly in the terminal
+
+This recipe is an example on how to leverage hyperlinks directly in the terminal.
+Hyperlinks are generated by terminal applications by using the [OSC-8 escape sequences](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda).
+
```
bew, line 5: I think you should add another paragraph describing what the experience of having this setup is.
For example mentioning that when you (ctrl/super-)click on a folder path in ls output, the shell will jump to that directory. Etc..

docs/recipes/hyperlinks.md
digitallyserviced: This does work over ssh, if the `file` check is bypassed (I checked). Since it utilizes just passing cmds to the pane, it should work regardless. so maybe also mentioning this would be helpful

bew:
Looking good now I think 🤔

docs/recipes/hyperlinks.md
```diff
@@ -0,0 +1,92 @@
...
+        return false
+      end
+    end
+  end
+
+  -- without a return value, we allow default actions
+end)
+
+config.mouse_bindings = {
+    -- Trigger open selection (default is without a modifier)
```
wez, line 63: I'm unclear why a ctrl-click binding is needed here; the text at the top of this example doesn't mention ctrl.
Should it?
If so, I think this comment should probably not mention the default behavior, as I think that makes this bit seem confusing to me: it implies that it is changing the default, but it isn't.

docs/recipes/hyperlinks.md
```diff
@@ -0,0 +1,92 @@
...
+return config
+```
+
+## Drawbacks
+
+This setup sends the text of the commands directly into the active pane which has some drawbacks:
+
+- Does not work inside of any terminal program, the pane must be in a shell prompt
+- Requires an empty command prompt before clicking a hyperlink; otherwise, the command may not execute correctly
+- Only works locally (not over SSH, docker, etc)
```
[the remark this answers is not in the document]
aorith: Great idea, this is what I came up with:

```lua
wezterm.on('open-uri', function(window, pane, uri)
  if uri:find('^file:') == 1 then
    -- uri format is: file://[HOSTNAME]/PATH[#linenr]
    local url = wezterm.url.parse(uri)

    local edit_cmd = url.fragment and 'nvim +' .. url.fragment .. ' "$_f"' or 'nvim "$_f"'
    local cmd = '_f="'
      .. url.file_path
      .. '"; { test -d "$_f" && { cd "$_f" ; ls -a -p --group-directories-first; }; } '
      .. '|| { test "$(file --brief --mime-type "$_f" | cut -d/ -f1 || true)" = "text" && '
      .. edit_cmd
      .. '; }; echo'

    pane:send_text(cmd .. '\r')

    return false
  end

  -- without a return value, we allow default actions
end)
```

I put an echo and the end to give some space after the cmd.
Regarding other filetypes, maybe it could fallback to `xdg-open` / `open`, or maybe that's overkill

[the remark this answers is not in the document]
digitallyserviced: Although I believe trying to ensure all these different methods and edge cases is good, they may be nice for a more robust plugin possibly. 

Adding more validation and fallbacks I would think for a simple copy and paste recipe I think adds way more complexity than what I thought was a simple and elegant functionality IMO. 

A simple implementation that is well documented makes for a great starter for anyone that could then tailor this to their systems and use cases.

One thing you could do is utilize the `InputSelector` and give the user options as to how to handle the operations that may not be so straightforward.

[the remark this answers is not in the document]
aorith: I like the original implementation too just because it doesn't bloat the prompt/history as much. But it just breaks badly over ssh/docker and might cause damage if run inside a terminal app or editor.

I'll check if there's a reliable way to tell if the pane is running a shell/ssh/app to try to make it more robust and fallback to the long command only over ssh

docs/recipes/hyperlinks.md
aorith: Yes, you're right, I added that because without the modifier I was opening hyperlinks when I wanted to select text.
Maybe I should just mention that they can reconfigure the mouse bindings

docs/recipes/hyperlinks.md
```diff
@@ -0,0 +1,126 @@
...
+    -- Also the pane is not in an alternate screen (an editor, less, etc)
+    local url = wezterm.url.parse(uri)
+    if is_shell(pane:get_foreground_process_name()) then
+      -- A shell has been detected. Wezterm can check the file type directly
+      -- figure out what kind of file we're dealing with
+      local success, stdout, _ = wezterm.run_child_process({ 'file', '--brief', '--mime-type', url.file_path })
+      if success then
+        if stdout:find('directory') then
+          pane:send_text(wezterm.shell_join_args({ 'cd', url.file_path }) .. '\r')
+          pane:send_text(wezterm.shell_join_args({ 'ls', '-a', '-p', '--group-directories-first' }) .. '\r')
```
gitworkflows, line 55:
```suggestion
          pane:send_text(wezterm.shell_join_args({ 'ls', '-a', '-p', '--hyperlink', '--group-directories-first' }) .. '\r')
```

docs/recipes/hyperlinks.md
```diff
@@ -0,0 +1,126 @@
...
+- Other hyperlinks like URLs remain unchanged and follow WezTerm's default behavior
+
+```lua
+local wezterm = require("wezterm")
+local act = wezterm.action
+local config = wezterm.config_builder()
+
+local function is_shell(foreground_process_name)
+  local shell_names = { 'bash', 'zsh', 'fish', 'sh', 'ksh', 'dash' }
+  local process = string.gsub(foreground_process_name, '(.*[/\\])(.*)', '%2')
```
gitworkflows, line 34:
```suggestion
  local process = string.match(foreground_process_name, "[^/\\]+$") or foreground_process_name
```

gitworkflows (changes requested):
# Add missing hyperlink flag
The code doesn't include the '--hyperlink' flag when executing 'ls' after 
changing directory, which is inconsistent with the recipe's purpose of enabling 
hyperlinks in the terminal.

# Fix path extraction regex
The current regex pattern for extracting the process name doesn't handle paths 
with multiple slashes correctly. It should be modified to extract just the 
filename from the full path.

docs/recipes/hyperlinks.md
```diff
@@ -0,0 +1,126 @@
...
+      action = act.Nop,
+    },
+}
+```
+
+## Drawbacks
+
+This setup sends the text of the commands directly into the active pane which has some drawbacks:
+
+- Does not work inside of any terminal program, the pane must be in a shell prompt
```
gitworkflows, line 116:
```suggestion
- Does not work inside of any interactive terminal program that is not a shell; the pane must be in a shell prompt
```

docs/recipes/hyperlinks.md
```diff
@@ -0,0 +1,126 @@
...
+- Requires an empty command prompt before clicking a hyperlink; otherwise, the command may not execute correctly
+- When a shell is not detected (see the `is_shell` function), WezTerm falls back to a long shell command, which may clutter the prompt slightly. This happens because there is no reliable way to determine a file's MIME type.
```
gitworkflows, line 118:
```suggestion
- When a shell is not detected (see the `is_shell` function), WezTerm falls back to a long shell command, which may clutter the prompt slightly. This happens because WezTerm cannot directly determine a file's MIME type when not connected to a local shell.
```

gitworkflows (changes requested):
Clarify the scope of the limitation & precision

docs/recipes/hyperlinks.md
aorith: The idea here is that the user will already have an alias for `ls` which is will be used in the local shell.

digitallyserviced:
@aorith 

When I suggested to create a recipe I really believed the simple and succinct one you originally posted was perfect. As far as being something that showed showcased a user created feature from the event, to a basic implementation, and did not complicate or try to handle all sorts of edge cases.

I commend your effort and apologize if this turned out much more than intended.

Looks great 👍 though.

I do believe it should be good for now. Anything  further should really be done with a plugin.

wez:
Thank you! I appreciate the effort!

PR #6801, opened 2025-03-15
Imgcat enhancements

mgpinf:
* Set raw mode when polling input in order to avoid local echo if `--hold` option is set
* Add key modifier combinations to exit polling input if `--hold` option is set:
  * `ctrl-c`
  * `ctrl-d`
  * `Escape`


wez:
Thank you!

PR #6805, opened 2025-03-17
Remove redundant character checks in Launcher, InputSelector

mgpinf:
We don't need to check if the alphabet includes `j`, `k` because if a character is included in the alphabet, the code corresponding to the first arm is executed


wez:
Thanks!

Issue #6341, opened 2024-10-31
Next Release Checklist


wez:
**BEFORE YOU COMMENT: please avoid duplicating information from other comments to avoid overwhelming this one.**
Prefer to use a thumbs up instead of a me-too comment.

We're overdue for making a new release.
Here's what's on my mind as release gating issues, and I need your help to check these off.
So if you are interested in having a new stable tag, please help by running through the list below and giving some feedback on the state of these things for your system:

* [ ] Verify that Wayland support isn't an obvious regression since the last stable release.
     *  Wayland support was essentially rewritten by the community, and since I don't use wayland, I don't have sufficient context now to understand where it has problems either in the code or due to compatibility.
     * If you are a Wayland user, and you are using a wezterm nightly build from October 2024 or later and having a good experience, please comment below with the name of your compositor and DE to say that it is as good as or better than the prior stable release.
     * If there is already a comment describing your environment, instead of posting a duplicate, please give it a thumbs-up reaction to "vote" for it if it is working well for you.
     * If it isn't working working as well or better than the prior stable release, please reference an issue that describes the problem and comment below.

* [ ] Verify that the default configuration doesn't crash out.
     * Run `wezterm -n` on your system to launch it without your config.  Does it successfully launch and display text with no artifacts or blocks for glyphs?
     * If not:
         *  try changing the [front_end](https://wezfurlong.org/wezterm/config/lua/config/front_end.html) setting to see if that is better:
        * `wezterm -n --config "front_end='WebGpu'"`
        * Please find/file an issue reporting the outcome of both the default and webgpu, along with information about the environment. **If the behavior is a regression since the prior stable release**, reference that issue here.
 
* [ ] Confirm that we're building for the current batch of OS/distros
    * https://github.com/wez/wezterm/pull/6156
    * https://github.com/wez/wezterm/pull/6200
    * Others? Please comment here 

* Other misc issues:
   * https://github.com/wez/wezterm/pull/6186

rootiest:
###  Specs
KDE Plasma Version: 6.2.2
KDE Frameworks Version: 6.7.0
Qt Version: 6.8.0
Kernel Version: 6.11.5-zen1-1-zen (64-bit)
Graphics Platform: Wayland

Processors: 16 x AMD Ryzen 9 4900HS with Radeon Graphics
Memory: 22.9 GiB of RAM
Graphics Processor: AMD Radeon Graphics (+ Nvidia RTX 2060)
[...]

PR #6821, opened 2025-03-24
lua-api-crates/mux: expose swap_active_with_index

SandWoodJones:
This is related to [this discussion of swapping panes ](https://github.com/wezterm/wezterm/discussions/3331). For an usage example:
```lua
local function swap_panes(dir)
  return wezterm.action_callback(function(win, pane)
    local tab = win:active_tab()
    local target_pane = tab:get_pane_direction(dir)
    if target_pane then
      local panes = tab:panes()
      local target_index = nil
      
      for i, p in ipairs(panes) do
        if p.pane_id == target_pane.pane_id then
          target_index = i - 1
          break
        end
      end
      
      if target_index then
        tab:swap_active_with_index(target_index, true)
      end
    end
  end)
end

config = {
  keys = {
    { key = 'h', mods = 'CTRL|ALT', action = swap_panes("Left") },
    { key = 'j', mods = 'CTRL|ALT', action = swap_panes("Down") },
    { key = 'k', mods = 'CTRL|ALT', action = swap_panes("Up") },
    { key = 'l', mods = 'CTRL|ALT', action = swap_panes("Right") },
  }
}
```

### DIFF
--- a/docs/config/lua/MuxTab/swap_active_with_index.md
+++ b/docs/config/lua/MuxTab/swap_active_with_index.md
@@ -1,0 +1,54 @@
+# `tab:swap_active_with_index(pane_index, keep_focus)`
+
+Swaps the position of the active pane with the specified pane index. If `keep_focus` is false, focus will change to the selected pane, otherwise focus stays on the currently active pane in its new position.
+
+# Example: tmux-like pane marking and swapping
+
+```lua
+local marked_pane = nil
+
+wezterm.on('mark', function(window, pane)
+  id = pane:pane_id()
+  tab = window:active_tab()
+
+  marked_pane = { id, tab }
+end)
+
+wezterm.on('swap-pane', function(window, pane)
+  if not marked_pane then
+    wezterm.log_error 'No marked pane'
+    return
+  end
+
+  local cur_tab = window:active_tab()
+
+  -- Do nothing if marked pane is same as current
+  if marked_pane[1] == cur_tab:active_pane():pane_id() then
+    return
+  end
+
+  if marked_pane[2]:tab_id() ~= cur_tab:tab_id() then
+    wezterm.log_error 'Pane is on another tab'
+    return
+  end
+
+  local target_index = nil
+  for i, p in ipairs(cur_tab:panes()) do
+    if p:pane_id() == marked_pane[1] then
+      target_index = i - 1
+      break
+    end
+  end
+
+  if target_index then
+    cur_tab:swap_active_with_index(target_index, true)
+  end
+end)
+
+config = {
+  keys = {
+    { key = 'm', mods = 'CTRL', action = wezterm.action.EmitEvent 'mark' },
+    { key = 'm', mods = 'ALT', action = wezterm.action.EmitEvent 'swap-pane' },
+  },
+}
+```
--- a/docs/config/lua/MuxTab/swap_active_with_index.md
+++ b/docs/config/lua/MuxTab/swap_active_with_index.md
@@ -32,17 +32,7 @@ wezterm.on('swap-pane', function(window, pane)
     return
   end
 
-  local target_index = nil
-  for i, p in ipairs(cur_tab:panes()) do
-    if p:pane_id() == marked_pane[1] then
-      target_index = i - 1
-      break
-    end
-  end
-
-  if target_index then
-    cur_tab:swap_active_with_index(target_index, true)
-  end
+  cur_tab:swap_active_with_id(marked_pane[1], true)
 end)
 
 config = {
--- a/lua-api-crates/mux/src/tab.rs
+++ b/lua-api-crates/mux/src/tab.rs
@@ -155,5 +155,14 @@ impl UserData for MuxTab {
             }
             Ok(())
         });
+
+        methods.add_method("swap_active_with_index", |_, this, (pane_index, keep_focus): (usize, bool)| {
+            let mux = get_mux()?;
+            let tab = this.resolve(&mux)?;
+
+            let success = tab.swap_active_with_index(pane_index, keep_focus).is_some();
+
+            Ok(success)
+        });
     }
 }
--- a/lua-api-crates/mux/src/tab.rs
+++ b/lua-api-crates/mux/src/tab.rs
@@ -156,13 +156,19 @@ impl UserData for MuxTab {
             Ok(())
         });
 
-        methods.add_method("swap_active_with_index", |_, this, (pane_index, keep_focus): (usize, bool)| {
+        methods.add_method("swap_active_with_id", |_, this, (pane_id, keep_focus): (Value, bool)| {
             let mux = get_mux()?;
             let tab = this.resolve(&mux)?;
 
-            let success = tab.swap_active_with_index(pane_index, keep_focus).is_some();
+            let target_id: PaneId = from_lua(pane_id)?;
 
-            Ok(success)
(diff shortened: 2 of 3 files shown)


digitallyserviced:
@SandWoodJones 

While I would argue that such a method in the API would be beneficial, the example use case is rather redundant when comparing to the `rotate_panes_clockwise` set of methods already available.

This originally confused me because I thought such a method already exists. It does but is limited to adjacent panes, which your example simply duplicates albeit with a different API call.

For this truly to be seen as an enhancement and not just another terminology for an existing feature, this should be able to swap panes with others that are not simply adjacent, and showcase this with your example.

Perhaps similar to how tmux you can `mark` a pane and move to another tab and then call `swap-pane` to switch the current focused pane with the marked one.

Also you need documentation and description of this method. This should explain as to why this is different, as well as provide an example like your OP but expand beyond adjacent panes.

SandWoodJones:
> This originally confused me because I thought such a method already exists. It does but is limited to adjacent panes, which your example simply duplicates albeit with a different API call.

What method is that?

digitallyserviced:
@SandWoodJones 

I mentioned it in my reply..  `RotatePanes`

https://wezterm.org/config/lua/keyassignment/RotatePanes.html

Like I said, your example simply deals with adjacent panes which technicaly is covered by this event already.

All i'm saying is you need documentation added to this PR, as well as an example that would handle swapping the pane with an actual index of a pane NOT adjacent to it.

Simply make a new event that stores the active pane id/index on key bind. Then make another event that would swap the "stored" pane id/index with the newly focused/active one...

digitallyserviced:
@SandWoodJones 

So I was not aware of this, and not sure how this changes what you have madee available within your API changes. I believe your API changes should still be available to users, however I just noticed that the PaneSelect action, seems to provide this functionality already.

https://wezterm.org/config/lua/keyassignment/PaneSelect.html?h=pane

```lua
    action = act.PaneSelect {
      mode = 'SwapWithActive',
    },
```

SandWoodJones:
I was aware of this action and I even referenced it as I was figuring out swap_active_with_index. It didn't fit my needs however. I can see it overlapping with pane marking as I wrote in the documentation, but I feel like it still stands on its own.

nonolai:
Hey there, I was tinkering around on my own PR for this when I realized this one already existed, and had a couple thoughts on this:

- I think the Lua function would be cleaner if it took Pane ID instead of Pane index. Indices seem to be somewhat ephemeral, but Pane ID consistently identifies a given pane. Additionally, for the purposes of both the marking example and getting the directional swapping of `tmux`, the relevant functions (i.e. `pane:pane_id` and `tab:get_pane_direction` respectively) return Pane ID.
- `Tab` and `TabInner` (in `mux/src/tab.rs`) should have a `idx_from_id` function analogous to the one in `Window` (in `mux/src/window.rs`).

When combined, these two recommendations (imo) greatly simplify the Lua code and give us more options on the Rust side, both for error handling on this feature and in the future.

SandWoodJones:
Thanks for the feedback. My goal here was to expose the method for my own use in my config and so my changes already provide what I need. I do think pane ID would be more straightforward but I feel like that requires Rust work and falls a bit out of scope of this PR.
I'd love to get this merged

nonolai:
I have a [commit](https://github.com/wezterm/wezterm/commit/4781d18f32401ab4870a00bb531a529df43125dc) I was using for my local build for a while that I think does what you'd need if you want to attempt it.

I'm not a maintainer, so I'm not blocking here, just another user coming from `tmux` who wants this feature. Either way (if this gets merged as is or tweaked to use pane ID) I'll get to use this feature :)

SandWoodJones:
I've implemented your code as a separate commit and changed the documentation to reflect that.

nonolai:
Awesome! Looking forward to seeing this feature merged

SandWoodJones:
It seems like the reviewer for this PR is no longer active. Could we get some eyes on this and get this merged?

docs/config/lua/MuxTab/swap_active_with_index.md
```diff
@@ -0,0 +1,44 @@
+# `tab:swap_active_with_index(pane_index, keep_focus)`
```
bew, line 1:
```suggestion
# `tab:swap_active_with_id(pane_id, keep_focus)`
```

docs/config/lua/MuxTab/swap_active_with_index.md
```diff
@@ -0,0 +1,44 @@
...
+
+Swaps the position of the active pane with the specified pane index. If `keep_focus` is false, focus will change to the selected pane, otherwise focus stays on the currently active pane in its new position.
+
+# Example: tmux-like pane marking and swapping
+
+```lua
+local marked_pane = nil
```
bew, line 8: Note that since the Lua config can be reloaded at different times (which you do not control), using a global is not recommended, you should use `wezterm.GLOBAL.marked_pane` instead.

docs/config/lua/MuxTab/swap_active_with_index.md
```diff
@@ -0,0 +1,44 @@
...
+    wezterm.log_error 'Pane is on another tab'
+    return
+  end
+
+  cur_tab:swap_active_with_id(marked_pane[1], true)
+end)
+
+config = {
+  keys = {
+    { key = 'm', mods = 'CTRL', action = wezterm.action.EmitEvent 'mark' },
```
bew, line 40: note: I'd recommend using `wezterm.action_callback` instead of manual on/EmitEvent

docs/config/lua/MuxTab/swap_active_with_index.md
```diff
@@ -0,0 +1,44 @@
+# `tab:swap_active_with_index(pane_index, keep_focus)`
+
+Swaps the position of the active pane with the specified pane index. If `keep_focus` is false, focus will change to the selected pane, otherwise focus stays on the currently active pane in its new position.
```
bew, line 3:
```suggestion
Swaps the position of the active pane with the specified pane ID.
If `keep_focus` is false, focus will change to the selected pane, otherwise focus stays on the currently active pane in its new position.
```

docs/config/lua/MuxTab/swap_active_with_index.md
```diff
@@ -0,0 +1,44 @@
...
+wezterm.on('swap-pane', function(window, pane)
+  if not marked_pane then
+    wezterm.log_error 'No marked pane'
+    return
+  end
+
+  local cur_tab = window:active_tab()
+
+  -- Do nothing if marked pane is same as current
+  if marked_pane[1] == cur_tab:active_pane():pane_id() then
```
bew, line 26: I don't think you need this check, the API function should already have this check

docs/config/lua/MuxTab/swap_active_with_index.md
```diff
@@ -0,0 +1,44 @@
...
+    return
+  end
+
+  if marked_pane[2]:tab_id() ~= cur_tab:tab_id() then
```
bew, line 30: What do you think about implementing support to swap with a pane ID from another tab 🤔

otherwise, please add a log line in the Lua-API function to ensure we at least get a log if Lua-side doesn't check.

bew (changes requested):
A few comments 😉

👉 I'm interested in starting a community fork, see https://github.com/wezterm/wezterm/issues/6341#issuecomment-4483047032, so this might gets merged there once ready.

PR #6849, opened 2025-04-03
Remove `mode` from `PushKittyState` CSI representation

the-mikedavis:
The [Kitty Keyboard document] specifies

    CSI = flags ; mode u

in the "Progressive enhancement" section stating that the `mode` (which controls how the flags are set) is optional. It then talks about pushing and popping flags - useful for applications using an alternate screen - with the sequence

    CSI > flags u  # for push, if flags omitted default to zero
    CSI < number u # to pop number entries, defaulting to 1 if unspecified

Termwiz previously also emitted and accepted the `mode` for the "push" command, but I believe it is a mistake. The docs don't explicitly mention the mode here - which you might think is because it is optional - but also the Kitty codebase does not accept a mode:

* [`screen_push_key_encoding_flags`] does not accept a `how` parameter like the definition of `screen_set_key_encoding_flags` above it.
* the Python client used to record and replay commands has a function of the same name which also lacks a `how` parameter: <https://github.com/kovidgoyal/kitty/blob/b0371d970fac15a414ea15fcf25c20269b4b4f65/kitty/client.py#L184-L185>

[Kitty Keyboard document]: https://sw.kovidgoyal.net/kitty/keyboard-protocol/
[`screen_push_key_encoding_flags`]: https://github.com/kovidgoyal/kitty/blob/b0371d970fac15a414ea15fcf25c20269b4b4f65/kitty/screen.c#L1692-L1704

### DIFF
--- a/term/src/terminalstate/performer.rs
+++ b/term/src/terminalstate/performer.rs
@@ -537,17 +537,13 @@ impl<'a> Performer<'a> {
                         .push(KeyboardEncoding::Kitty(flags));
                 }
             }
-            CSI::Keyboard(Keyboard::PushKittyState { flags, mode }) => {
+            CSI::Keyboard(Keyboard::PushKittyState(flags)) => {
                 if self.config.enable_kitty_keyboard() {
                     let current_flags = match self.screen().keyboard_stack.last() {
                         Some(KeyboardEncoding::Kitty(flags)) => *flags,
                         _ => KittyKeyboardFlags::NONE,
                     };
-                    let flags = match mode {
-                        KittyKeyboardMode::AssignAll => flags,
-                        KittyKeyboardMode::SetSpecified => current_flags | flags,
-                        KittyKeyboardMode::ClearSpecified => current_flags - flags,
-                    };
+                    let flags = current_flags | flags;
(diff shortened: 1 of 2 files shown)


term/src/terminalstate/performer.rs
```diff
@@ -537,17 +537,13 @@ impl<'a> Performer<'a> {
...
                     let current_flags = match self.screen().keyboard_stack.last() {
                         Some(KeyboardEncoding::Kitty(flags)) => *flags,
                         _ => KittyKeyboardFlags::NONE,
                     };
-                    let flags = match mode {
-                        KittyKeyboardMode::AssignAll => flags,
-                        KittyKeyboardMode::SetSpecified => current_flags | flags,
-                        KittyKeyboardMode::ClearSpecified => current_flags - flags,
-                    };
+                    let flags = current_flags | flags;
```
the-mikedavis, line 546: I'm not totally confident about this part. To me the `screen_push_key_encoding_flags` function from Kitty is a litle cryptic so I'm not sure which mode it's emulating. I think it's merging with the top of the stack :sweat_smile:

PR #6850, opened 2025-04-03
fix: honor environment variables and cwd unconditionally for SpawnCommand

### LINKED ISSUE #6845: SwitchToWorkspace does not pass through environment variables set in SpawnCommand

macOS

_No response_

### WezTerm version

20250320-072107-a8735851

### Describe the bug

When passing a `SpawnCommand` to the `SwitchToWorkspace` function, the `set_environment_variables` field is ignored.

### To Reproduce

1. Use the configuration below (from https://wezterm.org/config/lua/keyassignment/InputSelector.html), where the `set_environment_variables` has been set.
2. Switch to a workspace using Ctrl-Shift-S.
3. `echo $TEST` prints nothing.

### Configuration

```
local act = wezterm.action

config.keys = {
  {
    key = 'S',
    mods = 'CTRL|SHIFT',
    action = wezterm.action_callback(function(window, pane)
      -- Here you can dynamically construct a longer list if needed

      local home = wezterm.home_dir
      local workspaces = {
        { id = home, label = 'Home' },
        { id = home .. '/work', label = 'Work' },
        { id = home .. '/personal', label = 'Personal' },
        { id = home .. '/.config', label = 'Config' },
      }

      window:perform_action(
        act.InputSelector {
          action = wezterm.action_callback(
            function(inner_window, inner_pane, id, label)
              if not id and not label then
                wezterm.log_info 'cancelled'
              else
                wezterm.log_info('id = ' .. id)
                wezterm.log_info('label = ' .. label)
                inner_window:perform_action(
                  act.SwitchToWorkspace {
                    name = label,
                    spawn = {
                      label = 'Workspace: ' .. label,
                      cwd = id,
                      set_environment_variables = { TEST = "OK" },
                    },
                  },
                  inner_pane
                )
              end
            end
          ),
          title = 'Choose Workspace',
          choices = workspaces,
          fuzzy = true,
          fuzzy_description = 'Fuzzy find and/or make a workspace',
        },
        pane
      )
    end),
  },
}

return config
```

### Expected Behavior

`echo $TEST` shall return `OK`.

### Logs

_No response_

### Anything else?

- Setting `args` to `{ '/bin/zsh' }' makes it work.
- Setting `default_prog` to `/bin/zsh` does not solve the issue.

vincentbesanceney:
The SpawnCommand specifies information about a new command to be spawned. If the args field is omitted, the default program for the target domain is spawned. However, this also causes other fields like environment variables and cwd to be wiped out.

The command builder now behaves consistently for all fields, whether the args field is set or omitted.

Fixes #6845

### DIFF
--- a/wezterm-gui/src/spawn.rs
+++ b/wezterm-gui/src/spawn.rs
@@ -68,19 +68,26 @@ pub async fn spawn_command_internal(
         None
     };
 
-    let cmd_builder = {
-        let mut builder = if let Some(args) = spawn.args {
-            CommandBuilder::from_argv(args.iter().map(Into::into).collect())
-        } else {
-            CommandBuilder::new_default_prog()
-        };
-        for (k, v) in spawn.set_environment_variables.iter() {
-            builder.env(k, v);
+    let cmd_builder = match (
+        spawn.args.as_ref(),
+        spawn.cwd.as_ref(),
+        !spawn.set_environment_variables.is_empty(),
+    ) {
+        (None, None, false) => None,
+        _ => {
+            let mut builder = spawn
+                .args
+                .as_ref()
+                .map(|args| CommandBuilder::from_argv(args.iter().map(Into::into).collect()))
+                .unwrap_or_else(CommandBuilder::new_default_prog);
+            for (k, v) in spawn.set_environment_variables.iter() {
+                builder.env(k, v);
+            }
+            if let Some(cwd) = &spawn.cwd {
+                builder.cwd(cwd);
+            }
+            Some(builder)
         }
-        if let Some(cwd) = spawn.cwd {
-            builder.cwd(cwd);
-        }
-        Some(builder)
     };
 
     let workspace = mux.active_workspace().clone();
--- a/wezterm-gui/src/spawn.rs
+++ b/wezterm-gui/src/spawn.rs
@@ -68,8 +68,12 @@ pub async fn spawn_command_internal(
         None
     };
 
-    let cmd_builder = if let Some(args) = spawn.args {
-        let mut builder = CommandBuilder::from_argv(args.iter().map(Into::into).collect());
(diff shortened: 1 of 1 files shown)


wezterm-gui/src/spawn.rs
```diff
@@ -68,17 +68,19 @@ pub async fn spawn_command_internal(
...
+        };
         for (k, v) in spawn.set_environment_variables.iter() {
             builder.env(k, v);
         }
         if let Some(cwd) = spawn.cwd {
             builder.cwd(cwd);
         }
         Some(builder)
-    } else {
-        None
```
wez, line 81: there's some nuance to this; some domains (eg: serial) will refuse to spawn with anything other than `None` because they cannot build any kind of command.

Could you make another pass at this to ensure that we end up with `None` if none of the spawn-provided fields are set?

Thanks!

vincentbesanceney: Oh ok, I didn't know about serial domains. Here you go. 😊

wezterm-gui/src/spawn.rs
```diff
@@ -68,17 +68,26 @@ pub async fn spawn_command_internal(
...
-            builder.env(k, v);
-        }
-        if let Some(cwd) = spawn.cwd {
-            builder.cwd(cwd);
+    let cmd_builder = match (
+        spawn.args.as_ref(),
+        spawn.cwd.as_ref(),
+        !spawn.set_environment_variables.is_empty(),
+    ) {
+        (None, None, false) => None,
```
wez, line 76: Negative logic can be harder to grok, let's flip this around so that it is a bit easier to follow!

```suggestion
        spawn.set_environment_variables.is_empty(),
    ) {
        (None, None, true) => None,
```

wez:
Thank you!

PR #6856, opened 2025-04-06
Support SGR DECRQSS

the-mikedavis:
The SGR DECRQSS/DECRPSS request/response is useful for detecting SGR features like true-color and extended underlines even when terminfo is not installed properly. The response to the DECRQSS "DCS $ q m ST" prints the current SGR (`self.pen` of the `Performer`) in the form "DCS 1 $ r D...D m ST" where the SGR sequences match regular CSI SGR but without the CSI or 'm' bytes, semicolon separated.

I know that Neovim already uses this query as part of its detection routine - https://github.com/neovim/neovim/pull/28052 - and I'd like to add the same in Helix to reduce false negative detections of true-color and squiggly underlines.


term/src/terminalstate/performer.rs
```diff
@@ -334,6 +334,20 @@ impl<'a> Performer<'a> {
...
+                                write!(self.writer, "{}1$r", DCS).ok();
+                                let mut first = true;
+                                for sgr in self.pen.sgrs() {
+                                    if !first {
+                                        write!(self.writer, ";").ok();
+                                    }
+                                    first = false;
+                                    write!(self.writer, "{}", sgr).ok();
+                                }
+                                write!(self.writer, "m{}", ST).ok();
```
j4james, line 348: If I'm reading this correctly, I think you might me missing a `0` parameter at the start of the report. For example, if the bold and underline attributes were set, the response should be `DCS 1$r0;1;4m ST`.

The reason for this is that an application should be able to "play back" the content of any `DECRPSS` report to restore the terminal to the state it had at the time of the query. So for `SGR`, that requires you to clear any attributes that may already be set, before setting the ones that need to be restored.

For the default state, i.e. when no attributes are set, it doesn't really matter, but it's simpler to use a `0` there too, i.e. `DCS 1$r0m ST`. And as far as I'm aware, that was the way the original DEC terminals implemented it.

the-mikedavis: Thanks! I've seen some discussion about whether the zero is required but not a clear consensus. The vt100.net docs use it and I've seen some terminals like Foot use it. Kitty does not though. It simplifies this section of code here (we can do away with the `first` check) so I'm happy to add it. Plus it doesn't change how I would handle it on the application side.

PR #6865, opened 2025-04-10
Make some termwiz dependency optional

quark-zju:
Makes some `termwiz` dependencies optional so it's lighter weight for users.

Context: https://discord.com/channels/968932220549103686/1358968929414021190


wez:
Looks good, thank you!

PR #6868, opened 2025-04-10
Parse all permutations of optional `P s` parameters for `CSI P s ; P s`

### LINKED ISSUE #6860: Cursor position (CUP) Xterm control sequence parameters aren't optional

macOS

_No response_

### WezTerm version

20250320-072107-a8735851

### Describe the bug

In the Cursor Position (CUP) Xterm control sequence `CSI P s ; P s H`, both `P s` arguments should be optional, and should default to 1 if not present. They are currently mandatory in wezterm.

### To Reproduce

Run `printf '\033[;H'` and observe that the cursor's position isn't changed, but is changed with e.g. `printf '\033[1;1H'`

### Configuration

no config

### Expected Behavior

_No response_

### Logs

_No response_

### Anything else?

The correct behaviour can be seen in Xterm, and is described in https://www.xfree86.org/current/ctlseqs.html

wojciech-graj:
The following commands previously failed to parse (and therefore didn't work), but work in Xterm.
```
printf '\033[;H'
printf '\033[2;H'
```

Closes #6860

### DIFF
--- a/termwiz/src/escape/csi.rs
+++ b/termwiz/src/escape/csi.rs
@@ -1265,8 +1265,8 @@ impl ParseParams for OneBased {
 impl ParseParams for (OneBased, OneBased) {
     fn parse_params(params: &[CsiParam]) -> Result<(OneBased, OneBased), ()> {
         match params {
-            [] => Ok((OneBased::new(1), OneBased::new(1))),
-            [p] => Ok((OneBased::from_esc_param(p)?, OneBased::new(1))),
+            [] | [CsiParam::P(b';')] => Ok((OneBased::new(1), OneBased::new(1))),
+            [a] | [a, CsiParam::P(b';')] => Ok((OneBased::from_esc_param(a)?, OneBased::new(1))),
             [a, CsiParam::P(b';'), b] => {
                 Ok((OneBased::from_esc_param(a)?, OneBased::from_esc_param(b)?))
             }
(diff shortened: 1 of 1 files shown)


digitallyserviced:
@wojciech-graj 

I'm glad someone knew what they were doing, I looked at the tests and wondered what I was missing because it seemed like they were comparing the wrong parts, an already tokenized array of params, vs. the actual bytestring being parsed.

I had the same fix but the tests gave me a bit of a mind fuck.

Looks good. When you mentioned `CUP` i looked at the specs on the docs you listed and realized there's a ton of shit not setup for default values properly.

bew (approved):
This is indeed some interesting tests, but it actually makes sense once you get past the fact that you give the parse function some csi-specific lexer tokens 😅

digitallyserviced:
> This is indeed some interesting tests, but it actually makes sense once you get past the fact that you give the parse function some csi-specific lexer tokens 😅

@bew

After digging deeper through some more tests throughout the escape mod and other parser code, it would seem that some of those tests are actually geared for an intermediary decoding step; parsed some bytes to a vec of csiparam tokens(tested elsewhere) and now we want to make sure they end up filling/actionifying what they should. As well as being able to encode those csiparam tokens to a proper bytes representation.

It would seem other places may be also having issues with the defaults and/or lack of params, like the sgr/dec querying, and not to mention the many escapes that are just left out because of either barely used, or "fuck that, aint nobody got time for that".

I didn't really believe that what the big comment block I read on iTerm2's VT100 handler, and some of the long winded descriptions of some stuff on ctlseq about these standards impl and the many interpretations of the docs being out of whack. Now, after going through wezterm, alacritty, and iterm2 briefly, it made me realize how true it is and the difficulty these standards ended dealing with so many additions and changes over time (vt9999) have done.

hell, apparently arch's ncurses package has a wezterm terminfo/cap included that completely fucks alt screen rendering up with scroll regions destroying and artifacting everything. not sure why they do include it.

PR #6872, opened 2025-04-11
Fix Kitty keyboard encoding of Escape with 'disambiguate' flag enabled

the-mikedavis:
This fixes a small bug in the `KeyEvent::encode_kitty` function for the Escape key. When the `DISAMBIGUATE_ESCAPE_CODES` flag is set the representation of escape should use the form `CSI 27 u` rather than the raw char. (See the quoted part of the new test.)

### DIFF
--- a/wezterm-input-types/src/lib.rs
+++ b/wezterm-input-types/src/lib.rs
@@ -1723,6 +1723,9 @@ impl KeyEvent {
             match &self.key {
                 Char('\x08') => return '\x7f'.to_string(),
                 Char('\x7f') => return '\x08'.to_string(),
+                Char('\u{1b}') if flags.contains(KittyKeyboardFlags::DISAMBIGUATE_ESCAPE_CODES) => {
+                    return "\u{1b}[27u".to_string()
+                }
                 Char(c) => return c.to_string(),
                 _ => {}
             }
@@ -3206,6 +3209,28 @@ mod test {
             }
             .encode_kitty(flags),
             "\u{1b}[102;14u".to_string()
+        );
+    }
+
+    #[test]
+    fn encode_escape_disambiguated() {
+        // See <https://sw.kovidgoyal.net/kitty/keyboard-protocol/#disambiguate-escape-codes>:
+        // > Turning on this flag will cause the terminal to report the `Esc`, ..., keys using
+        // > `CSI u` sequences instead of legacy ones.
+        let flags = KittyKeyboardFlags::DISAMBIGUATE_ESCAPE_CODES;
+        assert_eq!(
+            KeyEvent {
+                key: KeyCode::Char('\u{1b}'), // Escape
+                modifiers: Modifiers::NONE,
+                leds: KeyboardLedStatus::empty(),
+                repeat_count: 1,
+                key_is_down: true,
+                raw: None,
+                #[cfg(windows)]
+                win32_uni_char: None,
+            }
+            .encode_kitty(flags),
+            "\u{1b}[27u"
         );
     }
 }


rawsp33d:
I encountered [what I believe to be] the same issue. I found that in neovim 0.12 with `enable_kitty_keyboard` set, I have to double tap escape to get it to send once. This does not happen in other terminals I had lying around, and does not happen if I disable the kitty keyboard option. And doesn't happen in older versions of neovim... I don't know why all those stars have to align :shrug:

Felixoid:
Dear @bew, I've seen your message in the #6341 PR, and would like to point to this issue in case you're collecting the PRs worth merging ASAP into the community fork.

bew:
> Dear @bew, I've seen your message in the #6341 PR, and would like to point to this issue in case you're collecting the PRs worth merging ASAP into the community fork.

Yes! Thanks a lot for pointing that PR to me!
I've added it to the Test&Review pile for the fork 👍

Do you know if there is an issue about this bug? 🤔

👉In the meantime if you and maybe a few others can check it out, build it and validate that it solves the problem for them it would also help out a lot!

Felixoid:
> Do you know if there is an issue about this bug? 🤔

I found it in https://github.com/earendil-works/pi/issues/4323 and then pushed two fixes, there in https://github.com/earendil-works/pi/pull/4482 and in https://github.com/wezterm/wezterm/pull/7787

the-mikedavis:
(Force-push was a rebase.) I last tried this a while ago but it did work correctly for me on Windows. I can reboot into Windows and give it another try. @Rahmyeon checked #7787 which is partially the same change - would you mind checking again?

wezterm-input-types/src/lib.rs
```diff
@@ -1723,6 +1723,9 @@ impl KeyEvent {
             match &self.key {
                 Char('\x08') => return '\x7f'.to_string(),
                 Char('\x7f') => return '\x08'.to_string(),
+                Char('\u{1b}') if flags.contains(KittyKeyboardFlags::DISAMBIGUATE_ESCAPE_CODES) => {
+                    return "\u{1b}[27u".to_string()
+                }
```
Felixoid, line 1728: I think, for consistency with other cases, we should produde `\u{1b}[27u;1u` here as other keys.

[Spec](https://sw.kovidgoyal.net/kitty/keyboard-protocol/#disambiguate-escape-codes) doesn't require it, but the form `CSI number; modifier u` is used everywhere else.


```suggestion
                // Skip ESC here to fall through to the CSI-u path below
                Char('\u{1b}') if flags.contains(KittyKeyboardFlags::DISAMBIGUATE_ESCAPE_CODES) => {}
```

wezterm-input-types/src/lib.rs
```diff
@@ -3208,4 +3211,26 @@ mod test {
...
+                modifiers: Modifiers::NONE,
+                leds: KeyboardLedStatus::empty(),
+                repeat_count: 1,
+                key_is_down: true,
+                raw: None,
+                #[cfg(windows)]
+                win32_uni_char: None,
+            }
+            .encode_kitty(flags),
+            "\u{1b}[27u"
```
Felixoid, line 3233:
```suggestion
            "\u{1b}[27u;1u"
```

And, for the best, case when there's no `KittyKeyboardFlags::DISAMBIGUATE_ESCAPE_CODES`

Felixoid:
I reopened #7787, I feel it's more consistent with other `DISAMBIGUATE_ESCAPE_CODES`

the-mikedavis:
Sure, let's continue in #7787

Issue #6328, opened 2024-10-26
Interest in custom Branch drawing symbols 🤔


bew:
Hello Wez o/

I found out about [this PR](https://github.com/kovidgoyal/kitty/pull/7681) (and [its followup](https://github.com/kovidgoyal/kitty/pull/7805)) to kitty terminal to add custom branch-drawing symbols, and was wondering if you'd accept a PR to add similar symbols to Wezterm 🤔

![](https://github.com/user-attachments/assets/4eb16021-7169-4ada-99e5-672a062e52f2)

(repeating part of that PR here…) These symbols are on an arbitrary range (0xf5d0 - 0xf5fb) in a Unicode Private Use Area that is a reasonable distance from the previous icon set in Nerd icons.

cc @rbong
Would you say the set of chars is final and ready to be impl in other terminals? Or are there more to be added soon™?
I see some discussion about these symbols [here](https://github.com/SuperBo/fugit2.nvim/discussions/88), are there other?

---
related: _I'm thinking about first implementing the custom chars 'selector' we talked about [a long time ago in 2022…](https://github.com/wez/wezterm/issues/1957#issuecomment-1120399027), and these new chars could then be easily disabled if needed_ 🤔

rbong:
>Would you say the set of chars is final and ready to be impl in other terminals? Or are there more to be added soon™?

You can draw any kind of directed acyclic graph in any orientation using these characters, including in orientations not typically used by Git branch viewers, and with connections working in a couple different ways... so I would say they're ready and future proof.

However one type of symbol I have not added is branches "jumping over" other branches. This could be up to 4 new symbols. These are not a necessity - you can always just draw "disconnected" branches, which is what every terminal branch viewer does right now and is much less visually cluttered. I have no explicit plans to add these symbols - however since some graphical branch viewers use these types of symbols, they're in the back of my mind and I wonder if they're needed.

There has also been discussion about different types of commit symbols. I am not convinced we need any more and have no plans to add more myself... The only proposed symbols right now are not typically required in a graph view, just a linear view, so I remain unconvinced that they're useful. I believe there will always be ideas for more commit symbols, and increasing the total number of symbols we have now by nearly 50% of the current total is not worth it. What we have is sufficient.

Feedback very much welcome.

wez:
Yeah, I'll take a PR for this! While you're in this mode, the spinner glyphs from https://github.com/ryanoasis/nerd-fonts/issues/1345 are appealing to me in order to support #6581 
[...]

PR #6873, opened 2025-04-12
Add new graph branch glyphs

BenBergman:
This is a WIP for #6328. Everything is working fairly well except I haven't figured out how to get the glyphs with dots and lines to connect to the dots correctly on all edges.

Screenshot 2025-04-12 at 1 22 20 PM

I also started adding descriptions for each symbol, but they were basically the same as the bitflags so I stopped part way through - I'm open to recommendations for what to do there.

@rbong I believe you were the one that created these unicode mappings to begin with. I was wondering if these two glyphs were left out on purpose:

Screenshot 2025-04-12 at 12 57 53 PM

The same glyphs with horizontal or vertical lines are in the list so it seems to me like these might be useful too.

image


BenBergman:
Is there a way to "erase" a poly shape? It seems like the easiest way to do the commit dots with lines would be draw the lines to the center of the circle and then erase a circle from the center. I've been poking at it for a bit now but can't seem to find a way to do it.

BenBergman:
Adding a blend mode I got some nice commit symbols.

image

I decided to just finish with the label comments. Other glyphs in the file have similarly redundant naming, so it seems like as good a plan as any. Feel free to adjust as needed.

rbong:
>I was wondering if these two glyphs were left out on purpose

There was. It is extremely hard to make out on small font sizes and has only a niche use case. Let me go through my thinking on it and you can tell me what you think.

I'll cover the glyph that is relevant to the perspective of the alignment of most git graphs. The other one is just a rotation of that and wouldn't make sense in the typical orientation.

2025-04-12-203232_25x47_scrot

The above glyph would move the branch above to the left while also moving the branch to the right below.

In a vanilla `git --graph` style branch viewer that uses lots of space for merges, you don't need this. You would do it this way:

2025-04-12-204759_73x107_scrot

This is also the way `tig` does it.

My branch viewer will only move branches to the left after a commit, so I don't need it:

2025-04-12-205131_72x89_scrot

It will also fill blank spaces with new branches. This is a kind of "lazy" style of "compressing" the graph (filling empty space and reducing its width), rather than `git log --graph`'s "eager" style, and it is more computationally efficient.

However, you could still use the glyph in a similar visual style of branch viewer if you want to move all of your branches one space to the left (mockup):

left

As you can see, in smaller font sizes it's extremely hard to make out.

You also *cannot color the branches* to make it more clear since they use the same tile. This *can only be used in uncolored graphs*, which is a large part of why it's not included.

If you want to use a different style of "eager" compression that uses colors and is more focused around straight lines, you unfortunately have to use as much space as `git log --graph` and `tig` does:

2025-04-12-211541_105x148_scrot

So why is the below glyph (and its rotations, if you want to draw the graph in other orientations) still useful?

2025-04-12-211843_27x39_scrot

This glyph means "the branch above is merged into from the left, while also forking out from the right". (Side note: you wouldn't want to do the opposite - fork to the left and merge from the right - because that would make an empty space on the left.)

If we *didn't* have this glyph, the commit above that's being merged into would have to continue to the right, which creates empty space:

2025-04-12-212426_77x188_scrot

Both of the above graphs mean the same thing, just where we place the continuation of the branch changes. In my opinion, although these are both a little weird, the top one makes more sense. You can't use it in every case, but it's pretty common.

This can also be used in colored graphs, since you can just make it the color of the vertical line.

This glyph can also be a little hard to make out, but four-way connections are always possible to decode contextually - if a branch does not continue below a line, it's a fork, and if it does, it's a merge.

By contrast, the other glyph would just be easy to confuse for a four-way connection, but it is not - it's two separate two-way connections. If we added it, it would hurt our ability to decode other four-way connections contextually if they're hard to make out, since we could confuse it for another glyph.

---

Let me know what you think. If you think it's worth adding despite the fact that it has limited use (most notably cannot be used in colored graphs) and the fact that it's hard to make out and would be easy to confuse for other symbols, I can try to ask around for more feedback.

In my opinion it is not worth it because just by adding it, we make the existing four-way connections worse by hurting the ability to decode them contextually.

BenBergman:
Thanks for the thorough response! Yeah, the lack of separate colours is certainly a killer, and the small font size doesn't do it any favours. Really my only argument back is that someone down the line might want to use these symbols for a non-vcs-graph use case (much like how we used box-drawing characters for graphs before these symbols) and that future use case might be frustrating without a complete set of combinations. I definitely see the argument that, for what they *are* designed for, they are more confusing than they are worth and leaving them out can be the safest option.

----

A question about your 4-way example though. I trust you are correct with your explanation since you've spent way more time in this stuff than me, but something isn't quite clicking for me.

image

The bottom image is very clear to me (at the expense of the space), but the top with the small-font-size issue isn't super intuitive to me. If we say that it is rendered small enough that we can't distinguish it from the other 4-way connectors, I'm not sure it is clear that the lower right branch is coming up and to the right as well as up and into the merge - it kind of looks like the branch from the lower left is coming up into the merge and also continuing up and to the right (ie. the non-existent  `╯` + `─` + `│` symbol) or the lower right branch is *only* going up and right while the lower left is going up to the commit as well as continuing up to the right (ie. image)

Or is the issue that I'm looking at it upside down? I'm used to the initial commit being at the bottom of the screen and newer commits above, but the circle symbol you used appears to be a commit not a merge maybe...?

Untitled

^ This image flowing from initial commit below to newer commits above does somehow seem less ambiguous to me, though arguably the small-text issue should still apply basically the same here. Perhaps colour is the missing link for clarity.

----

While you are here, are you able to give my changes for Wezterm's implementation a once-over?

rbong:
> The bottom image is very clear to me (at the expense of the space), but the top with the small-font-size issue isn't super intuitive to me. If we say that it is rendered small enough that we can't distinguish it from the other 4-way connectors, I'm not sure it is clear that the lower right branch is coming up and to the right as well as up and into the merge - it kind of looks like the branch from the lower left is coming up into the merge and also continuing up and to the right (ie. the non-existent `╯` + `─` + `│` symbol) or the lower right branch is _only_ going up and right while the lower left is going up to the commit as well as continuing up to the right (ie. image)

You're not looking at it upside-down, I just failed to use the merge commit glyph.

It's disambiguated by which branch the commit is on for the previous line.

If it were this character:

2025-04-12-234940_27x49_scrot

We would be saying "the branch above AND a branch to the right is being merged into from the left". Showing a merge INTO two branches at once is just not something any branch viewer does or should do. A line describing a merge/forks only comes after a commit, and the branch which the commit is on along with whether the other branches are new or continuing is enough to disambiguate any merge line, even without extended glyphs.

BenBergman:
I also added the spinner glyphs mentioned in #6328 since I used my added blend mode for them too. Let me know if you'd rather they be broken out to a new PR (though I at least have them in their own commit).

rbong:
>While you are here, are you able to give my changes for Wezterm's implementation a once-over?

Looks better than my implementation in Kitty

wezterm-gui/src/customglyph.rs
```diff
@@ -533,44 +572,63 @@ pub enum PolyCommand {
...
     Close,
 }
 
 impl PolyCommand {
     fn to_skia(&self, width: usize, height: usize, underline_height: f32, pb: &mut PathBuilder) {
         match self {
             Self::MoveTo(x, y) => pb.move_to(
-                x.to_pixel(width, underline_height),
-                y.to_pixel(height, underline_height),
+                x.to_pixel(width, underline_height, cmp::min(width, height)),
```
wez, line 586: FWIW, you should be able to do `a.min(b)` instead of `cmp::min(a, b)` for a more compact version of the same effect. I do prefer `a.min(b)` over `cmp::min`!

```suggestion
                x.to_pixel(width, underline_height, width.min(height)),
```

wezterm-gui/src/customglyph.rs
```diff
@@ -4804,6 +5081,7 @@ impl GlyphCache {
                     }],
                     &mut buffer,
                     PolyAA::AntiAlias,
+                    None,
```
wez, line 5084: this sort of `None` parameter makes it hard to intuit the purpose; how about using `BlendMode::default()` here instead of passing an `Option`?

wezterm-gui/src/customglyph.rs
```diff
@@ -4725,6 +4998,7 @@ impl GlyphCache {
         polys: &[Poly],
         buffer: &mut Image,
         aa: PolyAA,
+        blend_mode: Option<BlendMode>,
```
wez, line 5001:
```suggestion
        blend_mode: BlendMode,
```

as mentioned elsewhere, let's make this required; callers can pass `BlendMode::default()` if they don't care about the value?

wez (commented):
Thank you for this, and the spinners too!
I have a couple of minor stylistic comments, but this otherwise looks good to me!
I wonder if @bew might have some thoughts?

wezterm-gui/src/customglyph.rs
BenBergman: Yup, I fully agree. I only get to Rust occasionally so I forgot about that shorthand.

wezterm-gui/src/customglyph.rs
BenBergman: Sure, works for me.

BenBergman:
Okay, those code style changes are fixed, and the merge conflict with my other merged PR is done too. Anything else I can improve?

wez:
Looks great, thank you!

PR #6875, opened 2025-04-13
Add progress bar glyphs

BenBergman:
Since I was already adding glyphs for #6873, and in #6328 you asked for progress bar glyphs, I've added those too. I might take a crack at the spinner symbols later too, but I think that needs a new arc type for PolyCommand first.

image

Edit: I just realized you asked for *only* the spinners, not the progress bar. Woops! Well, do what you want with this PR then. 😅


wez:
This is great, thank you!

Issue #6571, opened 2025-01-10
Incomplete sentence in "Key Binding" docs


jwodder:
**What page or section of the docs have an issue?**

https://wezfurlong.org/wezterm/config/keys.html

**Describe the issue**

The last paragraph of the first section — just above the start of "Physical vs Mapped Key Assignments" — currently reads:

> Pay attention to the case of the text that you use and the state of the `SHIFT` modifier, as `key="A"` will match

This sentence isn't finished!  What will `key="A"` match?  *What will it match?!*

eugenesvk:
It will match Shift+a

Issue #4203, opened 2023-08-27
Unclear documentation regarding key bindings with SHIFT and single unicode character keys


adrianchong518:
**What page or section of the docs have an issue?**
Link: https://wezfurlong.org/wezterm/config/keys.html#configuring-key-assignments
> Pay attention to the case of the text that you use and the state of the `SHIFT` modifier, as `key="A"` will match 

**Describe the issue**
This line of the documentation seems to be cut off mid-sentence.

I assume this should say that:
> ..., as `key="A"` will match `SHIFT-a` despite not specifying `SHIFT` in `mods`

PR #6878, opened 2025-04-15
feat: support left/right mods for keybindings

### LINKED ISSUE #3710: Keybind: differentiate between left and right modifiers
<kbd>⌘</kbd> key is the most convenient modifier on a Mac since it's a thumb key, so it would be nice to use it as much as possible instead of using other less convenient modifiers
One way this could be achieved within WezTerm is if you could map left/right modifiers 

(as a side note: I see that <kbd>⎇</kbd> keys have separate debug modifier messages `LEFT_ALT` and `RIGHT_ALT` (although I can't seem to be able to use these in key bindings))

It would be great if `LEFT_SUPER` and `RIGHT_SUPER` modifier events (as well as all the others) could similarly be exposed and usable in user config in the `mods` field

I'm using the great Karabiner Elements for some of the functionality, but it suffers from various limitations:
  - it's not x-platform
  - it can't deal with various stack modes within WezTerm and 
  - it can't deal with nested apps like terminal text editors (though this is related to ↑ as I'm switching WezTerm key table on editor detection)

**Additional context**
None

ian-h-chamberlain:
Closes #3710 

This is my first time contributing so please let me know if there's anything I missed here or changes you'd like me to make! 

The basic implementation is this (new logic in **bold**):

During configuration:
1. For non-positional keybindings, add one mapping entry (same as current implementation)
2. **For "fully-positional" keybindings (all modifiers have left/right), add one mapping entry**
3. **For "mixed-positional" keybindings, register one binding for each possible left/right combination**

During lookup:
1. First try the common case of non-positional modifiers (same as current implementation)
2. **If that fails, try looking up the positional binding (matches steps 2 or 3 above)**
3. Finally perform fallback (same as current implementation)

I would like ask for help testing a few more scenarios since this is touching some fairly "core" keybinding code:

- [x] macOS
- [ ] Windows
- [ ] Linux

Are there any other scenarios I'm not thinking of that need testing? E.g. mapped vs physical vs raw keys, or X11 vs Wayland, etc.?


eugenesvk:
Hey, thanks for doing this! Here a couple of issues I've identified in my brief testing on Windows

- Alt is not differentiated, the key event only has ALt, not `ALT|LEFT_ALT` like for controls
- Shift doesn't match against the unmodified key, so `Shift+1` fails, but `Shift+!` works (but this is like a "double" shift in your config)

```lua
{mods = "LEFT_CTRL"           	,key="1"	,action="DecreaseFontSize"   	}, -- ✓
{mods = "RIGHT_CTRL"          	,key="1"	,action="IncreaseFontSize"   	}, -- ✓
{mods = "LEFT_CTRL|LEFT_ALT"  	,key="1"	,action="DecreaseFontSize"   	}, -- ✓
{mods = "RIGHT_CTRL|RIGHT_ALT"	,key="1"	,action="IncreaseFontSize"   	}, -- ✓
{mods = "LEFT_ALT"            	,key="1"	,action="DecreaseFontSize"   	}, -- ✗ WRONG key event, no alt sides in the main modifier
{mods = "RIGHT_ALT"           	,key="1"	,action="IncreaseFontSize"   	}, -- ✗
-- KeyEvent{key:Char('1'),modifiers:ALT,leds:NUM_LOCK,repeat_count:1,key_is_down:false,raw:Some(RawKeyEvent{key:Physical(K1),modifiers:ALT,leds:NUM_LOCK,phys_code:Some(K1),raw_code:49,scan_code:2,repeat_count:1,key_is_down:false,handled:Handled(false)}),win32_uni_char:Some('1')}
{mods = "LEFT_SHIFT"            	,key="1"	,action="DecreaseFontSize"	}, -- ✗ should work instead of 1
{mods = "RIGHT_SHIFT"           	,key="1"	,action="IncreaseFontSize"	}, -- ✗
{mods = "LEFT_SHIFT"            	,key="!"	,action="DecreaseFontSize"	}, -- ✓
{mods = "RIGHT_SHIFT"           	,key="!"	,action="IncreaseFontSize"	}, -- ✓
{mods = "LEFT_CTRL|LEFT_SHIFT"  	,key="1"	,action="DecreaseFontSize"	}, -- ✗
{mods = "RIGHT_CTRL|RIGHT_SHIFT"	,key="1"	,action="IncreaseFontSize"	}, -- ✗
-- KeyEvent{key:Char('1'),modifiers:CTRL|LEFT_CTRL ,leds:NUM_LOCK,repeat_count:1,key_is_down:false,raw:Some(RawKeyEvent{key:Physical(K1),modifiers:CTRL|LEFT_CTRL ,leds:NUM_LOCK,phys_code:Some(K1),raw_code:49,scan_code:2,repeat_count:1,key_is_down:false,handled:Handled(false)}),win32_uni_char:Some('\0')}
-- KeyEvent{key:Char('1'),modifiers:CTRL|RIGHT_CTRL,leds:NUM_LOCK,repeat_count:1,key_is_down:false,raw:Some(RawKeyEvent{key:Physical(K1),modifiers:CTRL|RIGHT_CTRL,leds:NUM_LOCK,phys_code:Some(K1),raw_code:49,scan_code:2,repeat_count:1,key_is_down:false,handled:Handled(false)}),win32_uni_char:Some('\0')}
```

eugenesvk:
> Windows in particular has some special handling for AltGr (aka right-alt).

AltGr is simply Alt+Ctrl, you can have a separate Right Alt that is a single key. Though not sure how this messes up with the regular non AltGr Alts I was testing with. Or do you mean there is no differentiation between AltGr and Right Alt in wezterm's code?

https://en.wikipedia.org/wiki/AltGr_key

> Windows interprets Ctrl+Alt as AltGr, to accommodate some compact keyboards like those of [netbooks](https://en.wikipedia.org/wiki/Netbook) which have neither the AltGr key nor a right-hand Alt key. Thus Ctrl+Alt+a has the same effect as AltGr+a. Because of this feature

> Hmm, I was under the impression this was already how bindings [normally work]https://github.com/wezterm/wezterm/blob/3bfe6df191af37bc33fe5fe6de12a40f7067f101/termwiz/src/input.rs#L216-L221) (except for ASCII letters (like k), which get normalized to uppercase (like K).

Well, maybe it's somehow fixed despite the comment saying it's not yet

>  In theory we should give the same treatment to keys like `[` -> `{`

Personally, I don't see the point of ASCII conversion either in shortcuts, there should never be confusing double Shift, `q` to me is simply a layout-dependent key label (so different physical in qwerty and dvorak) and then `phys:q` is layout independent key label (so same key in both layouts), this basic labeling is its only purpose, not to signal what your layout converts output to depending on mod.

So in both cases Shift+q should be bound as Shift+q and not Shift+Q. Likewise, "in theory" `q` should get the same treatment as `[` and stay `q` and `[` with Shift

But if it's too much change then maybe it's just a matter of docs/examples, I'll just stick with `phys:x` everywhere anyway since these don't break on layout changes

bew:
> > Windows in particular has some special handling for AltGr (aka right-alt).
> 
> AltGr is simply Alt+Ctrl, you can have a separate Right Alt that is a single key. Though not sure how this messes up with the regular non AltGr Alts I was testing with. Or do you mean there is no differentiation between AltGr and Right Alt in wezterm's code?

Wezterm has this option (disabled by default):
https://wezterm.org/config/lua/config/treat_left_ctrlalt_as_altgr.html

> https://en.wikipedia.org/wiki/AltGr_key
> 
> > Windows interprets Ctrl+Alt as AltGr, to accommodate some compact keyboards like those of [netbooks](https://en.wikipedia.org/wiki/Netbook) which have neither the AltGr key nor a right-hand Alt key. Thus Ctrl+Alt+a has the same effect as AltGr+a.

This isn't true on computers with left/right alt keys.
With full keyboards the Right Alt is AltGr, and it is different from Ctrl+Alt.

eugenesvk:
> This isn't true on computers with left/right alt keys.
> With full keyboards the Right Alt is AltGr, and it is different from Ctrl+Alt.

That's my keyboard, and with an AltGr-enabled layout like US International Autohotkey shows it's exactly the same:
  - pressing right AltGr sends LeftCtrl+RightAlt
  - pressing LCtrl+LeftAlt and key inserts the same symbol as AltGr and key

bew:
> That's my keyboard, and with an AltGr-enabled layout like US International Autohotkey shows it's exactly the same

Surprising then, I'm using a french layout where the AltGr key is used to access many symbols, and I could have sworn they were different (since I remap a lot of my keyboard, I'm pretty sure I'd have noticed. But maybe not `¯\_(ツ)_/¯`).
I don't use Windows anymore so won't be able to test again there.

ian-h-chamberlain:
> I don't use Windows anymore so won't be able to test again there.

I figured out how to switch to international US on my Windows machine, which as far as I can tell has a "real" AltGr key (even though my physical keyboard is still standard US), which behaved differently than LeftCtrl+LeftAlt. I've been too busy to work on this recently but will try to get back to it in a few weeks, hopefully.

ian-h-chamberlain:
Okay, it took me longer than I thought but I finally got a chance to work on this again for Windows! 

I added a line to documentation about the `AltGr` caveat mentioned (namely, `RIGHT_ALT|CTRL` is not usable for mapping on keyboard layouts that use `AltGr`).

If anyone else can test on various keyboard layouts / platforms that would be great, I am hopeful that the implementation is a bit more complete now but there are tons of combinations and platform-specific behavior so a lot of testing would be ideal.

----

Re: the `SHIFT` normalization, I am not completely sure what the original intent of the behavior is, and the documentation is a bit unclear as well: #6571 #4203

My intent with this PR is to keep the behavior the same as much as possible (particularly, to avoid breaking existing keybinds), but some direction from maintainers and ideally a corresponding docs update would be really helpful.

eugenesvk:
The docs mean what the issue you linked clarifies: `A` matches `⇧a`. My guess the goal is to follow this design in some other keybinding configs where instead of the explicit `⇧a` apps use `A`, and for a few keys that even makes sense (where you want `<` in your config instead of `⇧,` to signal that this key will move left.

But the native wezterm also accepts `⇧a`, so you can use both (though it doesn't seem to accept `<`?)

Re. the AltGr note I'd maybe change 
 > Note that in some keyboard layouts,  `CTRL|RIGHT_ALT` will be mapped to `AltGr` and therefore not be usable for keyboard shortcuts.

to
> Note that in some keyboard layouts the physical Right Alt key is treated as a special AltGr key and generates a combination of  `LEFT_CTRL|RIGHT_ALT` key events and therefore will not match standalone `RIGHT_ALT` shortcuts.

PR #6895, opened 2025-04-21
feat(mark-last-tab): append last tab name with '*'

masriomarm:
I think there should be a better method instead of make the save function so crowded. But this works for now similar to `tmux` windows. But, `tmux` is using '\*' for current and '-' for last Hardcoded '\*'

### DIFF
--- a/docs/config/lua/TabInformation.md
+++ b/docs/config/lua/TabInformation.md
@@ -10,6 +10,7 @@ The `TabInformation` struct contains the following fields:
 * `tab_id` - the identifier for the tab
 * `tab_index` - the logical tab position within its containing window, with 0 indicating the leftmost tab
 * `is_active` - is true if this tab is the active tab
+* `is_last` - is true if this tab is the previous active tab - last tab.
 * `active_pane` - the [PaneInformation](PaneInformation.md) for the active pane in this tab
 * `window_id` - the ID of the window that contains this tab {{since('20220807-113146-c2fee766', inline=True)}}
 * `window_title` - the title of the window that contains this tab {{since('20220807-113146-c2fee766', inline=True)}}
--- a/docs/config/lua/TabInformation.md
+++ b/docs/config/lua/TabInformation.md
@@ -10,7 +10,7 @@ The `TabInformation` struct contains the following fields:
 * `tab_id` - the identifier for the tab
 * `tab_index` - the logical tab position within its containing window, with 0 indicating the leftmost tab
 * `is_active` - is true if this tab is the active tab
-* `is_last` - is true if this tab is the previous active tab - last tab.
+* `is_last_active` - is true if this tab is the previously active tab. {{since('nightly', inline=True)}}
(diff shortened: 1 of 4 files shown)


masriomarm:
Thanks, I will be working on it.

masriomarm:
I can squash and rebase on mater once all clear.

wezterm-gui/src/termwindow/mod.rs
```diff
@@ -212,6 +212,7 @@ pub struct TabInformation {
     pub tab_id: TabId,
     pub tab_index: usize,
     pub is_active: bool,
+    pub is_last: bool,
```
wez, line 215: let's rename this to something a little more descriptive, because the default connotation of `last` in my native English is "the last one to be enumerated" which implies the right-most tab, but here it is really "the previously active tab" and the ambiguity is potentially confusing!

I realize that https://wezterm.org/config/lua/keyassignment/ActivateLastTab.html sets a precedent for using `last`.

How about `is_last_active` for this and the field method below?

wezterm-gui/src/termwindow/mod.rs
```diff
@@ -3460,6 +3462,11 @@ impl TermWindow {
                     tab_index: idx,
                     tab_id: tab.tab_id(),
                     is_active: tab_index == idx,
+                    is_last: if let Some(tab_index_last) = window.get_last_active_idx() {
+                        tab_index_last == idx
+                    } else {
+                        false
+                    },
```
wez, line 3469: I think we can make this a bit more succinct:

```suggestion
                    is_last: window.get_last_active_idx().map(|last| last == idx).unwrap_or(false),
```

docs/config/lua/TabInformation.md
```diff
@@ -10,6 +10,7 @@ The `TabInformation` struct contains the following fields:
 * `tab_id` - the identifier for the tab
 * `tab_index` - the logical tab position within its containing window, with 0 indicating the leftmost tab
 * `is_active` - is true if this tab is the active tab
+* `is_last` - is true if this tab is the previous active tab - last tab.
```
wez, line 13:
```suggestion
* `is_last_active` - is true if this tab is the previously active tab. {{since('nightly', inline=True)}}
```

docs/config/lua/window-events/format-tab-title.md
```diff
@@ -46,6 +46,13 @@ wezterm.on(
         { Text = ' ' .. title .. ' ' },
       }
     end
+    if tab.is_last then
+      -- Green color and append '*' to last tab.
```
wez, line 50:
```suggestion
    if tab.is_last_active then
      -- Green color and append '*' to previously active tab.
```

wez (commented):
Thanks! I'd like to rename the property, but otherwise I think the approach looks good.

regarding squashing, don't worry about it: github gives me a squash button that makes it convenient for me to squash and tweak the comment description when I'm ready to merge.

wez:
Thank you!

PR #6905, opened 2025-04-24
feat: Added support for window blur on KDE Wayland

psomani16k:
KDE Wayland background blur added, triggered by adding somthing similar to the configuratin:

```
config.window_background_opacity = 0.2
config.kde_window_background_blur = true
```

Default behavior is:
```
config.kde_window_background_blur = false
```

### Blur Example:
with_blur

### DIFF
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -21,11 +21,11 @@ use crate::units::Dimension;
 use crate::unix::UnixDomain;
 use crate::wsl::WslDomain;
 use crate::{
-    default_config_with_overrides_applied, default_one_point_oh, default_one_point_oh_f64,
-    default_true, default_win32_acrylic_accent_color, GpuInfo, IntegratedTitleButtonColor,
-    KeyMapPreference, LoadedConfig, MouseEventTriggerMods, RgbaColor, SerialDomain, SystemBackdrop,
-    WebGpuPowerPreference, CONFIG_DIRS, CONFIG_FILE_OVERRIDE, CONFIG_OVERRIDES, CONFIG_SKIP,
-    HOME_DIR,
+    default_config_with_overrides_applied, default_false, default_one_point_oh,
+    default_one_point_oh_f64, default_true, default_win32_acrylic_accent_color, GpuInfo,
+    IntegratedTitleButtonColor, KeyMapPreference, LoadedConfig, MouseEventTriggerMods, RgbaColor,
+    SerialDomain, SystemBackdrop, WebGpuPowerPreference, CONFIG_DIRS, CONFIG_FILE_OVERRIDE,
+    CONFIG_OVERRIDES, CONFIG_SKIP, HOME_DIR,
 };
 use anyhow::Context;
 use luahelper::impl_lua_conversion_dynamic;
@@ -563,6 +563,10 @@ pub struct Config {
     /// Only works on MacOS
     #[dynamic(default)]
     pub macos_window_background_blur: i64,
+
+    /// Only works on MacOS
+    #[dynamic(default = "default_false")]
+    pub kde_window_background_blur: bool,
 
     /// Only works on Windows
     #[dynamic(default)]
--- a/config/src/lib.rs
+++ b/config/src/lib.rs
@@ -821,3 +821,7 @@ fn default_one_point_oh() -> f32 {
 fn default_true() -> bool {
     true
 }
+
+fn default_false() -> bool {
+    false
+}
--- a/window/src/os/wayland/state.rs
+++ b/window/src/os/wayland/state.rs
@@ -3,6 +3,7 @@ use std::collections::HashMap;
 use std::rc::Rc;
 use std::sync::{Arc, Mutex};
 
+use libc::bind;
 use smithay_client_toolkit::compositor::{CompositorState, SurfaceData};
 use smithay_client_toolkit::data_device_manager::data_device::DataDevice;
 use smithay_client_toolkit::data_device_manager::data_source::CopyPasteSource;
@@ -26,12 +27,13 @@ use smithay_client_toolkit::{
     delegate_compositor, delegate_data_device, delegate_output, delegate_pointer, delegate_primary_selection, delegate_registry, delegate_seat, delegate_shm, delegate_subcompositor, delegate_xdg_shell, delegate_xdg_window, registry_handlers
 };
 use wayland_client::backend::ObjectId;
-use wayland_client::globals::GlobalList;
+use wayland_client::globals::{BindError, GlobalList};
 use wayland_client::protocol::wl_keyboard::WlKeyboard;
 use wayland_client::protocol::wl_output::WlOutput;
 use wayland_client::{delegate_dispatch, Connection, QueueHandle};
 use wayland_protocols::wp::text_input::zv3::client::zwp_text_input_manager_v3::ZwpTextInputManagerV3;
 use wayland_protocols::wp::text_input::zv3::client::zwp_text_input_v3::ZwpTextInputV3;
+use wayland_protocols_plasma::blur::client::org_kde_kwin_blur_manager::OrgKdeKwinBlurManager;
 
 use crate::x11::KeyboardWithFallback;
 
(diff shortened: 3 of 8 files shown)


john-okeefe:
Thank you so much for this. I can't wait for it to be merged and released!

config/src/config.rs
```diff
@@ -564,6 +564,10 @@ pub struct Config {
     #[dynamic(default)]
     pub macos_window_background_blur: i64,
 
+    /// Only works on MacOS
```
wez, line 567:
```suggestion
```

config/src/lib.rs
```diff
@@ -821,3 +821,7 @@ fn default_one_point_oh() -> f32 {
 fn default_true() -> bool {
     true
 }
+
+fn default_false() -> bool {
+    false
+}
```
wez, line 827: This is actually the default for bool, so it is not necessary to add this

```suggestion
```

config/src/config.rs
```diff
@@ -564,6 +564,10 @@ pub struct Config {
...
+    #[dynamic(default = "default_false")]
```
wez, line 568:
```suggestion
    #[dynamic(default)]
```

docs/config/lua/config/kde_window_background_blur.md
```diff
@@ -0,0 +1,26 @@
+---
+tags:
+  - appearance
+---
+# `kde_window_background_blur = false`
+
+When combined with `window_background_opacity`, enables background blur
+using the default KDE Wayland blur protocol.
```
wez, line 8:
```suggestion
using the KDE Wayland blur protocol.
```

docs/screenshots/wezterm-kde-background-blur.png
wez, line 1: This doesn't appear to be using blur?

window/src/os/wayland/state.rs
```diff
@@ -82,6 +84,14 @@ impl WaylandState {
...
             SubcompositorState::bind(compositor.wl_compositor().clone(), globals, qh)?;
 
+        let blur_manager: Result<OrgKdeKwinBlurManager, BindError> =
+            globals.bind(qh, 1..=1, GlobalData);
+
+        let blur_manager = match blur_manager {
+            Ok(manager) => Some(manager),
+            Err(_) => None,
+        };
+
```
wez, line 94: Can simplify this, I think:

```suggestion
        let blur_manager: Result<OrgKdeKwinBlurManager, BindError> =
            globals.bind(qh, 1..=1, GlobalData).ok();
```

window/src/os/wayland/window.rs
```diff
@@ -1229,6 +1243,31 @@ impl WaylandWindowInner {
...
+                    kde_blur.set_region(None);
+                } else {
+                    kde_blur.release();
+                }
+                kde_blur.commit();
+            }
+            None => {
+                // Can't blur if there is no blur manager
+            }
+        }
```
wez, line 1269:
```suggestion
        if let Some(manager) = &wayland_state.kde_blur_manager {
                let kde_blur = manager.create(self.surface(), &qh, GlobalData);
                if self.config.kde_window_background_blur {
                    kde_blur.set_region(None);
                } else {
                    kde_blur.release();
                }
                kde_blur.commit();
        }
```

window/src/os/wayland/window.rs
```diff
@@ -1383,6 +1422,47 @@ impl WindowHandler for WaylandState {
...
+    fn event(
+        _state: &mut Self,
+        _proxy: &OrgKdeKwinBlurManager,
+        _event: <OrgKdeKwinBlurManager as Proxy>::Event,
+        _data: &GlobalData,
+        _conn: &WConnection,
+        _qhandle: &wayland_client::QueueHandle<Self>,
+    ) {
+        // No events from OrgKdeKwinBlurManager...
+        unreachable!();
```
wez, line 1435: I'd rather not have wezterm panic if events are introduced in the future!

```suggestion
```

window/src/os/wayland/window.rs
```diff
@@ -1383,6 +1422,47 @@ impl WindowHandler for WaylandState {
...
+    fn event(
+        _state: &mut Self,
+        _proxy: &OrgKdeKwinBlur,
+        _event: <OrgKdeKwinBlur as Proxy>::Event,
+        _data: &GlobalData,
+        _conn: &WConnection,
+        _qhandle: &wayland_client::QueueHandle<Self>,
+    ) {
+        // No events from OrgKdeKwinBlur...
+        unreachable!();
```
wez, line 1449:
```suggestion
```

window/src/os/wayland/window.rs
```diff
@@ -1383,6 +1422,47 @@ impl WindowHandler for WaylandState {
...
+impl Dispatch<WlRegion, GlobalData> for WaylandState {
+    fn event(
+        _state: &mut Self,
+        _proxy: &WlRegion,
+        _event: <WlRegion as Proxy>::Event,
+        _data: &GlobalData,
+        _conn: &WConnection,
+        _qhandle: &QueueHandle<Self>,
+    ) {
+        unreachable!();
```
wez, line 1462:
```suggestion
```

window/src/os/windows/window.rs
```diff
@@ -678,7 +678,7 @@ impl WindowInner {
         }
     }
 
-    fn config_did_change(&mut self, config: &ConfigHandle) {
+    fn config_did_change(&mut self, config: ConfigHandle) {
```
wez, line 681: why are these changes required on Windows?

wez (commented):
Thank you! I have a few comments/questions!

config/src/config.rs
psomani16k: That's my bad, but maybe adding "Only works on KDE wayland" would be a better option?

config/src/lib.rs
psomani16k: Got it!

window/src/os/windows/window.rs
psomani16k: This removes an extra (and seemingly unnecessary) clone of the ConfigHandle object, however this wasn't supposed to be added in this PR, my bad again.

psomani16k: reverted this change in the latest commit

window/src/os/windows/window.rs
```diff
@@ -887,7 +887,7 @@ impl WindowOps for Window {
     fn config_did_change(&self, config: &ConfigHandle) {
         let config = config.clone();
         Connection::with_window_inner(self.0, move |inner| {
-            inner.config_did_change(&config);
+            inner.config_did_change(config);
```
wez, line 890: Seems like this shouldn't be here either?

```suggestion
            inner.config_did_change(&config);
```

wez:
Thank you!

superruzafa:
Hi, I ended here because I was trying to apply the `kde_window_background_blur` option in my config, as it was documented [here](https://wezterm.org/config/lua/config/kde_window_background_blur.html).

Shouldn't this option be hidden from the docs until it's included in a release?

wez:
@superruzafa the docs for that config item were missing the `since` marker; I've added and pushed that now

superruzafa:
@psomani16k, can you share what configuration you used (either wezterm or KDE) to achieve such effect? 🙏🏻

The background gets transparent only moving the window, but no blur in any case.

```lua
config.enable_wayland = true
config.window_background_opacity = 0.5
config.text_background_opacity = 0.5
config.kde_window_background_blur = true
```

```sh
$ wezterm --version
wezterm 20250604-060157-5106c8c1
```

psomani16k:
Hi @superruzafa , 
Your wezterm config seems to be alright, maybe it is something on the KDE part, can you tell me what version of kwin you are using? (i think the blur is supported kwin 6 onwards)

Also make sure blur is enabled in kwin:
System Settings -> Window Management -> Desktop Effects -> Blur

superruzafa:
@psomani16k, yes, it working by activating the Blur settings 💯 
The blur force is completely controlled by the global settings.

Also, it needed a lower background opacity value (0.9 -> 0.3) to notice the effect, specially with dark backgrounds.

The only thing I observed is that the background becomes fully transparent when moving the window. I completely disabled the KDE's windows transparency setting.

Good work, BTW!

[Videocaptura de pantalla_20250605_120211.webm](https://github.com/user-attachments/assets/fcbbc43e-ece1-4094-88e3-083022d5038b)

Issue #4899, opened 2024-01-29
wezterm cli set-window-title does not change the displayed window title


andresrinivasan:
### What Operating System(s) are you seeing this problem on?

macOS

### WezTerm version

20240128-202157-1e552d76

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

With no config, I can change the tab title with `wezterm cli set-tab-title`. When I try `wezterm cli set-window-title` the window title does not change. I then tried with a minimal config that with `config.hide_tab_bar_if_only_one_tab = true` thinking the tab title was in control. Even with no tabs I could not change the window title via the CLI. Using the escape sequence worked of course.

### To Reproduce

1. Start a wezterm with a minimal or no config
2. Invoke `wezterm cli set-window-title foo`

### Configuration

No config

### Expected Behavior

I expected the window title to change

### Logs

No errors shown

wez:
potentially resolved by b0682e957f5786ef1cf2705d3e63c3f6f4945574.
Please try the latest nightly and let me know!

andresrinivasan:
With `wezterm 20240203-110809-5046fc22`, I still could not set the window title, even when wezterm was started with `-n`.

osievert-gpfw:
I came across wezterm today and it ticks all my boxes EXCEPT for this issue. My workflow is that I have many terminal windows open at once, each dedicated to a specific topic. Within each of these windows I almost always have multiple tabs, and each of these tabs might be split one or more times. This creates (for me, anyway) a sane way to organize my work. I can easily use the macOS window selection facilities to bring a "topic window" to be my top window, and then I can productively work on that topic until I context switch to another in-flight topic.

So for me, I want the ability to set a window title that binds to that window regardless of what tab is selected in that window, and regardless of what program I run in that tab. I want the ability to set the title of each tab in the same way.

This bug was in the way of that, so I submitted [a PR](https://github.com/wezterm/wezterm/pull/6913) that fixes it. I have only used wezterm for a couple hours now, and have looked at this code base for even less. I would not be surprised if there are some other use cases that I did not consider in this fix. Please advise.

What this fix does is to change the logic used when actually setting the window title, to consider these three values (in priority order):
1. If the user has configured a lua handler for the `format-window-title` event, that is used.
2. If the user has explicitly set a window title for example by calling `wezterm cli set-window-title foo`, that is used.
[...]

PR #6913, opened 2025-04-27
fix: wezterm cli set-window-title #4899

osievert:
Fixes issue where the window title is not set when calling, e.g.

  `wezterm cli set-window-title foo`

This fix alters the window title logic, setting the window title in priority order:
- a title produced from the lua `format-window-title` event handler
- a title attached to the window object (this is what is set when the set-window-title call above is made)
- a default window title

As such, if you do not install your own lua format-window-title handler, then the cli command now works as expected.


johanbelin:
Has this been merged? Iit is still not working in the nightly build

PR #6919, opened 2025-04-28
Advertize the Su capability too

krobelus:
I'm not sure if applications are supposed to use "Smulx" and "Setulc" or
"Su". Among the terminals that implement XTGETTCAP, contour, foot, ghostty
and kitty advertize both (and XTerm neither).

"Su" is the one mentioned in the original documentation
https://sw.kovidgoyal.net/kitty/underlines/

It's a boolean capability but I'd argue that's even better since I don't
think the flexibility of string capabilities is needed.  (Personally, all
I want to use it for is to answer "will this terminal break if I emit the
hardcoded \e[4:3m and \e[58:2::1:2:3m".

Include "Su" in wezterm's terminfo, for the benefit of XTGETTCAP users..
Updated the compiled database with

	tic -x termwiz/data/wezterm.terminfo -o termwiz/data

Alternatively we could change the kitty documentation.


wez:
Thanks!

Issue #6673, opened 2025-02-13
window_decorations = "NONE" not working


dougg0k:
### What Operating System(s) are you seeing this problem on?

Linux Wayland

### Which Wayland compositor or X11 Window manager(s) are you using?

It's KDE, so Kwin. Latest.

### WezTerm version

 20250212-113941-11505b70

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

Before I was using the latest stable released to arch and the above bar were hidden. But after installing the latest through `wezterm-git`, it's showing even when `config.window_decorations` set to `"NONE"`.

```
❯ paru -Ss wezterm
extra/wezterm 20240203.110809.5046fc22-2 [23.87 MiB 105.39 MiB]
    A GPU-accelerated cross-platform terminal emulator and multiplexer
aur/wezterm-git 20240922.091625.5a594f51-3 [+11 ~0.20] [Out-of-date: 2024-09-25] [Installed: 20250212.113941.11505b70-1]
    A terminal emulator implemented in Rust, using OpenGL ES 2 for rendering.
```

`wezterm-git` shows a different version from the ones that is actually installed, which is the latest.

![Image](https://github.com/user-attachments/assets/cceab890-0684-4198-8f66-a599e5936751)

### To Reproduce

Install `wezterm-git` if you are on arch linux.

### Configuration

Only one that matters to the issue, as before, it was working. 

```
config.window_decorations = "NONE"
```

### Expected Behavior

To hide the title bar.

dougg0k:
After confirming that is on wayland only, I updated the information above.

Also confirmed that it happens when setting `config.enable_wayland` to `true`. Even if in the stable version, being set to `true`, worked fine.

However, if I set to `false`, even on wayland, which maybe turns to Xwayland, the issue are fixed.

wez:
Will likely be resolved by this ongoing PR:
 
* https://github.com/wezterm/wezterm/pull/4970

dougg0k:
@wez When there is going to be a next stable release with all the commits?

bew:
@dougg0k You can subscribe to #6341 to follow progress on the lockers for the next release 😉

tjex:
Same here on aarch linux (fedora 41). Can confirm setting `enable_wezterm = false` removes the titlebar (as desired through `config.window_decorations = "NONE"`.

[...]

Issue #6578, opened 2025-01-15
#4963 happening in NixOS in the latest nightly (nixos-24.11)


Martin1887:
### What Operating System(s) are you seeing this problem on?

Linux Wayland

### Which Wayland compositor or X11 Window manager(s) are you using?

Mutter

### WezTerm version

0-unstable-2025-01-03

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

#4963 is happening with `window_decorations = "RESIZE"`.

### To Reproduce

Open WezTerm, a bar with weird buttons appear at top. This also causes that the bottom of the terminal content is not shown.

### Configuration

```
window_decorations = "RESIZE"
```

### Expected Behavior

No bar is shown but the window can be resized and moved.

Martin1887:
It keeps happening in unstable-2025-02-23.

Issue #6318, opened 2024-10-24
The window overflows bottom of screen


metal3d:
### What Operating System(s) are you seeing this problem on?

Linux Wayland

### Which Wayland compositor or X11 Window manager(s) are you using?

Gnome Shell (Wayland) - Fedora 40

### WezTerm version

20241020_002442_2381b8ac

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

When I maximize the window, it overflows the bottom of the screen (like it is the size of the title bar). So, the last line of the terminal is not visible.

![image](https://github.com/user-attachments/assets/4727992f-cc8e-43e7-a264-c3acb2914fdf)

If I don't set a font size, or not using the shortcut to reduce the font size, it is "OK".

### To Reproduce

Set a "font size" in the configuration 

### Configuration

```lua
local wezterm = require("wezterm")
local config = {}
config.font_size = 10
config.hide_tab_bar_if_only_one_tab = true

-- necessary and I don't know why
config.default_prog = { "bash", "-i" }

config.window_decorations = "RESIZE"
config.enable_wayland = true

return config
```

### Expected Behavior

Actually, IMHO, the font size should not change the window scaling, only the "font". And so, the last line will not overlap the screen.

### Anything else?

I love this terminal 😄

metal3d:
Ho and it seems it overflows on the right too

jankatins:
I can confirm this with `config.window_decorations = "INTEGRATED_BUTTONS"` as well
![image](https://github.com/user-attachments/assets/c9f23114-ce94-46af-b7b5-5153fc96e352)
-> there should ba slight lighter grey under the last line similar to what is on the left side. E.g. this is in windowed mode:
![image](https://github.com/user-attachments/assets/e889d418-cca7-4c5e-94ac-fc60c7d98aba)

mistyfiky:
Fedora 43, Wayland, Gnome 49, `enable_wayland=true`, Wezterm (from Copr) 20260117_154428_05343b38 - same issue here, but no font_size change needed (maybe this has something to do with Cascadia font used). Maximized window overflows screen on right and bottom (both on screen with 1.25 scaling and without any).

A little workaround (in my case for nvim in maximized Wezterm window):

```lua
local padding_unit = "pt"
local padding_size = 6
local wl_titlebar_height = 21
local wl_windowbar_width = 3
[...]

Issue #6296, opened 2024-10-19
Title Bar is missing if config.enable_wayland = true


sandikata:
### What Operating System(s) are you seeing this problem on?

Linux Wayland

### Which Wayland compositor or X11 Window manager(s) are you using?

Mutter

### WezTerm version

wezterm 20241015-083151-9ddca7bd

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

If config.enable_wayland = true titlebar is missing at all, and terminal window can't be moved or resized at all.

### To Reproduce

Install git version, enable wayland (it is enabled by default)

Start wezterm

### Configuration

-- Pull in the wezterm API
local wezterm = require 'wezterm'

-- This table will hold the configuration.
local config = {}

-- In newer versions of wezterm, use the config_builder which will
-- help provide clearer error messages
if wezterm.config_builder then
  config = wezterm.config_builder()
end

-- terminfo
-- config.term = "wezterm"

-- Renderer
config.enable_wayland = false
config.front_end = "WebGpu" -- default OpenGL
-- config.max_fps = 165
config.window_background_opacity = 0.98

-- This is where you actually apply your config choices
-- For example, changing the color scheme:
--config.color_scheme = 'Spacemacs (base16)'
-- config.color_scheme = 'Andromeda'
config.color_scheme = 'Relaxed'

-- fonts configuration
config.font = wezterm.font_with_fallback {
--  { family = 'JetBrainsMono', weight = 'Light', scale = 1.15 },                        -- line_height = 1.1
  { family = 'Dank Mono', weight = 'Regular', scale = 1.3 }, -- line_height = 1.1
--  { family = 'Fira Code', weight = 'Regular', scale = 1.1 }, -- line_height = 1.1
--  { family = 'Hasklig', weight = 'Regular', scale = 1.2 }, -- line_height = 1.1
--  { family = 'Iosevka Term', weight = 'Regular', scale = 1.20, stretch = 'Expanded', },            -- line_height = 1.05
--  { family = 'Victor Mono', weight = 'Regular', scale = 1.20 }, -- line_height = 1.1
  { family = 'Cascadia Code', weight = 'Light', scale = 1.20 }, -- line_height = 1.1
--  { family = 'Agave', weight = 'Regular', scale = 1.20 },
[...]

PR #6923, opened 2025-04-30
Adopt sctk-adwaita as our Wayland window frame

### LINKED ISSUE #3858: window_decorations='RESIZE' results in no border for floating windows on Sway

Linux Wayland

Sway

### WezTerm version

wezterm unstable-2023-05-30

### Describe the bug

With window_decorations set to RESIZE, I expected floating windows to have borders drawn, but that doesn't appear to be the case.

Three screenshots:

1. wezterm fullscreen, wezterm floating on top (focused or unfocused, it's the same)
  * screenshot-1686683704
1. alacritty, with floating window unfocused
  * screenshot-1686683719
1. alacritty, with floating window focused
  * screenshot-1686683720

As you can see, with Alacritty, I can see where my floating windows are, whereas in Wezterm, it's very difficult to tell where the edges of a floating window are.

### To Reproduce

_No response_

### Configuration

```

wezterm.add_to_config_reload_watch_list("/home/cole/.config/wezterm")

local config = {
  default_prog = { "zsh" },
  enable_tab_bar = false,
  use_fancy_tab_bar = false,
  front_end = "WebGpu",
  initial_rows = 24,
  initial_cols = 120,
  font_size = 13,
  window_background_opacity = 1.0,
  enable_csi_u_key_encoding = true,
  default_cursor_style = 'BlinkingBar',
  colors = {
    foreground = "#ebdbb2",
    background = "#000000",
    ansi = {
      "#282828",
      "#cc241d",
      "#98971a",
      "#d79921",
      "#458588",
      "#b16286",
      "#689d6a",
      "#a89984"
    },
    brights = {
      "#928374",
      "#fb4934",
      "#b8bb26",
      "#fabd2f",
      "#83a598",
      "#d3869b",
      "#8ec07c",
      "#ebdbb2"
    },
  }
}

if wezterm.target_triple == "x86_64-pc-windows-msvc" then
  config.default_prog = { "powershell.exe" }
else
  config.enable_wayland = true
  config.window_decorations = "RESIZE"
  config.window_close_confirmation = "NeverPrompt"
  -- config.freetype_load_target = "Light"
  -- config.freetype_render_target = "HorizontalLcd"
  local f = wezterm.font_with_fallback({
    {family="Iosevka Comfy Fixed", weight="Regular"},
    {family="Font Awesome", weight="Regular"},
  })
  config.font = f;
end

-- issue#3142 workaround START
local wezterm_action = wezterm.action
local act = wezterm.action
config.mouse_bindings = {
  {
    event = { Down = { streak = 1, button = { WheelUp = 1 } } },
    mods = 'NONE',
    action = wezterm_action.ScrollByLine(-1),
  },
  {
    event = { Down = { streak = 1, button = { WheelDown = 1 } } },
    mods = 'NONE',
    action = wezterm_action.ScrollByLine(1),
  },
}
-- issue#3142 workaround END

return config

```

### Expected Behavior

_No response_

### Logs

_No response_

### Anything else?

_No response_

valpackett:
Should fix #3858, #4948, #4962, #5332, #6296, #6318, #6578, #6673  
Supersedes #4970, #5390

As mentioned in https://github.com/wezterm/wezterm/pull/4970#discussion_r1954569013 sctk-adwaita removes the burden of maintaining a custom decoration frame :) and it allows us to finally fix various frame related bugs once and for all :rocket: 

- [ ] wait for https://github.com/PolyMeilex/sctk-adwaita/pull/69 (nice) to land and release (or ship a git dep for now?)
- [ ] fix `window_decorations = "NONE"` (`set_hidden(true)` errors with shm thingy for me on niri) – actually skip the frame completely in that case i guess?
- [ ] test on more compositors

https://github.com/user-attachments/assets/3cebf78a-9876-43a5-a2a8-3552ff95388f


valpackett:
@BoltsJ It supports custom colors just fine and I'll definitely look into providing them.

Either way, this is kind of a "we have to have something" situation and current code is using a very broken fallback frame.

The CSD titlebar is, IMO, not the "intended" experience anyhow: if you like CSDs you should be using `INTEGRATED_BUTTONS`, if you don't you should be using SSDs. But we need to have *something* in the `TITLE` + CSD case, and it should be decent and polished above all, and not cause maintenance problems (the old custom frame was disabled with the sctk updates and no one showed up to restore it).

wez:
Thank you for this; I'm excited to see it! 
I've only very briefly skimmed this PR so far, but wanted to chime in and say that I'm totally fine with a git based dep.

piedoom:
excited to maybe be able to use wezterm on gnome again :sob: thank you :heart:

wez:
Just FYI, running `wezterm -n` with this PR under KDE plasma/wayland seems to run fine, but I see this in the stderr:

```console
$ ./target/debug/wezterm -n
12:06:12.105  WARN   sctk_adwaita::buttons > Ignoring unknown button type: icon
12:06:12.115  WARN   sctk_adwaita::buttons > Ignoring unknown button type: icon
12:06:12.189  WARN   sctk_adwaita::buttons > Ignoring unknown button type: icon
12:06:19.739  WARN   sctk_adwaita::buttons > Ignoring unknown button type: icon
```

not sure if you have control over that; a new one of those lines is logged each time window focus changes.  It wouldn't be a huge deal, but it does also clutter up the debug overlay (ctrl-shift-L to see it).

wez:
Running: `./target/debug/wezterm -n --config "window_decorations='RESIZE'"` doesn't allow dragging by the tab status area like I would expect, but resizing does work.

Same with adding in `INTEGRATED_BUTTONS`.  The buttons do behave correctly to maximize etc.

(this is still under KDE/plasma)

wez:
re: drag by tab bar status area, just verified with the mainline and it doesn't work there under wayland either, so it isn't specific to your PR. That does work on other systems though. If while you're looking into this you can make that work under wayland, that would be much appreciated!

valpackett:
Right, dragging by the tab bar definitely should be on my TODO list.

valpackett:
> 12:06:12.105  WARN   sctk_adwaita::buttons > Ignoring unknown button type: icon

That comes from the button layout which sctk-adwaita by default tries to read from `org.gnome.desktop.wm.preferences` via the `org.freedesktop.portal.Desktop` portal.

I'm not sure why KDE puts `icon` into *GNOME's* settings, when GNOME hasn't had title bar icons since the original 3.0 (I think), but the settings schema says unknown buttons should be [silently ignored](https://github.com/GNOME/gsettings-desktop-schemas/blob/e8599d1666999db7313fefe0456b871faa4580b4/schemas/org.gnome.desktop.wm.preferences.gschema.xml.in#L34-L35) and this is rather loud :) reported as https://github.com/PolyMeilex/sctk-adwaita/issues/72

teshst:
I encountered an issuing when running this new window frame implementation. Which by the way looks really good and I am excited that this is getting implemented. The issue I encountered was that when I would go to drag the window it would work the first time but when I would attempt to drag it a second time it would instead maximize the window or unmaximize it. It seemed to be taking my input as a double click. The only way to stop this was to right click then I could move the window again using the left click but it would do the same thing again after I finished dragging the window.


I found out this happens because in sctk-adwaita in pointers.rs they define the logic for the duration between clicks to decide if to move the window or maximize it. The way that it is currently implemented in this pull request is setting the duration to zero which makes every click after the first click look like a double click.

Original

In order to fix this I had to make the following changes

Location:  window.rs

Step 1:

Declared a struct field for start_time at line 587 in the screenshot

Step_1

Step 2:

Initialized the struct field at line 316 in the screenshot

Step_2

Step 3:

Stored the time since the click in the timestamp variable at line 716

Changed Duration::ZERO to timestamp at line 713

Commented out lines 722-724 because it stops double click from working consistently

Step_3

I am very new to rust so I am not sure if this is the best solution to this but it does fix the issue for me.

valpackett:
@teshst thanks a lot for the report! I haven't yet tested it myself in a non-tiling environment, so of course I missed what happens when dragging the window a lot.

> lines 722-724

Oh, haha, those are totally bogus. I think I might've been looking at some place in the old code when porting it over, but I (and/or the previous authors) somehow interpreted the "left" there as in "has *left* the window" and not "the left button has been clicked" :D

teshst:
@valpackett  No problem! I recently started using wezterm on gnome. Biggest issue for me is the window frame on gnome wayland. I compiled your fork to test the changes. So far the dragging is the only issue I have encountered.

Cargo.toml
```diff
@@ -196,6 +196,7 @@ shlex = "1.1"
 signal-hook = "0.3"
 siphasher = "1.0.1"
 smithay-client-toolkit = {version = "0.19", default-features=false}
+sctk-adwaita = { git = "https://github.com/PolyMeilex/sctk-adwaita", branch = "hide-titlebar" }
```
bew, line 199: i think you can use the main branch instead as the pr to hide titlebar was merged

valpackett: Of course! Sorry, I've been really busy with other stuff lately, I'll come back to this PR soon!

ksimuk: works well for me with `sctk-adwaita` as master branch. (Gnome, Wayland)

bew:
This would also solve #6025 as @MasonStooksbury [reported](https://github.com/wezterm/wezterm/issues/6025#issuecomment-3020473534)

HearseDev:
@wez @bew @ksimuk @tehst do you guys mind trying out #7095 to test out tab bar dragging and if there are any more issues i can check out since @valpackett has been busy.

valpackett:
Superseded by #7095, thanks @HearseDev \o/

PR #6412, opened 2024-11-19
Add font and window geometry settings to "Quick Start" example config

ernstki:
No matter what search terms I tried, for the life of me I couldn't find the config setting for initial window geometry using the docs site's built-in search.

Really the config settings to set the initial window geometry are quite straightforward, they're just not surfaced in the docs (IMO) in a place where a new user would expect to see them.

What this PR does is put those config settings front and center, right in [the "Quick Start"](https://wezfurlong.org/wezterm/config/files.html#quick-start) section.

As an entertaining aside, I tried asking a chatbot LLM and it hallucinated some very-plausible-but-nonexistent wezterm config settings. So eventually I turned to the issue tracker, discovered #256 with a search, where others mentioned similiar problems finding these settings in the docs.

PR #6937, opened 2025-05-06
WIP: Makefile `servedocs` target with livereload; reorder subsections in "Config" and "Colors & Appearance"

ernstki:
Marked "WIP" since it stacks on the commits from #6412 

Assuming that gets merged, this PR will add:

- a `make servedocs` target that live-reloads, with the mkdocs `--dirtyreload` option, which reduces build time from ~30 seconds to just seconds
  * this required removing `strict: true` from `docs/mkdocs-base.yml`, but adding `--strict` to the `make build` target (:eyes: not sure if this will break CI, though)
- better subsection structure for "Config" and "Colors & Appearance," so that similar config settings are grouped together
- many small stylistic fixes elsewhere, including the index

### DIFF
--- a/docs/mkdocs-base.yml
+++ b/docs/mkdocs-base.yml
@@ -50,8 +50,6 @@ theme:
     - toc.follow
 plugins:
   - include-markdown
-  - tags:
-      tags_file: tags.md
   - macros:
       module_name: mkdocs_macros
   - search:
--- a/docs/mkdocs-base.yml
+++ b/docs/mkdocs-base.yml
@@ -8,7 +8,6 @@ repo_name: wez/wezterm
 edit_uri: edit/main/docs/
 docs_dir: docs
 site_dir: gh_pages
-strict: true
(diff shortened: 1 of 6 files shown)


docs/mkdocs-base.yml
```diff
@@ -8,7 +8,6 @@ repo_name: wez/wezterm
 edit_uri: edit/main/docs/
 docs_dir: docs
 site_dir: gh_pages
-strict: true
```
wez, line 11: FWIW, I want to keep strict enabled as catching errors without it is very difficult

wez:
Please separate out the makefile and doc build changes from the doc contents changes, as blending them together makes it difficult to review!

ernstki:
> Please separate out the makefile and doc build changes from the doc contents changes, as blending them together makes it difficult to review!

Gotcha, that's a valid point, and will do. Thanks for coaching me on it!

Issue #6955, opened 2025-05-11
spawnCommand new tab/split does not use default_prog


adpce:
### What Operating System(s) are you seeing this problem on?

Linux X11

### Which Wayland compositor or X11 Window manager(s) are you using?

KDE Plasma 5.24.7

### WezTerm version

wezterm 20250511-100117-a841a72c

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

When running action.SplitHorizontal or action.SpawnCommandInNewTab, the new tab/pane opens running bash instead of the default_prog.

EDIT: clicking the new tab button in the non-fancy tab bar does respect default_prog.

### To Reproduce

1. Open wezterm (this does respect default_prog).
2. Spawn new tab/split.
3. It is running bash.

### Configuration

default_prog = { '/usr/local/bin/zsh', '--login' },

keys = {
		{ key = 'F1',  mods = 'SUPER',      action = act.SpawnCommandInNewTab { cwd = wezterm.home_dir } },
		{ key = 'F2',  mods = 'SUPER',      action = act.SplitHorizontal { cwd = wezterm.home_dir } },
}

### Expected Behavior

Open new tab/split, it runs default_prog. 

### Logs

No logs.

wez:
This sort of behavior was recently tweaked in #6850 
@vincentbesanceney would you mind taking taking a look at this?

vincentbesanceney:
I have been able to reproduce the issue. This is due to the use of `CommandBuilder::new_default_prog` in #6850. I'm gonna fix that.

PR #6972, opened 2025-05-19
fix: use config.default_prog for SpawnCommand (#6955)

vincentbesanceney:
SpawnCommand implementation was revised as part of #6850 to support environment variables and cwd unconditionally. This introduced an issue where CommandBuilder::new_default_prog was used in lieu of config.default_prog.


vincentbesanceney:
I think you are right :smile:. You didn't regress anything: my config and the config from #6955 both work. 👍

Issue #5952, opened 2024-08-13
Search/Copy Mode enhancements


mikatpt:
**Is your feature request related to a problem? Please describe.**

Hi! Thanks for such a lovely project. I'm coming from tmux land and 90% of things are peachy, but there's a few things with search that I'm having a lot of difficulty working around.

1. If you run a command (like a test), then go into `Copy Mode`, the stdout output will cause your cursor position to move with the output—this happens regardless of scrollback buffer size.
2. Transitioning from `Search Mode` to `Copy Mode` is a pain—it always throws you straight into selection (`Visual` mode for vim users), and there is no way to clear it. i.e. `ClearSelectionMode` and `SetSelectionMode` _only_ work if you're already in `Copy Mode`, running those after `ActivateCopyMode` will fail.
3. Search _always_ begins from the bottom of the terminal. If I am 6000 lines up in `CopyMode`, it's pretty unexpected to go into search and be thrown all the way back.

**Describe the solution you'd like**

I saw there was a request for a unified search/copy mode ( https://github.com/wez/wezterm/issues/1592 ), but am unsure that the solution came to totally did this—I think that removing search mode entirely, in favor of combing it as functionality in copy mode, is a more ergonomic solution here.

For us vim users, it's very natural to feel like you are simply in normal mode and you can navigate around, search using `/` and `?`, then select/copy with `v` and `y`—tmux did this quite well and it's the one thing I miss from it a lot!

Aside from that I'm very pleased with everything—thanks again for your hard work.

**Describe alternatives you've considered**

Here are my copy mode bindings. It's imperfect, allowing me to sort of mimick vim "normal mode"—again, there's no solution I found to going `Search Mode -> Normal Copy Mode`, only `Search -> Selection Copy Mode`

```lua
local wezterm = require('wezterm')
local function close_copy_mode()
    return act.Multiple({
        act.CopyMode('ClearSelectionMode'),
        act.CopyMode('ClearPattern'),
        act.CopyMode('Close'),
    })
end
local function copy_to()
    return act.Multiple({
        act.CopyTo('Clipboard'),
        act.CopyMode('ClearSelectionMode'),
    })
end
local function next_match(int)
    local m = act.CopyMode('NextMatch')
    if int == -1 then
        m = act.CopyMode('PriorMatch')
    end
    return act.Multiple({ m, act.CopyMode('ClearSelectionMode') })
end

local function extend_keys(target, source)
    -- snip. ... overrides duplicate `key` values
    return target
end

[...]

Issue #5236, opened 2024-03-29
Add search feature in copy mode


zaiqiang-dong:
**Describe the solution you'd like**
I really like copy mode, but I would like to add a search function because we usually need to find the string we need before copying. Like vim operations.
If we have this feature, we can copy it like the following:
1. CTRL-SHIFT-X enters copy mode
2. Press "/" to search for the target string, and the cursor will move to the target string while searching
3. Press "w" to select string, then press "y" to copy

PR #6999, opened 2025-06-01
copy-mode: allow Search from cursor (not from scrollback bottom)

mgpinf:
When in CopyMode, when we want to search for a pattern, it is useful to
activate the first result before or after the current cursor position.
This is similar to the behavior in vim.
Resolves:
* Point 3 mentioned in https://github.com/wezterm/wezterm/issues/5952#issue-246440134
* https://github.com/wezterm/wezterm/issues/5236


bew:
Hello!
What do you think about having a single `SearchRelativeToCursor` with parameter for direction, or even merge all into the existing `Search` action with new parameters?

mgpinf:
@bew I have updated the PR. Can you check? I have used a separate KeyAssignment `ExtendedSearch` because the `Search` KeyAssignment currently accepts only one parameter

9999years (approved):
Looks pretty good to me, but I'm not super familiar with this codebase so my approval is of limited value here :)

9999years:
I've built and tested this patch locally and it works very nicely. I'm not 100% sure about the API (I'd probably prefer to have this as a new mode for the existing `act.Search`) but it's a big improvement from the current state.

mgpinf:
I have merged all the new parameters into the existing `Search` action

 Now, the `Search` action takes one more pattern called `Extended` where we can
 specify the match position to activate and all options previously
 specified against Search

 The old API introduced in this PR looked like below:

 ```lua
 local act = wezterm.action

 act.ExtendedSearch {
   pattern = { CaseSensitiveString = selection_text },
   activate_match = activate_match,
 }
 ```

 The new API is as shown below:
 ```lua
 local act = wezterm.action

 act.Search {
   Extended = {
     pattern = 'CurrentSelectionOrEmptyString',
     activate_match = 'AfterCursor',
   }
 }
 ```

bew:
Hello! Could you describe what this `Extended` 'mode' (?) does is for? When should I use it compared to normal search?

Now that it's integrated in `Search`, would it be possible to merge the 2 instead of making a kind of additional search system in `Search`?

mgpinf:
I had used `Extended` because I wanted to be able to specify multiple parameters like pattern and match to activate. Can you suggest an example of how the API might look like if they are merged?

9999years:
@mgpinf I can't speak for @bew, but I'd like `act.Search` to support multiple parameter conventions simultaneously, e.g.:

```
-- Old style:
act.Search("CurrentSelectionOrEmptyString")

-- New style:
act.Search {
  pattern = "CurrentSelectionOrEmptyString",
  activate_match = "AfterCursor",
}
```

But I'm not sure if the WezTerm Lua bindings support this sort of thing currently (I think it would be fine to add that as an improvement in a later PR).

mgpinf:
The Search action now accepts a table format with an optional `activate_match` field:

```lua
wezterm.action.Search {
  pattern = { Regex = '[a-f0-9]{6,}' },
  activate_match = 'AfterCursor',
}
```

Values for `activate_match` values include:
* `First`: Activate the first match in the scrollback (default)
* `AfterCursor`: Activate the first match after the current cursor position
* `BeforeCursor`: Activate the first match before the current cursor position

The simple syntax continues to work:

```lua
wezterm.action.Search { Regex = '[a-f0-9]{6,}' }
```

9999years:
Looks super slick, love the new design!

config/src/keyassignment.rs
```diff
@@ -111,30 +111,153 @@ pub enum SelectionMode {
...
     Regex(String),
     CurrentSelectionOrEmptyString,
 }
 
+/// Pattern for search operations
+#[derive(Debug, Clone, PartialEq, Eq, ToDynamic)]
+pub struct Pattern {
+    pub pattern: InnerPattern,
+    #[dynamic(default)]
+    pub activate_match: Option<ActivateMatchPosition>,
```
bew, line 135: I think this field would be clearer if renamed to `start_from`
So it reads like e.g.: `start_from = "AfterCursor"`

Related: `First` could be renamed `ScrollbackBottom` for clarity.

And I don't think this needs to be `Option`, you can have a default on `ActivateMatchPosition`!

config/src/keyassignment.rs
```diff
@@ -111,30 +111,153 @@ pub enum SelectionMode {
...
+            "CaseSensitiveString",
+            "CaseInSensitiveString",
+            "CaseSmartString",
+            "Regex",
+            "CurrentSelectionOrEmptyString",
+        ]
+    }
+}
+
+impl FromDynamic for Pattern {
```
bew, line 150: FYI: #8039 will allow to avoid this entire boilerplate :)

docs/config/lua/keyassignment/Search.md
```diff
@@ -55,3 +60,43 @@ config.keys = {
 You may now use `wezterm.action.Search("CurrentSelectionOrEmptyString")` to have the search take the currently selected text as the item to search.
 
 The selection text is adjusted to be a single line.
+
+{{since('nightly')}}
+
+`Search` now accepts a table with additional configuration options. The simple
```
bew, line 66: I think it would be good to mention how it is useful when used from copy mode.

docs/config/lua/keyassignment/Search.md
```diff
@@ -55,3 +60,43 @@ config.keys = {
...
+local act = wezterm.action
+
+config.keys = {
+  -- search for git hashes, activating the first match after cursor
+  {
+    key = 'H',
+    mods = 'SHIFT|CTRL',
+    action = act.Search {
+      pattern = { Regex = '[a-f0-9]{6,}' },
+      activate_match = 'AfterCursor',
```
bew, line 89: and here, put this key in the `copy_mode` key table where it is most useful

bew (commented):
Welp this is an old PR!
Thanks everyone for sticking around 🙃

I didn't review the impl in the overlay yet, but here are a few direction comments already :)

I think we'll wait for next release before merging this.

---
FYI: Once #8039 lands the complexity of the dual config format will be completely nullified 😉

Issue #6685, opened 2025-02-16
Clipboard (most times) not working between windows


theNoof:
### What Operating System(s) are you seeing this problem on?

Linux Wayland

### Which Wayland compositor or X11 Window manager(s) are you using?

niri (niri 25.01)

### WezTerm version

wezterm 0-unstable-2025-02-10

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

i cant copy a selected text in one window and paste it in another, except sometimes after waiting ~10 minutes. copy/pasting from/to other windows works. also tried on build from 3.1.2025 (i believe).

i am working on nixos with the unstable channel and home-manager.

### To Reproduce

use the software mentioned above (it maybe occurs with any compositor, havent tested it).

### Configuration

no clipboard-related config

### Expected Behavior

copy-paste between windows working

### Logs

Debug Overlay
wezterm version: 0-unstable-2025-02-10 x86_64-unknown-linux-gnu
Window Environment: Wayland
Lua Version: Lua 5.4
OpenGL: Mesa Intel(R) UHD Graphics 620 (KBL GT2) 4.6 (Compatibility Profile) Mesa 24.3.4
Enter lua statements or expressions and hit Enter.
Press ESC or CTRL-D to exit
16:24:25.546 ERROR wezterm_mux_server_impl::local > writing pdu data buffer: Broken pipe (os error 32)

bew:
Since you edited the title, is it an issue only between terminals? (which ones?)
Does it work with other windows?

> it maybe occurs with any compositor, havent tested it

I don't know niri, would be interesting to check another one, maybe it's a niri issue?

w-jablonski:
Here also copy in Niri broken. Unable to copy to _other_ windows at all, incl. Wezterm windows. Each Wezterm window has kinda it's own clipboard. I copy to _micro_ as a workaround -- from micro it copies okay. All _paste_ works okay. On Sway no issue with copy. Arch Linux stable.

samuela:
I'm also on niri and having issues pasting to WezTerm. Alacritty works without issue. Unclear why. Most recently I found that copying from Firefox and pasting into WezTerm wasn't working.

For me I can reproduce by attempting to copy the SSH path of a GitHub repository from the repository overview page (dropping down from the green button on the right hand side) and then trying to paste that into WezTerm.

filtsin:
[...]

PR #7034, opened 2025-06-17
fix: save the wayland selection offer to every CopyAndPaste

XeroOl:
I use the niri compositor, and I'm affected by #6685. This patch fixes the issue for me.

I'm not very familiar with how WezTerm is structured, so I don't understand why WezTerm has a separate `CopyAndPaste` per window, but it makes sense to me that when something is copied, it should be able to be pasted in all windows, not just the one that was the active_surface_id at the time when the clipboard was filled.

I've tested this in niri, but not in other compositors.

### DIFF
--- a/window/src/os/wayland/copy_and_paste.rs
+++ b/window/src/os/wayland/copy_and_paste.rs
@@ -99,15 +99,6 @@ impl CopyAndPaste {
     }
 }
 
-impl WaylandState {
-    pub(super) fn resolve_copy_and_paste(&mut self) -> Option<Arc<Mutex<CopyAndPaste>>> {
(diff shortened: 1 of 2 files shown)


TGlide:
any update on this?

abatkin:
@wez Do you think that this PR (or something similar) can make it into an upcoming release? Paste from the system clipboard is mostly unusable from Wezterm on modern Linux/Wayland when there are multiple Wezterm windows. There are steps to reliably reproduce in #6685

bew:
Hello, I've taken this branch and updated it with latest `main`, see updated PR @ #7863. 
I'm planning to merge it fast.

PR #7035, opened 2025-06-18
Add copy mode cursor configuration options

mgpinf:
Add `copy_mode_cursor_fg`, `copy_mode_cursor_bg` configuration options

### DIFF
--- a/docs/config/appearance.md
+++ b/docs/config/appearance.md
@@ -71,6 +71,13 @@ config.colors = {
   -- Bar or Underline.
   cursor_border = '#52ad70',
 
+  -- Overrides the cell background color when copymode is enabled and the current cell is
+  -- occupied by the cursor and the cursor style is set to Block
+  copy_mode_cursor_bg = '#52ad70', -- {{since('nightly', inline=True)}}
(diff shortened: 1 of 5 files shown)


docs/config/appearance.md
```diff
@@ -71,6 +71,13 @@ config.colors = {
   -- Bar or Underline.
   cursor_border = '#52ad70',
 
+  -- Overrides the cell background color when copymode is enabled and the current cell is
+  -- occupied by the cursor and the cursor style is set to Block
+  copy_mode_cursor_bg = '#52ad70', -- {{since('nightly', inline=True)}}
+  -- Overrides the text color when copymode is enabled the current cell is occupied by
```
bew:
```suggestion
  -- Overrides the text color when copymode is enabled and the current cell is occupied by
```

bew (changes requested):
Hello!
Adding copy-mode specific handling in termwindow is the wrong way to go, you should instead see how to update the cursor color in the copy-mode overlay

Source code:
https://github.com/wezterm/wezterm/blob/main/wezterm-gui/src/overlay/copy.rs

PR #7046, opened 2025-06-21
term: indicate support for OSC 52 in Primary DA response

j4james:
This is an extension agreed upon by modern terminals to indicate that
they support copying to the clipboard with XTerm's OSC 52 sequence.

https://github.com/contour-terminal/vt-extensions/blob/master/clipboard-extension.md


j4james:
When we discussed this in https://github.com/contour-terminal/contour/issues/1761 you seemed to be happy to support this extension in WezTerm, so I thought I'd push things forward by creating a PR. As far as I can tell, WezTerm's `OSC 52` implementation already meets all the requirements of the extension, and I couldn't see any way to disable clipboard access, so it's just a matter of adding the `52` parameter to the DA response.

wez:
Thank you!

PR #7049, opened 2025-06-24
improve pct_to_glyph algorithm for progress reporting

sunshowers:
The current `pct_to_glyph` algorithm would unintentionally produce `\u{f0aa6}` for percentages in the range 96..=100.

Use a different algorithm that:

* uses an empty circle for the first 5%
* uses a filled circle for the last 5%
* has a roughly even distribution for the rest of the range


wez:
Thank you!

PR #7059, opened 2025-06-28
Disable '*_continuous' workflows for forks

tombh:
I was surprised to get a tranche of failed CI runs last night on my fork 😅

### DIFF
--- a/.github/workflows/gen_centos9_continuous.yml
+++ b/.github/workflows/gen_centos9_continuous.yml
@@ -28,6 +28,7 @@ on:
 jobs:
   build:
     runs-on: "ubuntu-latest"
+    if: github.repository == 'wezterm/wezterm'
     container: "quay.io/centos/centos:stream9"
     env:
       BUILD_REASON: "Schedule"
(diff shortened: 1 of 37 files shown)


.github/workflows/gen_centos9_continuous.yml
```diff
@@ -27,6 +27,7 @@ on:
 
 jobs:
   build:
+    if: github.repository == 'wezterm/wezterm'
```
wez: these `gen_xxx.yml` files are (re)generated by running `ci/generate-workflows.py`, so you'll need to add this change there for it to stick

wez (changes requested):
Thanks for this!

I can see that harfbuzz is being updated here; that's probably a mistake and it shouldn't be included in this PR.

Most of the workflow files are generated by a script, so that script should be updated so that the changes will persist.

tombh:
Oops sorry I missed the harfbuzz change. Removed that.

And used the generator to make the changes. But that changes all the workflow files. But I think that's okay? I don't think forks want to be automatically running any of them?

bew (approved):

Issue #6787, opened 2025-03-12
Unicode combining characters don't render when input directly


aurelivia:
### What Operating System(s) are you seeing this problem on?

Windows

### WezTerm version

wezterm 20240203-110809-5046fc22

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

I might be the only person in the world who does this, but I use a lot of unicode, and I have an XCompose equivalent set up to allow me to input various characters directly. While Wezterm has rendered all the characters I've used so far just fine, when I input them directly, they don't show up, even though they have been received.

### To Reproduce

Type (or paste) `e` `c` `h` `o` ` ` `'` `a` (U+0302 Combining Acute) `'`

Terminal renders `echo 'a'` (no acute)

Hit enter, and you get
```
$ echo 'a'
á
```
Thus demonstrating that the character is there, it's just not being rendered.

### Configuration

no config

### Expected Behavior

I'm able to see everything I type

### Logs

Is this helpful? Seems like a rendering bug... Will provide if asked.

digitallyserviced:
@aurelivia 

What shell are you using within wezterm?

I have noticed that not all shells seem to work as you intend. Below you can see zsh when pasting the character causes a unicode block to render. But bash will combine the characters. The screen is wezterm, initially in zsh, and then in bash.

I have pasted the character and that is the result I get. 



![Image](https://github.com/user-attachments/assets/f4613d14-c93b-4803-96cf-b1f167a5135b)

aurelivia:
Interesting. I am using MSYS bash by default, and I tried just using the native command prompt to attempt to rule out it as the culprit, which had the same behaviour, however I just tried using powershell and it worked. I'm not sure if it works in the native command prompt, as it doesn't render combining characters at all, however in the built-in MSYS bash shell, which I guess is Cygwin, works as expected. Additionally the new Windows Terminal does render combining characters on input for any of the shells, but has issues spacing the characters out.

Issue #6817, opened 2025-03-21
Wezterm hangs on language change postmessage


eugenesvk:
### What Operating System(s) are you seeing this problem on?

Windows

### Which Wayland compositor or X11 Window manager(s) are you using?

none

### WezTerm version

wezterm 20250320-072107-a8735851

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

Hangs with `WebGpu` (`OpenGL` works) when you send a language change message

```ahk
layout := DllCall("LoadKeyboardLayout"	, "str","00000419"	, "uint",1)
PostMessage(msg:=0x50, w:=0, l:=layout, winID:=WinActive("A"))
```

### To Reproduce

- Open WezTerm with the config from below
- use e.g. Autohotkey to send a language changing message
- see WezTerm hang

### Configuration

```lua
return {
  hide_tab_bar_if_only_one_tab = true,
  check_for_updates = false,
  front_end = "WebGpu", 
  adjust_window_size_when_changing_font_size = false,
}
```

### Expected Behavior

No hanging on getting a windows message

### Anything else?

The latest released version [20240203-110809-5046fc22](https://github.com/wezterm/wezterm/releases/tag/20240203-110809-5046fc22) doesn't hang, only nightly, not sure which version started it since couldn't easily find previous nightly versions

eugenesvk:
after another search, found that this is a dupe of https://github.com/wezterm/wezterm/issues/6786

God-damnit-all:
> after another search, found that this is a dupe of [#6786](https://github.com/wezterm/wezterm/issues/6786)

You forgot to close this issue.

PR #7066, opened 2025-07-01
fix: hang when receiving WM_INPUTLANGCHANGEREQUEST message

TrafalgarSX:
## Problem
Wezterm hangs when switching input methods using tools like `im-select` that send `WM_INPUTLANGCHANGEREQUEST` messages via `PostMessage`. This happens because the message is not explicitly handled and falls through to `DefWindowProcW`, which can cause deadlocks. 

The link below might explain why #6787 #6817 happens:
https://marc.durdin.net/2012/08/waitforsingleobject-why-you-should-never-use-it/

## Solution
- Added explicit handling for `WM_INPUTLANGCHANGEREQUEST` in `do_wnd_proc`
- Call `ActivateKeyboardLayout` directly and return `Some(0)` to prevent the message from reaching `DefWindowProcW`
- This maintains input method switching functionality while avoiding the deadlock

## Testing
- [x] Tested with `im-select.exe` switching between English and Chinese input methods(cmd: im-select.exe 1033 or 2052)
- [x] Verified wezterm no longer hangs during input method switches
- [x] Confirmed input method switching still works correctly
- [x] Built and tested release version

## Changes
- Modified `window/src/os/windows/window.rs` to handle `WM_INPUTLANGCHANGEREQUEST`

Fixes: #6787 #6817


God-damnit-all:
@wez This works for me, I'd appreciate if you would merge this soon.

Issue #1034, opened 2021-08-16
macOS: Disable titlebar but keep rounded corners


cwqt:
**Is your feature request related to a problem? Please describe.**
Currently `window_decorations` `TITLE` hides not only the titlebar but all other window decorations

**Describe the solution you'd like**
Extend `window_decorations` to allow hiding of titlebar decoration while keeping rounded corners, see pic, bottom is kitty & top is wezterm

in kitty it's possible to keep the corner decorations via:

```
hide_window_decorations titlebar-only
```

<img width="852" alt="Screenshot 2021-08-16 at 22 37 36" src="https://user-images.githubusercontent.com/61694629/129632707-145a6892-50b2-4acb-83f8-785b47c6918b.png">

wez:
Did you try `window_decorations = "RESIZE"` ?

nklmilojevic:
Just got into the same issue, I have tried resize as window_decorations on macos but it is still not a rounded corner.

nklmilojevic:
@wez can you reopen this one?

cwqt:
Sorry forgot about this one, & yes `RESIZE` has no effect

<img width="863" alt="Screenshot 2021-11-28 at 4 28 02 pm" src="https://user-images.githubusercontent.com/61694629/143776947-39ecd1d0-2292-40b5-ab06-3662fd019716.png">

nklmilojevic:
Any idea where is this coming from? I'd gladly help with a PR if you just point me where to look :)

wez:
I would gladly accept a PR for this!

All we do is apply a style mask based on the requested decorations:

https://github.com/wez/wezterm/blob/main/window/src/os/macos/window.rs#L713-L717

(there are a couple of places that call `decoration_to_mask` in there that do similar but slightly different things in different contexts,)

I haven't had the time to dig into why that stops the corners being rounded... I would suggest stackoverflow/google searching, and/or taking a peek in the kitty code (which might just delegate through glfw) to see how it makes this work!

nklmilojevic:
Thanks! I will also check at Alacritty because it has rounded edges.

nklmilojevic:
My rust is rusty (😅) but shouldn't we add https://docs.rs/cocoa/0.20.2/cocoa/appkit/enum.NSWindowTitleVisibility.html somewhere?

nklmilojevic:
[...]

PR #7081, opened 2025-07-10
Remove deprecated `NSWindowFullScreenButton`

### LINKED ISSUE #7080: Debug builds of `wezterm-gui` panic on startup on macOS

macOS

_No response_

### WezTerm version

Git commit 0198d23a

### Describe the bug

On macOS (Sequoia 15.5 aarch64) debug builds of `wezterm-gui` panic on startup:
```sh
$ cargo run --bin wezterm-gui
...
   Compiling wezterm-gui v0.1.0 (/Users/cpick/src/wezterm/wezterm-gui)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 5.83s
     Running `target/debug/wezterm-gui`
18:54:27.152  ERROR  env_bootstrap > panic at window/src/os/macos/window.rs:1394:25 - null pointer dereference occurred	
   0: backtrace::backtrace::libunwind::trace
...
```

That corresponds to the `msg_send!` invocation below:
```rust
for titlebar_button in &[
    appkit::NSWindowButton::NSWindowFullScreenButton,
    appkit::NSWindowButton::NSWindowMiniaturizeButton,
    appkit::NSWindowButton::NSWindowCloseButton,
    appkit::NSWindowButton::NSWindowZoomButton,
] {
    let button = window.standardWindowButton_(*titlebar_button);
    let _: () = msg_send![button, setHidden: hidden];
}
```

Some debugging revealed that `button.is_null() == true` any time `titlebar_button == NSWindowFullScreenButton`.
It looks like this is expected behavior, from [the docs](https://developer.apple.com/documentation/appkit/nswindow/buttontype/fullscreenbutton?changes=__6_5&language=objc):
> Deprecated
> The standard window button for NSWindowFullScreenButton is always nil; use NSWindowZoomButton instead

`NSWindowZoomButton` is already covered in the loop so it should be possible to fix this by removing `NSWindowFullScreenButton` without any loss of functionality.

This hasn't caused a problem so far because `wezterm-gui` is always built in release mode and (at least on macOS) dereferencing a null pointer won't panic, a quick test program:
```sh
$ cat src/main.rs
fn main() {
    let null_ptr: *const i32 = std::ptr::null();
    unsafe {
        let _value = *null_ptr; // undefined behavior
    }
    println!("no panic");
}

$ cargo run
   Compiling delme-rust-null v0.1.0 (/Users/cpick/src/delme-rust-null)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.11s
     Running `target/debug/delme-rust-null`

thread 'main' panicked at src/main.rs:4:22:
null pointer dereference occurred
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread caused non-unwinding panic. aborting.

$ cargo run --release
   Compiling delme-rust-null v0.1.0 (/Users/cpick/src/delme-rust-null)
    Finished `release` profile [optimized] target(s) in 0.13s
     Running `target/release/delme-rust-null`
no panic
```

### To Reproduce

Run `cargo run --bin wezterm-gui` on macOS

### Configuration

no config

### Expected Behavior

A clean startup

### Logs

_No response_

### Anything else?

_No response_

cpick:
Fixes #7080 to prevent null pointer dereference (and crash in debug mode) on macOS.

`NSWindowFullScreenButton`
[is deprecated](https://developer.apple.com/documentation/appkit/nswindow/buttontype/fullscreenbutton?changes=__6_5&language=objc). Removing it prevents a null pointer from being dereferenced and doesn't otherwise impact behavior.

Confirmed that `window_decorations = "RESIZE"` sill works [with rounded corners](https://github.com/wezterm/wezterm/issues/1034).

### DIFF
--- a/window/src/os/macos/window.rs
+++ b/window/src/os/macos/window.rs
@@ -1385,7 +1385,6 @@ fn apply_decorations_to_window(
         };
 
         for titlebar_button in &[
-            appkit::NSWindowButton::NSWindowFullScreenButton,
             appkit::NSWindowButton::NSWindowMiniaturizeButton,
             appkit::NSWindowButton::NSWindowCloseButton,
             appkit::NSWindowButton::NSWindowZoomButton,


wez:
Thanks!

Issue #7090, opened 2025-07-14
WezTerm shell integration conflicts with other terminal programs (wezterm.sh)


jobbler:
### What Operating System(s) are you seeing this problem on?

Linux Wayland

### Which Wayland compositor or X11 Window manager(s) are you using?

Gnome 48 with Wayland

### WezTerm version

wezterm 20250713_135109_85c587f9

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

No, and I'll explain why below

### Describe the bug

The wezterm.sh script added to the /etc/profiles.d directory always runs. This interferes with other terminal programs such as kitty. I propose adding a line in the wezterm.sh script to prevent it from running if the terminal is not WezTerm.

I did not try the nightly build because this script is the same in it.

### To Reproduce

Install WezTerm
Install Kitty
Open a kitty terminal and issue `echo $PS1`
Issue the `echo $PS1` command a second time and you will see the PS1 string has increased in size.
Every command causes the PS1 string to increase.

### Configuration

no config

### Expected Behavior

I would expect the wezterm.sh script to not execute when logging in unless I am running WezTerm.

jobbler:
I am moving comments I made in the [PR #7091](https://github.com/wezterm/wezterm/pull/7091) to the issue where I think they are better discussed.

jobbler:
Maybe just checking if any variable with "WEZTERM" in its name would work?

`env | grep -qE "^WEZTERM_"
if [ $? -eq 1 ]
then
echo "WEZTERM No" > /tmp/wez
return 0
fi

echo "WEZTERM Yes" > /tmp/wez
`

That works with tmux as well. I kept it POSIX compliant for shell compatibility. The echo statements are just an easy way for me to see how it behaves for the different terminals and tmux. I will take those out and submit an updated PR if you think this will work.

jobbler:
One option for passing the WEZTERM variables would be to use AcceptEnv and SendEnv in the sshd and ssh configuration files. It does require privileges to edit the remote hosts sshd configuration file.

Adding AcceptEnv WEZTERM_* to the remote hosts sshd_config files and restarting the sshd server instructs the sshd server to accept forwarded variables.
Adding the following to the users .ssh/config file:
Host HOSTNAME_OR_WILDCARD SendEnv WEZTERM_*

will send all the WezTerm variables to the remote host.
You can also add the SendEnv lines to the system wide ssh_config file.

[...]

PR #7091, opened 2025-07-14
Prevent shell integration from interfering with other terminal programs

jobbler:
The following changes the wezterm.sh script so that it only runs if the terminal is WezTerm. Thus preventing from interfering with other terminal programs.

I tested this on my local system and did not run the rust and cargo tests since it is not a change to any of the rust code.

This should address #7090

### DIFF
--- a/assets/shell-integration/wezterm.sh
+++ b/assets/shell-integration/wezterm.sh
@@ -14,6 +14,11 @@
 # WEZTERM_SHELL_SKIP_USER_VARS - disable user vars that capture information
 #                                about running programs
 
+if [ "${TERM_PROGRAM}" != "WezTerm" ]
+then
+  return 0
+fi
+
 # shellcheck disable=SC2166
 if [ -z "${BASH_VERSION-}" -a -z "${ZSH_NAME-}" ] ; then
   # Only for bash or zsh


assets/shell-integration/wezterm.sh
```diff
@@ -14,6 +14,11 @@
 # WEZTERM_SHELL_SKIP_USER_VARS - disable user vars that capture information
 #                                about running programs
 
+if [ "${TERM_PROGRAM}" != "WezTerm" ]
```
bew, line 17: I'm not sure checking `TERM_PROGRAM` is wise here, as for example when I check its value in tmux it is `tmux` and not `Wezterm` but I might want to have the shell integrations to work there. (although it won't be perfect)
Maybe just check if variables like `WEZTERM_PANE` exist?

Hmm but this might not work on remote servers unless that server is configured to forward `WEZTERM_*` vars 🤔

jobbler: Maybe just checking if any variable with "WEZTERM" in its name would work?

`env | grep -qE "^WEZTERM_"
if [ $? -eq 1 ]
then
  echo "WEZTERM No" > /tmp/wez
  return 0
fi

echo "WEZTERM Yes" > /tmp/wez
`

That works with tmux as well. I kept it POSIX compliant for shell compatibility. The echo statements are just an easy way for me to see how it behaves for the different terminals and tmux. I will take those out and submit an updated PR if you think this will work.

jobbler: One option for passing the WEZTERM variables would be to use AcceptEnv and SendEnv in the sshd and ssh configuration files. It does require privileges to edit the remote hosts sshd configuration file.

Adding `AcceptEnv WEZTERM_*` to the remote hosts sshd_config files and restarting the sshd server instructs the sshd server to accept forwarded variables.
Adding the following to the users .ssh/config file:
`Host HOSTNAME_OR_WILDCARD
     SendEnv WEZTERM_*`

will send all the WezTerm variables to the remote host. 
You can also add the SendEnv lines to the system wide ssh_config file.

This is probably not ideal. But would work. If you want something more automated, I guess an escape code could be implemented that could return the information. And then check that in the script. Or maybe the code just gets the emulators attention and it accepts the next string as a command that returns whatever (or does whatever).

Example: 
`echo "\003[Y"` returns "wezterm"
or `echo "\003[Y` grabes the terminals attention to accept a command like "term" or "status" or whatever is decided to implement in the future.

If you want to go the ssh configuration route and you are busy, I can add that to the PR as well since they tie together. But I do not know anything about rust, so I would not be much help in implementing anything like an escape code.

jobbler: @bew I am moving my comments over to the issue. It might be better discussed there. [Issue #7090](https://github.com/wezterm/wezterm/issues/7090)

PR #7093, opened 2025-07-16
Open in new tab setting in Nautilus integration

tmrlvi:
I was missing a "new-tab" option for the nautilus "Open in Wezterm" plugin. This PR adds this options through the `GSettings` interface. As a result, it also adds the schema `org.wezfurlong.wezterm` to the glib schema.

I am not sure if this is the proper way we want to configure this setting, or should we control `wezterm start` behavior in the wezterm configuration instead. I'm hoping setting this PR as draft could use as a basis for discussion. 

After updating the schemas, you can activate the "new-tab" option by running
```
gsettings set org.wezfurlong.wezterm new-tab true
```

To update the schemas for only a specific user, you can run 
```
cp assets/wezterm.gschema.xml ~/.local/share/glib-2.0/schemas/
glib-compile-schemas ~/.local/share/glib-2.0/schemas/
```

to update it system-wide, run
```
sudo cp assets/wezterm.gschema.xml /usr/share/glib-2.0/schemas/
sudo glib-compile-schemas /usr/share/glib-2.0/schemas/
```

### DIFF
--- a/README-DISTRO-MAINTAINER.md
+++ b/README-DISTRO-MAINTAINER.md
@@ -44,7 +44,7 @@ string.
 ## Additional Resources
 
 * `assets/shell-integration`, `assets/shell-completion`: should be deployed along with the `wezterm` executable
-* `assets/wezterm.desktop`, `assets/wezterm.appdata.xml`, `assets/wezterm-nautilus.py`: should be deployed along with `wezterm-gui`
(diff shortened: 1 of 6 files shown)


tmrlvi:
I saw now there was a past attempt to tackle this problem at https://github.com/wezterm/wezterm/pull/1775. I think the `GSettings` solution avoids changing the behavior of `wezterm start` and does not require additional integration with the DBus.

Issue #6738, opened 2025-02-28
Infinite loop when searching over system menu


bazhenov:
### What Operating System(s) are you seeing this problem on?

macOS

### Which Wayland compositor or X11 Window manager(s) are you using?

Don't thunk it is important, but technically speaking I'm using Rectangle for macOS

### WezTerm version

20250223-153857-fd532a8c

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

app became non responsive to any input, 100% CPU usage, Activity monitor reports that _Application is not responding_

### To Reproduce

1. launch wezterm
2. press <kbd>⌘</kbd><kbd>⇧</kbd><kbd>/</kbd> (system shortcut to search application menu)
3. search and confirm any item (for example, "New tab")
4. press <kbd>⌘</kbd><kbd>⇧</kbd><kbd>/</kbd> again

### Configuration

```lua
local wezterm = require 'wezterm'

local config = wezterm.config_builder()

config.front_end = "WebGpu"
config.color_scheme = 'rose-pine'
config.colors = {
  copy_mode_active_highlight_bg = { Color = '#ffffff' },
  copy_mode_active_highlight_fg = { Color = '#ffffff' },
  copy_mode_inactive_highlight_bg = { Color = '#ffffff' },
  copy_mode_inactive_highlight_fg = { Color = '#ffffff' },
  selection_bg = 'rgba(50% 50% 50% 50%)',
}
config.font = wezterm.font 'JetBrains Mono'
config.font_size = 13.0
config.quick_select_patterns = {
  '[a-zA-Z0-9-./_]{10,}',
}

return config
```

### Logs

When sampling Activity Monitor shows following abuser:

```
    2441 Thread_5135455   DispatchQueue_1: com.apple.main-thread  (serial)
    + 2441 start  (in dyld) + 2840  [0x182578274]
    +   2441 main  (in wezterm-gui) + 52  [0x100b15f88]
    +     2441 std::rt::lang_start_internal::hacda2dedffd2edb4  (in wezterm-gui) + 1092  [0x102100004]
    +       2441 std::rt::lang_start::_$u7b$$u7b$closure$u7d$$u7d$::h48f6b4e5859b9547  (in wezterm-gui) + 16  [0x100a3cdd8]
[...]

PR #7098, opened 2025-07-22
Prevent infinite loop in macOS' "Service" menu validation

cpick:
The `WezTermAppDelegate` class was setting `NSWindow` as its parent.

This caused an infinite loop any time macOS's "Service" menu was expanded becuase it would pass [`validRequestorForSendType:returnType:`](https://developer.apple.com/documentation/appkit/nsresponder/1524638-validrequestor) messages along the full responder chain to [check for objects that are eligible for services offered by other applications](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/EventOverview/EventArchitecture/EventArchitecture.html#//apple_ref/doc/uid/10000060i-CH3-148895).

The [responder chain](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/EventOverview/EventArchitecture/EventArchitecture.html#//apple_ref/doc/uid/10000060i-CH3-SW9) would walk up to the `WezTermAppDelegate` which, since it was an `NSWindow`, would walk back to `NSApp` which would walk to its delegate (`WezTermAppDelegate`) and so on forever.

Change `WezTermAppDelegate`s parent class `NSObject` to prevent this
loop.

Fixes: https://github.com/wezterm/wezterm/issues/6738
Fixes: https://github.com/wezterm/wezterm/issues/6833


wez:
Thank you!

joshmedeski:
@wez are we going to get a release soon so we can start using Apple Intelligence on a stable release of Wezterm? 🙏

wez:
@joshmedeski see https://github.com/wezterm/wezterm/issues/6341
As a life update: I'm in the middle of moving countries (USA -> UK) and am only just starting to come up for air and get back into work that will pay my mortgage. I haven't had head-space for wezterm, and haven't reviewed that issue myself in a while, so I'd appreciate patience while I get re-situated!

joshmedeski:
@wez totally understand, good luck with everything! ❤️

PR #7102, opened 2025-07-23
Remove $RUST_BACKTRACE from Nix devShell

cpick:
Make ~two~ one change~s~ to the Nix devShell environment:
~1. Install `lldb`: this allows `rust-lldb` to run properly.~
2. Don't set `$RUST_BACKTRACE`: prevents two backtraces from being printed

env-bootstrap/src/lib.rs's [`register_panic_hook()`](https://github.com/wezterm/wezterm/blob/d1bdb8a350907d080af8914f938d70c9e80cf2bb/env-bootstrap/src/lib.rs#L166-L186) already prints a backtrace unconditionally, setting `RUST_BACKTRACE=1` causes a duplicate to be printed. (CC @happenslol who, I believe, added this setting originally.)

I found ~both of these~ this change~s~ useful when debugging #7081 and #7098 so I think it'll ~they'll~ be generally helpful.

### DIFF
--- a/nix/flake.nix
+++ b/nix/flake.nix
@@ -235,7 +235,6 @@
             ]);
 
           LD_LIBRARY_PATH = libPath;
-          RUST_BACKTRACE = 1;
         };
 
         formatter = pkgs.nixfmt-rfc-style;


happenslol:
Unsetting the env var is fine, in my opinion. I'm not sure if I'd want to include lldb in the flake - that's usually something the user would bring with their IDE or editor config. Would like to hear other opinions, though!

cpick:
> I'm not sure if I'd want to include lldb in the flake - that's usually something the user would bring with their IDE or editor config. Would like to hear other opinions, though!

In my case I use `rust-lldb` from the command line. That script is already installed in the devShell (via one of the rust dependencies), but it depends on the `lldb` executable, which is not yet installed.

(If it matters I use nix-direnv to activate this flake automatically.)

cpick:
> I'm not sure if I'd want to include lldb in the flake

I've thought about this a bit more and I agree: it doesn't make much sense to add a general-purpose debugging tool like `lldb` to WezTerm's devShell.  (If it was somehow built into the repo's tests, scripts, or something else I would feel differently.) I've removed `lldb` from this pull request, reducing the scope to just undefining `$RUST_BACKTRACE`.

Thanks for the feedback @happenslol

happenslol:
Then this is a pretty easy approve from me :+1:

happenslol (approved):

wez:
Thanks!

PR #7131, opened 2025-08-04
Add function to percent encode string

mgpinf:
This is helpful to encode a string in order to be able to pass it as query string.

e.g. Using `InputSelector` to read a line and apply url encoding on the line and append it to a search engine's search url and open the resulting url

### DIFF
--- a/docs/config/lua/wezterm.url/encode.md
+++ b/docs/config/lua/wezterm.url/encode.md
@@ -1,0 +1,8 @@
+# `wezterm.url.encode(string)`
+
+{{since('nightly')}}
+
+Returns the percent encoded version of the provided string.
+This is helpful to encode a string in order to be able to pass
+it as query string.
+
--- a/lua-api-crates/url-funcs/src/lib.rs
+++ b/lua-api-crates/url-funcs/src/lib.rs
@@ -1,7 +1,10 @@
 use crate::mlua::UserDataFields;
 use config::lua::get_or_create_sub_module;
 use config::lua::mlua::{self, Lua, MetaMethod, UserData, UserDataMethods};
-use percent_encoding::percent_decode;
+use percent_encoding::{percent_decode, utf8_percent_encode, AsciiSet, CONTROLS};
+
+/// https://url.spec.whatwg.org/#query-percent-encode-set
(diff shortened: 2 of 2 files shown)


docs/config/lua/wezterm.url/encode.md
```diff
@@ -0,0 +1,8 @@
+# `wezterm.url.encode(string)`
+
+{{since('nightly')}}
+
+Returns the percent encoded version of the provided string.
+This is helpful to encode a string in order to be able to pass
+it as query string.
+
```
bew, line 6: Could you add a small but useful example using this function ?

mgpinf:
@bew I have updated the PR. Can you check?

docs/config/lua/wezterm.url/encode.md
```diff
@@ -0,0 +1,60 @@
...
+    action = wezterm.action_callback(function(window, pane)
+      local state = {}
+      state.search_engine = 'Youtube'
+      window:perform_action(prompt_for_line(state), pane)
+    end),
+  },
+}
+
+return config
+```
```
bew: Here is a slight simplification to your example, using less intermediate functions, uppercase for constants and without 'state' (just pass the search engine!)
````suggestion
```lua
local act = wezterm.action

local SEARCH_ENGINES = {
  ['Google'] = 'https://www.google.com/search?q=',
  ['Youtube'] = 'https://www.youtube.com/results?search_query=',
}

local function get_action_to_search(search_engine)
  return act.PromptInputLine {
    description = 'Search with ' .. search_engine,
    action = wezterm.action_callback(function(_win, _pane, line)
      if line and #line > 0 then
        local final_url = SEARCH_ENGINES[search_engine] .. wezterm.url.encode(line)
        wezterm.open_with(final_url)
      end
    end),
  }
end

config.keys = {
  {
    key = 'g',
    mods = 'CTRL',
    action = get_action_to_search('Google'),
  },
  {
    key = 'y',
    mods = 'CTRL',
    action = get_action_to_search('Youtube'),
  },
}

return config
```
````

bew (approved):
Otherwise it looks good, Wez will probably merge this when he has some availability ;)

PR #7137, opened 2025-08-06
fix: Detect if running in WSL with case insensitive 'microsoft' check

### LINKED ISSUE #7136: wsl2: running_under_wsl() is false

Linux Wayland

In WSL2

### WezTerm version

20250730-195751-6a493f88

### Describe the bug

I tried using wezterm.running_under_wsl() in my wezterm.lua, but its value was false. Instead, I set an environment variable `IN_WSL` in my launch scripts and use that.

### To Reproduce

Ctrl+Shift+L
Type at the prompt:
```
> local wezterm = require 'wezterm'
> wezterm.running_under_wsl()
```
Response: `false`

### Configuration

Works with no configuration

### Expected Behavior

I think it should be true when running under WSL.

### Logs

```
Debug Overlay
wezterm version: 20250730-195751-6a493f88 x86_64-unknown-linux-gnu
Window Environment: Wayland
Lua Version: Lua 5.4
OpenGL: D3D12 (AMD Radeon(TM) Graphics) 4.6 (Compatibility Profile) Mesa 24.2.8-1ubuntu1~24.04.1
Press ESC or CTRL-D to exit
> local wezterm = require 'wezterm'

> wezterm.running_under_wsl()
false
```

### Anything else?

```
% cat /proc/version
Linux version 6.6.87.2-microsoft-standard-WSL2 (root@439a258ad544) (gcc (GCC) 11.2.0, GNU ld (GNU Binutils) 2.37) #1 SMP PREEMPT_DYNAMIC Thu Jun  5 18:30:46 UTC 2025
```
My wsl is fully updated.

bew:
Close #7136

### DIFF
--- a/config/src/version.rs
+++ b/config/src/version.rs
@@ -25,10 +25,14 @@ pub fn running_under_wsl() -> bool {
     unsafe {
         let mut name: libc::utsname = std::mem::zeroed();
         if libc::uname(&mut name) == 0 {
-            let version = std::ffi::CStr::from_ptr(name.version.as_ptr())
-                .to_string_lossy()
-                .into_owned();
-            return version.contains("Microsoft");
+            // 'microsoft' is usually in version, in some cases it can be in release instead
+            // (see #7136)
+            let version = format!(
+                "{} {}",
+                std::ffi::CStr::from_ptr(name.version.as_ptr()).to_string_lossy(),
+                std::ffi::CStr::from_ptr(name.release.as_ptr()).to_string_lossy()
+            );
+            return version.to_lowercase().contains("microsoft");
         }
     };
 


jberkenbilt:
Here's a diff against the PR. The string `microsoft` appears in release, not version. Obviously tweak as needed.

```
diff --git a/config/src/version.rs b/config/src/version.rs
index 62a8ace44..12673bb88 100644
--- a/config/src/version.rs
+++ b/config/src/version.rs
@@ -25,9 +25,11 @@ pub fn running_under_wsl() -> bool {
     unsafe {
         let mut name: libc::utsname = std::mem::zeroed();
         if libc::uname(&mut name) == 0 {
-            let version = std::ffi::CStr::from_ptr(name.version.as_ptr())
-                .to_string_lossy()
-                .into_owned();
+            let version = format!(
+                "{} {}",
+                std::ffi::CStr::from_ptr(name.version.as_ptr()).to_string_lossy(),
+                std::ffi::CStr::from_ptr(name.release.as_ptr()).to_string_lossy()
+            );
             return version.to_lowercase().contains("microsoft");
         }
     };
```

PR #7140, opened 2025-08-07
copy_mode: implement `MoveToBlankLine`

### LINKED ISSUE #7079: Copy mode: Move to next/previous paragraph
I'd like to be able to jump between 'sections' of output that are smaller than "an entire command" (e.g. the semantic prompts feature) but larger than "a single line" and more precise than "a page".

In tmux's copy-mode, you can use `{` and `}` to move between paragraphs (as delimited by blank lines). This is very convenient!

You can add a key binding like this to poorly simulate the behavior:

```lua
config.key_tables.copy_mode = {
  {
    key = "{",
    action = act.Multiple {
      act.Search {
        Regex = "^\\s*$",
      },
      act.CopyMode("PriorMatch"),
      act.CopyMode("AcceptPattern"),
      act.CopyMode("ClearSelectionMode"),
    },
  },
}
```

This has several drawbacks:
1. It highlights all blank lines on the screen.
2. It leaves the search bar open (but not focused) at the bottom of the screen.
3. It destroys the user's current search pattern, if any.
4. It destroys the user's current selection, if any. (As a result, you cannot use paragraph movements to add to a selection.)
5. It is very finicky to figure out the right invocation to use here. (E.g. I tried adding `act.CopyMode("ClearPattern")` to hide the search UI, but that results in the entire key binding turning into a no-op, presumably because leaving the search UI resets the viewport to where it was before...?)

**Additional context**
Just jumping to blank lines would be 95% of the value here, but it would be nice if this feature was extensible, e.g. an action to jump to the next line matching a regex would be really useful!

Related:
- [ ] #5952

9999years:
This is equivalent to the `previous-paragraph` and `next-paragraph` bindings in `tmux`.

Closes #7079

If this looks good, let me know and I can write some tests and docs for it! I'm not sure if we want this to be a more general mechanism like jumping to a line that matches a pattern.

TODO:
- [x] Docs
- [ ] Tests

Usage:

```lua
local act = wezterm.action

return {
  key_tables = {
    copy_mode = {
      {
        key = "{",
        action = act.CopyMode { MoveToBlankLine = "Up" },
      },
      {
        key = "}",
        action = act.CopyMode { MoveToBlankLine = "Down" },
      },
    },
  },
}
```

### DIFF
--- a/mux/src/pane.rs
+++ b/mux/src/pane.rs
@@ -160,6 +160,24 @@ impl LogicalLine {
         }
         (y - 1, x - idx + self.physical_lines.last().unwrap().len())
     }
+
+    /// Determine if this is a single empty line.
+    ///
+    /// To be visually empty, it must contain one physical line containing only whitespace.
+    /// [`Line`]s don't track which cells are empty, so we can't distinguish between
+    /// whitespace and empty cells, so we just look at whitespace. This is also how Wezterm
+    /// deals with whitespace when copying to the clipboard, so this should be fine.
+    pub fn is_visually_empty(&self) -> bool {
+        if self.physical_lines.len() != 1 {
+            false
+        } else {
+            let physical_line = &self.physical_lines[0];
+            physical_line
+                .columns_as_str(0..usize::MAX)
+                .trim_end()
+                .is_empty()
+        }
+    }
 }
 
 /// A Pane represents a view on a terminal
--- a/wezterm-gui/src/overlay/copy.rs
+++ b/wezterm-gui/src/overlay/copy.rs
@@ -3,7 +3,7 @@ use crate::termwindow::keyevent::KeyTableArgs;
 use crate::termwindow::{TermWindow, TermWindowNotif};
 use config::keyassignment::{
     ClipboardCopyDestination, CopyModeAssignment, KeyAssignment, KeyTable, KeyTableEntry,
-    ScrollbackEraseMode, SelectionMode,
+    ScrollDirection, ScrollbackEraseMode, SelectionMode,
 };
 use mux::domain::DomainId;
 use mux::pane::{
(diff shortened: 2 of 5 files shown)


bew:
Hello, thanks for the contribution!

What's your reasoning for allowing any number instead of just `-1`/`1` for a direction?
If you want to allow skipping multiple paragraphs this can already be done using the `Multiple` action if the user wants.

9999years:
@bew No real reason, happy to cut that feature if you'd like.

bew:
I this it'll be easier to make sense of and describe, and it's still easy to add back if needed

9999years:
@bew Done, now it takes a direction parameter: `act.CopyMode { MoveToBlankLine = "Up" }`.

wezterm-gui/src/overlay/copy.rs
```diff
@@ -822,6 +823,38 @@ impl CopyRenderable {
...
+            }
+
+            let lines = self.delegate.get_logical_lines(line_index..line_index + 1);
+
+            if lines.len() == 1 && lines[0].is_empty() {
+                break;
+            }
+        }
+
+        // Note: We jump to the `line_index` even if we didn't find enough lines. This lets you use
```
bew:
```suggestion
        // Note: We jump to the `line_index` even if we didn't find a blank line. This lets you use
```

mux/src/pane.rs
```diff
@@ -160,6 +160,15 @@ impl LogicalLine {
         }
         (y - 1, x - idx + self.physical_lines.last().unwrap().len())
     }
+
+    pub fn is_empty(&self) -> bool {
```
bew: Maybe we could rename this to `is_visually_empty` ?
Also please add a small documentation line describing the function's behavior

bew (commented):
Looking good :+1: 
Few wording comments only

If you're willing I think you can proceed by adding documentation for this new action!

9999years:
Added the requested doc comments, I still need to add docs to the manual + tests. Thanks for the review!

9999years:
@bew Added docs, PTAL :)

PR #7146, opened 2025-08-11
FEATURE: Add metadata for cargo-deb build

oleid:
When running `cargo deb -p waze`, this will produce a debian package below `target/debian`. It can be installed the usual way via `deb -i <path>`


stappersg (approved):
_Nice_

Issue #7117, opened 2025-07-29
"Unrecognized tmux cc line" error for %unlinked-window-renamed (when using multiple sessions)


depds:
### What Operating System(s) are you seeing this problem on?

macOS

### WezTerm version

20250725-073938-dd6e5bd2

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

```
17:10:42.627  INFO   mux::localpane > tmux -CC mode requested
17:10:42.628  INFO   mux::tmux      > tmux session changed:0
17:10:42.719  INFO   mux::tmux      > Tmux create window id 0
17:10:42.862  INFO   mux::tmux_commands > new pane synced, id: 0
17:10:44.256  INFO   mux::tmux_commands > new pane synced, id: 1
17:10:44.348  INFO   mux::tmux_commands > new pane synced, id: 2
17:10:44.448  INFO   mux::tmux_commands > new pane synced, id: 3
17:10:44.613  INFO   mux::tmux_commands > new pane synced, id: 5
17:10:44.657  INFO   mux::tmux_commands > new pane synced, id: 4
17:28:58.565  ERROR  wezterm_escape_parser::tmux_cc > Unrecognized tmux cc line:  --> 1:28
  |
1 | %unlinked-window-renamed @6 bash
  |                            ^---
  |
  = expected EOI
17:29:04.649  ERROR  wezterm_escape_parser::parser  > impossible C0/C1 control code '\u{80}' 0x80 was dropped
17:29:04.649  ERROR  wezterm_escape_parser::parser  > impossible C0/C1 control code '\u{80}' 0x80 was dropped
17:29:04.650  ERROR  wezterm_escape_parser::parser  > impossible C0/C1 control code '\u{80}' 0x80 was dropped
17:29:04.653  INFO   mux                            > domain detached panes: [3, 4, 5, 6, 2, 1]
17:29:04.676  ERROR  wezterm_escape_parser::parser  > impossible C0/C1 control code '\u{80}' 0x80 was dropped
17:29:04.676  ERROR  wezterm_escape_parser::parser  > impossible C0/C1 control code '\u{80}' 0x80 was dropped
17:29:04.676  ERROR  wezterm_escape_parser::parser  > impossible C0/C1 control code '\u{80}' 0x80 was dropped
```

[...]

PR #7148, opened 2025-08-12
fix: tmux -CC "Unrecognized tmux cc line error for %unlinked-window-r…

joexue:
…enamed"

refs: #7117

Fix the “unlinked-window-renamed” event parsing issue


bew (approved):
Looks good, thanks!

(you'll be notified if so)~~

FYI Development/Maintenance on Wezterm is slowly coming back on track, Wez being a bit more available and me having new github powers! 💪 ([ref](https://github.com/wezterm/wezterm/issues/6341#issuecomment-4620396384))

gulp:
The grammar change here makes the trailing name **required**:

```
unlinked_window_renamed = { "%unlinked-window-renamed " ~ window_id ~ " " ~ any_text }
```

`man tmux` (3.6b) documents the id-only form, `%unlinked-window-renamed window-id` with no name, and it contrasts deliberately with `%window-renamed window-id name` a few lines below it. Under this rule a line without a trailing name becomes a parse error.

That is worse than a normal parse failure. `Parser::parse` sets `tmux_state = None` on any error from the tmux_cc parser, which ends control mode for the life of the process; the window stops responding permanently, with no message in the UI. That is the same symptom this PR is fixing, just reached from the other direction.

Making the name optional (the precedent `exit` already sets) accepts both forms:

```
unlinked_window_renamed = { "%unlinked-window-renamed " ~ window_id ~ (" " ~ any_text)? }
```

and in the handler, treat a missing name as an empty string rather than an error.

I have this as a patch at gulp/wezterm@324c380, with three regression tests covering the id-only form, a name containing spaces, and a vis-encoded name, if it is useful to fold in here.

One more thing worth recording, relevant to how the incomplete rule in #6773 reached a release: `wezterm-escape-parser` gates `tmux_cc` behind an off-by-default Cargo feature, so

```
cargo test -p wezterm-escape-parser
```

compiles none of the control-mode parser and still reports ok. Only `termwiz` turns the feature on, via its default features. A name filter matching zero tests also exits 0, so `cargo test -p wezterm-escape-parser tmux` reports ok having run nothing at all. Anyone touching this module wants:

```
cargo test -p wezterm-escape-parser --features tmux_cc,std
```

---

*Investigated and written by Claude Opus 5 (Claude Code), posted from @gulp's account with their review and approval. The findings above were reproduced against a local build; the account owner did not author this text.*

PR #7153, opened 2025-08-14
Fix `mismatched_lifetime_syntaxes` warnings

9999years:
Fixes a bunch of warnings like this:

```
warning: hiding a lifetime that's elided elsewhere is confusing
   --> window/src/os/macos/window.rs:670:23
    |
670 |     fn display_handle(&self) -> Result<DisplayHandle, HandleError> {
    |                       ^^^^^            ------------- the same lifetime is hidden here
    |                       |
    |                       the lifetime is elided here
    |
    = help: the same lifetime is referred to in inconsistent ways, making the signature confusing
    = note: `#[warn(mismatched_lifetime_syntaxes)]` on by default
help: use `'_` for type paths
    |
670 |     fn display_handle(&self) -> Result<DisplayHandle<'_>, HandleError> {
    |                                                     ++++
```

This lint was added in Rust 1.89: https://github.com/rust-lang/rust/issues/141787

### DIFF
--- a/async_ossl/src/lib.rs
+++ b/async_ossl/src/lib.rs
@@ -21,7 +21,7 @@ impl AsyncSslStream {
 
 #[cfg(unix)]
 impl std::os::fd::AsFd for AsyncSslStream {
-    fn as_fd(&self) -> std::os::fd::BorrowedFd {
+    fn as_fd(&self) -> std::os::fd::BorrowedFd<'_> {
         self.s.get_ref().as_fd()
     }
 }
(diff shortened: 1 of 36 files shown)


bew:
Why is it still draft? Are you missing a few warnings or don't know how to resolve them?

9999years:
@bew It does fix all the warnings, I just didn't know if it was wanted + forgot to mark it ready. Thanks!

wez:
Thank you!

Issue #7144, opened 2025-08-11
Cannot input Chinese characters in control mode


zeyugao:
### What Operating System(s) are you seeing this problem on?

macOS

### WezTerm version

wezterm 20250730-195751-6a493f88

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

It can display and input the chinese characters in wezterm when not in control mode or in traditional tmux

<img width="453" height="362" alt="Image" src="https://github.com/user-attachments/assets/4b81f701-fc30-47f5-aa56-3e8fe3116769" />

But it cannot input the chinese characters with ime in control mode

<img width="246" height="44" alt="Image" src="https://github.com/user-attachments/assets/b8cf875a-18a4-4b41-94b6-f58d15c1c1ce" />

https://github.com/user-attachments/assets/23f1b3a6-da40-4d92-842a-d0781dcc2fcc

The display is correct

<img width="223" height="64" alt="Image" src="https://github.com/user-attachments/assets/47dc972f-d443-43eb-8078-c38f2e35f52d" />

Copy chinese characters into the terminal is also not working

### Configuration

no config

joexue:
Since my wezterm gui cannot display chinese characters correctly, could you help to do two tests?
1 just put in one character “你”
2 run “ tmux send-keys -t %{pane id} 你 

Thanks

zeyugao:
here is the recording

https://github.com/user-attachments/assets/adf18fb3-e319-4a91-a998-b16d6cd04b0e

1. one character, “你”, is also broken
2. tmux send-keys -t %{pane id} 你 works fine

joexue:
Thanks for your testing.
It is very interesting. Let me think about it.

chengjianhua:
@joexue hello, when would you merge this change? I ran into the same issue here.

joexue:
> [@joexue](https://github.com/joexue) hello, when would you merge this change? I ran into the same issue here.

I don’t think I have the permission to do so, it’s project owner’s call.

connormclaud:
It would be nice to have https://github.com/wezterm/wezterm/pull/7154 merged.

A lot of unicode simply don't work in tmux CC mode, including cyrilics

bew:
Hello everyone, #7154 is finally merged!

This should be fixed in the next nightly!

PR #7154, opened 2025-08-15
fix: tmux -CC unicode input

### LINKED ISSUE #7162: UTF8 problem with tmux control mode

Windows

_No response_

### WezTerm version

20250713-135109-85c587f9

### Describe the bug

When I use "plain" tmux on a WSL tab, when I type a key which generates an utf-8 unicode characters display correctly.
In tmux command mode they are treated as ansi caracters:
ie:
I press the  "é" key, "Ã©" is printed
Same behavior when I launch tmux remotely, ie: ssh -At host: OK without "-CC", NOK with "-CC"

### To Reproduce

N.B.: To get the "é" character you can press the main keyboard 2 key with a French layout.

A file containing the utf character "é" was previously created , ie;
echo -n é >OK

Launch tmux in command mode:
tmux -CC

If you press "é" you get "Ã©"
Create a file with what is outputted when you press the "é" key.
echo -n Ã©  >NOK

$ cat OK;echo
é
$ cat NOK;echo
Ã©
$ hexdump OK
0000000 a9c3
0000002
$ hexdump NOK
0000000 83c3 a9c2
0000004
(the "OK" file was created during the first step)

I think this is not a character display problem, as a file containing utf8 caracters display correctly

With "plain" tmux I get "é" character everywhere.

### Configuration

no config

### Expected Behavior

Get a "é" when I press the corresponding key.

### Logs

_No response_

### Anything else?

_No response_

joexue:
refs: #7144

Fix the Unicode input issue.

### DIFF
--- a/mux/src/tmux_commands.rs
+++ b/mux/src/tmux_commands.rs
@@ -1024,7 +1024,7 @@ impl TmuxCommand for SendKeys {
         for &byte in self.keys.iter() {
             write!(&mut s, "0x{:X} ", byte).expect("unable to write key");
         }
-        format!("send-keys -t %{} {}\r", self.pane, s)
+        format!("send-keys -H -t %{} {}\r", self.pane, s)
     }
 
     fn process_result(&self, domain_id: DomainId, result: &Guarded) -> anyhow::Result<()> {


bew (approved):

bew:
For reference, from `man tmux` (3.6a)

```
send-keys [-FHKlMRX] [-c target-client] [-N repeat-count] [-t target-pane] [key ...]
        (alias: send)
    Send a key or keys to a window or client.  Each argument key is the name of the key (such as ‘C-a’ or ‘NPage’) to send; if the string is not recognised as a key, it is sent as a series of characters.  If -K is given, keys are sent to target-client, so they
    are looked up in the client's key table, rather than to target-pane.  All arguments are sent sequentially from first to last.  If no keys are given and the command is bound to a key, then that key is used.

    The -l flag disables key name lookup and processes the keys as literal UTF-8 characters.  The -H flag expects each key to be a hexadecimal number for an ASCII character.

    The -R flag causes the terminal state to be reset.

    -M passes through a mouse event (only valid if bound to a mouse key binding, see MOUSE SUPPORT).

    -X is used to send a command into copy mode - see the WINDOWS AND PANES section.  -N specifies a repeat count and -F expands formats in arguments where appropriate.
```
Especially: `The -H flag expects each key to be a hexadecimal number for an ASCII character.`

bew (approved):

PR #7178, opened 2025-08-28
pty: Fix typo psuedo => pseudo

fasterthanlime:
Title says it all.

### DIFF
--- a/pty/src/lib.rs
+++ b/pty/src/lib.rs
@@ -1,5 +1,5 @@
 //! This crate provides a cross platform API for working with the
-//! psuedo terminal (pty) interfaces provided by the system.
+//! pseudo terminal (pty) interfaces provided by the system.
 //! Unlike other crates in this space, this crate provides a set
 //! of traits that allow selecting from different implementations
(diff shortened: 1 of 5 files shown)


bew (approved):
Lowl, nice one 😅

Wez isn't very responsive at the moment while he's moving to another country and settling, but it'll be get merged some day!

bew:
@wez I don't think this one is easy to merge since it changes types/module names and so would be a breaking change 🤔

I let you merge it / handle it

wez:
I'm not sure if these are actually exported from the crate.  I don't see them in the docs, but I'm not sure how much divergence we have there from what's currently in `main`.  If they're not public then we can merge this at will.
However, if they are public, then we'd need to do a semver bump in the Cargo.toml as part of this change.

bew:
> I'm not sure if these are actually exported from the crate.

Indeed they aren't, I think the `pub` in the `pseudocon` module lead me in error, it's only public for that private module, not public for the crate..!

Let's merge (finally!) once CI is green!

PR #7195, opened 2025-09-03
Fix mermaid newlines in what-is-a-terminal.md

Ricky-Blackman:
Mermaid doesn't convert `\n` to newlines; you need to use literal new lines in the diagram. Fortunately, mermaid *does* ignore (some?) whitespace, so you don't have to ruin your lovely indentation.

For example, this is incorrect:
~~~
```mermaid
flowchart LR
    A["First line\nSecond line"]
```
~~~

```mermaid
flowchart LR
    A["First line\nSecond line"]
```

I have converted most cases of this to something more like the following:
~~~
```mermaid
flowchart LR
    A["
        First line
        Second line
    "]
```
~~~
```mermaid
flowchart LR
    A["
        First line
        Second line
    "]
```
The leading and trailing newlines, as well as the leading spaces, seem to be ignored by mermaid.

I did say "most" before; the exception is the `input\noutput` labels that I changed to `input/output` so they'd match the rest of the page.

There may be more instances of this across the documentation; I only noticed this one because I was looking at it.

### DIFF
--- a/docs/what-is-a-terminal.md
+++ b/docs/what-is-a-terminal.md
@@ -57,13 +57,16 @@ kernel, which manages the actual communication with the terminal.
 flowchart LR
     subgraph Kernel
         direction LR
-        TTY["TTY device\n(e.g. /dev/tty/0)"] 
+        TTY["
+            TTY device
+            (e.g. /dev/tty/0)
+        "] 
     end
     
     subgraph Userspace
-    SHELL["Shell Program (e.g. zsh)"] <-- "input\noutput" --> TTY
+    SHELL["Shell Program (e.g. zsh)"] <-- "input/output" --> TTY
     end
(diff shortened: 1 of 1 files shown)


bew (approved):

bew:
Note that using `<br>` instead of plain multiline labels also work and avoid making the source mermaid code much larger with more lines (subjective though..).

---

> With traditional strings, you needed to add <br> tags for text to wrap in nodes.

REF: https://mermaid.js.org/syntax/flowchart.html#markdown-strings

<table>

<tbody>
  <tr>
    <td>

````
```mermaid
flowchart LR
    A["First line<br>Second line"]
```
````

</td>
    <td>

```mermaid
flowchart LR
    A["First line<br>Second line"]
```

</td>

</tr>
</tbody>
</table>

Issue #7197, opened 2025-09-04
Left status bar overlaps with window buttons with `window_decorations = "INTEGRATED_BUTTONS|RESIZE"`


cppbear:
### What Operating System(s) are you seeing this problem on?

macOS

### WezTerm version

20250730-195751-6a493f88

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

When using WezTerm on macOS with `config.window_decorations = "INTEGRATED_BUTTONS|RESIZE"`, the left status bar overlaps with the native macOS window control buttons (close, minimize, maximize). Additionally, there appears to be an unnecessarily large gap between the left status bar content and the first tab. This issue makes the status bar content partially unreadable and affects the overall UI aesthetics.

### To Reproduce

1. Set `window_decorations = "INTEGRATED_BUTTONS|RESIZE"`.
2. Use `window:set_left_status` to set left status bar.
3. Launch WezTerm on macOS.

### Configuration

```lua
local wezterm = require("wezterm")
local config = wezterm.config_builder()

config.window_decorations = "INTEGRATED_BUTTONS|RESIZE"

wezterm.on("update-status", function(window, pane)
    local workspace = window:active_workspace()
    local cells = {
        { Text = workspace },
    }

    window:set_left_status(wezterm.format(cells))
end)

return config

```

### Expected Behavior

The left status bar should be correctly positioned, respecting the space taken by the macOS window control buttons, and should have a reasonable, consistent padding from the first tab.

### Anything else?

<img width="520" height="90" alt="Image" src="https://github.com/user-attachments/assets/787ca629-25fd-475d-9759-979c15cdb0d2" />

alberti42:
I created a PR #7199 that fixes the issue. It would be great if you could test it. Thanks!

cppbear:
Hello @alberti42 ,

I have tested the code from PR #7199 locally by compiling and running it, and I can confirm that the issue is now resolved. This is exactly what I was hoping for!

Attached is a screenshot demonstrating the fix.

<img width="1150" height="102" alt="Image" src="https://github.com/user-attachments/assets/2f91d951-c83e-4eca-bd06-1aed8db7b73a" />

Thank you for the swift fix and excellent work!

PR #7199, opened 2025-09-05
Fix: left status bar overlaps with window buttons (#7197)

alberti42:
#### Problem

On macOS, the left status bar content would render underneath the OS "traffic light" window controls when using integrated title bar buttons (`window_decorations = "INTEGRATED_BUTTONS"`).

This issue affected both the fancy and classic (text-based) tab bars, which relied on incorrect, hardcoded margins (`FIXME`s in the code) that were insufficient to clear the button area.

This bug is related to issue #7197, which this PR completely resolves.

#### Solution

This PR resolves the issue by replacing the hardcoded values with a robust, dynamic calculation of the button area's width. The fix is applied consistently to both tab bar renderers.

The changes are broken down as follows:

**1. Dynamic Width Calculation in the Windowing Layer**
* The `get_os_parameters` function for macOS now uses FFI to call native AppKit APIs (`standardWindowButton:`).
* It queries the `frame` of the native window buttons to compute the exact physical pixel width required for padding, correctly handling Retina (high-DPI) displays.
* This logic is now centralized in the windowing layer, providing a clean abstraction for the UI layer to consume.

**2. Refactored Fancy Tab Bar Renderer (`use_fancy_tab_bar = true`)**
* The fancy tab bar renderer now consumes the dynamic pixel value from `os_parameters`.
* It uses `Dimension::Pixels` to create a correctly sized spacer element, avoiding a "double scaling" issue and ensuring pixel-perfect alignment.
* Redundant hardcoded padding was removed to establish a single source of truth for the margin.

**3. Updated Classic Tab Bar Renderer (`use_fancy_tab_bar = false`)**
* The classic (text-based) tab bar renderer was also updated to use the dynamic pixel value.
* It now calculates the correct number of padding *cells* by dividing the required pixel width by the width of a single character cell.
* This makes the padding for the classic bar fully dynamic and accurate, adapting correctly to different font sizes.

The final result is a clean, pixel-perfect alignment of the tab bar content for both tab bar styles.

#### How to Test

**Test Case 1: Fancy Tab Bar**
1. On a macOS system, configure `wezterm.lua` with:
   ```lua
   config.use_fancy_tab_bar = true
   config.window_decorations = "INTEGRATED_BUTTONS"
   ```
2. Add a left status bar element using the `update-status` event.
   ```lua
   wezterm.on("update-status", function(window, pane)
    window:set_left_status(wezterm.format {
      { Text = 'My status bar text' },
    })
   end)
   ```
4. Launch WezTerm and confirm the left status text is positioned correctly.
5. Enter and exit fullscreen to confirm the layout remains correct.

**Test Case 2: Classic Tab Bar**
1. On a macOS system, configure `wezterm.lua` with:
   ```lua
   config.window_decorations = "INTEGRATED_BUTTONS"
   ```
2. Add a left status bar element using the `update-status` event.
3. Launch WezTerm and confirm the left status text is positioned correctly.
4. Enter and exit fullscreen to confirm the layout remains correct.

### DIFF
--- a/window/src/os/macos/window.rs
+++ b/window/src/os/macos/window.rs
@@ -19,7 +19,7 @@ use cocoa::appkit::{
     self, CGFloat, NSApplication, NSApplicationActivateIgnoringOtherApps,
     NSApplicationPresentationOptions, NSBackingStoreBuffered, NSEvent, NSEventModifierFlags,
     NSOpenGLContext, NSOpenGLPixelFormat, NSPasteboard, NSRunningApplication, NSScreen, NSView,
-    NSViewHeightSizable, NSViewWidthSizable, NSWindow, NSWindowStyleMask,
+    NSViewHeightSizable, NSViewWidthSizable, NSWindow, NSWindowButton, NSWindowStyleMask,
 };
 use cocoa::base::*;
 use cocoa::foundation::{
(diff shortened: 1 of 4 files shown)


window/src/os/macos/window.rs
```diff
@@ -920,9 +920,47 @@ impl WindowOps for Window {
...
+            // This unsafe block calls the native macOS APIs
+            unsafe {
+                let window_id: id = self.ns_window;
+
+                // 1. Get each of the three "traffic light" buttons
+                let close_button: id =
+                    msg_send![window_id, standardWindowButton: NSWindowButton::NSWindowCloseButton];
+                let zoom_button: id =
+                    msg_send![window_id, standardWindowButton: NSWindowButton::NSWindowZoomButton];
+
```
bew: hum, you say 'three' but you only get two buttons.. Did you miss the minimize button?

window/src/os/macos/window.rs
```diff
@@ -920,9 +920,47 @@ impl WindowOps for Window {
...
+                // 2. If we can't find the buttons (eg: fullscreen), fall back to a safe value.
+                if close_button == nil || zoom_button == nil {
+                    64.0 // Fallback value
+                } else {
+                    // 3. Get the frame (position and size) of the leftmost and rightmost buttons.
+                    let zoom_frame: NSRect = msg_send![zoom_button, frame];
```
bew: comment is wrong, you only get the frame for the rightmost one 🤔

PR #7200, opened 2025-09-05
feat(terminal): Implement dynamic DECRQSS and DECRQM cursor queries

alberti42:
**Description:**

This PR implements the correct handling for terminal cursor queries to ensure they reflect the true state of the cursor, including user-configured defaults and runtime changes.

Specifically, this addresses three issues:
1.  **DECRQSS Cursor Shape Query (`DCS $q "q" ST`)**: Previously, this was unrecognized and returned a failure. It is now correctly handled, reporting the cursor's current shape.
2.  **DECRQM Blinking Cursor Query (`CSI ?12$p`)**: Previously, this was hardcoded to always report "not blinking". It now correctly reports the cursor's actual blinking status.
3.  **DECSM Blinking Cursor Set/Reset (`CSI ?12h`/`l`)**: Adds support for applications to toggle the cursor between its blinking and steady variants at runtime, overriding the configured default for the current session.

The implementation follows the established architectural pattern of dispatching actions from the `Performer` to `perform_*` methods in `TerminalState` for clean separation of concerns.

**Follow-up Required:** The logic for resolving `CursorShape::Default` is currently commented out in both new query handlers. It needs to be connected to the main `wezterm_config` to dynamically query the user's `default_cursor_style`, especially to handle cases where the configuration is reloaded at runtime. A core developer with knowledge of the configuration-passing mechanism is best suited to complete this final step.

### DIFF
--- a/term/src/terminalstate/mod.rs
+++ b/term/src/terminalstate/mod.rs
@@ -20,9 +20,9 @@ use wezterm_cell::UnicodeVersion;
 use wezterm_escape_parser::csi::{
     Cursor, CursorStyle, DecPrivateMode, DecPrivateModeCode, Device, Edit, EraseInDisplay,
     EraseInLine, Mode, Sgr, TabulationClear, TerminalMode, TerminalModeCode, Window, XtSmGraphics,
-    XtSmGraphicsAction, XtSmGraphicsItem, XtSmGraphicsStatus, XtermKeyModifierResource,
+    XtSmGraphicsAction, XtSmGraphicsItem, XtSmGraphicsStatus, XtermKeyModifierResource
 };
-use wezterm_escape_parser::{OneBased, OperatingSystemCommand, CSI};
+use wezterm_escape_parser::{OneBased, OperatingSystemCommand, ShortDeviceControl, CSI};
 use wezterm_surface::{CursorShape, CursorVisibility, SequenceNo};
 
 mod image;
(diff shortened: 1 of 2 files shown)


term/src/terminalstate/mod.rs
```diff
@@ -1431,17 +1431,52 @@ impl TerminalState {
...
-            | Mode::ResetDecPrivateMode(DecPrivateMode::Code(
-                DecPrivateModeCode::StartBlinkingCursor,
-            )) => {}
-            Mode::QueryDecPrivateMode(DecPrivateMode::Code(
-                DecPrivateModeCode::StartBlinkingCursor,
-            )) => {
-                self.decqrm_response(mode, true, false);
-            }
+			// --> START OF MODIFICATION <--
+	        Mode::SetDecPrivateMode(DecPrivateMode::Code(
```
bew: Remove that `START OF MODIFICATION` comment

Also you seem to have a mix of tabs & spaces for indentation in all of your changes
:point_right: Please always use spaces (or use `rustfmt` to do that automatically)

term/src/terminalstate/mod.rs
```diff
@@ -1431,17 +1431,52 @@ impl TerminalState {
...
+	                CursorShape::BlinkingBar => CursorShape::SteadyBar,
+	                // If it's already steady, do nothing
+	                _ => self.cursor.shape,
+	            };
+	        }
+
+	        Mode::QueryDecPrivateMode(DecPrivateMode::Code(
+	            DecPrivateModeCode::StartBlinkingCursor,
+	        )) => {
+	            // This part remains the same as our previous fix
```
bew: unnecessary comment :thinking:

term/src/terminalstate/performer.rs
```diff
@@ -289,77 +289,22 @@ impl<'a> Performer<'a> {
...
+	                    self.state.perform_decrqss(s);
+	                }
+	                _ => {
+	                    if self.config.log_unknown_escape_sequences() {
+	                        log::warn!("unhandled {:?}", s);
+	                    }
+	                }
+	            }
+	        }
+	        _ => { /* ... existing logic ... */ }
```
bew: The LLM doesn't do everything for you here :grimacing: 

please restore the warning log as before!

term/src/terminalstate/performer.rs
```diff
@@ -289,77 +289,22 @@ impl<'a> Performer<'a> {
...
-                        log::warn!("unhandled {:?}", ctrl);
-                    }
-                }
-            },
-        }
+          match &ctrl {
+	        DeviceControlMode::ShortDeviceControl(s) => {
+	            match (s.byte, s.intermediates.as_slice()) {
+	                (b'q', &[b'$']) => {
+	                    // This is now clean and consistent with the rest of the code
```
bew: unnecessary comment..

term/src/terminalstate/mod.rs
```diff
@@ -2595,6 +2630,69 @@ impl TerminalState {
...
+                // DECSCL - select conformance level
+                write!(self.writer, "{}1$r65;1\"p{}", DCS, ST).ok();
+            }
+            &[b' ', b'q'] => {
+                // DECSCUSR - cursor style query (xterm extension)
+                let shape = self.cursor.shape;
+
+                // if shape == CursorShape::Default {
+                //     // Query the live configuration via the trait
+                //     shape = wezterm.config.default_cursor_style().into();
```
bew, line 2652: To access the config you should be able to simply do `self.config`

bew (changes requested):
A few comments.

I'm wondering, how did you find out about this thing needing changes?
Are you working on or using a project using these feature?

---
Also, I believe you used an AI/LLM to generate some (if not all?) of these changes.
:point_right: I think you should say it EXPLICITLY at the start or end of the PR's description, mentioning what part (code/docs/PR) it was used for.
Also mention how much review & manual testing you did yourself.

alberti42:
Hi Benoit,

Thank for the review! I am sorry; I feel embarrassed for the sloppy PR and the remnants from chatGPT/Gemini. I don't recall which one I was using a few days ago.
 
> I'm wondering, how did you find out about this thing needing changes? Are you working on or using a project using these feature?

The whole thing started from a PR for yazi (https://github.com/sxyazi/yazi/pull/3118). `yazi` and supposedly other popular software like `neovim` query the cursor state, change it in the alternate screen, and then restore it back to the value they queried.

Originally I made a patch to `yazi` to work around limitations in `tmux`. But then we decided with the developer of `yazi` that it was better to patch `tmux` rather than creating a workaround. This is why I filed a PR to `tmux` (https://github.com/tmux/tmux/pull/4618), which is now accepted in upstream.

Since I read in the past days quite a many old pages of documentation about the terminal queries, I thought I could use this fresh knowledge to implement it as well for `wezterm`. I personally would not need it because I run `tmux` in `wezterm`, but I thought it may be useful to other people running pure `wezterm`.
 
> Also, I believe you used an AI/LLM to generate some (if not all?) of these changes. 👉 I think you should say it EXPLICITLY at the start or end of the PR's description, mentioning what part (code/docs/PR) it was used for. Also mention how much review & manual testing you did yourself.

Sorry, normally I use LLM but do thorough testing and make sure I understand each line. Also, normally code is not just copied and paste without further modification / extensive refactoring from the LLM first version. I say normally because this example clearly proves the opposite.

I also use LLM to write PR description. I start from a short draft in bullet points and showing a diff of the code, and let the LLM write a more human-friendly version. I personally find it much more helpful than cryptic comments, often only understandable by few people expert on the subject.

I will mention it explicitly in the future; though I am wondering whether there are still people out there avoiding using LLM. In my view, they should not. In my experience, iterating with LLM takes sometimes even more time because of the numerous iterations, but the the final result is of higher quality, certainly when writing plain English.

PS: This text was written without LLM ;-)

alberti42:
Hey Benoit, I made a new commit that should resolve all issues you raised. If you want to test the changes, I wrote when working on the patch for `tmux` a small bash utility (https://github.com/alberti42/term_query) for querying the cursor shape. I am sure there are tons out there that do the same, but in case you do not know any better, you can use this one.

PS: I think I know what happened with the LLM. I thought it was an easy fix and small lines of code were required, so I lowered my guard. I checked the original chat. I had asked gemini to refactor the code from `performer.rs` into a new function `perform_decrqss` to be inserted in `mod.rs` to parallel `perform_csi_cursor` function in the same file. Because it was supposed to be an easy fix, I literally copied & paste everything without carefully checking—which is always bad—as proven here.

The only thing I am not able to fix is related to your comment:

> To access the config you should be able to simply do `self.config`

`self.config` is an instance of `wezterm_term::terminalstate::TerminalState`, and not of `config::Config`, which exposes:

```rust
pub default_cursor_style: DefaultCursorStyle,
```

The LLM was suggesting to add a trait and expose `default_cursor_style()`, so that I could reach it from `perform_decrqss` but I did not want to do it without first consulting with someone more knowledgeable of the inner structure of `wezterm`. This is why I left this part non-implemented yet.

#### Rationale: Why do we need `default_cursor_style()`?

When the cursor style is not initialized explicitly by the user yet, `wezterm` uses default values.

When querying:

```bash
echo -ne '\e[0 q' # set the cursor style to default value
echo -ne '\eP$q q\e\\' # query the cursor style
```

we could in principle return: `1$r0 q` when the cursor style is set to the default value. However, this is not what xterm does. Generally, xterm is considered the gold standard for what concerns the expected reply.

Moreover, when you probe the blinking state alone with:

```bash
 echo -ne '\e[?12$p'
```

there is no default value for the blinking state like there is for the cursor shape. This forces the terminal emulator to always choose between blinking or not blinking.

What xterm does in both cases (cursor style and blinking state) is to always report what the user sees on the screen. To achieve this, it is necessary to access the default cursor style used by `wezterm` when the cursor style is not explicitly set by the user or when the user restores it to the default value.

PR #7202, opened 2025-09-05
fix(units): allow GuiPosition to parse pt and cell units

alberti42:
#### Rationale
The `Dimension` type in wezterm already supports parsing and serializing four unit suffixes:

- `px` → pixels  
- `pt` → points (72 pt = 1 inch)  
- `%` → percentage  
- `cell` → multiples of font cell size  

However, `GuiPosition::from_str` had a custom `parse_dim` function that only recognized `px`, `%`, and bare numbers (pixels). This meant that when `GuiPosition` was parsed from strings such as `"10pt,20pt"`, it would fail to parse and silently fall back to interpreting values as pixels.

This discrepancy caused issues in user configs where window positions were provided in logical **points** rather than **pixels**. For example, on HiDPI screens, the window would not appear centered because points were treated as pixels.

#### Change
- Refactored `GuiPosition::parse_dim` to reuse the existing robust parser in `DefaultUnit::Pixels.from_dynamic_impl`, ensuring full support for `px`, `pt`, `%`, and `cell`.

#### Example
Before this change, the following Lua snippet did **not** properly center the wezterm window when using `"pt"`:

```lua
wezterm.on("gui-startup", function(cmd)
  local screen = wezterm.gui.screens().active
  local ratio_x = 0.70
  local ratio_y = 0.70

  local width_pt, height_pt = screen.width * ratio_x, screen.height * ratio_y

  local x_pos_pt = (screen.width - width_pt) / 2
  local y_pos_pt = (screen.height - height_pt) / 2

  local tab, pane, window = wezterm.mux.spawn_window {
    position = {
      x = tostring(x_pos_pt) .. "pt",
      y = tostring(y_pos_pt) .. "pt",
      origin = 'ActiveScreen'
    }
  }

  window:gui_window():set_inner_size(width_pt, height_pt)
end)
```

- **Before:** `"123pt"` was rejected and treated as pixels, so the window was misaligned.  
- **After:** `"123pt"` is correctly parsed as `Dimension::Points`, converted to pixels using screen DPI, and the window centers properly on all monitors.

#### Outcome
- `GuiPosition` parsing is now consistent with other `Dimension` usages in wezterm.  
- Users can safely use `"pt"` and `"cell"` in their configs for positioning windows.  

#### Future steps

- Consider adding tests for `pt` and `cell` suffixes to guarantee consistent parsing.
- Adjust documentation to reflect all supported units


bew:
Hello!
I'm surprised that the change in this PR give the result you're describing..

The PR changes `parse_dim`, which is used in `parse_x_y`, which is used in `GuiPosition::from_str`.
But as far as I can see, `GuiPosition::from_str` is _NOT_ used at all when converting from a Lua value to a Rust' `GuiPosition`..

The derive `FromDynamic` on the struct uses the fields attribute `#[dynamic(try_from = "crate::units::PixelUnit")]`, which in turns already uses `DefaultUnit::Pixels.from_dynamic_impl`.. :
https://github.com/wezterm/wezterm/blob/e95b37134a3e3f19dd9cb2f88d6875c66614c3db/config/src/units.rs#L34-L41

So I think this PR should be closed, your config issue is elsewhere..

Happy to re-open if you can explain how this fixes your config 🤔

alberti42:
I feel sorry that you closed the PR without even discussing the issue. The bug was real but it was many months ago.

I think the description of the bug was clear. Could you not reproduce it? But perhaps you are right that I did not fix it at the correct level in the code. You are much more knowledgeable than me.

I also fully understand you need to do triage to find out what PR to prioritize.

Thanks for your work!

bew:
You’re right, I closed this a bit too quickly 😬, and I’m sorry I did so without giving you space to clarify or for us to find a solution together..
But we can continue talking, the PR being closed shouldn't prevent discussion from happening 🙃

So I did a bit of digging, and using this diff on current `main`:
```diff
diff --git a/lua-api-crates/mux/src/lib.rs b/lua-api-crates/mux/src/lib.rs
index b02db3a90..01ae9b707 100644
--- a/lua-api-crates/mux/src/lib.rs
+++ b/lua-api-crates/mux/src/lib.rs
@@ -241,6 +241,7 @@ impl SpawnWindow {
             _ => config::configuration().initial_size(0, None),
         };
 
+        println!("Spawning window at position: {:#?}", self.position);
         let (cmd_builder, cwd) = self.cmd_builder.to_command_builder();
         let (tab, pane, window_id) = mux
             .spawn_tab_or_window(
diff --git a/window/src/connection.rs b/window/src/connection.rs
index 9564d4999..3a88a5dc0 100644
--- a/window/src/connection.rs
+++ b/window/src/connection.rs
@@ -82,6 +82,7 @@ pub trait ConnectionOps {
     }
 
     fn resolve_geometry(&self, geometry: RequestedWindowGeometry) -> ResolvedGeometry {
+        let origin = geometry.origin.clone();
         let bounds = match self.screens() {
             Ok(screens) => {
                 log::trace!("{screens:?}");
@@ -127,6 +128,8 @@ pub trait ConnectionOps {
             .y
             .map(|y| y.evaluate_as_pixels(height_context) as i32 + bounds.origin.y as i32);
 
+        println!("resolve_geometry: origin={:?} x={:?} y={:?} -> px x={:?} y={:?}", origin, geometry.x, geometry.y, x, y);
+
         ResolvedGeometry {
             x,
             y,
```
and running your example config:
```lua
local wezterm = require"wezterm"

wezterm.on("gui-startup", function(cmd)
  local screen = wezterm.gui.screens().active
  local ratio_x = 0.70
  local ratio_y = 0.70

  local width_pt, height_pt = screen.width * ratio_x, screen.height * ratio_y

  local x_pos_pt = (screen.width - width_pt) / 2
  local y_pos_pt = (screen.height - height_pt) / 2

  print(x_pos_pt, y_pos_pt)

  local tab, pane, window = wezterm.mux.spawn_window {
    position = {
      x = tostring(x_pos_pt) .. "pt",
      y = tostring(y_pos_pt) .. "pt",
      origin = "ActiveScreen"
    }
  }

  window:gui_window():set_inner_size(width_pt, height_pt)
end)

return {}
```

Shows that the position is correctly converted to Points:
```
12:22:40.435  INFO   logging > lua: 691.2 388.80000000000007
Spawning window at position: Some(
    GuiPosition {
        x: Points(
            691.2,
        ),
        y: Points(
            388.8,
        ),
        origin: ActiveScreen,
    },
)
resolve_geometry: origin=ActiveScreen x=Some(Points(691.2)) y=Some(Points(388.8)) -> px x=Some(650) y=Some(2741)
```
(running on MacOS) 🤔

PR #7216, opened 2025-09-11
Implement get_semantic_zones() for ClientPane

### LINKED ISSUE #5141: Jump to semantic regions not working through non-local domains

Linux X11

_No response_

### WezTerm version

wezterm 20240102-011725-2f7ca41a

### Describe the bug

In non-local domains (i.e., ssh or mux domains, at least), jumping with semantic regions is not working. The jumps simply do nothing.

### To Reproduce

1. Enable semantic region in shell and wezterm
2. Start mux domain with e.g. unix domain
3. Goto copy mode and jump

### Configuration

Since by default moving by zone type is not accessible from the command palette: 
```lua
local act = wezterm.action
return {
    key_tables = {
        copy_mode = {
            { key = '(', mods = 'SHIFT', action = act.CopyMode { MoveBackwardZoneOfType = 'Prompt' } },
            { key = ')', mods = 'SHIFT', action = act.CopyMode { MoveForwardZoneOfType = 'Prompt' } },
        }
    }
}
```

### Expected Behavior

_No response_

### Logs

_No response_

### Anything else?

_No response_

Osse:
By adding a new request and response we can query the remote terminal
for its semantic zones. Thus handy features like the ScrollToPrompt and
CopyMode actions work as expected in other domains than the local
domain.

Fixes #5141

---------------------------------------------------

I implemented this fairly quickly buy just basing it on the patterns in
surrouding code, and it seems to work exactly as I had hoped. Some
concerns/points:

- As far as I can tell the RPC stuff needs to be `async` functions while
  `get_semantic_zones()` from the `Pane` trait is not. I worked around this by
  using `smol::block_on()`. I see that is used many places in the code base but I'm
  not sure it's acceptable here. An alternative is to make the function `async`
  in the trait, but that seemed messy for other reasons.
- I couldn't find any places to add a test for this.
- There is no new documentation. Imo. that's justified as I am making an
  existing feature work in more contexts which the existing documentation did
  not clarify that it _didn't_ work.

### DIFF
--- a/codec/src/lib.rs
+++ b/codec/src/lib.rs
@@ -34,7 +34,7 @@ use termwiz::image::{ImageData, TextureCoordinate};
 use termwiz::surface::{Line, SequenceNo};
 use thiserror::Error;
 use wezterm_term::color::ColorPalette;
-use wezterm_term::{Alert, ClipboardSelection, StableRowIndex, TerminalSize};
+use wezterm_term::{Alert, ClipboardSelection, SemanticZone, StableRowIndex, TerminalSize};
 
 #[derive(Error, Debug)]
 #[error("Corrupt Response: {0}")]
@@ -441,7 +441,7 @@ macro_rules! pdu {
 /// The overall version of the codec.
 /// This must be bumped when backwards incompatible changes
 /// are made to the types and protocol.
-pub const CODEC_VERSION: usize = 45;
+pub const CODEC_VERSION: usize = 46;
 
 // Defines the Pdu enum.
 // Each struct has an explicit identifying number.
@@ -502,6 +502,8 @@ pdu! {
     GetPaneDirection: 60,
     GetPaneDirectionResponse: 61,
     AdjustPaneSize: 62,
+    SemanticZonesRequest: 63,
+    SemanticZonesResponse: 64,
 }
 
 impl Pdu {
@@ -1113,6 +1115,16 @@ pub struct GetLinesResponse {
 pub struct EraseScrollbackRequest {
     pub pane_id: PaneId,
     pub erase_mode: ScrollbackEraseMode,
+}
+
+#[derive(Deserialize, Serialize, PartialEq, Debug)]
+pub struct SemanticZonesRequest {
+    pub pane_id: PaneId,
+}
+
+#[derive(Deserialize, Serialize, PartialEq, Debug)]
+pub struct SemanticZonesResponse {
+    pub results: Vec<SemanticZone>,
 }
 
 #[derive(Deserialize, Serialize, PartialEq, Debug)]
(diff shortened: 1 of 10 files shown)


codec/src/lib.rs
```diff
@@ -502,6 +502,8 @@ pdu! {
     GetPaneDirection: 60,
     GetPaneDirectionResponse: 61,
     AdjustPaneSize: 62,
+    SemanticZonesRequest: 63,
```
wez, line 506: Please increment `CODEC_VERSION` (on line 444 above this) when adding new PDUs; that will enable the client to realize when the server doesn't support the PDU and error out and inform the user.

wezterm-client/src/pane/clientpane.rs
```diff
@@ -425,6 +425,20 @@ impl Pane for ClientPane {
         Ok(())
     }
 
+    fn get_semantic_zones(&self) -> anyhow::Result<Vec<SemanticZone>> {
+        match smol::block_on(async {
```
wez: we must not use block_on here, as it will block the GUI thread that may be calling into this code, and can lead to deadlock in various scenarios

The proper fix would be to change the signature of `get_semantic_zones` in the trait definition in `mux/src/pane.rs` to be async, then to address the fan-out from that to add appropriate awaits in the all the places that the compiler will identify

wez (changes requested):
Thanks for diving in!
I think the approach looks fine, but we have to implement this without using `block_on`.
I think you can probably get quite far (maybe even 100%) with making this async just by changing the signature of the trait method and seeing what the compiler complains about; likely just then adding some `.awaits`.

Could you give that a try and update the PR?
Thanks!

codec/src/lib.rs
Osse: Done!

wezterm-client/src/pane/clientpane.rs
Osse: I suspected this was the case. I pushed 0cdf12ab5 to make it async, but I ended up adding even more `smol::block_on()` because of it. It's WIP. Hopefully I can spend some more time on it today.

Osse: Made some progress and have pushed 2452d2764, replacing the previous one. I assume this will all be squashed in the end (either by you or me). This just shows the changes needed from my first proposal to making it async.

There are three `smol::block_on()` calls left that I don't know what to do with. Looking at the nearby code `promise::spawn::spawn(...).detach()` is used a lot. It seems  to be used in cases where you just "fire and forget" something, which doesn't work in this case I believe.

Osse:
I have been using Wezterm with this patch for a few months now and everything seems to work very well. I have not noticed any regressions, locally nor over SSH. I took it upon myself to rebase and squash the changes as I mentioned above.

I haven't had much time to improve the patch, though, and there are still a few `block_on()` that I don't know how to get rid of.

PR #7266, opened 2025-09-29
termwiz: make ScreenSize and Blocking serde serializable

quark-zju:
Respect the use_serde feature widely used elsewhere.


wez:
Thank you!

PR #7274, opened 2025-10-01
fix: missing `equals` cases in check condition.

I-Info:
Currently, the boundary conditions in `extend_with()` are inconsistent with those in `need_more_quads()`:
https://github.com/wezterm/wezterm/blob/bf9a2aeebacec19fd07b55234d626f006b22d369/wezterm-gui/src/renderstate.rs#L309-L324
https://github.com/wezterm/wezterm/blob/bf9a2aeebacec19fd07b55234d626f006b22d369/wezterm-gui/src/renderstate.rs#L386-L393

Note that there is an unhandled "equals" edge case here.

A straightforward demonstration is to keep typing characters into the terminal, and when the number of vertices is exactly equal to the buffer size, the cursor will disappear (cursor vertices is not copied to the buffer). The cursor will not return until more characters are entered, forcing the number of vertices to exceed the buffer size.

### DIFF
--- a/.github/workflows/gen_windows.yml
+++ b/.github/workflows/gen_windows.yml
@@ -18,7 +18,7 @@ on:
 
 jobs:
   build:
-    runs-on: "windows-latest"
+    runs-on: "windows-2022"
     
     env:
       CARGO_INCREMENTAL: "0"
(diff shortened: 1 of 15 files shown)


wez:
Thanks!

PR #7276, opened 2025-10-02
chore: bump thiserror to 2.0

tisonkun:


tisonkun:
cc @wez

wez:
What's the motivation for this change? I typically do dependency updates like this as part of a larger sweep. This sort of change has no visible effect for wezterm users, so it doesn't seem to be urgent.

tisonkun:
OK. That makes sense.

When as a maintainer, when there is a patch avaiable and could be a net win, I typically merge since it contributes without downside. But if your style prevents you from doing this, it's understandable and I'd leave it to your preference.

tisonkun:
That said, projects would gradually move to thiserror 2.0 and if there is only thiserror 2.0, we don't need to compile thiserror 1.0.

This crate is the last one in one of my project's dep tree that refers thiserror 1.0. But that extra compiile step takes seconds to complete, so it is not an urgent issue, yes.

wez:
> But if your style prevents you from doing this,

it's not a matter of style, it's a matter of time available to release all of the crates that are impacted by this dep upgrade.
While it's one-click to merge this PR, unless you are using a git ref, the change won't make it to you until after the various wezterm crates have gone through their own release processes.

I don't have time for that right now.

tisonkun:
OK. I probably understand your preference to upgrade deps just before cutting a release. Hopefully this is a hint that thiserror is to be upgraded at that time :D

Issue #7156, opened 2025-08-15
Wezterm Only Using Quarter of Window


MrMiu:
### What Operating System(s) are you seeing this problem on?

Linux Wayland

### Which Wayland compositor or X11 Window manager(s) are you using?

Arch Linux with hyprland-git WM (version 0.5)

### WezTerm version

wezterm 20250730-195751-6a493f88

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

When starting up Wezterm, the terminal itself gets maximized as per my configurations in Hyprland. However, Wezterm itself only appears to be using a quarter of its own window. 

<img width="1921" height="1006" alt="Image" src="https://github.com/user-attachments/assets/5e6c0cd0-6b16-4a74-a689-2e41fd621d35" />
<img width="1909" height="1006" alt="Image" src="https://github.com/user-attachments/assets/da4abab4-6ab3-4f41-9d37-78270e47585a" />

### To Reproduce

Opening Wezterm in Hyprland with the command "wezterm"

### Configuration

local wezterm = require("wezterm")
local config = wezterm.config_builder()

config.color_scheme = "Catppuccin Mocha"
config.window_background_opacity = 0.8

return config

### Expected Behavior

Wezterm to be able to utilize the entire window it is opened in.

### Logs

11:30:38.209  WARN   window::os::wayland::connection > Unable to resolve appearance using xdg-desktop-portal: get_appearance.read_setting: Reading xdg-portal org.freedesktop.appearance color-scheme: org.freedesktop.DBus.Error.UnknownMethod: No such interface “org.freedesktop.portal.Settings” on object at path /org/freedesktop/portal/desktop
11:30:38.262  WARN   wezterm_gui::termwindow::resize > cannot resize window to match Some(RowsAndCols { rows: 28, cols: 62 }) because window_state is MAXIMIZED

### Anything else?

The issue does not appear when disabling Wayland in the Wezterm config file (however I would perfer to run Wezterm with Wayland natively).

I've tried forcing Hyprland to open it as a float window at the start as well as manually setting the size of Wezterm through initial_rows and initial_cols to no avail. In addition, I have uninstalled and reinstalled both hyprland-git (and all of its supporting packages) as well as wezterm-git.

On another note, I am assuming that the nightly version of Wezterm is the version inside of the AUR (wezterm-git).

bew:
[...]

PR #7095, opened 2025-07-17
adopt sctk-adwaita as our frame on wayland

HearseDev:
This bumps sctk-adwaita to use the main branch that allows to hide the title bar. Allows dragging via the tab_bar, and adopts the changes in https://github.com/wezterm/wezterm/pull/6923#issuecomment-2853352203.

PR #7277, opened 2025-10-03
Fix fractional scaling issues in hyprland 0.51.0

kalebo:
In hyprland 0.51.0 wezterm behaves badly as seen at https://github.com/hyprwm/Hyprland/discussions/11727 when fractional scaling is enabled. 

Wezterm is the only application I've seen that behaves this way on hyprland 0.51.0. Poking through the code with the help of Claude helped surface the attached changes which fix the rendering issues.

Specifically, the call to `set_window_geometry`  in `WaylandWindow::new_window` expects the surface coordinates. It's simply incorrect to be passing the pixel width and height here. There is an existing instance method in `WaylandWindowInner` for converting pixels to surface coordinates, but we can't call that in the `new_window` constructor. To avoid duplicating the `pixels_to_surface` calculations, I hoisted them up into a `WaylandDimensions` trait.

### DIFF
--- a/window/src/os/wayland/window.rs
+++ b/window/src/os/wayland/window.rs
@@ -58,6 +58,29 @@ use crate::{
     WindowEventSender, WindowKeyEvent, WindowOps, WindowState,
 };
 
+/// Wayland-specific coordinate conversion methods for Dimensions
+trait WaylandDimensions {
+    fn dpi_factor(&self) -> f64;
+    fn pixels_to_surface(&self, pixels: i32) -> i32;
+    fn surface_to_pixels(&self, surface: i32) -> i32;
+}
+
+impl WaylandDimensions for Dimensions {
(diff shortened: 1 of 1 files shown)


Rolv-Apneseth:
Can confirm this works for me, thanks for this. I believe the relevant issue on this repo is #7156

rudrabhoj:
Thanks, this fixes wezterm for me on Hyprland 0.51.1 on FreeBSD 14.3!

bew:
@wez hey, sorry for the direct ping but I think this would benefit users for your attention:
Could you take a look at / merge this PR if it makes sense, so hyprland users can have a non-broken experience for the time being ?

wez:
Looks good, thank you!

HearseDev:
@kalebo Hi, since #7095 might get merged, could you please try it out and see if this issue is not present in my PR. Thank you

PR #7283, opened 2025-10-06
fix: memory leak in macOS MetalLayer management

I-Info:
When using `WebGPU` front end, creating and closing windows will cause `VM:IOSurface` memory leak, due to an unreleased object reference.

Also with some other trivial memory leak prevention measures.


wez:
Thank you!

PR #7292, opened 2025-10-10
tmux-CC: suppress capability handshakes in control mode

tekacs:
## Summary
- suppress OSC/CSI/DCS capability probes when writing through tmux control mode so we stop asking tmux to echo them back at the pane
- classify the same sequences on the read side, buffering per pane so fragmented replies are dropped cleanly and flushing that state when tmux tears a window down
- keep regular application output untouched; only handshake traffic is intercepted

## Background & Motivation
When tmux is driven via `tmux -CC`, it relays *every* byte it sees back to the control client. WezTerm emits colour and capability probes whenever a pane starts up or focus changes; in control mode those queries and their answers were forwarded straight into the running program. Codex (and even tools like `jj diff`) ended up seeing raw OSC/CSI text such as `10;rgb:…`, which broke layout and disabled spinners.

## Implementation Notes
- `classify_handshake_sequence` recognises the capability probes we originate (OSC 10/11/12, `CSI ?/>/=…c`, `CSI 4/8/14/16 … t`, `DCS tmux;… ST`) and is shared by both the writer and reader paths.
- `queue_send_keys` now drops those probes before they hit tmux, while `strip_handshake_sequences` buffers `%output` per pane and discards any recognised replies—even if tmux split them across multiple events—then clears the buffer when the pane disappears.
- All other output continues to pass straight through to the application.

## Testing
- macOS 15.1, nightly build `20251008-100947-abcdef12`, running `wezterm` → `tmux -CC new -As work` → launch Codex / focus panes / split windows: no stray OSC/CSI text, UI renders immediately, spinner blinks.
- `tmux -CC attach` into existing sessions with Codex, vim, `jj diff`: panes stay interactive and `jj diff` no longer prints raw `CSI ?…c` tails.

## Without this fix
image

### DIFF
--- a/mux/src/tmux.rs
+++ b/mux/src/tmux.rs
@@ -52,6 +52,155 @@ pub(crate) struct TmuxRemotePane {
 
 pub(crate) type RefTmuxRemotePane = Arc<Mutex<TmuxRemotePane>>;
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum HandshakeMatch {
+    Complete(usize),
+    Incomplete,
+    None,
+}
+
+fn osc_body_len(data: &[u8]) -> Option<usize> {
+    for (idx, byte) in data.iter().enumerate() {
+        match byte {
+            0x07 => return Some(idx + 1),
+            0x1b if data.get(idx + 1) == Some(&b'\\') => return Some(idx + 2),
+            _ => {}
+        }
+    }
+    None
+}
+
+fn csi_body_len(data: &[u8]) -> Option<(usize, u8)> {
+    for (idx, byte) in data.iter().enumerate() {
+        if (0x40..=0x7e).contains(byte) {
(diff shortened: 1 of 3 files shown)


tekacs:
Apologies if this isn't the best way to fix this! I've just remedied the issue for my own use illustratively, so that I can actually benefit from this mode – and it's amazing, thank you! Happy to retool it if it'll help.

tekacs:
Apologies, the commits of this series are still clean, but the PR as a whole now also includes syncing title + tab order across to tmux – I'm slowly improving the state of `tmux -CC` every time I run into an issue.

I'm keeping the actual probe filtering commit at the end of the series, since it's the one that I'm the least confident in this being the correct implementation of.

LeszekSwirski:
This is great, I was getting these "junk" characters when using nvim inside tmux -CC in Wezterm, and with this PR I don't anymore.

PR #7296, opened 2025-10-11
Add PowerShell integration script

rashil2000:
This PR adds the following shell integration features to Wezterm via a PowerShell script:
- OSC 133 (semantic prompt A/C/D)
- OSC 7 (cwd)
- OSC 1337 user-vars: WEZTERM_PROG (full command line), WEZTERM_CMD (resolved command), WEZTERM_USER, WEZTERM_HOST, WEZTERM_IN_TMUX
- Toggleable via functions (enable/disable)

I have updated the docs with instructions as well. 

Tested on Windows and macOS and with PowerShell 5.1 as well as PowerShell 7.

### DIFF
--- a/assets/shell-integration/wezterm.ps1
+++ b/assets/shell-integration/wezterm.ps1
@@ -99,7 +99,7 @@ $null = New-Module wezterm {
             if ($gc.CommandType -eq 'Alias') {
               $resolved = $gc.Definition
             } else {
-              $resolved = $gc.Name
+              $resolved = $gc.Name -replace '(?i)\.exe$', ''
             }
           } else {
             $resolved = $candidate


assets/shell-integration/wezterm.ps1
```diff
@@ -0,0 +1,271 @@
+#!/usr/bin/env pwsh
```
bew, line 1: Does this actually work on Windows, there is no `/usr` there 👀

rashil2000: It _technically_ will work inside Git Bash/MSYS2/Cygwin, but there's no use of that 😅

Elsewhere it'll simply be ignored so I left it there.

PR #7338, opened 2025-10-31
Removed "start --cwd ." from Exec=wezterm in wezterm/assets/wezterm.desktop

sunbearc22:
This pull request is a follow up action to Discussion [#7252](https://github.com/wezterm/wezterm/discussions/7252). 

**Summary of Issue:**
Presently for Linux desktops, whenever the WezTerm desktop icon is clicked on, WezTerm will be started with the `wezterm start --cwd .` command. However doing so overrides the `config.default_gui_startup_args` option declared in the `wezterm.lua` file, which in turn causes the wezterm Window to not appear and function as is defined in the `wezterm.lua` file. Users of wezterm would then end up wasting their time "debugging" why the wezterm window is not responding to their `config.default_gui_startup_args` specifications , which seems unnecessary.

I am proposing that `Exec=wezterm` should be used in the `wezterm.desktop` file instead of `Exec=wezterm start --cwd .`.


bew (commented):
I'm not sure this is still the right solution 🤔

The `start --cwd .` was added for a good reason: #2271 aka being able to start Wezterm from a gui app, in a specific directory (`wezterm` alone would always start in $HOME)

There was also an attempt to remove the `--cwd .` in #6148 but was quickly closed.

I think we'd need a bit more change to continue supporting the GUI-open-in-dir use-case 🤔

sunbearc22:
> I'm not sure this is still the right solution 🤔
> 
> The `start --cwd .` was added for a good reason: #2271 aka being able to start Wezterm from a gui app, in a specific directory (`wezterm` alone would always start in $HOME)
> 
> There was also an attempt to remove the `--cwd .` in #6148 but was quickly closed.
> 
> I think we'd need a bit more change to continue supporting the GUI-open-in-dir use-case 🤔

@bew After reading the other discussion posts that you have share here, below are my thoughts:
1.  Regarding #2103, such users can and should just declare an alias in their `~/.bashrc` to achieve what they want. For example, 
    ` alias wezterm="wezterm start --cwd ." `
     So whenever they type `wezterm` in the terminal, WezTerm would just start from their current working directory. 
2. Regarding #2271, I did not really understand the issue raised by @Anomalocaridid. I tried running  `wezterm start --cwd .` in gnome terminal and that opened wezterm in the cwd of the gnome terminal. Since this command worked, the use of an _alias_ as is mentioned in my point 1 should address his needs too. If he is saying that when he issues the wezterm command from the terminal, and his linux system would run the `wezterm.desktop` file to run wezterm instead of the wezterm binary found in his $PATH, then that situation is new to me. I would think most Linux distributions don't do that. 
3.  Regarding #6148 , although @rindeal articulated concerns were direct and rejected, it does appear to echo one of the concerns that I have mentioned in my PR & Discussion #7252.  

I urge reviewers of my PR to see the evidences that I have presented in Discussion #7252.  I have provided insight as why it is bad for linux desktop users to have `Exec=wezterm start --cwd .` declared in their `wezterm.desktop` file; it breaks the `config.default_gui_startup_args` option declared in the `wezterm.lua` file. Furthermore, whenever the WezTerm desktop icon is clicked on, wezterm should always open either from $HOME (which is achieved by `Exec=wezterm`) or in the directory path  declared by  `config.default_gui_startup_args` option in the `wezterm.lua` file.  Please also consider the alternative solution that I have proposed for #2013 (and possibly for #2271).  I do think accepting my PR can benefit more linux users of wezterm than harm them. 

Best regards.

PR #7345, opened 2025-11-03
fix omitted color params in sixel parser

### LINKED ISSUE #7344: SIXEL: omitted color parameters between semicolons are not treated as 0 but as 100

Linux X11

_No response_

### WezTerm version

20251025-070338-b6e75fd7

### Describe the bug

According to the VT330/VT340 Programmer Reference Manual (chapter 14, [Parameter Separator (;)](https://vt100.net/docs/vt3xx-gp/chapter14.html#S14.3.6)):

> If there is no number before the separator, the terminal assumes that parameter is 0. If there is a number after the separator, the terminal assumes that parameter is 0.

This rule applies to all DCS sequences, including SIXEL color definitions (Pc;Pu;Px;Py;Pz).
Therefore, the following should be equivalent:

`#1;2;;50;50   ≡   #1;2;0;50;50`

Other terminals like xterm and alacritty render it well. But wezterm currently treats the ommited 0 as a maximum value (100) instead.

### To Reproduce

This should print a 12x12 teal rectangle but wezterm renders it light coral pink:
```sh
echo -e '\033P;1;q#;2;;;#1;2;;50;50#1!12~-#1!12~\033\\'
```

Image

Adding a 0 for the red channel makes wezterm render it as it's supposed to:
```sh
echo -e '\033P;1;q#;2;;;#1;2;0;50;50#1!12~-#1!12~\033\\'
```

Image

### Configuration

no config

### Expected Behavior

Wezterm should interpret missing numbers before semicolons as 0, not as 100.

### Logs

_No response_

### Anything else?

_No response_

joseluis:
Fixes #7344

Omitted parameters in color definition sequences (e.g. `#1;2;;50;`) were incorrectly treated as 100 due to -1 sentinel values wrapping during type conversion. The DEC Sixel specification states that any missing parameter between semicolons must default to 0.

This change normalizes unset parameters to 0 when interpreting both RGB and HLS color definitions, matching the behavior of other terminals and the Sixel standard.

### DIFF
--- a/wezterm-escape-parser/src/parser/sixel.rs
+++ b/wezterm-escape-parser/src/parser/sixel.rs
@@ -97,11 +97,15 @@ impl SixelBuilder {
         match self.current_command {
             b'#' if self.param_no >= 4 => {
                 // Define a color
-                let color_number = self.params[0] as u16;
-                let system = self.params[1] as u16;
-                let a = self.params[2] as u16;
-                let b = self.params[3] as u8;
-                let c = self.params[4] as u8;
+                let param = |i| match self.params.get(i).copied().unwrap_or(-1) {
+                    -1 => 0,
+                    v => v.max(0),
+                };
+                let color_number = param(0) as u16;
+                let system = param(1) as u16;
+                let a = param(2) as u16;
+                let b = param(3) as u8;
+                let c = param(4) as u8;
 
                 if system == 1 {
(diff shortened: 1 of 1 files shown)


joseluis:
Linking to https://github.com/wezterm/wezterm/issues/6341 given the current maintenance status, and the fact this is a complete fix for the related issue.

PR #7360, opened 2025-11-09
wayland: Show client-side decorations when compositor doesn't provide server-side

andrewcarson:
With the default Debian 13 GNOME Wayland, the compositor does not provide server side window decorations.

Previously, WezTerm would:
1. Request server-side decorations for window_decorations = "TITLE | RESIZE"
2. Hide client-side frame assuming the request would be honored
3. Never check if compositor actually provided server-side decorations

This resulted in windows with no decorations on Gnome, which always responds with client-side decoration mode.

Now we check the decoration_mode field in WindowConfigure events and:
- Show client-side frame if user wants decorations AND compositor sets client-side mode
- Hide frame immediately if compositor provides server-side OR user set window_decorations = "NONE"
- Defer showing frame until we have valid dimensions

This respects both the user's preference and the compositor's capabilities.

### DIFF
--- a/window/src/os/wayland/window.rs
+++ b/window/src/os/wayland/window.rs
@@ -697,6 +697,12 @@ impl WaylandWindowInner {
         self.dimensions.pixels_to_surface(pixels)
     }
 
+    fn should_show_client_side_frame(&self, window_config: &WindowConfigure) -> bool {
+        let user_wants_decorations = self.config.window_decorations != WindowDecorations::NONE;
+        let compositor_selected_client = window_config.decoration_mode == DecorationMode::Client;
+        user_wants_decorations && compositor_selected_client
+    }
+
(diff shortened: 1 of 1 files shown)


RoloEdits:
Am I correct is guessing that this fixes the regressions wezterm has seen where you cannot resize/drag? Or is this simply a fallback when that happens so that some functionality remains?

andrewcarson:
> Am I correct is guessing that this fixes the regressions wezterm has seen where you cannot resize/drag? Or is this simply a fallback when that happens so that some functionality remains?

I was focussing mostly on title bar. I have noticed that resizing is a not working great (I this is mostly that the mouse pointer is not changing), but really I was waiting to see what happens with this PR before I investigate further.

bew:
Have you seen the work started at https://github.com/wezterm/wezterm/pull/7095 ?
Afaiu it aims to fix tje situation for Wayland regarding the titlebar + resize

Issue #5500, opened 2024-06-01
Undocumented maximum value for max_fps


rpdelaney:
**What page or section of the docs have an issue?**
Docs for [max_fps](https://wezfurlong.org/wezterm/config/lua/config/max_fps.html)

**Describe the issue**
When setting `max_fps = 360` I am met with an error:

```lua
error converting Lua table to Config (Config::from_dynamic: Error processing
Config::max_fps: out of range integral type conversion attempted.
{
    "max_fps": 360,
})
stack traceback:
        [C]: in metamethod 'newindex'
        [string "/home/ryan/.config/wezterm/wezterm.lua"]:18: in main chunk
```
Based on the error, I gather that 360 is too high, but the limits are not documented.

beckend:
Can this be raised? The gaming monitors are way above this limit, I tried to use 300 for mine and the same error as above.

PR #7366, opened 2025-11-11
Fix: config max_fps

beckend:
Fixes https://github.com/wezterm/wezterm/issues/5500

We are already using it as u64 here and there:

image

Now gaming monitors can use their desired refresh rate.


wez:
Thank you!

PR #7374, opened 2025-11-13
Remove initial_value param from an example of PromptInputLine.md

romusz:
Since initial_value is only available in nightly builds the example in the current form errors out for regular builds


romusz:
Hi there, I won't insist, though I would recommend that examples are not using params that only work in nightly builds. It's better IMO to have smaller working examples than to show all/many options. Also worth considering what the semantics of 'nightly' is (I'm new to Wez) - is it experimental, not guaranteed to ever make into a release or pre-release.
Anyway, I can close the PR if the author/contributors/community feels strongly that it does not help people looking to quickly copy/paste an example.

bew:
Alright, let's close this PR then

Docs changes are always welcome but I guess this is an exception `¯\_(ツ)_/¯`

PR #7385, opened 2025-11-18
feat: smart case search

### LINKED ISSUE #6018: Option for "smart" case sensitive search

When configuring search, I can [choose between case sensitive or case insensitive](https://github.com/wez/wezterm/blob/a471618397fb4231be1c1f10bd69ea0c20740b00/config/src/keyassignment.rs#L112-L117), however I'm interested in a third option that I grew accustomed to as a former iTerm2 user -- "smart" case sensitivity, [as described in its docs](https://iterm2.com/documentation-one-page.html):

> The default case sensitivity option of "Smart Case Sensitivity" performs a case-sensitive search if the search query contains any upper case letters. Otherwise, a case-insensitive search is performed.

Perhaps something like this:

```lua
config.keys = {
  {
    key = "f",
    mods = "SUPER",
    action = wezterm.action.Search({ SmartCaseSensitivity = "" })
  },
}
```

I'm relatively new to wezterm and haven't come up with any workarounds, but if there's a way to do this through the existing API then I'd be happy to use that approach.

chelmertz:
Add "smart-case" into the cycling of already existing search modes.

Fixes https://github.com/wezterm/wezterm/issues/6018


chelmertz:
Sorry for a drive-by PR, maybe it's good enough to warrant some eyeing.

It works when testing manually, and when configuring a key to `wezterm.action.Search { CaseSmartString = "someString" }`.

I did not update docs/config/lua/keyassignment/CopyMode/CycleMatchType.md nor docs/config/lua/keyassignment/Search.md -- unsure if I should do this (and if so - how to get the correct refs in the `since()` notation, before merging).

Best regards,
Calle

bew:
Hello o/


> I did not update docs for CycleMatchType / Search actions -- unsure if I should do this

You can update them yes, we'll merge both the code and the docs updates

For the `since()` notation you can use: `{{since('nightly')}}` (exactly)

docs/scrollback.md
```diff
@@ -71,8 +71,10 @@ When the search overlay is active the behavior of wezterm changes:
 * `CTRL-N` and `DownArrow` will cause the selection to move to any next matching text.
 * `PageDown` will traverse to the next match one page at a time.
 * `CTRL-R` will cycle through the pattern matching mode; the initial mode is case-sensitive
-  text matching, the next will match ignoring case and the last will match using the
-  [regular expression syntax described here](https://docs.rs/regex/1.3.9/regex/#syntax).
+  text matching, the next will match ignoring case, the next is "smart" and will be
+  case-sensitive if any input is uppercase, otherwise it will match ignoring case, and the
+  last will match using the [regular expression syntax described
+  here](https://docs.rs/regex/1.3.9/regex/#syntax).
```
bew: ⚠️ I think that breaking the link on multiple lines will break it
```suggestion
  last will match using the
  [regular expression syntax described here](https://docs.rs/regex/1.3.9/regex/#syntax).
```

chelmertz: Thanks, adjusted.

chelmertz:
Added the doc updates (thanks for the pointer about "nightly"!) and rebuilt it on top of latest master - for me, this works fine when manually testing it, but I wouldn't mind an extra pair of eyes since it's my first contribution :)

The doc updates could be written differently, but I opted for "let's mention all options in the top of each doc section, and have a footnote/"since" mentioning the addition of smart case. This way, there's still only one list of valid options (for HEAD/recent versions), rather than two places to mentally combine into the full set of options.

bew (approved):

mrdziuban:
Hi @bew, is there anything preventing this from being merged?

bew:
Hello @chelmertz, thanks for the ping this PR had slipped my mind since I took over maintenance of the repo 😬

Tested locally, works great thank you!

Issue #7393, opened 2025-11-22
Split RPM build into separate client/server packages


jsgf:
**Is your feature request related to a problem? Please describe.**
I use Wezterm with the mux server, but the server is on a headless system with no GUI/font/etc libraries installed. When I install the wezterm .rpm package it brings in all those dependencies as well. I'd like to just install the server and command line tools without the gui.

**Describe the solution you'd like**
Break the package up into
- wezterm-common - the CLI tools
- wezterm-mux-server - the mux server, depends on common
- wezterm-gui  - the UI side, depends on common
- wezterm - a virtual package which depends on all of the above

**Describe alternatives you've considered**
I've also just been using `cargo install` to install parts, but I don't want a full Rust/C toolchain on the server either (including all the build deps).

**Additional context**
I have an implementation for this; I'll put up a related PR shortly.

PR #7395, opened 2025-11-23
Split RPM package into separate subpackages for wezterm-common, wezterm-gui and wezterm-mux-server

jsgf:
Uses RPM subpackages (%package -n) with a single spec file - the standard approach for generating multiple binary RPMs from one build. This allows wezterm-mux-server to be installed without GUI dependencies (X11, Wayland, fontconfig, mesa-libEGL, etc).

Changes:
- wezterm-common: CLI launcher and utilities, requires only openssl
- wezterm-gui: GUI terminal with desktop integration, requires GUI libraries
- wezterm-mux-server: Headless multiplexer server, requires only openssl
- wezterm: Metapackage requiring all three for backward compatibility

Single %build section compiles all binaries together.

Also fixes handling of TARGET_DIR, which was being partially ignored.

I'm not sure what the equivalent for other packaging schemes looks like.

Fixes https://github.com/wezterm/wezterm/issues/7393

### DIFF
--- a/ci/deploy.sh
+++ b/ci/deploy.sh
@@ -40,15 +40,15 @@ case $OSTYPE in
     for bin in wezterm wezterm-mux-server wezterm-gui strip-ansi-escapes ; do
       # If the user ran a simple `cargo build --release`, then we want to allow
       # a single-arch package to be built
-      if [[ -f target/release/$bin ]] ; then
-        cp target/release/$bin $zipdir/WezTerm.app/Contents/MacOS/$bin
+      if [[ -f $TARGET_DIR/release/$bin ]] ; then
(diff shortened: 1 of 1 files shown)


wez:
Looks good, thank you!

wez:
Ah, and of course the workflows are different for main vs. prs, and there's a failure:
https://github.com/wezterm/wezterm/actions/runs/19613326694/job/56162331223#step:22:16

wez:
I think that's fixed up, now the question is whether the installation instructions in the docs still hang together; there's the copr installation and manual installation stuff that should be run through.  I'm not within easy reach of a Fedora box to try this at the moment.

jsgf:
The COPR ones should be OK since wezterm depends on all the others. The Fedora manual download won't work without also downloading the other rpms.

Issue #1432, opened 2021-12-20
Allow changing Bar cursor's width


eugenesvk:
**Is your feature request related to a problem? Please describe.**

I like the `Bar` cursor style instead of the `Block` one, but it's a bit too narrow, so visibility suffers more than it should

**Describe the solution you'd like**

A configuration option that would allow me to change the width of a `Bar` cursor

**Describe alternatives you've considered**
Going back to the `Block`

**Additional context**

wez:
I'm not excited to add a very specific configuration option for this, because I think it would, over time as more options are added, make the implementation more complex and harder to change without breaking someone's carefully crafted config on an upgrade.

I've been considering exposing the [custom block glyph styling datatypes](https://github.com/wez/wezterm/blob/main/wezterm-gui/src/customglyph.rs#L249-L258) to lua to allow for defining frames in [animated cursors](https://github.com/wez/wezterm/discussions/1136) and I think that same mechanism could be used here for this; while it is a more complex configuration than just adding a width control, the overall result would simplify the logic for selecting a cursor shape and make a very powerful configuration for end users.

I don't know exactly what this would look like yet, but the general requirements are:

* Ability to override the possible cursor shapes individually.  eg: do all of the below one way for `SteadyBlock` and a different way for `SteadyBar` and so on.
* Ability to specify either:
   * an image file to use for the shape
   * use the same set of drawing commands used for internal custom glyphs
* For animations:
   * If using animated gif/png then simply use those frames and timing info 
   * Otherwise, ability to specify multiple frames (eg: discrete image files, or discrete drawing commands per frame), as well as the frame delay
* For inactive/unfocused display:
  * Specify a single frame to use
  * Consider adding a color scheme option for the inactive cursor color

klamonte:
OSC 22 is supported by several terminals for the application to select a cursor. I have not yet looked into it deeply, but apparently xterm and mintty at least do not agree on what those cursors should be.

If this feature is implemented, one might also get OSC 22 for nearly free.

bertin0:
> Consider adding a color scheme option for the inactive cursor color

So at the moment how is the inactive cursor color chosen? Is it hardcoded? Is it linked to any other color in the color scheme?

wez:
There is currently no distinction for color based on the active pane. The shape does change to an outline shape when inactive.

bertin0:
[...]

PR #7420, opened 2025-12-07
Cursor trail

### LINKED ISSUE #7387: feat: add opt-in cursor trail / smear effect (the last thing keeping many users on Kitty)
### Feature request: opt-in cursor trail / motion smear effect

Dear @wez and the WezTerm team,

I've been trying to fully switch to WezTerm from Kitty for literally years.  
Every single time I set it up perfectly, fall in love with the Lua config, the multiplexing, the performance, the fonts, the everything… and then after 2–3 days I crawl back to Kitty with tears in my eyes.

**The only reason** — literally the one and only missing feature — is the cursor trail/smear effect that Kitty added recently (and that Neovide had forever).

I completely understand and respect that you personally find cursor animations distracting (you mentioned it in https://github.com/wez/wezterm/discussions/1136). But for a surprising number of us this tiny visual cue is not a gimmick — it's a real productivity and comfort feature:

- When pairing, teaching, or streaming, viewers can instantly see where the cursor jumped
- On high-DPI / large / ultrawide monitors it's way easier to track fast `hjkl`, mouse movements, or `Ctrl-f`/`Ctrl-b` jumps
- It simply feels smoother and more responsive when typing or navigating shell output

This is **the single remaining excuse** many of us have for not making WezTerm our permanent daily driver. Everything else is already objectively better.

#### Existing related issues (so you see we're not alone):
- #6492 — Possibility for adding cursor trail in Wezterm?
- #6529 — cursor animations
- https://github.com/wez/wezterm/discussions/1136 — old discussion where the topic was dismissed

#### Proposed solution (as minimal as possible):
Even a simple opt-in option would make hundreds (maybe thousands) of people switch instantly and never look back:

```lua
-- wezterm.lua
config.enable_cursor_trail = true        -- false by default, respects your taste!
config.cursor_trail_duration = 0.3       -- seconds, optional
config.cursor_trail_opacity = 0.5        -- optional
```

Or even just `config.cursor_trail = true` with hard-coded sane defaults.

It could be disabled by default, hidden behind `unstable_features = true`, or whatever — just give us the choice.

Please, please consider adding this one tiny feature. It would finally remove the very last reason for many of us to keep Kitty installed. I (and many others silently watching these issues) would be eternally grateful and would evangelize WezTerm even harder than we already do.

Thank you for creating what is already the best terminal on the planet.  
This one little thing would make it undeniably perfect for everyone.

With hope and a lot of emotion,  
a very sad almost-full-time WezTerm user who just wants to come home 🥲

flowchartsman:
Initial cursor trail implementation based off of Kitty's

sample config:

```lua
local config = {}

config.cursor_trail = {
  enabled = true,
  -- dwell_threshold controls the amount of time the cursor must sit still
  -- to draw a trail.
  dwell_threshold = 80,
  -- distance_threshold defines the radius in cells the cursor must jump before
  -- a trail is drawn.
  distance_threshold = 5,
  -- duration is the animation time in milliseconds for leading edges of the trail to reach the cursor
  duration = 300,
  -- spread is a multiplier for duration applied to the trailing edges of the quad used to render the trail. This increases the apparent smear.
  spread = 2,
  opacity = 0.6,
}
config.animation_fps = 60

return config
```

closes #7387 

**NOTE:** Given the discussion in #1136, #1432 and elsewhere, I'm leaving this PR as a draft and a discussion piece for now, since it effectively reduces cosmetic animations to a single effect with a very specific configuration, and would make the process of migrating to a more general approach more difficult if that's the direction the project wants to go.


flowchartsman:
@wez or any other contributors: feel free to chime in if this is way off-base.

There are definitely some rough edges that need sanded down even now (hence the draft), and if a complete gutting or rewrite is required, it won't hurt my feelings.

wezterm-gui/src/termwindow/render/pane.rs
```diff
@@ -569,6 +572,46 @@ impl crate::TermWindow {
...
+                    let (r, g, b, _) = cursor_border_color.tuple();
+                    let trail_color =
+                        LinearRgba::with_components(r, g, b, config.cursor_trail.opacity);
+
+                    quad.set_fg_color(trail_color);
+                    quad.set_hsv(None);
+                    quad.set_texture(white_space);
+                    quad.set_is_background();
+                }
+            }
```
flowchartsman, line 614: metrics and trace for paint_pane.lines is AFTER this, so, provided this implementation stands, open questions are:

- should the histogram macro and the TODO be moved above the cursor trail?
- does cursor trail need a metric?

wezterm-gui/src/termwindow/render/pane.rs
```diff
@@ -569,6 +572,46 @@ impl crate::TermWindow {
...
+            // Update and render cursor trail (if enabled)
+            if config.cursor_trail.enabled
+                && pos.is_active
+                && self
+                    .cursor_trail
+                    .tick(TickContext::from_cursor(cursor, &config.cursor_trail))
+            {
+                let mut quad_impl = layers.allocate(0)?;
+
+                if let QuadImpl::Vert(quad) = &mut quad_impl {
```
flowchartsman, line 585: From what I've seen in the code, it doesn't ever look like this would be boxed, but if that ends up not being the case, is there more defensive work that should be done here?

wezterm-gui/src/termwindow/render/pane.rs
```diff
@@ -569,6 +572,46 @@ impl crate::TermWindow {
...
+                    let frame_interval =
+                        std::time::Duration::from_millis(1000 / config.animation_fps as u64);
+                    self.update_next_frame_time(Some(start + frame_interval));
```
flowchartsman, line 588: I'm not sure if this is correct; feedback welcome.

wezterm-gui/src/termwindow/cursor_trail.rs
```diff
@@ -0,0 +1,274 @@
...
+            quad: TrailQuad::default(),
+            target: TrailTarget::default(),
+            last_cursor_pos: Pos::default(),
+            cursor_last_moved: now,
+            updated_at: now,
+        }
+    }
+
+    /// Update the trail state and return true if the quad should be rendered.
+    pub fn tick(&mut self, ctx: TickContext) -> bool {
```
flowchartsman, line 189: The logic here probably needs some cleanup/clarification. It's currently doing multiple duty for:

1. Did the cursor move since the last invocation and, if so, did it move far enough?
2. Has the cursor sat still long enough for the quad target to be updated?
3. Is the quad "close enough" to the cursor position to stop rendering?

And I think it's a little confusing.

zhecks:
will this capability be released in new release version?

bew:
> will this capability be released in new release version?

If it gets merged, then yes I guess, subscribe to this PR to get notified

emacsomancer:
Maybe affected by my running it on old laptop with integrated graphics, but, using the suggested settings above, to my eyes, the cursor itself immediately jumps, with the trail following behind:


https://github.com/user-attachments/assets/c3e3d97c-5f93-4f99-ad3b-864dc75ccc63



I have a sense that kitty's cursor takes a moment to appear in the new position, creating more of an impression (to my eye) of movement:


https://github.com/user-attachments/assets/1d6d7edb-248b-455e-aa42-426be48350c1



Maybe my computer or eyes are off, I'm not sure.

But wondering if there's a way to delay the display of the cursor in the new position (turn it invisible? change it to the background colour?)

In any case, very cool work!

flowchartsman:
> Maybe affected by my running it on old laptop with integrated graphics, but, using the suggested settings above, to my eyes, the cursor itself immediately jumps, with the trail following behind:

This is probably a combination of a couple factors. First would be the dwell time, which exists to keep the trail from spasmodically jumping around on every small cursor hop, since there is no way in Wezterm that I could find to distinguish cursor movement driven by user input from that which happens as a consequence of command output or by direct repositioning. [N]vim, for example, tends to briefly hop the cursor to the upper left of the terminal before the TUI is rendered and it finds its home in the editor window. Without a dwell time, you'd see a distracting series of hops every time it started. So, instead, the animation only starts after the cursor has not moved for a moment, which might contribute to the perception of lag. Maybe try lowering the dwell time a little and see if that changes your perception. It's a bit of a hack, and I'm not entirely satisfied with it.

The other factor is probably a consequence of how the animation is rendered: the quad used to draw the trail is initialized in the position of the most recent "significant" jump and with the same dimensions as a block cursor. Then the animation code starts to interpolate the vertices towards their destination.  The "leading" edge transitions more quickly than the trailing edge to create the effect, but it still needs to transit the entire distance. If the leading edge were instead initialized at the destination (or some fraction of the distance thereto), that might decrease the perception of lag. I'll play around with it some, but I'm reluctant to add it as another setting. Overall, my preference is to find some satisfying combination of settings and then just hard code them, rather than provide too many knobs to twiddle.

zhecks:
will this be approved?

estin:
Thanks for your effort! In my case, the cursor trail doesn't follow to the new pane. Also, there are some artifacts when typing in a new pane.

<details><summary>video</summary>

https://github.com/user-attachments/assets/46de469d-b057-4ccf-8046-4dd2fd74d2fa
</details>


<details><summary>sys info</summary>

```console
fastfetch
                  -`                     user@archlinux
                 .o+`                    --------------
                `ooo/                    OS: Arch Linux x86_64
               `+oooo:                   Host: H510M H (-CF)
              `+oooooo:                  Kernel: Linux 6.18.3-arch1-1
              -+oooooo+:                 Uptime: 12 hours, 38 mins
            `/:-:++oooo+:                Packages: 1800 (pacman), 146 (nix-user), 46 (nix-default), 8 (flatpak), 15 (snap)
           `/++++/+++++++:               Shell: nushell 0.109.1
          `/++++++++++++++:              Display (P24FBA-RAGL): 1920x1080 in 24", 100 Hz [External]
         `/+++ooooooooooooo/`            DE: COSMIC
        ./ooosssso++osssssso+`           WM: cosmic-comp (Wayland)
       .oossssso-````/ossssss+`          Cursor: Adwaita
      -osssssso.      :ssssssso.         Terminal: WezTerm fe53678
     :osssssss/        osssso+++.        Terminal Font: Comic Code
    /ossssssss/        +ssssooo/-        CPU: Intel(R) Core(TM) i5-10400F (12) @ 4.30 GHz
  `/ossssso+/:-        -:/+osssso+-      GPU: AMD Radeon RX 550 / 550 Series [Discrete]
 `+sso+:-`                 `.-/+oso:     Memory: 6.53 GiB / 31.21 GiB (21%)
`++:.                           `-/+/    Swap: 4.00 KiB / 4.00 GiB (0%)
.`                                 `/    Disk (/): 362.75 GiB / 438.55 GiB (83%) - ext4
                                         Disk (/home): 733.79 GiB / 937.82 GiB (78%) - ext4
                                         Local IP (enp3s0): 192.168.31.72/24
                                         Locale: en_US.UTF-8
```

</details>


<details><summary>config</summary>

```lua
cursor_trail = {
		enabled = true,
		-- dwell_threshold controls the amount of time the cursor must sit still
		-- to draw a trail.
		dwell_threshold = 80,
		-- distance_threshold defines the radius in cells the cursor must jump before
		-- a trail is drawn.
		distance_threshold = 5,
		-- duration is the animation time in milliseconds for leading edges of the trail to reach the cursor
		duration = 100,
		-- spread is a multiplier for duration applied to the trailing edges of the quad used to render the trail. This increases the apparent smear.
		spread = 2,
		opacity = 0.6,
	},
```
</details>


I am happy to provide logs or debug information to help resolve this issue

P.S. the following patch fixes it

```diff
diff --git a/wezterm-gui/src/termwindow/render/pane.rs b/wezterm-gui/src/termwindow/render/pane.rs
index c89b44f9c..c967ef2c1 100644
--- a/wezterm-gui/src/termwindow/render/pane.rs
+++ b/wezterm-gui/src/termwindow/render/pane.rs
@@ -590,7 +590,9 @@ impl crate::TermWindow {
                     let left_pixel_x =
                         padding_left + border.left.get() as f32 + (pos.left as f32 * cell_width);
                     let px_x = (self.dimensions.pixel_width as f32 / -2.0) + left_pixel_x;
-                    let px_y = (self.dimensions.pixel_height as f32 / -2.0) + top_pixel_y;
+                    let px_y = (self.dimensions.pixel_height as f32 / -2.0)
+                        + top_pixel_y
+                        + (pos.top as f32 * cell_height);

                     self.cursor_trail.apply_to_quad(
                         quad,
```

bew:
Hello guys, quick heads up, I'm considering going with #7737 PR instead since it has a more complete feature-set, see comparison done by @gottz: https://github.com/wezterm/wezterm/pull/7737#issuecomment-4253989987

👉 Prefer to follow that PR instead if you're interested in this feature 😉


Thanks for the PR anyway @flowchartsman , it's always useful to get alternative implementations :)
👉 Feel free to chime in and do a review pass on his PR if you have the time, you may find corner (no pun intended ^^) cases I won't think of during my review! 🙏

Issue #6920, opened 2025-04-28
Wezterm interpret config.window_decoration option value "RESIZE" as "RESIZE|TITLE"


ce7er2s:
### What Operating System(s) are you seeing this problem on?

Linux Wayland

### Which Wayland compositor or X11 Window manager(s) are you using?

WM: Mutter (Wayland)
DE: GNOME 48.1

### WezTerm version

wezterm 20250424-220948-1439661d

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

No, and I'll explain why below

### Describe the bug

My old config set config.window_decorations="RESIZE" which was fine.
After update it also adds a window bar above like "RESIZE|TITLE" would.

Here's wezterm with RESIZE:

![Image](https://github.com/user-attachments/assets/4cf0c527-bb9e-4821-9a63-eaf98b8aabad)

Here's wezterm with TITLE:

![Image](https://github.com/user-attachments/assets/5d1c0b0b-93a9-4ec1-a22d-de182e665cff)

Here's wezterm with NONE:

![Image](https://github.com/user-attachments/assets/9ca5b282-2dbb-4af5-bf3a-114972647df7)

I didn't try nightly build because I just built this version (it's from HEAD at this moment).

### To Reproduce

Set config.window_decorations="RESIZE"

### Configuration

```lua
-- Pull in the wezterm API
local wezterm = require 'wezterm'

-- This will hold the configuration.
local config = wezterm.config_builder()

-- This is where you actually apply your config choices

-- For example, changing the color scheme:

local act = wezterm.action

config.window_frame = {
  -- The font used in the tab bar.
  -- Roboto Bold is the default; this font is bundled
  -- with wezterm.
  -- Whatever font is selected here, it will have the
  -- main font setting appended to it to pick up any
  -- fallback fonts you may have used there.
  font = wezterm.font { family = 'Roboto', weight = 'Bold' },

  -- The size of the font in the tab bar.
  -- Default to 10.0 on Windows but 12.0 on other systems
  font_size = 12.0,

  -- The overall background color of the tab bar when
  -- the window is focused
  active_titlebar_bg = '#333333',

  -- The overall background color of the tab bar when
  -- the window is not focused
  inactive_titlebar_bg = '#333333',
}

config.harfbuzz_features = { 'calt=0', 'clig=0', 'liga=0' }

config.colors = {
  tab_bar = {
    -- The color of the inactive tab bar edge/divider
    inactive_tab_edge = '#575757',
[...]

PR #7429, opened 2025-12-10
 Don't draw window decorations in Wayland if `TITLE` is not enabled

luben:
The idea of the `INTEGRATED_BUTTONS` is to replace the window close/minimize/maximize buttons in the window title bar. With this change the Wayland mode matches the X11 behaviour.


luben:
Addresses multiple issues, e.g.
- https://github.com/wezterm/wezterm/issues/6920
- https://github.com/wezterm/wezterm/issues/6578

Issue #7409, opened 2025-12-01
links are highlighted when the mouse pointer is in a different pane


Infinoid:
### What Operating System(s) are you seeing this problem on?

Linux X11

### Which Wayland compositor or X11 Window manager(s) are you using?

xfce

### WezTerm version

20240203-110809-5046fc22 (debian forky on arm64), 20251123-154500-ac08cdf6 (nightly on x86-64)

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

When multiple panes are visible, and the bottom left pane is active, moving the mouse over the top left pane highlights links in the bottom right pane.

### To Reproduce

* Start wezterm
* Split vertically
* Split each half horizontally
* Open a file containing URLs with `less` in the bottom left pane
* Move mouse over the top left pane
* Witness the devastation

![Image](https://github.com/user-attachments/assets/5971d7d7-d561-4cda-871e-a7e6fa7f3e61)
(I don't know why everything is pink; it's some artifact of the screen recording.)

### Configuration

I don't know if layout/appearance settings matter, I'll include them just in case.

```lua
config.keys = {
  {
    key = '\\',
    mods = 'CTRL',
    action = wezterm.action.SplitHorizontal { domain = 'CurrentPaneDomain' },
  },
  {
    key = '-',
    mods = 'CTRL',
    action = wezterm.action.SplitVertical { domain = 'CurrentPaneDomain' },
  },
  {
    key = 'LeftArrow',
    mods = 'ALT',
    action = wezterm.action.ActivatePaneDirection "Left",
  },
  {
    key = 'RightArrow',
    mods = 'ALT',
    action = wezterm.action.ActivatePaneDirection "Right",
  },
  {
    key = 'UpArrow',
    mods = 'ALT',
    action = wezterm.action.ActivatePaneDirection "Up",
  },
  {
    key = 'DownArrow',
    mods = 'ALT',
    action = wezterm.action.ActivatePaneDirection "Down",
  },
}
config.hide_tab_bar_if_only_one_tab = true
config.tab_bar_at_bottom = false
config.use_fancy_tab_bar = false
config.window_padding = {
        left = 0,
        right = 0,
        top = 0,
        bottom = 0,
}
config.inactive_pane_hsb = {
        brightness = 0.4,
}
config.font = wezterm.font 'Noto Sans Mono'
```

### Expected Behavior

[...]

PR #7434, opened 2025-12-15
Detect mouse over the wrong panel during URL highlighting

Infinoid:
In the `termwindow` crate, mouse-over URL highlighting is done by `mouse_event_terminal()` using the "row" number determined by `mouse_event_impl()`.  It assumes that "row" is relative to the currently active pane, but that assumption isn't always true.

Add a check for whether the mouse is over the currently active pane.

Fixes: #7409

### DIFF
--- a/wezterm-gui/src/termwindow/mouseevent.rs
+++ b/wezterm-gui/src/termwindow/mouseevent.rs
@@ -654,6 +654,7 @@ impl super::TermWindow {
         capture_mouse: bool,
     ) {
         let mut is_click_to_focus_pane = false;
+        let mut is_over_active_pane = true;
 
         let ClickPosition {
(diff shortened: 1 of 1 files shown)


Infinoid:
I'm still learning rust, so please let me know if there's a better way to implement this.

I think it might be a good idea to move the `FindCurrentLink` stuff into the condition too, as there's no point searching for a URL when we won't highlight it anyway.  But I wanted to start with a minimal diff, to make it clear where the issue is.

Infinoid:
Tested on linux aarch64 (x11/xfce), the same machine I noticed this bug on.

Using `cargo run --bin wezterm-gui` without this patch:

7434-before

Same test, with patch applied:

7434-after

PR #7445, opened 2025-12-20
Update linuxbrew installation instructions

tommica:
There original instructions resulted in this error:

```
$ brew install wezterm
Warning: Treating wezterm as a cask. For the formula, use wezterm/wezterm-linuxbrew/wezterm or specify the `--formula` flag. To silence this message, use the `--cask` flag.
Error: macOS is required for this software.
```

### DIFF
--- a/docs/install/linux.md
+++ b/docs/install/linux.md
@@ -281,7 +281,7 @@ hide:
 
     ```console
     $ brew tap wezterm/wezterm-linuxbrew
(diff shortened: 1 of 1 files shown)


bew:
Related to #7457 which goes the other way

bew:
I've merged a similar PR with the alternative solution (full formula 'path' mention) https://github.com/wezterm/wezterm/pull/7768

Issue #4592, opened 2023-11-18
Support voice input on macOS


nekowasabi:
### What Operating System(s) are you seeing this problem on?

macOS

### Which Wayland compositor or X11 Window manager(s) are you using?

No.

### WezTerm version

Nightly 20231117-151105-60065258

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

No, and I'll explain why below

### Describe the bug

As a result of trying to activate the MacOS voice input function in Wezterm, the voice input icon does not appear and no text is entered on the wezterm screen when I vocalize. I am able to use the standard MacOS Terminal and iTerm2.

### To Reproduce

1. Open Wezterm 2. Focus Cursor 3. Type F2(default mac os voice input shortcut key

### Configuration
local wezterm = require("wezterm")
local act = wezterm.action

return {
  use_ime = true,
  hide_tab_bar_if_only_one_tab = true,
  enable_tab_bar = true,
	disable_default_key_bindings = true,
  keys = {
    { key = 'v', mods = 'CMD',  action=wezterm.action{PasteFrom="Clipboard"}},
  },
}

### Expected Behavior

Show voice input icon.

jackfranklin:
I ended up here after googling as I'd love to have this enabled, particularly given that it's quite fun to talk to AI tools rather than type!

https://github.com/wezterm/wezterm/issues/5180 sounds like a duplicate of this. It links to [this PR in iTerm2](github.com/gnachman/iTerm2/pull/456/files) where they added support (I think?). 

Do you have any sense of how hard this would be to support? Presumably we mostly "just" need to give it permission to hook into the OS dictation permissions system to allow it to accept the voice input. I would be happy to attempt to contribute here if possible.

BarakXYZ:
It would be amazing to have this feature!

Edit:
For now, I've worked around this using this very nice tool called [hear](https://github.com/sveinbjornt/hear?tab=readme-ov-file).
It gives a CLI to the built-in Speech Recognition API Apple has. We can then automate dictation to places Mac won't support with this vibe-coded script (lol):
```bash
#!/bin/bash

PID_FILE="/tmp/hear-pid.txt"
OUTPUT_FILE="/tmp/dictation-output.txt"

# Check if hear is running (by checking if PID file exists and process is alive)
if [ -f "$PID_FILE" ] && kill -0 $(cat "$PID_FILE" 2>/dev/null) 2>/dev/null; then
[...]

PR #7453, opened 2025-12-23
macos: fix dictation support, with new configs for preview text

### LINKED ISSUE #5180: Allow Speech Recognition (dictate) on macOS

- As a documentation writer using Vim, I want to use the "Speech Recognition" feature of macOS with wezterm.
- People with carpal tunnel syndrome may appreciate this feature.

- Allow Speech Recognition 

- https://developer.apple.com/documentation/bundleresources/information_property_list/nsspeechrecognitionusagedescription
- Working in iterm2, here is the PR  (I think) https://github.com/gnachman/iTerm2/pull/456/files

Erog38:
This updates the macOS bundle to contain the required update for `NSSpeechRecognitionUsageDescription` to enable access to macOS dictation tools. 

Additionally, color configurations were updated to allow the customization of the preview text input. 

Changes here are a little more robust that I was hoping for, but I had to also update render logic to appropriately differentiate the text input cursor from the preview text, allowing for the cursor and the preview text to all be customized independently. 

Lastly, a change was required [here](https://github.com/wezterm/wezterm/compare/main...Erog38:feat/macos-dictation?expand=1#diff-8d5deb0ebb3ad681f41c38cd78277ee084fdde60e0402a803c4cb3f730760d66R1981) which provides macOS a much-needed location for where to put dictated text. 

This built and operates locally, and I'll be daily driving this build for a while to see what issues I can find with things like Claude Code wanting dictation input.

This should fix https://github.com/wezterm/wezterm/issues/5180 and https://github.com/wezterm/wezterm/issues/4592 as they're effectively duplicates.

### DIFF
--- a/wezterm-gui/src/termwindow/render/mod.rs
+++ b/wezterm-gui/src/termwindow/render/mod.rs
@@ -571,45 +571,32 @@ impl crate::TermWindow {
                 self.dead_key_status != DeadKeyStatus::None || self.leader_is_active();
 
             if dead_key_or_leader && params.is_active_pane {
-                let (fg_color, bg_color) = {
-                    let compose_fg = params
+                let (fg_color, bg_color) = (
+                    params
                         .config
                         .resolved_palette
                         .compose_fg
-                        .map(|c| c.to_linear());
-                    let compose_bg = params
-                        .config
-                        .resolved_palette
-                        .compose_bg
-                        .map(|c| c.to_linear());
-
-                    match (compose_fg, compose_bg) {
-                        (Some(fg), Some(bg)) => (fg, bg),
-                        (Some(fg), None) => {
-                            let bg = if self.use_reverse_video_cursor(&params) {
-                                params.fg_color
-                            } else {
-                                params.cursor_bg
-                            };
-                            (fg, bg)
-                        }
-                        (None, Some(bg)) => {
-                            let fg = if self.use_reverse_video_cursor(&params) {
+                        .map(|c| c.to_linear())
+                        .unwrap_or_else(|| {
+                            if self.use_reverse_video_cursor(&params) {
                                 params.bg_color
                             } else {
                                 params.cursor_fg
-                            };
-                            (fg, bg)
-                        }
-                        (None, None) => {
-                            if self.use_reverse_video_cursor(&params) {
-                                (params.bg_color, params.fg_color)
-                            } else {
-                                (params.cursor_fg, params.cursor_bg)
                             }
-                        }
-                    }
-                };
+                        }),
+                    params
+                        .config
+                        .resolved_palette
+                        .compose_bg
+                        .map(|c| c.to_linear())
+                        .unwrap_or_else(|| {
+                            if self.use_reverse_video_cursor(&params) {
+                                params.fg_color
+                            } else {
+                                params.cursor_bg
+                            }
+                        }),
+                );
 
                 let fg_color = self.ensure_min_contrast(fg_color, bg_color);
 
--- a/wezterm-gui/src/termwindow/render/mod.rs
+++ b/wezterm-gui/src/termwindow/render/mod.rs
@@ -571,10 +571,44 @@ impl crate::TermWindow {
                 self.dead_key_status != DeadKeyStatus::None || self.leader_is_active();
 
             if dead_key_or_leader && params.is_active_pane {
-                let (fg_color, bg_color) = if self.use_reverse_video_cursor(&params) {
-                    (params.bg_color, params.fg_color)
-                } else {
-                    (params.cursor_fg, params.cursor_bg)
+                let (fg_color, bg_color) = {
+                    let compose_fg = params
+                        .config
+                        .resolved_palette
+                        .compose_fg
+                        .map(|c| c.to_linear());
+                    let compose_bg = params
+                        .config
+                        .resolved_palette
+                        .compose_bg
+                        .map(|c| c.to_linear());
+
+                    match (compose_fg, compose_bg) {
+                        (Some(fg), Some(bg)) => (fg, bg),
+                        (Some(fg), None) => {
+                            let bg = if self.use_reverse_video_cursor(&params) {
+                                params.fg_color
+                            } else {
+                                params.cursor_bg
+                            };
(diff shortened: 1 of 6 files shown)


Erog38:
Just adding that I've been driving this version daily and it's been great 😄

Erog38:
I would prefer to not @ a maintainer as I know folks are super busy, so this is just my update to say that this is still working great for me. I'm trying to keep on top of merge conflicts as well.

planteman:
I just closed my PR for this. I will be using this branch until the maintainers get around to merging this. Thank you!

window/src/os/macos/window.rs
```diff
@@ -1979,7 +1979,8 @@ impl WindowView {
     }
 
     extern "C" fn selected_range(_this: &mut Object, _sel: Sel) -> NSRange {
-        NSRange::new(NSNotFound as _, 0)
+        // Required for macOS dictation support; see #4592
+        NSRange::new(0, 0)
```
bew, line 1990: Could you mention why we use 0 0 instead of something more useful ?

wezterm-gui/src/termwindow/render/mod.rs
```diff
@@ -571,10 +571,44 @@ impl crate::TermWindow {
...
+                            (fg, bg)
+                        }
+                        (None, None) => {
+                            if self.use_reverse_video_cursor(&params) {
+                                (params.bg_color, params.fg_color)
+                            } else {
+                                (params.cursor_fg, params.cursor_bg)
+                            }
+                        }
+                    }
```
bew: I think you can simplify the match with this, to reduce logic duplication: 🤔
```suggestion
                    (
                        compose_fg.unwrap_or_else(||
                            if self.use_reverse_video_cursor(&params) {
                                params.bg_color
                            } else {
                                params.cursor_fg
                            }
                        ),
                        compose_bg.unwrap_or_else(||
                            if self.use_reverse_video_cursor(&params) {
                                params.fg_color
                            } else {
                                params.cursor_bg
                            }
                        )
                    )
```

bew (commented):
Looks nice!
Could you tell me how I can test this PR ? (never used the dictation feature)

It would also be awesome if you could add some documentation about this 🤔

window/src/os/macos/window.rs
Erog38: Simply because my experience in this space is very limited, if you have a recommendation on how to get the current cursor location and inject it here I'd love to take it into account! This simply adds the dictation input at the bottom of the terminal for preview which gets injected into the text location on keypress.

Erog38:
@bew Apologies on letting this sit a while, I've addressed the feedback and added to the appearance doc for the new color fields. To test, you need a macbook and to turn on dictation in System Settings > Keyboard > Dictation. I recommend also setting the "Shortcut" to something you can be familiar with. 

Then build this with cargo and open the new app. 

When it comes up, hit the shortcut you defined in Dictation settings and speak into your active microphone. you should see text be written directly on the terminal line, if you have a tui with text input like with Claude Code, the cursor doesn't align, but that's a known limitation since we can't grab the current cursor location from the app. 

This seems consistent with how emacs and kitty solved the problem, I updated the comment on that line to help explain.

wezterm-gui/src/termwindow/render/mod.rs
Erog38: Done, thanks!

BrennonTWilliams:
Is this going to get merged? Running into the same issue this PR addresses

window/src/os/macos/window.rs
bew: I guess it's good enough, and could be improved later on 🤔

bew (approved):
Hey @Erog38 !
I've (finally 👀) tested your changes and the dictation basically works perfectly for me 👍
Even the small mic icon was correctly at the location of the cursor in the terminal, even though we used 0,0.. 🚀

Erog38:
Thanks @bew! I've mostly switched to using wispr flow since this PR, but if there is anything I can do to bring this up to date with merge conflicts or similar I'm happy to!

bew:
No, it looks great thank you!

(you'll be notified if so 🙃)~~

Zeraphim:
Independent confirmation that the `selected_range` change here fixes third-party dictation, not just macOS's built-in voice input.

Environment: macOS 26.5 (Darwin 25.5.0, Apple Silicon), MacWhisper 14.8.1, which inserts dictated text at the cursor of the focused element.

MacWhisper logs the failure mode explicitly. On WezTerm `20240203-110809`, every attempt aborts ~2ms after the hotkey, before any recording happens:

```
[dictationController] -------------- 🎙️ STARTING DICTATION --------------
[dictationSession]    🗣️ (EB2E) Start was called
[dictationSession]    🗣️ (EB2E) noActiveTextFieldFound
[dictationController] handleNonRecoverableError (id: EB2E). Error: 'No Active Textfield'
```

38 consecutive failures. The same app dictates into iTerm2 and Terminal fine on the same machine, and MacWhisper's "allow dictation everywhere" setting (which skips its own focused-text-field check) does not help — the failure is that the focused element reports no selected range at all.

With `NSRange::new(0, 0)`, built from `main` @ `699fd77b4`: 3 dictation sessions, 0 errors, transcript inserted at the shell prompt. MacWhisper's history table records `com.github.wez.wezterm` as the target app for the first time.

So this fixes a broader class of breakage than #4592's title suggests — any accessibility-driven text insertion, not only Apple's dictation.

Note: I opened #8160 before finding my way fully around the history here. It carries only the `selected_range` fix plus the Info.plist key, without the `compose_fg`/`compose_bg` additions, in case a minimal diff is easier to land. This PR is approved and I have no interest in competing with it — happy for #8160 to be closed in favour of this one.

Zeraphim:
@bew — checking in on this one.

You approved it on April 5th, and your first merges on the repo landed June 4th, so I suspect this just predates you having merge access and never came back around. It still shows as mergeable and clean.

Is anything still blocking it? Happy to do the legwork if it needs a rebase or anything else.

Thanks @Erog38 for keeping it current this whole time.

bew:
Thanks for the ping, I just merged it! 🎉

PR #7455, opened 2025-12-24
feat: add Cairo2D software rendering backend

oetiker:
## Summary

Add a pure 2D software rendering backend using Cairo, designed for environments without GPU acceleration (VNC, remote desktop, terminal servers) where the llvmpipe software GPU emulation is too CPU-intensive.

### Motivation

In terminal server setups and remote desktop environments, GPU acceleration is often unavailable. The existing `Software` front end uses llvmpipe to emulate a full GPU in software, which is highly CPU-intensive for terminal rendering where most frames are largely static text. Cairo2D bypasses GPU emulation entirely and renders directly to a pixel buffer using efficient 2D operations.

### Key Features and Optimizations

- **Line-based dirty region tracking**: Per-line content hashing detects which terminal lines have changed, re-rendering only affected lines
- **Glyph caching**: Rendered glyphs cached by glyph_id, fg/bg colors, and cell dimensions for instant reuse
- **Partial X11 updates**: Targeted `PutImage` calls update only dirty regions (50-95% bandwidth savings)
- **Frame reuse**: Complete frame hash detection skips rendering when content is unchanged
- **Efficiency metrics**: Track bandwidth savings over 1s/10s/60s time windows

### Implementation

- New `FrontEndSelection::Cairo2D` variant in config
- `CairoTexture` implementing `Texture2d` for glyph atlas
- Cairo2D rendering path in `call_draw_cairo2d()`
- X11 `present_software_frame` and `present_software_frame_region` methods
- `srgb8_to_linear_u8` helper for gamma-correct glyph blending
- Documentation in `docs/config/lua/config/front_end.md`

### Platform Support

- **X11** (Linux): Full support including partial updates
- **Wayland**: Not planned, but possible to add

### Usage

```lua
return {
  front_end = "Cairo2D",
}
```

## Test Plan

- [x] Manual testing on X11 with VNC
- [x] Verified glyph rendering, colors, selection highlighting
- [x] Tested partial updates during scrolling and cursor movement
- [ ] Cargo tests pass (CI verification needed)

### DIFF
--- a/wezterm-gui/src/termwindow/render/draw_cairo.rs
+++ b/wezterm-gui/src/termwindow/render/draw_cairo.rs
@@ -551,7 +551,21 @@ impl crate::TermWindow {
                                 } else {
                                     0
                                 };
-                                (cache.cell_col, cache.cell_col + (cache.cell_num_cols.max(1) as usize), row, row + 1)
+                                let mut sc = cache.cell_col;
+                                let mut ec = cache.cell_col + (cache.cell_num_cols.max(1) as usize);
+                                // Expand range to cover actual pixel extent from vertex positions,
+                                // which include bearing offsets that can extend beyond cell boundaries
+                                if cell_width > 0 {
+                                    let vx1 = (quad_verts[V_TOP_LEFT].position[0] + half_width as f32)
+                                        .max(0.0) as usize;
+                                    let vx2 = (quad_verts[V_BOT_RIGHT].position[0] + half_width as f32)
+                                        .max(0.0) as usize;
+                                    let vx1_in_grid = vx1.saturating_sub(padding_left);
+                                    let vx2_in_grid = vx2.saturating_sub(padding_left);
+                                    sc = sc.min(vx1_in_grid / cell_width);
+                                    ec = ec.max((vx2_in_grid + cell_width - 1) / cell_width);
+                                }
+                                (sc, ec, row, row + 1)
                             } else {
                                 // Multi-cell quad (e.g., selection rectangle) - hash into all covered cells
                                 // Convert screen coords to pixel coords, then subtract padding to get


wezterm-gui/src/quad.rs
```diff
@@ -40,20 +40,43 @@ pub struct Vertex {
...
     pub has_color: f32,
     pub mix_value: f32,
+    // Stable glyph identifier for Cairo2D cache (ignored by GPU backends)
+    pub glyph_id: u32,
+    // Background color for Cairo2D cache (ignored by GPU backends)
+    pub bg_color: [f32; 4],
+    // Cell boundaries for Cairo2D background fill (ignored by GPU backends)
+    // cell_y is the top of the cell, cell_height is the full cell height
+    pub cell_y: f32,
+    pub cell_height: f32,
```
wez: This increases the size of the vertex buffer for all targets.
Have you measured the memory and performance overheads for this?  The increased bandwidth usage is likely very measurable and noticeable for users on 4k screens with full size windows.

wezterm-gui/src/quad.rs
```diff
@@ -398,6 +500,9 @@ impl<'a> TripleLayerQuadAllocatorTrait for TripleLayerQuadAllocator<'a> {
 #[cfg(test)]
 #[test]
 fn size() {
-    assert_eq!(std::mem::size_of::<Vertex>() * VERTICES_PER_CELL, 272);
-    assert_eq!(std::mem::size_of::<BoxedQuad>(), 84);
+    // Vertex: 96 bytes (includes cell_y, cell_height for Cairo2D)
+    // 4 vertices per cell = 384 bytes
+    assert_eq!(std::mem::size_of::<Vertex>() * VERTICES_PER_CELL, 384);
```
wez: This is a 40% memory increase with no benefit for the vast majority of users

wezterm-gui/src/quad.rs
```diff
@@ -398,6 +500,9 @@ impl<'a> TripleLayerQuadAllocatorTrait for TripleLayerQuadAllocator<'a> {
...
+    // BoxedQuad: 112 bytes (includes cell_y, cell_height for Cairo2D)
+    assert_eq!(std::mem::size_of::<BoxedQuad>(), 112);
```
wez: This is a 33% memory increase with no benefit for the vast majority of users

window/src/lib.rs
```diff
@@ -264,6 +264,36 @@ pub trait WindowOps {
...
+        &self,
+        _pixels: &[u8],
+        _width: u32,
+        _height: u32,
+    ) -> anyhow::Result<()> {
+        // Default implementation does nothing - backend-specific implementations
+        // should override this to actually present the frame
+        Ok(())
+    }
+
```
wez: This feels redundant with `present_software_frame_region` below; it's just two additional parameters that can be set to zero.  I'd rather not clutter up this interface or implementation with the repeated code

window/src/lib.rs
```diff
@@ -264,6 +264,36 @@ pub trait WindowOps {
         Ok(())
     }
 
+    /// Present a software-rendered frame to the window.
+    /// The pixels are in BGRA format (native byte order for most platforms).
+    /// This is used by the Cairo2D backend for GPU-less rendering.
+    fn present_software_frame(
+        &self,
+        _pixels: &[u8],
```
wez: My initial reaction to this is, why aren't we using one of the types in window::bitmaps for this?  I'm not hugely enthusiastic about having a raw byte buffer be part of a trait interface; it feels like an end-run around the intentional layered design.

color-types/src/lib.rs
```diff
@@ -979,6 +1004,27 @@ impl LinearRgba {
         Self(self.0, self.1, self.2, self.3 * alpha)
     }
 
+    /// Porter-Duff "over" compositing: self over other.
+    /// Blends self on top of other using self's alpha.
+    /// Blending is done in sRGB space to match Cairo's compositing behavior.
```
wez: I know that this particular module looks a bit messy, but I've spent a significant amount of time making these types and the various operations perform in the correct color spaces.  Having gone to the effort of building a `LinearRgba`, why would we want to now blend in srgb space?  It's wasteful of CPU and is not in the same color space as you'd expect to use on a `LinearRgba` instance.

I think that I'd feel happier about this if `LinearRgba::composite_over` was purely the "over" operation in linear space, and if there was also a `SrgbaTuple::composite_over` that performed the "over" operation in srgb space.

That would allow the caller to opt-in to the additional color space conversion if they required and make it more clear when a "surprise!" conversion might be happening.

wezterm-gui/src/termwindow/render/draw.rs
```diff
@@ -14,9 +14,919 @@ impl crate::TermWindow {
...
+        use crate::renderstate::VertexBuffer;
+        use crate::termwindow::cairo2d::CairoTexture;
+        use ::window::bitmaps::Texture2d;
+        use ::window::WindowOps;
+        use anyhow::Context;
+        use cairo::{Format, ImageSurface, Operator};
+        use std::cell::RefCell;
+        use std::collections::hash_map::DefaultHasher;
+        use std::hash::{Hash, Hasher};
+        use wezterm_color_types::srgb8_to_linear_u8;
```
wez: These `use` statements should all be at the top of the file

wezterm-gui/src/termwindow/render/draw.rs
```diff
@@ -14,9 +14,919 @@ impl crate::TermWindow {
         match frame {
             RenderFrame::Glium(ref mut frame) => self.call_draw_glium(frame),
             RenderFrame::WebGpu => self.call_draw_webgpu(),
+            RenderFrame::Cairo2D => self.call_draw_cairo2d(),
         }
     }
 
+    fn call_draw_cairo2d(&mut self) -> anyhow::Result<()> {
```
wez: This function is massive, with lots of types and inner functions.  Please refactor this into a separate file, eg: `draw_cairo.rs` to make it much easier to follow

wezterm-gui/src/termwindow/render/draw.rs
```diff
@@ -14,9 +14,919 @@ impl crate::TermWindow {
...
+            last_frame_hash: u64,
+            // Per-line bucket data for detecting which lines changed
+            line_buckets: Vec<LineBucket>,
+            // Efficiency tracking over time windows
+            efficiency_1s: EfficiencyWindow,
+            efficiency_10s: EfficiencyWindow,
+            efficiency_60s: EfficiencyWindow,
+        }
+
+        thread_local! {
```
wez: Please explain why thread_local (which is essentially a bunch of global variables) is being used so much in here when it would be simpler to manage via `self` or render state?

wezterm-gui/src/termwindow/render/draw.rs
```diff
@@ -14,9 +14,919 @@ impl crate::TermWindow {
...
+        // Convert linear RGB to sRGB (gamma correction)
+        #[inline]
+        fn linear_to_srgb(linear: f32) -> f64 {
+            let srgb = if linear <= 0.0031308 {
+                linear * 12.92
+            } else {
+                1.055 * linear.powf(1.0 / 2.4) - 0.055
+            };
+            srgb as f64
+        }
```
wez: Please don't add ad-hoc color conversion functions at this layer.  Please use the types from the color types crate instead; it is ok to augment that crate!

wezterm-gui/src/termwindow/render/draw.rs
```diff
@@ -14,9 +14,919 @@ impl crate::TermWindow {
...
+            v.fg_color[0].to_bits().hash(hasher);
+            v.fg_color[1].to_bits().hash(hasher);
+            v.fg_color[2].to_bits().hash(hasher);
+            v.fg_color[3].to_bits().hash(hasher);
+            v.bg_color[0].to_bits().hash(hasher);
+            v.bg_color[1].to_bits().hash(hasher);
+            v.bg_color[2].to_bits().hash(hasher);
+            v.bg_color[3].to_bits().hash(hasher);
+            v.has_color.to_bits().hash(hasher);
+        }
```
wez: Can we `derive(Hash)` on the `Vertex` type? If not, this function must not be here, and should be an explicit `impl Hash for Vertex` in `quad.rs` alongside the definition of that struct.

wezterm-gui/src/termwindow/render/draw.rs
```diff
@@ -14,9 +14,919 @@ impl crate::TermWindow {
...
+
+        // Hash vertex fields for line-level change detection (same as frame, for consistency)
+        #[inline(always)]
+        fn hash_vertex_for_line(v: &crate::quad::Vertex, hasher: &mut impl Hasher) {
+            hash_vertex_for_frame(v, hasher);
+        }
```
wez: Please remove this redundant extra function

wezterm-gui/src/termwindow/render/draw.rs
```diff
@@ -14,9 +14,919 @@ impl crate::TermWindow {
...
+                cell_width: u16,
+                cell_height: u16,
+            }
+
+            struct CachedGlyph {
+                pixels: Vec<u8>, // Pre-rendered BGRA pixels (full cell)
+            }
+
+            thread_local! {
+                static GLYPH_CACHE: RefCell<HashMap<GlyphCacheKey, CachedGlyph>> = RefCell::new(HashMap::new());
```
wez: how is the cache pruned? What keeps memory usage within reasonable levels over time?

wezterm-gui/src/termwindow/render/mod.rs
```diff
@@ -621,9 +621,11 @@ impl crate::TermWindow {
             visibility,
         ) {
             // Selected text overrides colors
+            // Composite selection_bg over the cell's original bg_color to handle
+            // selection backgrounds with alpha transparency (e.g., semi-transparent selection)
```
wez, line 629: Please explain this in more detail.
What impact does this have on the majority of users with a gpu?

wezterm-gui/src/termwindow/render/screen_line.rs
```diff
@@ -358,6 +358,11 @@ impl crate::TermWindow {
                     .with_context(|| format!("layers.allocate({cursor_layer})"))?;
                 quad.set_hsv(hsv);
                 quad.set_has_color(false);
+                // Cairo2D needs cell bounds and background color for proper rendering
+                quad.set_cell_bounds(pos_y, cell_height);
+                quad.set_bg_color(params.palette.background.to_linear());
+                // Use a special glyph_id for cursor sprites (high bit set)
+                quad.set_glyph_id(0x8000_0000 | (shape as u32));
```
wez, line 365: This needs more explanation. Why do we need an id at all, and why do we need to special case cursor sprites?

docs/config/lua/config/front_end.md
```diff
@@ -40,3 +41,101 @@ See also:
...
+* **VNC/Remote desktop** - Connections where X11 forwarding or remote display protocols are used
+* **Virtual machines** - Guests without GPU passthrough
+* **Headless servers** - Systems without display hardware
+
+In these scenarios, the `Software` front end forces the use of llvmpipe,
+which emulates a full GPU in software. This is highly CPU-intensive and
+inefficient for terminal rendering where most frames are largely static
+text. Cairo2D avoids this overhead by rendering only what has changed.
+
+### Performance Optimizations
```
wez: I appreciate this information, but I'm not sure that so much of it should be part of the `front_end` config option docs.  A lot of this feels like an architecture overview that should be part of the implementation/code rather than here in the user-facing docs.

wez (commented):
Thank you for this!

I do like the idea! I have a few questions and thoughts; I may sound critical, but it is not because I don't like what you're doing, I'm just trying to understand this better and figure out the right end state.

Firstly, the caching:

* Can you share the metrics that you're collecting?  I understand that they're likely experimental at this stage, but I'd like to see it. How effective is the caching; is it worth the maintenance overhead of that additional fork of the rendering logic?  How does performance compare to `Software`?
* How can we leverage the existing structures for the caching here, without inflating the size of data that we send to the GPU for the common case?  Can you wrap up the data into intermediate types so that the cairo caching doesn't impact the GPU rendering flow?
* Is there a simpler way to achieve a similar level of performance?  For example, the existing render logic already has some complex stuff of its own to cache by `Line` and avoid repainting unchanged lines.  Did you try using that and how effective was it?

Secondly, the interface:

* I'd appreciate you commenting here on this PR on the high level strategy for implementing this backend for Wayland, macOS and Windows. I don't require the full implementation for those platforms to be part of this PR, but I would like to see where it is going and it is better to figure out where the hard parts are likely to be up-front if that will help to steer where we're going here.
* The software frame presentation impl uses a raw byte buffer and exploits some implicit knowledge of both the termwindow impl and how we happen to set things up for X11, and also makes some assumptions about Cairo.  Those assumptions potentially will not apply for Wayland, macOS or Windows.  I'd feel happier if we at least re-used some of the types from the bitmaps module so that those other platform implementations could reason about the bitmap data with more confidence in the face of future code changes.

oetiker:
Hi @wez thanks for your comments and questions.

What prompted me to write this patch, was that I found that in our environment with a software GPU wezterm could easily go up to 300% CPU usage while running claude-code. While claude-code is inferencing, it runs little ASCII animations. 

Limiting the render pipes and optimizing the config settings I was able to get down the CPU load to 70%, but this is still huge for a terminal essentially just sitting still.

From my days writing rrdtool I had some experience with Cairo, so I wondered if we could do some optimizations by doing the rendering in a more conscientious way.  And the results are pretty nice, with my patch, the cpu load of a wezterm is now so low that it does not pop up on the radar anymore. :)

Do I get this right, that you might be interested in applying the optimizations in a way that all rendering backends could profit from them?

Will answer your questions in detail later.

wezterm-gui/src/quad.rs
oetiker: Fixed. Cairo2D cache data is now stored in parallel vectors (Cairo2DCacheData) rather than in the Vertex struct. The Vertex size remains 272 bytes (unchanged from main). Additionally, the HeapQuadAllocator now takes a uses_cairo2d flag so GPU backends don't allocate or process cache data at all.

wezterm-gui/src/quad.rs
oetiker: see above

window/src/lib.rs
oetiker: Consolidated to just `present_software_frame_region`. Callers pass `dst_x=0, dst_y=0` for full frame updates.

window/src/lib.rs
oetiker: Fair point. The raw buffer was chosen to avoid allocation overhead when presenting from Cairo's `ImageSurface::data()` directly. Would a wrapper type like `SoftwareFrameBuffer<'a>` that borrows the slice be acceptable? Or should this go through `Image/ImageTexture`?

color-types/src/lib.rs
oetiker: Fixed. Added `SrgbaTuple::composite_over` for sRGB-space blending and kept `LinearRgba::composite_over` purely in linear space. Callers can now choose the appropriate color space.

wezterm-gui/src/termwindow/render/draw.rs
oetiker: done

wezterm-gui/src/termwindow/render/draw.rs
oetiker: done

wezterm-gui/src/termwindow/render/draw.rs
oetiker: now uses wezterm_color_types::srgb8_to_linear_u8

wezterm-gui/src/termwindow/render/draw.rs
oetiker: The hash function intentionally includes `Cairo2DCacheData` alongside `Vertex` fields for frame-change detection. Since we can't derive Hash on types we don't own (floats) and the hash needs both structs, would it be acceptable to keep this as a free function but move it to `quad.rs` alongside the types?

wezterm-gui/src/termwindow/render/mod.rs
oetiker: The problem was that the selection did optically not look correct without this change, but maybe I missunderstood the intent.

wezterm-gui/src/termwindow/render/draw.rs
oetiker: Moved to `TermWindow.cairo2d_state: RefCell<Cairo2DRenderState>`. Each window now manages its own rendering state.

wezterm-gui/src/termwindow/render/screen_line.rs
oetiker: The `glyph_id` provides cache lookup stability for the glyph cache. Cursor sprites get special IDs (starting at `0x8000_0000`) because they're synthetic quads not from the font rasterizer.

Should I add an inline comment?

docs/config/lua/config/front_end.md
oetiker: trimmed

wezterm-gui/src/quad.rs
oetiker: see above

wezterm-gui/src/termwindow/render/draw.rs
oetiker: Added `LfuCache` with a configurable maximum entry count. The glyph cache uses least-frequently-used eviction to bound memory.

oetiker:
Pushed updates addressing the review feedback:

Memory/performance (GPU users):
  - Vertex struct unchanged at 272 bytes - Cairo2D cache data stored in parallel Cairo2DCacheData vectors
  - HeapQuadAllocator now backend-aware via uses_cairo2d flag
  - GPU path uses extend_with() and QuadImpl::Boxed - original code path unchanged
  - Cairo2D path uses extend_with_cairo2d() which delegates to extend_with() then handles cache

Code structure:
  - Refactored to draw_cairo.rs
  - thread_local replaced with TermWindow.cairo2d_state
  - Color conversions use wezterm_color_types crate
  - present_software_frame consolidated to single present_software_frame_region
  - Docs trimmed to focus on configuration
  - Glyph cache uses LfuCache for bounded memory

Still open:
  - Raw &[u8] buffer in present_software_frame_region vs window::bitmaps type
  - hash_vertex_for_frame location (currently in draw_cairo.rs)

oetiker:
I pushed my benchmarking tool

```
   Terminal load generator for measuring rendering performance with
   configurable update patterns, rates, and content variation.

   Includes benchmark-render.pl script for automated comparison between
   Software (llvmpipe) and Cairo2D backends.

   Benchmark results (8s duration, 120x40 terminal):

   Scenario             | Software | Cairo2D | CPU Saved | RAM Saved
   ---------------------|----------|---------|-----------|----------
   Idle (prefilled)     |      10% |      6% |     +41%  |     +55%
   Typing (1 char @30Hz)|     290% |     64% |     +78%  |     +50%
   Single cell @120Hz   |     348% |     66% |     +81%  |     +53%
   Full screen @30Hz    |     272% |     55% |     +80%  |     +34%
   Scrolling output     |     321% |     49% |     +85%  |     +40%

   Cairo2D averages 73% less CPU and 46% less memory than Software.
```

oetiker:
I found that the glyph cache was too small ... with 8196 things look just a bit better still

```
  Scenario             | Software | Cairo2D | CPU Saved | RAM Saved
  ---------------------|----------|---------|-----------|----------
  Idle (prefilled)     |      11% |      6% |     +44%  |     +55%
  Typing (1 char @30Hz)|     281% |     56% |     +80%  |     +50%
  Single cell @120Hz   |     338% |     59% |     +83%  |     +53%
  Full screen @30Hz    |     264% |     53% |     +80%  |     +35%
  Scrolling output     |     297% |     48% |     +84%  |     +37%
```

oetiker:
After implementing cell-level dirty tracking for Cairo2D, selection highlighting had two visual bugs:

1. Color mismatch: Selected cells with glyphs appeared a different shade than empty selected cells. This was caused by selection_bg having alpha=0.5, which produced different results when blended in linear color space (glyph backgrounds pre-composited in Rust) vs sRGB color space (Cairo's selection rectangle blending).
2. Update delay: When selecting text, glyph cells would highlight immediately while empty cells lagged behind (and vice versa when deselecting). This was because the selection rectangle quad was only hashed into its top-left corner cell for dirty tracking, leaving covered cells unaware of the change.

Solution

Selection color consistency:
  - Added is_selected: bool to Cairo2DCacheData to track selected cells
  - For selected glyph cells, skip the pre-composited background fill and let Cairo's selection rectangle provide the background
  - Skip the glyph cache for selected cells (cached pixels include background)
  - Result: both glyph and empty cells use consistent sRGB-space alpha blending

Multi-cell dirty tracking:
  - Selection rectangles (and other multi-cell quads) now hash into ALL covered cells, not just the corner
  - Cell range is calculated from the quad's bounding box with proper padding offset subtraction
  - Fixes the stray artifacts at line start and selection boundaries

Testing

  Verified with manual selection testing and the benchmark suite. Performance remains excellent:
  - CPU usage: 76.6% less than Software renderer (average)
  - Memory usage: 43.0% less than Software renderer (average)

oetiker:
replaced this with #7607,  #7608, #7609

PR #7457, opened 2025-12-25
docs(linuxbrew install): formula flag

cultab:
`brew install wezterm` attempts to install the macOS package and fails with `Error: macOS is required for this software.` because it treats `wezterm` as a cask, as per the warning displayed before the error:

 Warning: Treating wezterm as a cask. For the formula, use wezterm/wezterm-linuxbrew/wezterm or specify the `--formula` flag. To silence this message, use the `--cask` flag.

### DIFF
--- a/docs/install/linux.md
+++ b/docs/install/linux.md
@@ -281,13 +281,13 @@ hide:
 
     ```console
     $ brew tap wezterm/wezterm-linuxbrew
-    $ brew install wezterm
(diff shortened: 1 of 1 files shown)


bew:
Related to #7445 which goes the other way

bew:
I've merged a similar PR with this change https://github.com/wezterm/wezterm/pull/7768

Issue #6068, opened 2024-08-31
Glyphs look bad and box drawing


Kaikacy:
### What Operating System(s) are you seeing this problem on?

Linux X11

### Which Wayland compositor or X11 Window manager(s) are you using?

qtile

### WezTerm version

wezterm 20240812-215703-30345b36

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

No, and I'll explain why below

### Describe the bug

box drawing doesn't look that good and also some glyphs, too. like u+25e5 and some others that contain diagonal lines
look weird.  
wezterm:
![31-08-24_18:04:37_a](https://github.com/user-attachments/assets/af1b59e1-00bc-4a5a-b6d4-eee1b667a3a5)
kitty:
![31-08-24_18:04:50_a](https://github.com/user-attachments/assets/2e649685-b65c-44a0-9dab-5581f6d18da5)

wezterm line drawing (this might be intentional):
![31-08-24_18:08:12_a](https://github.com/user-attachments/assets/6b0188a0-a47f-49bf-b7ef-e5dd7fe63d34)
kitty line drawing:
![31-08-24_18:08:31_a](https://github.com/user-attachments/assets/79907523-d8bf-4277-a817-afc253480d62)

i am using jetbrains mono nerd font

### To Reproduce

print triangle unicode characters (like u+25e5) and box drawing glyphs (again it look weird too because i switched from kitty)

### Configuration

```lua
c.freetype_load_target = "HorizontalLcd" -- better anti-aliasing
c.custom_block_glyphs = true
c.anti_alias_custom_block_glyphs = true
```
i can't disable `custom_block_glyphs`, then it looks terrible and anti_aliasing doesn't do anything for triangle glyphs,
`freetype_load_target` has to be HorizontallLcd for some characters too look good (that depends on monitor)

### Expected Behavior

display triangle glyphs correctly like kitty does

Wukuyon:
I tested the rendering of U+25E2, U+25E3, U+25E4, and U+25E5 (`◢ ◣ ◤ ◥`) on wezterm 20241104-083404-51c794ac with `custom_block_glyphs = true` on macOS 14.7.1 (23H222) with the same NeoVim semigraphics art. WezTerm gave different but still bad results:

<img width="686" alt="Neovim semigraphics art on WezTerm" src="https://github.com/user-attachments/assets/787c8495-a040-4634-8628-f4067b1d658b">

[...]

PR #7466, opened 2026-01-02
add custom black triangles from geometric shapes

sh1boot:
Fix for #6068.  Add the solid black corner triangles from the Geometric Shapes Unicode block.


sh1boot:
How do I make something happen here?

PR #7469, opened 2026-01-02
wezterm-blob-leases: Use uuid v4 over v1

yiding:
Greetings. I'm looking at how much effort it is to make an ipad build, and one issue is dependencies that don't support ios targets. One of those is mac_address, which seems only to be used for creating uuids.

Unless we really want v1 uuids for some reason (e.g. ordering... but the LeaseId doesn't implement Ord), changing to v4 seems like a good way to get rid of the mac_address dependency.

Alternatively, we can use v7 uuid which is random + timestamp instead of mac + timestamp.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -358,15 +358,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "atomic"
-version = "0.6.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


wez:
Greetings! I hope you are well!
Yeah, this sounds fine, thank you!  Will be interested to see how far you get with an ipad build :-)

PR #7476, opened 2026-01-07
Fix install script copy buttons not stripping `$` prefix

spenserblack:
The JavaScript that makes the "copy to clipboard" buttons remove the `$` prefix was failing to select the buttons due to a mismatch in the expected buttons' class. This updates the selector in the JavaScript.

It also changes the BSD install code blocks from `bash` to `console`, since that's what the selector expects.

---

BTW, this terminal is fantastic :heart:

### DIFF
--- a/docs/install/freebsd.md
+++ b/docs/install/freebsd.md
@@ -4,7 +4,7 @@ WezTerm is available [via the ports system](https://www.freshports.org/x11/wezte
 
 The most common FreeBSD architectures have pre-built binaries which you can install via:
 
-```bash
+```console
 $ pkg install wezterm
 ```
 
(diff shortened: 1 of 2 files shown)


wez:
Thanks for this, but there is supposed to be some javascript that automatically removes the `$` when you copy.  Could you take a look a debugging why that is no longer functioning (for you?) instead of changing the docs?

Thanks!

spenserblack:
Oh, OK, I wasn't aware of that. Sure, that's a much smarter solution :+1:

spenserblack:
Well that was much simpler :sweat_smile:

I'm guessing that, previously, the buttons had the `.md-copy` class, but that's no longer the case. I just swapped out the classes.

Alternatively, we can select on the `data-md-type="copy"` attribute instead of the class, or perhaps even just naively assume that the only button inside a code block will be the copy button and make the selector less specific. Depends on how stable we expect the button's classes/attributes to be.

wez:
re: stability, at some point this year we'll be switching over to https://github.com/zensical/zensical and we can figure out what might need to be tweaked then!

wez:
Thank you!

PR #7481, opened 2026-01-08
ringlog: make printing to stderr faster

nagisa:
I was trying to debug an issue I'm seeing with wezterm last night and figured I would try to use the terminal for my day-to-day with trace logs enabled while I try to figure it out.

Unfortunately it turned out to be disruptive enough to make the terminal almost unusable. I narrowed at least *that* part down to the fact that printing out each log message would result in about screenful worth of `write(2)` syscalls, many of them printing just one space character each (padding? debug structure identation?)

Although it would be ideal to use a line-based buffered writer here, pre-formatting the log string to a String before printing seemed like a straightforward quick solution and it works amazingly well!


wez:
Thank you!

PR #7483, opened 2026-01-08
Fix macOS notifications to display as toast popups

### LINKED ISSUE #5476: OSC 9 escape code does not display a toast notification, but they are in the notifications list when checked.

macOS

_No response_

### WezTerm version

20240520-135708-b8f94c47

### Describe the bug

I've tried sending the OSC 9 escape code from multple shells and devices and get no toasts. The first time I did get the MacOS allow notifications prompt and chose yes. If I click on the clock to show notifications they are there, but they never pop up as expected. 

### To Reproduce

send a notification with the OSC 9 escape sequence.

### Configuration

no config

### Expected Behavior

The notification will pop up under the clock, sound an alert, and dwell for a few seconds like all other notifications.

### Logs

debug log is empty

### Anything else?

When using the version from homebrew I didn't even get the hidden notifications. They started appearing when I tried on the nightly.

nikhilm:
Add UNNotificationPresentationOptions::Banner and Badge flags to notification presentation options in macos.rs. This ensures that OSC 9 escape code notifications and window:toast_notification() calls display as visible toast popups on macOS with badge indicators, not just entries in the Notification Center.

Fixes issue where notifications were only shown in Notification Center with sound, but no visible banner popup appeared.

Thanks to @joshuataylor for identifying the issue and suggesting the fix: https://github.com/wezterm/wezterm/issues/5476#issuecomment-3652352095

Fixes #5476

### DIFF
--- a/wezterm-toast-notification/src/macos.rs
+++ b/wezterm-toast-notification/src/macos.rs
@@ -47,8 +47,10 @@ define_class!(
             completion_handler: &block2::Block<dyn Fn(UNNotificationPresentationOptions)>,
         ) {
             log::debug!("will_present");
-            let options =
-                UNNotificationPresentationOptions::List | UNNotificationPresentationOptions::Sound;
+            let options = UNNotificationPresentationOptions::List
+                | UNNotificationPresentationOptions::Sound
+                | UNNotificationPresentationOptions::Badge
+                | UNNotificationPresentationOptions::Banner;
             completion_handler.call((options,));
         }
 


joshuataylor:
I have tested this, and also figured [Badge](https://developer.apple.com/documentation/usernotifications/unnotificationpresentationoptions/badge) might be good too.

I also signed up for an Apple Developer account, as notifications won't show unless you've codesigned (it throws an error), and I couldn't be bothered figuring out self-signing.

`codesign -s target/release/wezterm`

```
-            let options =
-                UNNotificationPresentationOptions::List | UNNotificationPresentationOptions::Sound;
+            let options = UNNotificationPresentationOptions::List
+                | UNNotificationPresentationOptions::Sound
+                | UNNotificationPresentationOptions::Badge
+                | UNNotificationPresentationOptions::Banner;
```

I also find MacOS settings/permissions odd -- once I did this, it worked on the regular nightly Wezterm. :shrug:

(I never ended up raising a PR as I got distracted with other things, thanks for this!)

wez:
Thanks to you both for this!  I feel like Badge might be polarising for certain users.  For example, I tend to go out of my way to suppress badges for most apps, as I find it to be a tax on my attention.  Does setting Badge by default prevent the user from switching off the badge in their notification preferences?

josh-tlw:
On another computer right now, but no, and it lists what "permissions" it has on the Notifications System Settings.

MacOS 15.7.3

image

Wezterm nightly:
image

image

An application with more permissions:
image

nikhilm:
@wez could this be merged given Joshua has tested it already?

wez:
Thank you!

joshuataylor:
Can confirm, with the latest nightly as of 2026-01-18 (wezterm 20260117-153515-30ef869d), on MacOS 15.7.3 (24G419), I can see the notifications as expected!

Following https://wezterm.org/config/lua/config/notification_handling.html :

```sh
printf "\e]777;notify;%s;%s\e\\" "title" "body"
```

wez3


```sh
printf "\e]9;%s\e\\" "hello there"
```

image


:tada:

PR #7491, opened 2026-01-10
Fix open hyperlink bindings with mouse reporting

### LINKED ISSUE #4536: CTRL-click does not work for opening links when bypass_mouse_reporting_modifiers is also CTRL

Windows

_No response_

### WezTerm version

20230828-220414-99c96139

### Describe the bug

If both the bypass_mouse_reporting_modifiers and the OpenLinkAtMouseCursor modifiers are CTRL, I can't open links with CTRL-left in a terminal app with mouse reporting. It works fine otherwise.

I would like to open links with ctrl-left click, regardless of whether the app supports mouse reporting.

### To Reproduce

Apply config below and click a link with ctrl-left

### Configuration

  bypass_mouse_reporting_modifiers = 'CTRL',
  mouse_bindings = {
    -- Change the default click behavior so that it only selects
    -- text and doesn't open hyperlinks
    {
      event = { Up = { streak = 1, button = 'Left' } },
      mods = 'NONE',
      action = act.CompleteSelection 'ClipboardAndPrimarySelection',
    },
    -- and make CTRL-Click open hyperlinks
    {
      event = { Up = { streak = 1, button = 'Left' } },
      mods = 'CTRL',
      action = act.OpenLinkAtMouseCursor,
    },
  },

### Expected Behavior

Open link on ctrl-click in an app with mouse reporting

### Logs

_No response_

### Anything else?

_No response_

rkitover:
Support default and user bindings for `OpenLinkAtMouseCursor` and `CompleteSelectionOrOpenLinkAtMouseCursor` in mouse reporting mode using the same binding.

Fix #4536.

### DIFF
--- a/docs/recipes/hyperlinks.md
+++ b/docs/recipes/hyperlinks.md
@@ -141,8 +141,7 @@ This setup sends the text of the commands directly into the active pane which ha
 - When a shell is not detected (see the `is_shell` function), WezTerm falls back to a long shell command, which may clutter the prompt slightly. This happens because WezTerm cannot directly determine a file's MIME type when not connected to a local shell.
 
 If you're using [tmux](https://github.com/tmux/tmux), you'll need to enable the
-hyperlinks terminal feature and, depending in your configuration
-(see [bypass mouse reporting modifiers](../config/lua/config/bypass_mouse_reporting_modifiers.md)), add `Shift` when clicking an hyperlink.
+hyperlinks terminal feature.
 
 ```tmux
 set -sa terminal-features ",*:hyperlinks"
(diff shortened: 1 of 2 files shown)


bew:
This looks like a complex event handling change for something that can easily be done in config only (I've posted my impl of this in #4536)

rkitover:
That is a good solution indeed, but I think it may still be worthwhile to consider adding this behavior.

bew:
I don't think it's worth adding, the behavior can be done in config quite easily, there is no need to add a special handling for this specific thing in the core..

PR #7504, opened 2026-01-15
Fix parsing of partial SGR mouse sequences

### LINKED ISSUE #7076: Termwiz API change/addition request
In Zellij, we're using the termwiz InputParser to parse STDIN. We are experiencing some difficulty when handling segmented input. When termwiz identifies the input as ambiguous, this is (mostly - see below) not an issue, since we can debounce a timer and call the `parse` method again with `maybe_more = false` and have the ambiguous bytes in the buffer handled.

I've found at least one case in which input that is ambiguous is not marked as ambiguous:

```
SEGMENT 1: "\u{1b}[<0;42;17M\u{1b}[<0;"
SEGMENT 2: "42;17m"
```

Here we receive the first mouse event correctly, and the rest goes off the rails (iirc one ESC and the rest as bare text even before the second segment, I can reproduce the exact details if necessary).

An additional problem is that since we need to do some post-processing, we really need to know which raw bytes correspond to which event in the original buffer.

It would be great if the above case would be categorized as ambiguous. It would also be great if the parser returns if not the exact raw bytes that correspond to each event, then their length so that we can infer the bytes for post-processing.

N/A

**Additional context**
Additionally, I am wondering if there are other similarly ambiguous sequences that are not classified as such? Perhaps because of a different requirement I am not thinking about or aware of? If so, would be nice to have an additional stricter mode that classifies all as ambiguous unless they're an exact match.

If not a quick fix on your side, I'd be happy to throw together a PR (haven't looked at the relevant code in a while, so not sure the scope here).

jgiannuzzi:
Fixes #7076

### DIFF
--- a/termwiz/src/input.rs
+++ b/termwiz/src/input.rs
@@ -1376,6 +1376,13 @@ impl InputParser {
                             }
                             continue;
                         }
+
+                        // If we have a potential SGR mouse sequence prefix and more
+                        // data might come, wait for additional input before falling
+                        // through to keymap lookup
+                        if maybe_more && self.buf.as_slice().starts_with(b"\x1b[<") {
+                            return;
+                        }
                     }
 
                     match (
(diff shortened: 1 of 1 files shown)


wez:
Look good, thank you!

imsnif:
Hey @wez - apologies for the ping. Would you be open to cutting a new release with this fix? We depend on it in Zellij. I could always vendor termwiz if this is blocked by something or you don't have the capacity. Thanks!

garrettjavalia:
I want to see this in relrease, too!

Issue #81, opened 2019-12-24
Unable to compile freetype error


SevereOverfl0w:
## Is it a build problem?

```
   Compiling freetype v0.1.0 (/builddir/wezterm-20191218-101156-bf35707/deps/freetype)
error: failed to run custom build command for `freetype v0.1.0 (/builddir/wezterm-20191218-101156-bf35707/deps/freetype)`

Caused by:
  process didn't exit successfully: `/builddir/wezterm-20191218-101156-bf35707/target/release/build/freetype-2592e6c7ec0b4511/build-script-build` (exit code: 1)
--- stdout
TARGET = Some("x86_64-unknown-linux-gnu")
OPT_LEVEL = Some("3")
HOST = Some("x86_64-unknown-linux-gnu")
CC_x86_64-unknown-linux-gnu = None
CC_x86_64_unknown_linux_gnu = None
HOST_CC = None
CC = Some("cc")
CFLAGS_x86_64-unknown-linux-gnu = None
CFLAGS_x86_64_unknown_linux_gnu = None
HOST_CFLAGS = None
CFLAGS = Some("-fstack-clash-protection -D_FORTIFY_SOURCE=2 -mtune=generic -O2 -pipe   ")
CRATE_CC_NO_DEFAULTS = None
DEBUG = Some("true")
CARGO_CFG_TARGET_FEATURE = Some("fxsr,sse,sse2")
running: "cc" "-O3" "-ffunction-sections" "-fdata-sections" "-fPIC" "-g" "-fno-omit-frame-pointer" "-m64" "-fstack-clash-protection" "-D_FORTIFY_SOURCE=2" "-mtune=generic" "-O2" "-pipe" "-I" "zlib" "-DHAVE_SYS_TYPES_H" "-DHAVE_STDINT_H" "-DHAVE_STDDEF_H" "-D_LARGEFILE64_SOURCE=1" "-o" "/builddir/wezterm-20191218-101156-bf35707/target/x86_64-unknown-linux-gnu/release/build/freetype-f0a65fb7ff931d73/out/zlib-build/zlib/adler32.o" "-c" "zlib/adler32.c"
cargo:warning=cc: error: zlib/adler32.c: No such file or directory
cargo:warning=cc: fatal error: no input files
cargo:warning=compilation terminated.
exit code: 1
[...]

PR #7508, opened 2026-01-17
Start stable branch

abitrolly:
Run `apt-get update` and start stable branch

Stable branch is needed to update dependencies and track regressions, such as failure to build from source (https://github.com/wezterm/wezterm/issues/81) occurring in fresh Ubutu 24.04.

This PR should not be merged to `main`. Instead a new branch `stable` needs to be created from the tag `20240203-110809-5046fc22` and the PR needs to be rebased against it.

(t can be used then to test and patch regressions such as https://github.com/flathub/org.wezfurlong.wezterm/pull/23#issuecomment-3761813698 without waiting for `main` to become stable.


wez:
Thanks, but I explicitly do not want to maintain branches, it doesn't match the development model I use here.
You can maintain your own branches and forks, but I do not want them in the main wezterm project.

Issue #3098, opened 2023-02-12
Event on pane change


eugenesvk:
**Is your feature request related to a problem? Please describe.**
I'd like to activate a key table whenever I switch to a pane that has a specific process (and deactivate on switching to a pane without it).

**Describe the solution you'd like**
Being able to subscribe to an event that fires on pane switching (which should receive `window` and `pane` objects)

**Describe alternatives you've considered**
I know how to check the process and activate a key table and even how to add that check into my own pane/tab switching keybinds
However, I don't understand how to make it a general thing, e.g., when I switch tab with a mouse, obviously my hooks in a keybind don't fire

**Additional context**

b0o:
+1, this would be useful to be able to implement a key binding to focus the previously focused pane

PR #7509, opened 2026-01-17
termwindow: emit tab-focus-changed window event (solves #3098)

JornC:
## Summary

Emits a `tab-focus-changed` window event when a tab receives focus, enabling Lua scripts to respond to tab focus changes. For example, updating a diff viewer's working directory to match the newly focused tab.

Solves https://github.com/wezterm/wezterm/issues/3098

## Changes

- Emit tab-focus-changed window event in activate_tab() when switching to a different tab
- Added documentation for the new event

## Test plan

- Verified event fires when switching tabs via wezterm.on('tab-focus-changed', ...)
- Confirmed window and pane parameters are passed correctly to handler
- Verified event does not fire when clicking the already-active tab

### DIFF
--- a/docs/config/lua/window-events/tab-focus-changed.md
+++ b/docs/config/lua/window-events/tab-focus-changed.md
@@ -1,0 +1,31 @@
+# `tab-focus-changed`
+
+{{since('nightly')}}
+
+The `tab-focus-changed` event is emitted when switching to a different tab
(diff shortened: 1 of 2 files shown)


JornC:
Closed because I realised a "pane" focus event is more appropriate than a "tab" focus. See #7510

PR #7510, opened 2026-01-17
termwindow: emit pane-focus-changed window event

### LINKED ISSUE #3098: Event on pane change

JornC:
Emits a `pane-focus-changed` window event when the focused pane in a window changes.

Closes https://github.com/wezterm/wezterm/issues/3098

## Background

I originally wanted a "tab focus changed" event and went down a rabbit hole trying to pass a `MuxTab` to Lua. Then I realized wezterm is pane-centric: tabs are containers, panes are what matter, and every window event passes `(window, pane)` — so `pane-focus-changed` is the right abstraction.

## Approach

The gui window tracks its last focused pane id and emits `pane-focus-changed` whenever that changes, driven off the notifications that accompany a focus change (`PaneFocused`, `PaneAdded`, `PaneRemoved`, `TabAddedToWindow`, `WindowInvalidated`). This covers switching panes, switching tabs, and — per @bew's review — creating and destroying tabs and panes. `PaneAdded` is now forwarded to the window handler so splits are observed, and the tracker is primed at window construction so the first focus change isn't swallowed.

An earlier revision broadcast `MuxNotification::PaneFocused` from `activate_tab()` instead; that's dropped in favor of the change-detector, since tab switches already reach it via `WindowInvalidated`, and broadcasting an overlay pane id (target tab in copy-mode/search) caused spurious focus-reconcile errors.


JornC:
Is there anything I can do to help this PR get a review? My local workflow has successfully been using this patch for the past few weeks.

bew (approved):
no it looks good 👍
We're mainly waiting for Wez at this point, he has a huge backlog of catching up to do on the project 😬

JornC:
@bew since the repo seems to have begun moving along 👌🏻 I'd like to resurface this PR for review -- still using it for my daily driver, happy to address any feedback whenever someone has bandwidth. No rush, thanks!

bew:
Hello @JornC !

I updated the branch with latest `main` changes ;)

⚠️ I noticed that the event doesn't fire when creating a new tab or pane, nor when destroying a tab or pane.
👉 Could you look into this?

---

Small config used for testing:
```lua
local wezterm = require"wezterm"

wezterm.on("pane-focus-changed", function(win, pane)
  print("PANE FOCUS CHANGED! -- WIN:", win, "PANE:", pane)
end)

return {}
```

JornC:
Thanks @bew! Fixed - it now fires on tab/pane create and destroy too.

bew (approved):
~~Thank you for the update, the new impl looks good! 🚀~~

bew:
Actually thinking about it more, what do you think about only watching `PaneFocused` mux event at window-level, and add new places to emit this event 🤔

_I'd suggest trying this in a separate PR, so we can use this one if that other impl turns out to not be viable._

PR #7516, opened 2026-01-22
Update user-var-changed.md

adamh-oai:
Make it clearer that wrap is necessary for long values, in case someone (not me, mind you) didn't read carefully and spent too much time debugging.

### DIFF
--- a/docs/config/lua/window-events/user-var-changed.md
+++ b/docs/config/lua/window-events/user-var-changed.md
@@ -8,7 +8,7 @@ used to set a user var.
 You can use something like the following from your shell:
 
 ```bash
-printf "\033]1337;SetUserVar=%s=%s\007" foo `echo -n bar | base64`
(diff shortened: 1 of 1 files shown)


docs/config/lua/window-events/user-var-changed.md
```diff
@@ -8,15 +8,15 @@ used to set a user var.
 You can use something like the following from your shell:
 
 ```bash
-printf "\033]1337;SetUserVar=%s=%s\007" foo `echo -n bar | base64`
+printf "\033]1337;SetUserVar=%s=%s\007" foo `echo -n bar | base64 --wrap=0`
```
bew, line 11: Note that on macos the `--wrap` option doesn't exist, so I think it should stay only in the note below 🤔

adamh-oai: Fair point, will close.

PR #7526, opened 2026-01-25
fix(wezterm-gui): max tab title length

erics118:
Fixes the logic regarding tab title length.

This is a really weird detail that took a fair bit of debugging to find

The `format-tab-title` event currently receives an incorrect `max_width` parameter. Right now, `max_width` is basically the minimum title length possible. 

example config to reproduce: (it basically displays the max_width inside the tab title)

```lua
config.tab_max_width = 50
wezterm.on("format-tab-title", function(tab, tabs, panes, conf, hover, max_width)
    return {
        { Text = max_width .. tab.active_pane.title },
    }
end)
```

video:

https://github.com/user-attachments/assets/e0252da4-41e4-41b7-9092-e0c5bfbcd009

observe how the title for each tab is `5zsh`. The value of `tab.active_pane.title` was `" zsh "` (observe spaces on left/right for padding), which has a length of 5.

The expected tab title should be `50zsh`. this change fixes it.

Previously, the code was using the variable `tab_title_len`, which is the length of the previous tab title, while it should be using `tab_width_max`, which is the max width allowed for the tab title.

new video:

https://github.com/user-attachments/assets/936ed4e9-9f7a-47cb-adc5-2dd681f60e1a

### DIFF
--- a/wezterm-gui/src/tabbar.rs
+++ b/wezterm-gui/src/tabbar.rs
@@ -465,7 +465,7 @@ impl TabBarState {
                 pane_info,
                 config,
                 hover,
-                tab_title_len,
+                tab_width_max,
             );
 
             let cell_attrs = if active {


bew (approved):
Oh! Nice catch!

vjohansen:
What is the bug? 

It is documented as " the maximum number of cells available to draw this tab when using the retro tab bar style.". This works as expected. it would be nice if similar behavior was available when retro=false (return tab width)

https://wezterm.org/config/lua/window-events/format-tab-title.html

Do you want it pass the config.tab_max_width in the function call?

erics118:
WDYM works as expected?	

My understanding is that the correct value of `max_width` should be 50 in my example as there is more than enough space for the tabs to be 50 chars wide. (the screen recording was just a small part of the entire wezterm window)

Currently, it sets `max_width` to 5 due to using the variable `tab_title_len` (which is the current tab width) rather than `tab_width_max` (the max width of tab allowed).

ie, `max_width` is always equal to `#tab.active_pane.title`.

PR #7548, opened 2026-02-02
Treat SCTKWindowState::TILED as a maximized state

wmww:
I'm also encountering #7156. It looks like what happened is `SCTKWindowState::TILED` got added, and replaced the `SCTKWindowState::MAXIMIZED` in at least the tiling window manager case. Wezterm isn't handling this state, and so it was treating tiled windows as resizable, and then you end up rendering at the wrong size. There could be other problems, the full resizing logic is rather complex, but this seems more correct to me and fixes my usecase.


bew:
Alternative PR: #7966 by @hamano which adds a new TILED state for wezterm and excludes it in `can_resize`.

I like this and plan to merge it, let's close this PR.

Issue #489, opened 2021-02-17
Toast Notifications


wez:
This issue tracks the state of toast notification support.

## OS level support

Whether we can actually show a notification in the desktop environment.

|OS               | Basic Notification  | Click Handling  |
| --------------|------------|-------------|
|Linux           | Yes         |  Yes         |
|*BSD     | Yes, in nightly | <-- |
|macOS       | Yes          | Yes         |
|Windows    | Yes          | Yes  |

## Escape sequence support

Do we have plumbing to allow an app to show a notification.

|Sequence | Implemented | Reference |
|--------------|--------------------|---------------|
| OSC 9     | 20210314-114017-04b7cedd | `printf "\e]9;%s\e\\" "hello there"`  from original [iTerm](https://iterm2.com/documentation-escape-codes.html). Note that this conflicts with [ConEmu](https://conemu.github.io/en/AnsiEscapeCodes.html#ConEmu_specific_OSC) |
| OSC 99  | No                   | `printf "\e]99;;%s\e\\" "hello there"` [kitty spec](https://github.com/kovidgoyal/kitty/blob/master/docs/protocol-extensions.rst#desktop-notifications) |
| OSC 777 | 20210314-114017-04b7cedd | `printf "\e]777;notify;%s;%s\e\\" "title" "body"` [rxvt - call perl extension named "notify"](https://manpages.debian.org/testing/rxvt-unicode/urxvt.7.en.html) |

## Multiplexer support

Whether notifications can traverse multiplexer boundaries

|Protocol | Supported |
|-----------|----------------|
|wezterm mux codec | Nightly |
|tmux CC | No |

bew:
NOTE: OSC99 is not only kitty's spec, it is based on an ongoing work to properly define a spec for desktop notifications available here: https://gitlab.freedesktop.org/terminal-wg/specifications/-/issues/13

bew:
Could we have an option to configure the urgency of the raised notification? Currently it is ALWAYS at the `critical` level.
I can't find a way to configure this in the OSC's documentation so I guess the terminal should have that.

An option like `notification_urgency = "normal"`

On linux `notify-send` can specify urgency with `-u URGENCY`, like: `notify-send -u normal Hello!`

The urgency levels are: low, normal, critical (see https://developer.gnome.org/notification-spec/#urgency-levels)
It is passed in the hints dict: `"urgency" | byte | The urgency level`

wez:
[...]

PR #7555, opened 2026-02-03
fix: set DBus notification priority to normal

### LINKED ISSUE #7553: Toast notifications on KDE plasma do not expire

Linux Wayland

Wayland KWin on KDE Plasma

### WezTerm version

wezterm 20251014-193657-64f2907c

### Describe the bug

According to my investigations, this is not _really_ wezterm's bug, rather incongruency in expectations.

On dbus systems, Wezterm generates notification requests with Critical urgency:
https://github.com/wezterm/wezterm/blob/05343b387085842b434d267f91b6b0ec157e4331/wezterm-toast-notification/src/dbus.rs#L104-L104 

For some reason, plasma notifications override whatever timer is attached to critical notifications, [causing them to persist indefinitely](
https://github.com/KDE/plasma-workspace/blob/349804365ae31ea7e5ecf7a3566e835a8af693b0/libnotificationmanager/notification.cpp#L492).

It is somewhat weird of Plasma to assume notification timer should be superseded by urgency, but it is also weird of wezterm to default all notifications to Critical(2) urgency. If this is not exposed to be set by user, Normal should be default.

### To Reproduce

1. Launch wezterm on KDE Plasma with the config below
2. Open a new tab with the mouse click

### Configuration

```lua

wezterm.on('new-tab-button-click', function (window, _)
    window:toast_notification("New tab open", "and I am here forever", nil, 5000)
end)

return {}
```

### Expected Behavior

Wezterm emits dbus notifications with "Normal" urgency by default. Exposing desired urgency level as an argument in Lua function is optional. Even in that case, unconfigured notifications should default to "Normal".

### Logs

wezterm version: 20251014-193657-64f2907c x86_64-unknown-linux-gnu
Window Environment: Wayland
Lua Version: Lua 5.4
OpenGL: Mesa Intel(R) Iris(R) Xe Graphics (ADL GT2) 4.6 (Compatibility Profile) Mesa 25.3.4-arch1.1

### Anything else?

I did not test nightly since it builds forever on my laptop. Same goes for rebuilding plasma notification manager. I will, however, submit the report to KDE Plasma, too.

ekorchmar:
Closes #7553

### DIFF
--- a/wezterm-toast-notification/src/dbus.rs
+++ b/wezterm-toast-notification/src/dbus.rs
@@ -101,7 +101,7 @@ async fn show_notif_impl(notif: ToastNotification) -> Result<(), Box<dyn std::er
     }
 
     let mut hints = HashMap::new();
-    hints.insert("urgency", Value::U8(2 /* Critical */));
+    hints.insert("urgency", Value::U8(1 /* Normal */));
     let notification = proxy
         .notify(
             "wezterm",


bew (approved):
Hmmm I agree is probably better to use 'Normal' instead of 'Critical' as the urgency..

NOTE : I mentioned the urgency when the toast notification was implemented, and Wez couldn't really see any difference at least in GNOME. (https://github.com/wezterm/wezterm/issues/489#issuecomment-785252679)

What is your environment, do you have a problem with the urgency being 'Critical' ?

---

ekorchmar:
To sum up discussion in #7753, both KDE and GNOME ignore notification duration/lifetime setting if urgency is set to critical, which means that notification then never closes on its own.

PR #7556, opened 2026-02-04
fix(gui): handle IME committed text in prompt overlays

dyxushuai:
**Why**
- PromptInputLine overlays (e.g. tab rename) drop IME committed text (KeyCode::Composed), causing input to be lost when confirming or switching IME.

**How**
- Add Pane::send_composed_text to route committed text through the correct input path.
- Implement send_composed_text for TermWizTerminalPane via its input channel.
- Use send_composed_text in the Key::Composed dispatch path.

**Tests**

- cargo check -p wezterm-gui

https://github.com/user-attachments/assets/f31d5fb7-8f87-49c2-b904-e31d86043be7


bew (approved):
Change looks good thank you

We still need to wait for Wez to actually merge

shsaihdsaiudh:
I can confirm that the original issue is still reproducible on a recent Windows nightly.

### Environment

- Windows 11 Pro 64-bit, version `10.0.26200` (build `26200`)
- WezTerm `20260707-093716-fff02ca5`
- `zh-Hans-CN` with iFlytek Pinyin IME (`讯飞输入法`); Microsoft Pinyin is also installed

### Minimal configuration

```lua
local act = wezterm.action

config.keys = {
  {
    key = 'F2',
    mods = 'NONE',
    action = act.PromptInputLine {
      description = 'Rename tab:',
      action = wezterm.action_callback(function(window, _, line)
        if line ~= nil then
          window:active_tab():set_title(line)
        end
      end),
    },
  },
}

return config
```

### Result

1. Open `PromptInputLine` with `F2`.
2. Switch to the Chinese IME, compose Chinese text, and commit a candidate.
3. The committed Chinese text is dropped by the prompt overlay; ASCII input is accepted.

The same Chinese text works when entered through a native Windows input dialog and passed to `wezterm cli set-tab-title`, so tab-title rendering and the CLI Unicode path are both fine. This is specifically the prompt-overlay IME commit path described by this PR.

I have not built this PR locally yet, so this is a reproduction report against the current nightly, not a claim that the patch has been verified on Windows. I would be happy to test a Windows artifact if one becomes available. Thanks for working on this fix.

bew:
Tested locally (before/after) on macOS, works perfectly ✨

Let's merge now that I have that power 😜 

Thanks for the PR!

PR #7558, opened 2026-02-04
kitty: improve animated image handling

zouyonghe:
### Problem
  - `chafa -f kitty <animated.gif>` only shows a still frame in WezTerm (no animation).
  - Interrupting playback with `Ctrl-C` can emit a one-off `base64 decode` ERROR due to an incomplete final chunk.

### Root cause
  - `chafa -f kitty` drives animation by repeatedly sending full-frame `a=T` updates
  (often with `image_id=0` and no `placement_id`). WezTerm was accumulating multiple image attachments per cell and the stacking/order caused older frames to remain on top.
  - When the stream is interrupted mid-chunk, the coalesced base64 payload is incomplete and decode fails; leftover accumulator state can also affect subsequent kitty graphics input.

### Changes
  - Fix kitty image attachment behavior:
    - Replace existing attachments for the same `(image_id, placement_id)` instead of accumulating.
    - Keep attachments ordered by `(z_index, image_id)` to align with kitty layering expectations.
  - Add basic kitty animation control support (`a=a`) and plumb animation state/loop/pending-frame through rendering.
  - Mark kitty image placements dirty when frame/state changes so the screen is repainted.
  - Improve robustness for chunked base64:
    - Try per-chunk decode first; fall back to concatenated decode.
    - On interrupted/incomplete streams, clear the kitty accumulator and log at debug level instead of error.

  ### How to test
  - `chafa -f kitty path/to/animated.gif` should animate correctly.
  - Press `Ctrl-C` during playback: should exit without spurious ERROR logs.
  - Static kitty images should remain unchanged.

  ### Notes
  - Changes are scoped to kitty image/animation handling; iTerm/sixel paths are not affected.


zouyonghe:
Updated this PR against the latest upstream `main` (`8afe0ad30`). The merge was clean, and the PR diff is still scoped to the kitty/image handling files.

What this contributes:

- Improves Kitty graphics compatibility for clients such as `chafa -f kitty`, especially animated GIF playback driven by repeated full-frame `a=T` updates.
- Replaces existing cell image attachments for the same `(image_id, placement_id)` instead of accumulating stale frames, while preserving deterministic `(z_index, image_id)` ordering.
- Adds basic Kitty animation control handling (`a=a`) and carries animation state, loop count, and pending-frame information through rendering.
- Marks affected Kitty image placements dirty when frame/state changes so updated frames repaint.
- Makes chunked base64 handling more robust, including interrupted streams such as Ctrl-C during playback.

I also rechecked whether current upstream has an equivalent fix: `main` has some placement-related helpers, but I did not find the animation-control, pending-frame, dirty-repaint, or chunked-base64 robustness pieces from this PR.

Local checks after merging upstream:

- `cargo fmt --all -- --check` passed; only existing nightly-only rustfmt config warnings were emitted.
- `cargo check -p wezterm-term -p wezterm-gui -p wezterm-escape-parser -p wezterm-cell` passed with 0 errors; warnings appear unrelated/existing.

@wez would you be open to reviewing this direction for Kitty graphics compatibility, or would you prefer this to be split into smaller PRs such as attachment replacement, base64 robustness, and animation control/rendering state?

Issue #6986, opened 2025-05-23
Tabs do not use the full width of the tab bar


Bocom:
### What Operating System(s) are you seeing this problem on?

macOS

### WezTerm version

20250518-060218-5663e749

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

Tabs don't utilize the full width of the tab bar and start to shrink each tab further with each tab starting from the tenth, leaving a massive gap that grows with each subsequent tab.

### To Reproduce

Open a new window and start opening tabs. From the tenth tab onwards, they should start to shrink, even with a lot of space left in the tab bar.

### Configuration

No config

### Expected Behavior

Tabs should use the full width of the tab bar before shrinking and shouldn't leave a massive gap.

### Logs

Debug Overlay
wezterm version: 20250518-060218-5663e749 aarch64-apple-darwin
Window Environment: macOS 15.4.1 (24E263)
Lua Version: Lua 5.4
OpenGL: Apple M3 Pro 4.1 Metal - 89.4

### Anything else?

Up to nine tabs, all tab are normal sized
<img width="697" alt="Image" src="https://github.com/user-attachments/assets/9f3ad6cf-8bef-4dc4-b69e-5a472eccf9a0" /> 

From the the tenth tab, they start to shrink
<img width="697" alt="Image" src="https://github.com/user-attachments/assets/34042123-0e81-4d94-bad9-9f9eb569faac" />

Every subsequent tab opened shrinks each tab further
<img width="697" alt="Image" src="https://github.com/user-attachments/assets/08060a77-04db-481d-88f7-6f8d5522ea36" />

PR #7574, opened 2026-02-11
Add vertical tab bar support (Left/Right positioning)

### LINKED ISSUE #3537: Allow the tab bar to be rendered on the side
I'm completely in love with [Tree Style Tab](https://addons.mozilla.org/en-US/firefox/addon/tree-style-tab/) because it allows for high density while still being able to differentiate the tabs by more than favicon, or having to slowly scroll (ignore the "tree" part of its functionality in this case). Back in my macOS days I used [iTerm2's Tab Bar Location](https://iterm2.com/documentation-preferences-appearance.html) option to have a tab bar on the side, and it's one of the few things I still miss on Linux :penguin: 

I would be stoked if there was an option to render the tab bar on the left (especially if the width was adjustable) :blue_heart: 

**Additional context**
I imagine this is neither a popular request (judging by this being the first issue to mention it near as I can tell), nor easy to implement, so no hard feelings if this goes straight into the ice box :+1: 

I couldn't for the life of me find a screenshot of iTerm2 doing it, but here's one from the [hyper-vertical-tabs](https://github.com/randallsquared/hyper-vertical-tabs) Hyper plugin demonstrating the same thing:

hyper-vertical-tabs

lansator:
## Summary
      - Add `tab_bar_position` config option with `"Top"`, `"Bottom"`, `"Left"`, `"Right"` values
      - Add `tab_bar_width` config option (default `"200px"`) for controlling vertical tab bar width
      - Vertical positioning automatically uses the fancy tab bar; `tab_bar_at_bottom` continues to work for
      backward compatibility

Closes #3537

Disclosure: Built with Claude Code and checked with OpenAI Codex. 

PS: This is my first contribution to open source.

### DIFF
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -476,6 +476,21 @@ pub struct Config {
     #[dynamic(default)]
     pub tab_bar_at_bottom: bool,
 
+    /// Controls the position of the tab bar: Top, Bottom, Left, or Right.
+    /// When set to a non-default value, this takes precedence over tab_bar_at_bottom.
+    /// Left and Right positions require use_fancy_tab_bar = true (the default).
+    #[dynamic(default)]
+    pub tab_bar_position: TabBarPosition,
+
+    /// The width of the tab bar when positioned on the left or right side.
+    /// Ignored when the tab bar is at the top or bottom.
+    /// Defaults to 200 pixels.
+    #[dynamic(
+        try_from = "crate::units::PixelUnit",
+        default = "default_tab_bar_width"
+    )]
+    pub tab_bar_width: Dimension,
+
     #[dynamic(default = "default_true")]
     pub mouse_wheel_scrolls_tabs: bool,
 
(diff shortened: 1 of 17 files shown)


trousev:
@wez , what do you think about this PR?

TBH, I (as a user) would be extremely happy to have left aligned tabs =)

tyler-abovedesign:
Please add this!

gilescope:
I'm forking and stealing this. Have thrown #6527 on top: https://github.com/gilescope/wezterm/tree/giles-vertical-tabs-drag-reorder

Hope vertical tabs lands at some point.

jpdstan:
@wez

mmcc007:
Another use case for this, with numbers, because a side bar removes the problem
rather than easing it.

I run ~24 tabs in a 1360px window, each a different project folder. On the
horizontal bar that is ~56px per tab, and after per-tab padding there is room for
about 6 characters — so `orbsoft-corpus` and `proto-michael-moe` both render as
six letters and are indistinguishable at a glance.

Measured at 24 tabs (fancy bar, `show_close_tab_button_in_tabs = false`,
Liberation Sans Narrow):

| tab bar `font_size` | characters per tab |
|---|---|
| 7pt  | all names fit, none legible |
| 8pt  | ~6-7 |
| 10pt | ~4 |
| 14pt | zero — separators only |

There is no size that is both legible and informative, because the width per tab
is `window_width / tab_count` and shrinks as tabs are added.

A 200px vertical strip gives every tab the full 200px for its label — no division
by tab count at all. 24 tabs down a 768px-tall window is 32px per row, so the full
folder name is legible with room to spare, and the problem stops being a trade-off.

Worth noting the horizontal bar cannot be tuned into a solution: I also measured
that 33% of the bar (449px of 1360px) is left *unused* at 24 tabs, so the tabs are
not even consuming the width already available (#6986, #7591, #7809). Vertical
side-steps that entirely.

Happy to test a build if that is useful.

PR #7580, opened 2026-02-13
chore: use `LazyLock` instead of `lazy_static`

Expyron:


wezterm-gui/src/overlay/debug.rs
```diff
@@ -15,9 +15,7 @@ use termwiz::lineedit::*;
 use termwiz::surface::Change;
 use termwiz::terminal::Terminal;
 
-lazy_static::lazy_static! {
-    static ref LATEST_LOG_ENTRY: Mutex<Option<DateTime<Local>>> = Mutex::new(None);
-}
+static LATEST_LOG_ENTRY: Mutex<Option<DateTime<Local>>> = Mutex::new(None);
```
bew, line 18: Why do you use LazyLock for some but not for others (like here) ?

Expyron: Some constructors from std (such as `Mutex::new`, `Vec::new`, `AtomicBool::new`, ...) are `const`, so they can be assigned directly to a static and there is no need to lazily-initialize them.

I've removed the lazy-initialization where it was not needed, but many functions are not `const` and that's where I used `LazyLock`.

PR #7399, opened 2025-11-24
Introduce sequencing for 'pane focus' related events (Recreated)

Strackeror:
This is a rebase and an attempt to take into account wez's comments for #6349:
It's been a year without updates from the original PR author, and this issue has been a consistent thorn in my side for a while now, so I took the liberty of recreating the PR and try to get this in.
> Switching between panes (especially over domains where large latencies are involved) can cause a loop where we get stuck repeatedly switching back and forth between two panes. This happens because handlers of PaneFocused events can also produce these events, causing us to get stuck in a loop if we let the GUI process stale PaneFocused events.
> 
> The solution is to introduce a 'serial' (AKA 'sequence number') for GUI-initiated pane focus changes. This will be echo'd back by the server, which allows the GUI to avoid processing the response if it has sent a new request in the mean time. CLI-initiated pane focus changes (eg. via wezterm cli activate-pane) will continue to be unconditionally handled by the client.
> 
> Inspired by https://github.com/wezterm/wezterm/pull/4737 and https://github.com/wezterm/wezterm/pull/4737#issuecomment-1876173247
> 
> Verified activate-pane-direction, activate-pane and activate-tab still work correctly
> 
> closes: https://github.com/wezterm/wezterm/issues/4390
> closes: https://github.com/wezterm/wezterm/issues/4693

Also closes #7388

Issue #7388, opened 2025-11-19
Complete freeze when switching windows on a mux server


Strackeror:
### What Operating System(s) are you seeing this problem on?

Windows, mux-server on an openSUSE-Tumbleweed virtual machine

### WezTerm version

wezterm-mux-server 20251111-071056-118802c2;  wezterm 20251111-071056-118802c2

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

When running multiple windows with several panes (likelihood of occurence increases with amount of panes/tabs) on a single mux server, changing window focus can make the GUI *completely* freeze.

The logs get absolutely dumped with this in a very tight loop:
```
17:20:58.574  DEBUG  wezterm_client::client                 > decoded serial 0 PaneFocused
17:20:58.574  DEBUG  codec                                  > encode_async SetFocusedPane size=5
17:20:58.574  DEBUG  wezterm_client::client                 > decoded serial 0 PaneFocused
17:20:58.574  DEBUG  wezterm_client::client                 > decoded serial 381 UnitResponse
17:20:58.574  DEBUG  wezterm_client::client                 > decoded serial 0 PaneFocused
17:20:58.574  DEBUG  wezterm_client::client                 > decoded serial 0 PaneFocused
17:20:58.574  DEBUG  wezterm_client::client                 > decoded serial 382 UnitResponse
17:20:58.574  DEBUG  wezterm_gui::termwindow                > Notification(Any { .. })
17:20:58.574  DEBUG  wezterm_client::client                 > decoded serial 0 PaneFocused
17:20:58.574  DEBUG  wezterm_client::client                 > decoded serial 0 PaneFocused
17:20:58.574  DEBUG  wezterm_client::client                 > decoded serial 383 UnitResponse
17:20:58.574  DEBUG  wezterm_gui::termwindow                > Notification(Any { .. })
17:20:58.574  DEBUG  codec                                  > encode_async SetFocusedPane size=5
17:20:58.575  DEBUG  wezterm_client::client                 > decoded serial 0 PaneFocused
17:20:58.575  DEBUG  wezterm_client::client                 > decoded serial 0 PaneFocused
17:20:58.575  DEBUG  wezterm_gui::termwindow                > Notification(Any { .. })
17:20:58.575  DEBUG  wezterm_client::client                 > decoded serial 384 UnitResponse
[...]

PR #7584, opened 2026-02-14
FIX: infinite feedback loop on slow SSHMUX connections when pane-switching, which hangs wezterm

### LINKED ISSUE #4390: Switching panes in multiple directions at once causes the focused pane to rapidly switch when any key is pressed

stepelu:
### Issue:

When switching panes/tabs over high-latency SSHMUX connections, wezterm hangs due to an infinite feedback loop between client and server:

1. User switches pane focus → client sends `SetFocusedPane(B)` to server
2. Server applies the change and broadcasts `PaneFocused(B)` back to all clients, including the originator
3. Client receives the echo → `focus_pane_and_containing_tab()` → `set_active_pane()` → `advise_focus_change()` → `focus_changed(true)` → `advise_focus()` → sends another `SetFocusedPane(B)` back to server
5. Goto 2

With rapid switching (A→B→A), stale echoes cause oscillation between panes, flooding the event loop.
The existing guard in `advise_focus()` on `focused_remote_pane_id` does not prevent this because during oscillation the focus genuinely alternates between different pane IDs.

### Fix:

In `process_unilateral()`'s `PaneFocused` handler, pre-set `focused_remote_pane_id` to match the server's notification **before** calling `focus_pane_and_containing_tab`. 
This way, when the call chain reaches `advise_focus()`, the guard sees the pane as already focused and suppresses the echo `SetFocusedPane` — breaking the loop.

The change is entirely client-side, very short, and no protocol version bumps nor server changes are needed.
A limitation of this fix is that rapid switching can still cause brief visual glitches where focus momentarily lands the wrong pane (this is still much preferable to the current behaviour, that simply results in wezterm hanging and requiring a restart).

It is my understanding that this PR fixes: #4390, #4693, #7388

### Related:

These two PR #6349, #7399 propose fixes for the mentioned issue (here, and in the linked issues) via a counter.
It seems that there are some unresolved questions about a single location where this counter is incremented (`ClientPane::advise_focus` could be a good place, with the counter handled locally and then transmitted only, instead of having both client and server generating the counters), and these implementations necessarily are more complex.

It would be great if this PR could be accepted for improve the situation for now.

### DIFF
--- a/wezterm-client/src/pane/clientpane.rs
+++ b/wezterm-client/src/pane/clientpane.rs
@@ -226,6 +226,18 @@ impl ClientPane {
                 // it here.
                 log::trace!("advised of remote pane focus: {pane_id}");
 
+                // Pre-set focused_remote_pane_id to match what the server told us.
+                // This prevents advise_focus() from echoing SetFocusedPane back to
+                // the server when focus_pane_and_containing_tab triggers
+                // advise_focus_change -> focus_changed(true) -> advise_focus().
+                // Without this, the echo creates an infinite feedback loop that
+                // hangs the GUI, especially over high-latency connections where
+                // stale echoes cause focus oscillation between panes.
+                {
+                    let mut focused = self.client.focused_remote_pane_id.lock().unwrap();
+                    *focused = Some(self.remote_pane_id);
+                }
+
                 let mux = Mux::get();
                 if let Err(err) = mux.focus_pane_and_containing_tab(self.local_pane_id) {
                     log::error!("Error reconciling remote PaneFocused notification: {err:#}");


stepelu:
@wez any chance you can briefly review this short PR?

FramedStone:
same issue here.

bew:
Replaced by #7763

PR #7592, opened 2026-02-17
windows: add native ARM64 build, packaging, and runtime support

mshinjo:
Greetings! This PR adds CI, packaging, runtime default/fallback, and documentation changes to support Windows ARM64 (aarch64-pc-windows-msvc) builds and distribution.

I tested this patch series on a Lenovo T14s Snapdragon (Qualcomm Snapdragon X1E-78-100) with Windows 11 Home (version: 24H2, build: 10.0.26100.7462), and it seems to work fine. https://github.com/mshinjo/wezterm/actions/runs/22084570868

The debug overlay from the built binary is below:
```
Debug Overlay
wezterm version: 20260216-220912-60f8b0e2 aarch64-pc-windows-msvc
Window Environment: Windows
Lua Version: Lua 5.4
WebGPU: name=Qualcomm(R) Adreno(TM) X1-85 GPU, device_type=IntegratedGpu, backend=Vulkan, driver=Qualcomm Technologies Inc. Adreno Vulkan Driver, driver_info=Driver Build: 9876292e35, , 1748599450
                                    Date: 05/30/2025
                                                    Compiler Version: E031.50.05.00
   Driver Branch: PP137
                       , vendor=20803, device=909329200
Press ESC or CTRL-D to exit
```

I am new to this codebase, and I tried my best to keep each commit small and reviewable.
In some cases, I was not sure whether my design decisions were right. Please let me know if I missed anything.
Thanks!


ci/deploy.sh
```diff
@@ -116,16 +122,23 @@ case $OSTYPE in
...
+      cp $TARGET_DIR/release/mesa/opengl32.dll \
+          $zipdir/mesa
+    fi
     7z a -tzip $zipname $zipdir
-    iscc.exe -DMyAppVersion=${TAG_NAME#nightly} -F${instname} ci/windows-installer.iss
+    if [[ "$win_arch" == "arm64" ]] ; then
+      iscc.exe -DMyAppVersion=${TAG_NAME#nightly} -DArm64Package -DNoExtraDLLs -F${instname} ci/windows-installer.iss
+    else
+      iscc.exe -DMyAppVersion=${TAG_NAME#nightly} -F${instname} ci/windows-installer.iss
+    fi
```
wez: Could we avoid the repetition here by doing something like `-DWezTermArch=${win_arch}` and then putting the appropriate conditional inside the installer script?

ci/deploy.sh
```diff
@@ -116,16 +122,23 @@ case $OSTYPE in
...
       $TARGET_DIR/release/wezterm.pdb \
-      assets/windows/conhost/conpty.dll \
-      assets/windows/conhost/OpenConsole.exe \
-      assets/windows/angle/libEGL.dll \
-      assets/windows/angle/libGLESv2.dll \
       $zipdir
-    mkdir $zipdir/mesa
-    cp $TARGET_DIR/release/mesa/opengl32.dll \
-        $zipdir/mesa
+    if [[ "$win_arch" != "arm64" ]] ; then
```
wez: A couple of things this raises:

1. conpty and openconsole are things that we bundle and ship for ourselves because windows traditionally has been slow to ship those components; they generally only update with releases to the OS.  We currently only have the intel versions.  We have a PR open to update these to a more recent version.  The ideal situation is that we ship a consistent version across architectures for a given wezterm build/version.  So I think we should make appropriate arch subdirs under assets/windows/conhost and track from there.  This has fanout to the various build.rs's elsewhere
2. I'd be amazed if those angle pieces are still useful; we should probably consider removing them as I don't think anyone is running these in practice any more and I can't remember when they were last updated.  That's a separate concern from this PR though.
3. Similar thoughts about the mesa opengl32.dll: I'm unsure if this is live code in practice, but it may be more likely to be used than angle.

Since *this* particular PR shouldn't make a decision to remove bits, I propose that instead it could move those to architecture subdirs and change the logic here to copy things from the current architecture subdir, if it exists.

The goal is to make the logic here look a bit cleaner; ideally it would be structured so that if/when we decide to remove the angle/libEGL.dll for a specific arch, we wouldn't be required to change the logic here in the deploy script.

Now, the other thing this raises is: what does a multi-arch package look like on windows?  Is that even supportable via the .iss script?  The reason for asking is whether we should be steering towards that so that we don't have to completely redo the changes in this PR if we choose to tackle that in the future.  I don't think this PR should produce a multi-arch unless it turns out to be a trivial additional step, but we should be aware of what that state looks like

ci/generate-workflows.py
```diff
@@ -501,7 +504,11 @@ def package(self, trusted=False):
                 "MACOS_APPLEID": "${{ secrets.MACOS_APPLEID }}",
                 "MACOS_APP_PW": "${{ secrets.MACOS_APP_PW }}",
             }
-        steps = [RunStep("Package", "bash ci/deploy.sh", env=deploy_env)]
+        if self.is_windows_arm():
+            deploy_env = deploy_env or {}
+            deploy_env["WIN_ARCH"] = "arm64"
+        deploy_cmd = "bash ci/deploy.sh"
+        steps = [RunStep("Package", deploy_cmd, env=deploy_env)]
```
wez: These two lines don't appear to be a functional change; please revert them so that this PR shows only the essential functional changes

ci/deploy.sh
```diff
@@ -101,13 +101,19 @@ case $OSTYPE in
 
     ;;
   msys)
-    zipdir=WezTerm-windows-$TAG_NAME
+    win_arch=${WIN_ARCH:-x64}
```
wez: Can we find a way to avoid having the caller pass in the arch for arm/intel?

On unix `uname -m` returns an architecture.  Does cygwin/msys return an appropriate arch on arm?

When building under MSVC, ISTR that the compiler exports a set of env vars that informs which architecture will be used.

Maybe the highest signal for this is to run `rustup default` to see what rust is going to produce; that way we're unlikely to end up with a mismatch between what we built and what this script decides.

ci/generate-workflows.py
```diff
@@ -501,7 +504,11 @@ def package(self, trusted=False):
                 "MACOS_APPLEID": "${{ secrets.MACOS_APPLEID }}",
                 "MACOS_APP_PW": "${{ secrets.MACOS_APP_PW }}",
             }
-        steps = [RunStep("Package", "bash ci/deploy.sh", env=deploy_env)]
+        if self.is_windows_arm():
```
wez: I'd love to eliminate this logic; see my comments on arch detection in the deploy script

ci/generate-workflows.py
```diff
@@ -726,7 +733,7 @@ def create_flathub_pr(self):
 
     def create_winget_pr(self):
         steps = []
-        if "windows" in self.name:
+        if "windows" in self.name and not self.is_windows_arm():
```
wez: I'd prefer to add a `publish_winget` boolean, similar to how things like `app_image` work, so that we can decouple this from a specific architecture.  Presumably this is something we need to evaluate for multi-arch as well; does winget support multi-arch, and what does it look like?

ci/subst-release-info.py
```diff
@@ -56,6 +58,13 @@ def build_subst(subst, stable, categorized):
         subst[f"{kind}_asset"] = name
         subst[f"{kind}_dir"] = dir
 
+    # ARM64 Windows artifacts may not exist for older releases.
+    for kind in ("windows_arm64_zip", "windows_arm64_exe"):
+        kind = f"{kind}_{stable}"
+        subst.setdefault(kind, "")
+        subst.setdefault(f"{kind}_asset", "")
+        subst.setdefault(f"{kind}_dir", "")
```
wez: If a value doesn't exist, it seems better to me to not to set a dummy empty value.  Why do we need this?

config/src/frontend.rs
```diff
@@ -1,14 +1,27 @@
...
+            Self::WebGpu
+        }
+
+        #[cfg(not(all(windows, target_arch = "aarch64")))]
+        {
+            Self::OpenGL
+        }
+    }
+}
+
```
wez, line 26: There's no explanation for why, either in the PR or, more importantly, here in the code.
Why is this conditional?

deps/freetype/build.rs
```diff
@@ -74,13 +74,17 @@ fn libpng() {
...
+    let target = env::var("TARGET").unwrap();
     if let Ok(arch) = env::var("CARGO_CFG_TARGET_ARCH") {
         match arch.as_str() {
             "aarch64" | "arm" => {
                 cfg.file("libpng/arm/arm_init.c")
-                    .file("libpng/arm/filter_neon.S")
                     .file("libpng/arm/filter_neon_intrinsics.c")
                     .file("libpng/arm/palette_neon_intrinsics.c");
+                // MSVC cannot compile filter_neon.S, so skip on Windows targets.
+                if !target.contains("windows") {
```
wez: Should this more closely match the windows msvc specific target, rather than just all windows?

wezterm-gui/src/termwindow/mod.rs
```diff
@@ -844,19 +844,42 @@ impl TermWindow {
...
+                            WebGpuState::new(&window, dimensions, &config).await?,
+                        ));
+                    } else {
+                        return Err(err);
+                    }
+
+                    #[cfg(not(all(windows, target_arch = "aarch64")))]
+                    {
+                        return Err(err);
+                    }
```
wez: This is quite a complex change.  Why is it needed?
There's no commentary here, so it's hard to evaluate.  Ideally this commit doesn't need to change anything in this particular module in the source.

wez (commented):
Thanks for this!
I have a few questions and suggestions!

searleser97:
@mshinjo could you take a look at above comments from @wez ?

ci/deploy.sh
mshinjo: Yes, that makes sense. Done

ci/deploy.sh
mshinjo: I moved the existing bundled Windows assets under x64 sub directories and changed the copy logic to use the current arch dir if it exists.

ci/generate-workflows.py
mshinjo: Done. Sorry about that.

ci/deploy.sh
mshinjo: Yes, that makes sense and is cleaner. It now runs `rustup default` to derive the arch

ci/generate-workflows.py
mshinjo: Added a `publish_winget` flag. I'm not too familiar with winget, but I can maybe take a look later.

ci/subst-release-info.py
mshinjo: The empty values were there because there are no past ARM64 releases, and that would be problematic. I changed it so that the docs now check whether the ARM64 vals are defined before rendering

config/src/frontend.rs
mshinjo: The rationale was in the commit message, but I added a comment too. OpenGL driver support does not seem to be available by default on Windows on Arm (at least in my environment), so this prefers wgpu's DX12 backend.

deps/freetype/build.rs
mshinjo: Yes, that makese sense. Done.

wezterm-gui/src/termwindow/mod.rs
mshinjo: I originally added this as a safety net because OpenGL failed on my Windows on Arm machine, but after defaulting Windows ARM64 to WebGPU, this runtime fallback is not needed for this PR.
I dropped this patch, so termwindow is no longer changed.

mshinjo:
Hi Wez! Thanks very much for the review! I prepared a v2 patch series based on your feedback and suggestions. I realized after pushing that there is a conflict, so I'll rebase during the next review cycle.

Built artifacts: https://github.com/mshinjo/wezterm/actions/runs/27457530451

eniolasonowo:
I installed the binary, seems to be working fine.

PR #7605, opened 2026-02-20
feat: add option to disable usage of `PSUEDOCONSOLE_INHERIT_CURSOR`

amrbashir:
When cursor is inherited, I get the following when running the bash example:

```
You can now type commands for Bash (type 'exit' to quit):
^[[25;1R
```
and then it just freezes

I couldn't figure if there is a way to fix this so I thought I'd just make it configurable


bew:
NOTE: this flag was added in #4982

PR #7610, opened 2026-02-20
fix: sync user_vars on mux reconnect (Issue #5832)

### LINKED ISSUE #5832: User vars become unset after detach an re-attach to multiplexer Unix domain

macOS

_No response_

### WezTerm version

20240719-081740-643d85f8

### Context

- Using the multiplexer using Unix domains, exactly as described in the docs: https://wezfurlong.org/wezterm/multiplexing.html#unix-domains.
- Using the technique suggested by @wez in #3648  to use a user var called `IS_NVIM` (set/unset by Neovim on entry/exit) to control if a key should be relayed to Neovim
- Using the [`wezterm-move`](https://github.com/letieu/wezterm-move.nvim) plugin in Neovim with a simplified version of the suggested WezTerm Lua code in my WezTerm setup, using a user var instead of the process name to determine if a key should be relayed to Neovim instead of being consumed by WezTerm

### The problem

Using the IS_NVIM user var in my WezTerm Lua code to control if keys should be interpreted by WezTerm or relayed to the terminal (and by extension to Neovim) generally works like a charm. But if I have Neovim running when detaching from the multiplexer (or just quit WezTerm), and subsequently attach again, the IS_NVIM user var is no longer set, meaning that the keystrokes I want to pass through to the shell/Neovim are not passed through.

I have tried using the `WEZTERM_PROG` user var set by shell integration, and that exposes the same problem, confirmed by logging the value from my WezTerm Lua code.

This indicates - as suggested by @wez in [this comment](https://github.com/wez/wezterm/discussions/3648#discussioncomment-10103702) - user vars are not synchronized on attach after a detach.

### To Reproduce

1. Start WezTerm using the config below, and ensure that shell integration is set up
2. Set the `IS_NVIM` user var to `true`:  `__wezterm_set_user_var IS_NVIM "true"`
3. To ascertain thath the WezTerm config works: Run `cat` and type `^h` and `k` and see that these are sent to the shell/`cat`. Do not exit `cat`.
4. Detach from the muxer or quit WezTerm
5. Attach or start WezTerm again, going back to the running `cat` process. Type `h` and `k` again and observe that they are no longer sent to the shell

### Configuration

The smallest relevant subset of my config:

```lua

local function is_vim(pane)
  return pane:get_user_vars().IS_NVIM == "true"
end

local direction_keys = {
  h = "Left",
  j = "Down",
  k = "Up",
  l = "Right",
}

local function split_nav(key)
  return {
    key = key,
    mods = "CTRL",
    action = wezterm.action_callback(function(win, pane)
      if is_vim(pane) then
        win:perform_action({
          SendKey = { key = key, mods = "CTRL" },
        }, pane)
      else
        win:perform_action({ ActivatePaneDirection = direction_keys[key] }, pane)
      end
    end),
  }
end

config.keys = {
  split_nav("h"),
  split_nav("j"),
  split_nav("k"),
  split_nav("l"),
}

config.unix_domains = {
  {
    name = "unix",
  },
}
config.default_gui_startup_args = { "connect", "unix" }

return config
```

### Expected Behavior

I expect the user vars to be retained between detach/attach, making the UX consistent between newly started sessions and re-attached sessions.

### Logs

_No response_

### Anything else?

_No response_

yosagi:
Fixes #5832

## Problem

When reconnecting to a mux server (detach → reattach), user variables
set during the session are lost. This is because `GetPaneRenderChangesResponse`
does not include `user_vars` — only real-time changes via `NotifyAlert`
are propagated. On reconnect, the client starts with an empty HashMap.

## Solution

Include `user_vars` in `GetPaneRenderChangesResponse` so that they are
synced on every render update, including the first one after reconnect.

Changes (3 files, +4 lines):

- **codec/src/lib.rs**: Add `user_vars: HashMap<String, String>` field
  to `GetPaneRenderChangesResponse` with `#[serde(default)]`
- **wezterm-mux-server-impl/src/sessionhandler.rs**: Populate `user_vars`
  via `pane.copy_user_vars()` in the response
- **wezterm-client/src/pane/clientpane.rs**: Apply received `user_vars`
  to the client pane state

## Compatibility note

Adding a field to `GetPaneRenderChangesResponse` changes the bincode
serialization format. Although `#[serde(default)]` is present, bincode
does not support optional trailing fields the way JSON does, so mixed
versions will fail to communicate.

This may warrant a `CODEC_VERSION` bump. I've left it unchanged in this
PR — happy to add the bump if preferred.

## Testing

Verified with a local build: set a user var in a mux session, detach,
reattach, and confirm the user var is preserved.

### DIFF
--- a/codec/src/lib.rs
+++ b/codec/src/lib.rs
@@ -925,6 +925,8 @@ pub struct GetPaneRenderChangesResponse {
 
     pub input_serial: Option<InputSerial>,
     pub seqno: SequenceNo,
+    #[serde(default)]
+    pub user_vars: HashMap<String, String>,
 }
 
 #[derive(Deserialize, Serialize, PartialEq, Debug)]
--- a/wezterm-client/src/pane/clientpane.rs
+++ b/wezterm-client/src/pane/clientpane.rs
@@ -138,6 +138,7 @@ impl ClientPane {
         match pdu {
             Pdu::GetPaneRenderChangesResponse(mut delta) => {
                 *self.mouse_grabbed.lock() = delta.mouse_grabbed;
+                *self.user_vars.lock() = std::mem::take(&mut delta.user_vars);
 
                 let bonus_lines = std::mem::take(&mut delta.bonus_lines);
                 let client = { Arc::clone(&self.renderable.lock().inner.borrow().client) };
--- a/wezterm-mux-server-impl/src/sessionhandler.rs
+++ b/wezterm-mux-server-impl/src/sessionhandler.rs
@@ -139,6 +139,7 @@ impl PerPane {
             working_dir: working_dir.map(Into::into),
             input_serial: force_with_input_serial,
             seqno: self.seqno,
+            user_vars: pane.copy_user_vars(),
         })
     }
 }


kalebo:
I can confirm this fixes the issue. Thanks @yosagi! 

Is there a possibility of getting this mainlined @bew?  IMHO user_vars as a feature is kneecapped by not having this.

bew:
Hello o/ I'd love to, but only Wez has merge power at the moment...

PR #7618, opened 2026-02-25
Implement tab:move_to_window method to allow moving tabs between windows using Lua

### LINKED ISSUE #5976: After moving a tab to a new window, it is impossible to move it back to an existing window.
After moving a tab to a new window, it is impossible to move it back to an existing window.

I'd like to be able to revert the move tab to new window, and also to be able to move tabs across different windows.

I found no solutions, except to reopen the tab from scratch. But that is not what I have in mind. 

In case this feature is already implemented, it is difficult to find documentation about it. In this case this FR would become a DocBUG.

nickguletskii:
Potentially/partially closes #5976

Depends on #7617 

Implements `tab:move_to_window(target_window)`, which moves the tab (including all its panes and split layout) from its current window to the specified target window. The tab is appended to the end of the target window's
tab list.

I implemented this because I always have a gazillion WezTerm windows open and I wanted to implement an action that automatically combines all of them into one window.

### DIFF
--- a/mux/src/lib.rs
+++ b/mux/src/lib.rs
@@ -1317,7 +1317,7 @@ impl Mux {
         &self,
         tab_id: TabId,
         target_window_id: WindowId,
-    ) -> anyhow::Result<()> {
+    ) -> anyhow::Result<bool> {
         // Find which window currently contains this tab
         let src_window_id = self
             .get_window_containing_tab(tab_id)
@@ -1325,7 +1325,7 @@ impl Mux {
 
         // No-op if already in target window
         if src_window_id == target_window_id {
-            return Ok(());
+            return Ok(false);
         }
 
         // Verify target window exists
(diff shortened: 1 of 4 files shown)


mux/src/lib.rs
```diff
@@ -1304,6 +1304,82 @@ impl Mux {
...
+
+    pub fn move_tab_to_window(
+        &self,
+        tab_id: TabId,
+        target_window_id: WindowId,
+    ) -> anyhow::Result<()> {
+        // Find which window currently contains this tab
+        let src_window_id = self
+            .get_window_containing_tab(tab_id)
+            .ok_or_else(|| anyhow::anyhow!("tab {} not found in any window", tab_id))?;
```
ekorchmar, line 1324: Can this actually ever happen?

nickguletskii: Not as far as I know/understand (although I am not well familiar with wezterm's internals, I just recently started using it). However, I still generally prefer returning an `Err` instead of panicking using `expect` or `unwrap`.

Issue #7272, opened 2025-09-30
Pasting Images


sai-sy:
**Is your feature request related to a problem? Please describe.**
I want to paste images in opencode. I [can see others do it with some success on other systems ](https://github.com/sst/opencode/issues/906#issue-3224845889) but I'm having some issues [on my stack](https://github.com/sst/opencode/issues/2878#issuecomment-3352442222). I'm not actually sure if this is a wezterm issue, wsl2 issue, or an opencode issue so if someone could point me in the right direction that woudl grealy be appreciated.

**Describe the solution you'd like**
To be able to paste images, specifically in opencode but in general as well?

bew:
Hello, I think the main question I have would be: What is opencode expecting to get from the terminal when you paste an image 🤔
- A path? To some tmp file made by the terminal? (when should it be cleaned up?)
- The image content?
- other?

And how do you copy the image exactly? From your file explorer? (so it already has a path)
Or from the web?

sai-sy:
I'm personally copying from snipping tool. Which with my minimal understanding copies a bitmap? So I have to figure out how openterm can grab the clipboard contents in that form and pass it through the API

which seems like a solely independent opencode problem but I guess poses an interesting question about what if any handling wezterm has when it gets pasted something that's an image

alexismanuel:
@sai-sy Hello ! Did you manage to solve your issue ? I am having the exact same problem where I can't paste images from my clipboard (and copied from the snipping tool) into OpenCode TUI.

Miskamyasa:
Joining to the thread, cause it doesn't work for me as well. Have to use alacritty in this case

robertfall:
To answer the original question here, this is what clipboard inspector shows: 

<img width="1363" height="850" alt="Image" src="https://github.com/user-attachments/assets/31a279b8-64bd-4fd3-bea3-e479b6cbaeed" />

It appears to be an image in my clipboard, `image/png` to be specific. Though that is from the web API, but it's literal image data on the clipboard.

bew:
Getting the image data from the clipboard is easy, but this question remains:

> What is opencode expecting to get from the terminal when you paste an image 🤔
> 
>  * A path? To some tmp file made by the terminal? (when should it be cleaned up?)
>  * The image content?
>  * other?

And if it's a path, what should happen when the program is not running on the local computer (e.g. You're through ssh) and thus doesn't have access to local paths?

robertfall:
[...]

PR #7621, opened 2026-02-27
Add clipboard image paste support with Ctrl+V smart paste for all platforms

alarmz:
## Summary

- Add `PasteImageToSshUpload` key assignment: reads clipboard image, uploads to remote SSH server via SFTP (or SCP fallback), and pastes the remote file path into the terminal
- Add smart `PasteFrom` fallback: when clipboard has an image but no text, `Ctrl+V` / `Cmd+V` automatically saves the image locally and pastes the file path
- **Add `Ctrl+V` as default smart paste binding on Windows/Linux** — text is pasted directly, images are saved locally and the path is pasted. macOS is unaffected (`Ctrl+V` remains "verbatim insert", `Cmd+V` already handles smart paste)
- Add clipboard image reading support for **macOS** (NSPasteboard), **Linux X11** (xclip/xsel), **Linux Wayland** (wl-paste), and **Windows** (DIB→PNG conversion)
- Add startup check for required clipboard tools on Linux/FreeBSD with warning notification
- Add config options: `ssh_image_paste_remote_path`, `ssh_image_paste_enabled`, `image_paste_local_path`
- Fix SFTP O_CREAT flag for libssh-rs backend
- Fix `has_tool` to use pure Rust PATH lookup instead of external `which`
- Add comprehensive unit tests (including keybinding tests for Ctrl+V, Super+V, Ctrl+Shift+V, permute_keys)

## What changed since v1

**Fixed: Ctrl+V paste broken in local mode on Windows** — the previous version had several issues when pasting with `Ctrl+V` in a local (non-SSH) terminal on Windows:

1. **Hardcoded `/tmp/` path**: `image_paste_local_path` defaulted to `/tmp/clipboard_{timestamp}.png`, which doesn't exist on Windows. Now uses `std::env::temp_dir()` to resolve to the platform-correct temp directory (e.g., `C:\Users\<user>\AppData\Local\Temp\`).
2. **Missing parent directory creation**: Image save could fail if the target directory didn't exist. Now calls `create_dir_all()` before writing.
3. **"Not an SSH session" error in local mode**: `PasteImageToSshUpload` (bound to `Ctrl+V` on Windows) would error when not in an SSH domain. Now gracefully falls back to saving the image locally and pasting the path — same behavior as `PasteFrom` image fallback.
4. **No text fallback**: When clipboard contained only text (no image), `PasteImageToSshUpload` would show an error toast. Now falls back to pasting text from clipboard when no image is available.
5. **Updated docs and tests**: Added behavior-by-context table to `PasteImageToSshUpload` docs, platform-specific defaults to `image_paste_local_path` docs, and unit tests for temp dir, directory creation, and local vs remote path handling.

## Important: user config note

`PasteImageToSshUpload` is registered as a **built-in default binding** for `Ctrl+V` on Windows/Linux (in `commands.rs`). Users do **not** need to add it manually to their `.wezterm.lua`. If a user explicitly adds `wezterm.action.PasteImageToSshUpload` to their config and later runs an upstream wezterm binary (without this feature), they will see a configuration error:

> `PasteImageToSshUpload` is not a valid KeyAssignment variant

This is expected — the variant only exists in builds that include this PR. The default binding approach avoids this issue entirely: upstream builds simply won't register the binding, while builds with this PR will automatically include it.

## Motivation

When working with remote CLI tools (e.g., Claude Code over SSH), there's no easy way to share clipboard screenshots. This feature enables pasting clipboard images directly — either uploading to a remote server or saving locally — and inserting the file path into the terminal.

The smart paste behavior (`Ctrl+V` on Windows/Linux, `Cmd+V` on macOS) provides a seamless experience: text is pasted as usual, and when the clipboard only contains an image, the image is saved locally and its path is pasted instead.

**Related:** This addresses the use case discussed in #7272 (Pasting Images).

## Test plan

- [x] `cargo build --release` succeeds on Windows and macOS
- [x] `cargo test -p config` — tests pass (including image_paste_local_path platform defaults)
- [x] `cargo test -p wezterm-gui` — keybinding unit tests pass (Super+V, Ctrl+V, Ctrl+Shift+V, permute_keys)
- [x] Manual test on Windows: Ctrl+V pastes text normally in local terminal
- [x] Manual test on Windows: Ctrl+V with image-only clipboard saves to temp dir and pastes path
- [x] Manual test on macOS: Cmd+V smart paste works as before
- [x] Manual test: normal text paste (Cmd+V / Ctrl+V) still works as before
- [ ] CI build on Linux


alarmz:
Withdrawing this PR. Will resubmit after further updates.

Issue #4167, opened 2023-08-19
Characters in the command line after entering tmux


ziombo:
### What Operating System(s) are you seeing this problem on?

Windows

### WezTerm version

20230712-072601-f4abf8fd and 20230817-092749-87a00992 (nightly)

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

I'm using wezterm on windows and use it with WSL2 Ubuntu. I upgraded wezterm to the latest version and since then whenever I enter tmux (with 'tmux' command) I get some characters insterted by default: 
![image](https://github.com/wez/wezterm/assets/15613606/548cf1a0-c232-45c1-8ded-aaccb8de471f)
61;6;7;22;23;24;28;32;42c

It's always the same characters and this doesn't happen when I run tmux within other terminals like Windows Terminal or standard command line.

### Configuration

No changes in configuration, only updated wezterm to the latest version (also checked if something changes with nightly, but it doesn't)

wez:
Please capture a terminal recording:

* Launch wezterm.  If possible, please use the default terminal size of 80x24 cells as it helps to keep things smaller and easier to manage.
* Inside that terminal run `wezterm record` to start a recording session.
* Run through your reproduction steps
* Then type `exit`
* You should see a message like:

```
*** Finished recording to /var/tmp/wezterm-recording-sF6B3u.cast.txt
```

* Attach the file that it produced to this issue.

The file is an asciicast (compatible with https://asciinema.org/) and can also be replayed using `wezterm replay`.

The terminal recording allows me to replicate what is being sent to the terminal without requiring me to install the same applications as you and replicate your configuration for everything.

ferdinandyb:
[wezterm-recording-dtHqGH.cast.txt](https://github.com/wez/wezterm/files/12556742/wezterm-recording-dtHqGH.cast.txt)

I can repro a very similar issue, wezterm consistently prints it's current version number (20230712-072601-f4abf8fd) into the terminal with possibly some other non-printable characters. Reattaching to a tmux session with vim open is not fun :)

wez:
@ferdinandyb see https://github.com/wez/wezterm/issues/2060

ferdinandyb:
Hmm, I tried "reasonable" numbers, but even `set-option -g escape-time 10000` reproduces it.

wez:
@ferdinandyb please open your own separate issue

[...]

PR #7626, opened 2026-02-28
fix: suppress duplicate DA responses for tmux control mode panes

randalmurphal:
## Summary

When using wezterm with `tmux -CC` (control mode), applications that query terminal Device Attributes (DA1/DA2/DA3) — like neovim — cause garbage characters to appear in the shell after exiting:

```
^[[?65;4;6;18;22;52c
```

**Root cause:** Both tmux and wezterm respond to DA queries, producing duplicate responses:

1. **tmux's terminal emulator** (`input.c`) responds synchronously via the pane PTY — the application receives this immediately
2. **wezterm's terminal state machine** also generates a response, which travels back through the tmux control protocol as a `send-keys` command. This round-trip is slow enough that by the time it arrives, the querying application has exited and the shell receives it as garbage input

**Fix:** Add a `suppress_device_attributes` flag to `TerminalState` (following the existing pattern of `suppress_initial_title_change` and `enable_conpty_quirks`) and set it when creating tmux control mode panes. Since tmux already handles DA responses for pane applications, the duplicate response from wezterm is both redundant and harmful.

## Changes

- `term/src/terminalstate/mod.rs`: Added `suppress_device_attributes` field and setter; guarded DA1, DA2, and DA3 response paths
- `mux/src/tmux_commands.rs`: Call `suppress_device_attributes()` when creating tmux control mode panes

## Test plan

- [x] Verified fix locally: launched wezterm with `tmux -CC`, opened and exited neovim repeatedly — no garbage characters
- [x] Verified without fix: same scenario reproduces the garbage characters consistently
- [x] `cargo fmt` clean
- [x] `cargo test` passes (3 pre-existing SFTP test failures unrelated to this change)
- [x] Non-tmux-CC usage unaffected (flag only set for tmux control mode panes)

Relates to #4167

### DIFF
--- a/mux/src/tmux_commands.rs
+++ b/mux/src/tmux_commands.rs
@@ -211,13 +211,19 @@ impl TmuxDomainState {
             active_lock: active_lock.clone(),
         };
 
-        let terminal = wezterm_term::Terminal::new(
+        let mut terminal = wezterm_term::Terminal::new(
             size,
             std::sync::Arc::new(config::TermConfig::new()),
             "WezTerm",
             config::wezterm_version(),
             Box::new(writer.clone()),
         );
+        // Suppress DA responses for tmux control mode panes.
+        // tmux's own terminal emulator already responds to DA queries
+        // from pane applications. If wezterm also responds, the response
+        // travels back via send-keys and arrives after the querying
+        // application has exited, leaking as garbage in the shell.
+        terminal.suppress_device_attributes();
 
         Ok(Arc::new(LocalPane::new(
             local_pane_id,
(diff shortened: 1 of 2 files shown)


randalmurphal:
No tests added bc the existing TestTerm infrastructure looks like it verifies screen state but doesn't capture writer output (responses sent back to the PTY). DA suppression affects writer output only, similar to the untested suppress_initial_title_change flag. Happy to add tests if there's a preferred pattern for asserting writer behavior, just let me know.

bew:
NOTE (for me): related PR: #7292

PR #7643, opened 2026-03-09
Focus the originating pane/tab when a toast notification is clicked

### LINKED ISSUE #7171: Toast Notification Changing to Tab that sent it
I have started using the OSC codes to send notifications from my scripts which is amazing. The only thing is that when I click the notification (assuming it is not a URL notification) it just takes be back to WezTerm, not the tab/pane that sent the notification. 

I would be willing to contribute this feature, but I wanted to check in here first if it would be welcome and on a few details.

First off, I think the switching back to the window is a OS/WM specific thing, so I don't think that could be done in any reasonable way for cross platform.

With a quick look through the code it seems like we could pass along the (tab|pane)_id like we do with the url then switch back to those when we get the callback from the notification. Seems simple enough unless I am missing anything obvious.

The other question is whether this should be behind a config value. It seems like a pretty reasonable default action when it is basically doing nothing right now. It could annoy people who are not used to it though. Also if it gets put behind a config value, that could open up the future for a custom notification handler where WezTerm would execute a command instead/in addition to of switch to the pane.

I would love it to switch to the tab and pane that sent the notification.

Not sure of any better options I am not thinking of.

ferologics:
## Summary

When a terminal program sends an OSC 777 / OSC 9 / OSC 99 notification, clicking the resulting toast now activates the **pane and tab that emitted it**, instead of just bringing the WezTerm window to the foreground.

This resolves the FIXME that was already in `frontend.rs`:

```rust
// FIXME: if notification.focus is true, we should do
// something here to arrange to focus pane_id when the
// notification is clicked
```

Fixes #7171

## Approach

**`ToastNotification`** gets a new `on_click: Option<Box<dyn FnOnce() + Send + 'static>>` field — an opaque callback invoked when the user clicks the notification. This keeps the notification crate decoupled from the mux.

**`frontend.rs`** constructs the callback to call `Mux::focus_pane_and_containing_tab(pane_id)` on the main thread (same pattern used by `MuxNotification::PaneFocused`). Only set when `focus: true` in the alert.

**Platform backends** invoke the callback from their click handlers:

| Platform | Mechanism |
|----------|-----------|
| **macOS** (UNUserNotificationCenter) | Global `HashMap<UUID, callback>` registry. Stored on dispatch, invoked from `did_receive_notification`, cleaned up on click/timeout. |
| **Linux** (dbus) | Invoked inline in the existing `action_invoked` stream handler. |
| **Windows** (toast) | Invoked from the `Activated` handler, wrapped in a `Mutex` to satisfy the `Fn` (not `FnOnce`) bound on `TypedEventHandler`. |

## Changes

- `wezterm-toast-notification/src/lib.rs` — Add `on_click` field, manual `Debug` impl (replaces derive since `Box<dyn FnOnce>` isn't `Debug`/`Clone`)
- `wezterm-toast-notification/src/macos.rs` — Callback registry + delegate integration
- `wezterm-toast-notification/src/dbus.rs` — Invoke callback on action, also fixed a bug where clicking a notification without a URL would not close the listener streams
- `wezterm-toast-notification/src/windows.rs` — Invoke callback on activation
- `wezterm-gui/src/frontend.rs` — Wire up `focus_pane_and_containing_tab` via `on_click`
- `wezterm-gui/src/scripting/guiwin.rs`, `wezterm-font/src/lib.rs` — Add `on_click: None` to existing struct literals

## Testing

- Built and verified `cargo check` passes on macOS (all existing warnings unchanged)
- The macOS path is the primary one I can test; Linux/Windows changes follow the same pattern and are structurally simple

### DIFF
--- a/wezterm-font/src/lib.rs
+++ b/wezterm-font/src/lib.rs
@@ -440,6 +440,7 @@ impl FallbackResolveInfo {
                     ),
                     url: Some(url.to_string()),
                     timeout: Some(Duration::from_secs(15)),
+                    on_click: None,
                 }
                 .show();
             } else {
--- a/wezterm-gui/src/frontend.rs
+++ b/wezterm-gui/src/frontend.rs
@@ -15,7 +15,7 @@ use std::collections::{BTreeMap, HashSet};
 use std::rc::Rc;
 use std::sync::Arc;
 use wezterm_term::{Alert, ClipboardSelection};
-use wezterm_toast_notification::*;
+use wezterm_toast_notification;
 
 pub struct GuiFrontEnd {
     connection: Rc<Connection>,
@@ -100,7 +100,7 @@ impl GuiFrontEnd {
                         Alert::ToastNotification {
                             title,
                             body,
-                            focus: _,
+                            focus,
                         },
                 } => {
                     let mux = Mux::get();
@@ -124,10 +124,39 @@ impl GuiFrontEnd {
                             if show {
                                 let message = if title.is_none() { "" } else { &body };
                                 let title = title.as_ref().unwrap_or(&body);
-                                // FIXME: if notification.focus is true, we should do
-                                // something here to arrange to focus pane_id when the
-                                // notification is clicked
-                                persistent_toast_notification(title, message);
+
+                                // When focus is requested, arrange to activate
+                                // the pane/tab that sent the notification when
+                                // the user clicks on it.
+                                let on_click = if focus {
+                                    Some(Box::new(move || {
+                                        promise::spawn::spawn_into_main_thread(async move {
+                                            let mux = Mux::get();
+                                            if let Err(err) =
+                                                mux.focus_pane_and_containing_tab(pane_id)
+                                            {
+                                                log::error!(
+                                                    "Error focusing pane from notification \
+                                                     click: {err:#}"
+                                                );
+                                            }
+                                        })
+                                        .detach();
+                                    })
+                                        as Box<dyn FnOnce() + Send + 'static>)
+                                } else {
+                                    None
+                                };
+
+                                wezterm_toast_notification::show(
+                                    wezterm_toast_notification::ToastNotification {
+                                        title: title.to_string(),
+                                        message: message.to_string(),
+                                        url: None,
+                                        timeout: None,
+                                        on_click,
+                                    },
+                                );
                             }
                         }
                     }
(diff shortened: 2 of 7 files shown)


vekexasia:
I confirm this works on my machine (windows)

ycna07:
work on Linux with this patch
```diff
diff --git a/wezterm-toast-notification/src/dbus.rs b/wezterm-toast-notification/src/dbus.rs
index 090e0bafd..442f288ab 100644
--- a/wezterm-toast-notification/src/dbus.rs
+++ b/wezterm-toast-notification/src/dbus.rs
@@ -119,9 +119,6 @@ async fn show_notif_impl(notif: ToastNotification) -> Result<(), Box<dyn std::er
         )
         .await?;
 
-    // Move on_click out so we can consume it (FnOnce) in the action handler.
-    let mut on_click = notif.on_click;
-
     let (mut invoked_stream, abort_invoked) = abortable(proxy.receive_action_invoked().await?);
     let (mut closed_stream, abort_closed) = abortable(proxy.receive_notification_closed().await?);
 
@@ -133,7 +130,7 @@ async fn show_notif_impl(notif: ToastNotification) -> Result<(), Box<dyn std::er
                     if let Some(url) = notif.url.as_ref() {
                         wezterm_open_url::open_url(url);
                     }
-                    if let Some(callback) = on_click.take() {
+                    if let Some(callback) = notif.on_click {
                         callback();
                     }
                     abort_closed.abort();
```

yamam:
I tested this PR on Windows. The click callback correctly selected the
originating pane/tab, but the WezTerm GUI window remained in the background.

The callback only called:

```rust
mux.focus_pane_and_containing_tab(pane_id)
```

Adding an explicit focus call for the GUI window fixed the issue for me:

```diff
diff --git a/wezterm-gui/src/frontend.rs b/wezterm-gui/src/frontend.rs
index 86cd9c332..1eee4396e 100644
--- a/wezterm-gui/src/frontend.rs
+++ b/wezterm-gui/src/frontend.rs
@@ -140,6 +140,13 @@ impl GuiFrontEnd {
                                                      click: {err:#}"
                                                 );
                                             }
+
+                                            let front_end = crate::frontend::front_end();
+                                            if let Some(gui_window) =
+                                                front_end.gui_window_for_mux_window(window_id)
+                                            {
+                                                gui_window.window.focus();
+                                            }
                                         })
                                         .detach();
                                     })

```

It looks like Windows does not automatically activate the WezTerm window when
a URL-less toast notification is clicked, unlike the environment assumed in
the original issue.

Would it make sense to include the GUI window focus in the callback as well?
I confirmed that this brings WezTerm to the foreground and selects the
originating pane/tab on Windows.

PR #7673, opened 2026-03-19
feat: add per-pane title bars

fireboy1919:
## Summary

Adds a `pane_border_status` configuration option that renders a one-row title bar at the top or bottom of each split pane, with proper layout so terminal content never overlaps the bar and the terminal process sees the correct dimensions.

### Motivation

When working with multiple split panes it can be hard to identify what each pane is doing at a glance.  Many terminal multiplexers (tmux, Zellij, etc.) solve this with per-pane title bars; WezTerm currently has no equivalent.

### What this PR does

* **New config field** `pane_border_status: PaneBorderStatus` (enum `Off` | `Top` | `Bottom`, defaults to `Off`).  Existing configs are unaffected.

* **New Lua event** `format-pane-title` fired for each pane on every frame, mirroring `format-tab-title`.  The callback receives `(pane, panes, tabs, config)` and may return a plain string or a `FormatItem` table.  Falls back to the pane's window title when no callback is registered.

* **`paint_pane_title_bar()`** in a new `termwindow/render/pane_bar.rs` draws a filled background rect and calls `render_screen_line()` at the title bar's pixel row.

* **Correct layout — no overlap, correct terminal dimensions:**
  * `get_pos_panes_for_tab()` resizes each terminal process to `height - 1` rows (lazily, only when the current size differs).  The process therefore never writes into the title bar row.
  * For `Top`, `paint_pane` adds one cell of pixel offset so terminal lines render starting at the row below the title bar.
  * For `Bottom`, the title bar sits at `pos.top + pos.height` (the row immediately below the terminal content).
  * `PositionedPane.height` is updated to `height - 1` so all downstream rendering uses the correct cell count.

* **Close button** — a `×` rendered at the right edge of each pane title bar closes (kills) that pane on click.  Controlled by the new `show_close_pane_button_in_pane_bar` config option (defaults to `true`).  Implemented via a new `UIItemType::ClosePane(PaneId)` variant that plugs into the existing UI hit-test and mouse-dispatch infrastructure used by the tab close button.

* **6 unit tests** covering `parse_status_text` and `default_pane_title`.

### Example config

```lua
config.pane_border_status = "Top"

wezterm.on("format-pane-title", function(pane, panes, tabs, config)
  local zoomed = pane.is_zoomed and " [Z]" or ""
  local active = pane.is_active and " *" or ""
  return pane.title .. zoomed .. active
end)

-- optional: disable the close button
-- config.show_close_pane_button_in_pane_bar = false
```

### Testing

```
cargo test -p wezterm-gui
```

All 18 tests pass (6 new, 12 existing).

### DIFF
--- a/wezterm-gui/src/termwindow/resize.rs
+++ b/wezterm-gui/src/termwindow/resize.rs
@@ -2,6 +2,7 @@ use crate::resize_increment_calculator::ResizeIncrementCalculator;
 use crate::utilsprites::RenderMetrics;
 use ::window::{Dimensions, ResizeIncrement, Window, WindowOps, WindowState};
 use config::{ConfigHandle, DimensionContext};
+use config::PaneBorderStatus;
 use mux::Mux;
 use std::rc::Rc;
 use wezterm_font::FontConfiguration;
@@ -171,7 +172,7 @@ impl super::TermWindow {
 
         let border = self.get_os_border();
 
-        let (size, dims, ri_calc) = if let Some(cell_dims) = scale_changed_cells {
+        let (size, dims, ri_calc, window_base_rows) = if let Some(cell_dims) = scale_changed_cells {
             // Scaling preserves existing terminal dimensions, yielding a new
             // overall set of window dimensions
             let size = TerminalSize {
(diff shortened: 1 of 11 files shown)


fireboy1919:
wezterm-panes

wezterm-gui/src/termwindow/resize.rs
fireboy1919, line 1: I'm not 100% convinced resizing as it currently works is correct.  When you resize with a mouse drag, a bunch of events get sent, and some sort of debounce code might be needed to make it behave better.

PR #7674, opened 2026-03-19
Fix text_min_contrast_ratio

waj:
The current color shifts made to achieve min contrast ratio are quite flawed. First, I think there's a bug because in line [1085](https://github.com/wezterm/wezterm/blob/05343b387085842b434d267f91b6b0ec157e4331/color-types/src/lib.rs#L1085) is using `reduced_col` instead of `increased_col` but even fixing that the color adjust turns very quickly to just black.

This is mainly because the algorithm just applies the contrast value obtained from the WCAG formula and use it as the L value for OKlab. That's essentially wrong because those values are not compatible. There's actually no way to compute directly the required L value to achieve the desired contrast ratio, so instead I replaced that with a binary search.

Also I made a change to prevent color adjustments on block chars.

This is how it looks before and after the changes. Notice how the colors are preserved in the new version:
image
image

Also note how my prompt gets all broken with the current version:
image

While the new version keeps the block chars untouched:
image

Finally, the entire color theme, before and after. Note that not only some colors turn too quickly to black, but some combinations are actually not readable at all:
image

image

I'm using `text_min_contrast_ratio = 2`.

### DIFF
--- a/color-types/src/lib.rs
+++ b/color-types/src/lib.rs
@@ -1049,7 +1049,7 @@ impl LinearRgba {
         let g = -1.2684380046 * l_ + 2.6097574011 * m_ - 0.3413193965 * s_;
         let b = -0.0041960863 * l_ - 0.7034186147 * m_ + 1.7076147010 * s_;
 
-        Self(r, g, b, alpha)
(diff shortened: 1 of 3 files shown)


waj:
Hey can I have at least an opinion on this? ☺️
I've been using this branch since then and it's working great for me.

PR #7679, opened 2026-03-21
Add vertical tab bar with drag-to-resize and left/right positioning

### LINKED ISSUE #3537: Allow the tab bar to be rendered on the side

al3rez:
## Summary

- Adds a vertical tab bar that renders on the left or right side of the window
- Supports drag-to-resize via a separator edge handle (min 80px, max 50% of window width)
- Close button on tabs hidden by default, revealed on hover for a cleaner look
- Separator line color follows theme via `inactive_tab_edge`
- Full mouse event handling: click tabs, hover cursor changes, drag resize

## Configuration

```lua
config.tab_bar_vertical = true
config.tab_bar_vertical_width = 200  -- pixels, default 200
config.tab_bar_vertical_position = "Left"  -- "Left" or "Right"
```

## Implementation Details

- New `build_vertical_fancy_tab_bar` method renders tabs as stacked `Block` elements
- `VerticalTabBarPosition` enum controls left/right placement
- Right positioning uses `translate()` approach (compute at x=0, translate after) to ensure identical internal layout
- Content area (panes, splits, mouse coordinates) properly offset via `tab_bar_left_offset()`
- Resize increments account for vertical tab bar width
- Box model border workaround for separator line (documented at `box_model.rs:1236`)
- Extracted shared helpers: `tab_bar_colors()`, `bar_element_colors()`, `titlebar_bg_linear()`
- Constants: `MIN_VERTICAL_TAB_BAR_WIDTH`, `MAX_VERTICAL_TAB_BAR_WIDTH_RATIO`, `RESIZE_HANDLE_WIDTH`

## Files Changed (10 files, +590 -33)

- `config/src/config.rs` — New config fields and `VerticalTabBarPosition` enum
- `wezterm-gui/src/termwindow/mod.rs` — `UIItemType::VerticalTabBarResize`, constants, state
- `wezterm-gui/src/termwindow/mouseevent.rs` — Resize drag handling, cursor changes
- `wezterm-gui/src/termwindow/render/fancy_tab_bar.rs` — Vertical tab bar builder + helpers
- `wezterm-gui/src/termwindow/render/tab_bar.rs` — Height/width methods, resize handle registration
- `wezterm-gui/src/termwindow/render/pane.rs` — Content offset for vertical bar
- `wezterm-gui/src/termwindow/render/split.rs` — Split offset for vertical bar
- `wezterm-gui/src/termwindow/render/mod.rs` — Horizontal gap calculation
- `wezterm-gui/src/termwindow/resize.rs` — Terminal size accounting
- `wezterm-gui/src/resize_increment_calculator.rs` — Resize increment base width

## Test plan

- [ ] Verify vertical tab bar renders on left (default) and right
- [ ] Verify drag-to-resize works in both positions
- [ ] Verify close button appears on hover, hidden by default
- [ ] Verify pane content, splits, and mouse clicks are properly offset
- [ ] Verify hot-reload of `tab_bar_vertical_position` between Left/Right
- [ ] Verify horizontal tab bar still works when `tab_bar_vertical = false`

### DIFF
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -476,8 +476,7 @@ pub struct Config {
     #[dynamic(default)]
     pub tab_bar_at_bottom: bool,
 
-    /// If true, render the tab bar vertically on the left side of the window
-    /// instead of horizontally at the top or bottom.
+    /// If true, render the tab bar vertically instead of horizontally.
     #[dynamic(default)]
     pub tab_bar_vertical: bool,
 
@@ -485,6 +484,10 @@ pub struct Config {
     /// Defaults to 200.
     #[dynamic(default = "default_tab_bar_vertical_width")]
     pub tab_bar_vertical_width: usize,
+
+    /// Position of the vertical tab bar: "Left" or "Right". Defaults to "Left".
+    #[dynamic(default)]
+    pub tab_bar_vertical_position: VerticalTabBarPosition,
 
     #[dynamic(default = "default_true")]
     pub mouse_wheel_scrolls_tabs: bool,
@@ -1988,6 +1991,13 @@ pub enum VerticalWindowContentAlignment {
     Top,
     Center,
     Bottom,
+}
+
+#[derive(Debug, FromDynamic, ToDynamic, Clone, Copy, PartialEq, Eq, Default)]
+pub enum VerticalTabBarPosition {
+    #[default]
+    Left,
+    Right,
 }
 
 #[derive(FromDynamic, ToDynamic, Clone, Copy, Debug, PartialEq, Eq)]
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -476,6 +476,16 @@ pub struct Config {
     #[dynamic(default)]
     pub tab_bar_at_bottom: bool,
 
+    /// If true, render the tab bar vertically on the left side of the window
+    /// instead of horizontally at the top or bottom.
+    #[dynamic(default)]
+    pub tab_bar_vertical: bool,
+
+    /// The width in pixels of the vertical tab bar when tab_bar_vertical is true.
+    /// Defaults to 200.
+    #[dynamic(default = "default_tab_bar_vertical_width")]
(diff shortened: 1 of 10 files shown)


google-goldfish:
Instead of having vertical tab bar width would it be possible to have a percentage instead? Cover 5% or 10% or 15% of the screen so it's uniformly adjusted for monitor/laptop screens.

bew:
> Instead of having vertical tab bar width would it be possible to have a percentage instead? Cover 5% or 10% or 15% of the screen so it's uniformly adjusted for monitor/laptop screens.

Note that with the design note I posted here https://github.com/wezterm/wezterm/issues/3537#issuecomment-5302136011 adding such functionality would simply be adding percentage support to `tab_min/max_width` which could be done independently 🤔

google-goldfish:
Understood thanks, does that mean we accept the implementation done here and take it to `main`?

bew:
not sure about this impl exactly, but we can revisit this after the next release I think, where it should be easier to work / design & plan new features 🤔

PR #7680, opened 2026-03-21
feat: add workspace sidebar panel with tree view and Finder-style UX

Elias-Palmer:
Add a native, toggleable workspace sidebar that displays all workspaces and their tabs in a tree view. Toggle with Ctrl+Shift+E (configurable).

Features:
- Click to switch workspaces and tabs
- Double-click to inline rename workspaces and tabs
- Per-item hover close buttons (red danger zone for workspaces)
- Ghost buttons (+/✕) that reveal on hover
- Active tab indicator (green dot)
- Bottom-pinned "+ New Workspace" footer
- Theme-aware colors from window_frame and tab_bar config
- Configurable width, position (left/right), and default visibility
- Scroll support and persistent expand/collapse state

New config options: sidebar_width, sidebar_position, sidebar_default_visible
New key assignment: ToggleSidebar

Alright imma be real, this feature is 100% vibe coded. I wanted a better way to manage wor spaces and tabs and im a visual kinda guy. 100% reject this or redesign it if you want idrc, i just figure id share it.

### DIFF
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -512,6 +512,18 @@ pub struct Config {
     #[dynamic(try_from = "crate::units::PixelUnit", default = "default_half_cell")]
     pub min_scroll_bar_height: Dimension,
 
+    #[dynamic(
+        try_from = "crate::units::PixelUnit",
+        default = "default_sidebar_width"
+    )]
+    pub sidebar_width: Dimension,
+
+    #[dynamic(default)]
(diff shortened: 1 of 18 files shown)


bew:
It's an interesting idea, and I'm glad you made yourself a solution for your own needs, but we won't accept vibe-coded solutions without careful design and human thoughts.

PR #7682, opened 2026-03-21
Vim-like cursor navigations

sherocktong:
### Summary

This PR adds comprehensive Vim-style improvements to WezTerm's copy mode across two phases: first, a full set of Vim navigation motions and yank operations; second, always-on relative line numbers and a cleaner search workflow.

---

#### 1. Vim-like Navigations — *2026-03-22*

**What:**
- Added word motions: `w` / `W` (forward word/WORD), `b` / `B` (backward word/WORD), `e` / `E` (end of word/WORD)
- Added character jump: `f{c}` / `F{c}` (jump to char), `t{c}` / `T{c}` (jump before char), `;` / `,` (repeat jump)
- Added viewport jumps: `H` / `M` / `L` (top/middle/bottom of viewport), `gg` (scroll top), `G` (scroll bottom)
- Added selection modes: `v` (cell), `V` / `Shift+V` (line), `Ctrl+V` (block)
- Added yank operations: `y` (yank selection), `yiw` / `yiW` / `yie` / `yiE` (yank text objects), `yy` (yank line)
- Added text object support: `iw`, `iW`, `ie`, `iE` for inner word variants
- New `CopyModeAssignment` variants: `BeginTextObjectYank`, `BeginTextObjectSelection`, `BeginVimPrefixG`, `JumpForward`, `JumpBackward`, `JumpAgain`, `JumpReverse`

**Why:** Copy mode previously only supported basic `h`/`j`/`k`/`l` movement and simple selection. Vim users needed word-level navigation, character jumps, and efficient yank operations to make copy mode genuinely useful for selecting and copying terminal output.

---

#### 2. Fix `e` Navigation — *2026-03-22*

**What:** Fixed `e` / `E` to land on the *last character* of a word rather than one cell past it.

**Why:** The off-by-one made `e` inconsistent with Vim's behaviour, putting the cursor one column beyond the word end.

---

#### 3. Count Prefix for `j` / `k` — *2026-03-23*

**What:**
- Digit keys `1`–`9` now accumulate a pending count while in copy mode
- `j` / `k` (and arrow keys) consume the count and move that many rows at once (e.g. `10j` = move down 10 lines)
- Updated docs for `MoveUp` / `MoveDown`

**Why:** Navigating large terminal scrollbacks by holding `j`/`k` is slow and imprecise. Count prefixes are the standard Vim idiom for jumping directly to a known distance.

---

#### 4. Fix `yy` to Copy Full Line — *2026-03-23*

**What:** Fixed `yy` to correctly select the entire current line before yanking, then close copy mode.

**Why:** The previous implementation did not fully expand the line selection before copying, resulting in partial yanks.

---

#### 5. Relative Line Numbers Gutter — *2026-04-03*

**What:**
- An always-on gutter renders on the left of every line in copy mode — no toggle needed
- Cursor line shows `0`; all other lines show their absolute distance from the cursor
- Content is shifted right via `line.insert_cell(0, ...)` so the gutter never overwrites terminal text
- Gutter width scales with viewport height automatically
- Cursor `x` is offset by `gutter_width` in `get_cursor_position()` so it tracks over content correctly
- `line.clear_appdata()` busts the `line_quad_cache` after each render to prevent stale cached lines hiding the gutter

**Why:** Relative line numbers make count motions (`5j`, `12k`) practical — you can see at a glance exactly how far away any line is without counting.

---

#### 6. Search Navigation Overhaul — *2026-04-03*

**What:**
- `/` added to `copy_key_table` → opens the search box from copy mode
- `n` / `N` added to `copy_key_table` → navigate forward/backward through matches without re-opening the search box
- `Enter` in `search_key_table` changed from `PriorMatch` to `AcceptPattern` → exits the search text box, keeps results highlighted, and returns to copy mode navigation
- `Ctrl+P` / `Ctrl+N` in `search_key_table` now also bind raw control codes `\x10` / `\x0e` so they work regardless of terminal key normalisation

**Why:** Previously, pressing `Enter` in the search box triggered `PriorMatch` instead of exiting the box, and there was no way to leave the search input and use `hjkl` without losing results. The new flow mirrors Vim: `/pattern` → `Enter` → `n`/`N` to step through matches.

---

#### 7. Removed Inactive Match Highlights & Selection-on-Jump — *2026-04-03*

**What:**
- Inactive search matches no longer render with a purple background — only the active match is highlighted (yellow)
- Jumping to a match via `n`/`N` no longer creates a visual selection range; `activate_match_number` now calls `clear_selection()` and adjusts the viewport instead of `adjust_selection()`

**Why:** Purple highlights on every match were visually noisy and obscured the terminal content. The selection range created on jump looked like accidental text selection — it couldn't be dismissed without pressing `v` or closing copy mode.


bew:
Please describe what you're trying to fix, why you made these changes

sherocktong:
> Please describe what you're trying to fix, why you made these changes

This PR is intended to make copy mode behave more consistently with Vim. I found a few gaps where common motions/operators were either missing or behaved unexpectedly: e didn’t always move to the expected word end, count-based vertical motions like 10j/10k weren’t supported, and yy didn’t yank the current line. I added the missing Vim-like navigation/operator handling and fixed those behaviors so copy mode is more predictable for users relying on Vim muscle memory.

rudrabhoj:
Any update on this PR?

Btw @sherocktong, did you try copy mode in zellij? Wezterm's copy mode looks broken there.

PR #7683, opened 2026-03-22
fonts: add configurable font thickening

### LINKED ISSUE #3774: Font looks thin and spaced out compared to other terminals

macOS

_No response_

### WezTerm version

wezterm version: 20230524-062011-1cd340bb aarch64-apple-darwin

### Describe the bug

Not sure if this actually is a bug, but font rendering by default looks very different from other terminals I tried. Specifically, fonts look thinner and are more spaced out horizontally on WezTerm compared to Terminal.app, Alacritty and Kitty.

On default settings, here is what WezTerm looks like with Menlo 14pt:

image

And Alacritty for reference, Kitty looks the same:

image

I understand that font rendering is subjective, choices for anti-aliasing methods and such can be up to the user. I tried these tweaks and got pretty close to other terminals:
- To get thicker looking strokes, switch to `freetype_render_target = "HorizontalLcd"`
- To squeeze things horizontally, set `cell_width = 0.9`

The stroke thickness I understand might just be a different default choice, but the cell width feels like a bug to me. As far as I understand, it should choose the width of the monospace glyphs as defined in the font, which should be consistent across terminals, correct?

Here is the result after config tweaks in WezTerm:

image

Issue seems a bit similar to https://github.com/wez/wezterm/issues/689, but there wasn't really any result from that.

### To Reproduce

_No response_

### Configuration

```
config.font = wezterm.font('Menlo')
config.font_size = 14
config.freetype_load_target = "Light"
config.freetype_render_target = "HorizontalLcd"
config.cell_width = 0.9
```

### Expected Behavior

_No response_

### Logs

_No response_

### Anything else?

_No response_

### LINKED ISSUE #7192: Option to disable thin stroke

I am using WezTerm on Mac OS and I want to tweak its font rendering like the builtin Terminal.app or iTerm2 with "thin-stroke" turned off.

Here's how iterm2 with thin-stroke ON and Wezterm renders.

Image

iterm2 on the left, wezterm on the right. They looks basically the same.

While iterm2 with thin-stroke OFF and Wezterm renders.

Image

I prefer the to turn thin-stoke off and make text bolder and smooth. So I suggest maybe adding an option to disable thin stroke.

**Additional context**

canova:
Hi! First of all, thanks a lot for the project!

I was considering about switching to WezTerm and the major blocker that I noticed was the thin font rendering compared to other terminals like Alacritty or Ghostty on macOS. Then searching in the existing issues, I noticed some filed already that had some attention. So this fixes #3774 and fixes #7192.

I initially had the same issue when I first tried Ghostty, and I managed to fix it by using [`font-thicken = true`](https://ghostty.org/docs/config/reference#font-thicken) in their config. That's why I was looking at how Ghostty implemented it, and it was actually pretty straightforward. So I did something similar to WezTerm in this PR. I might be biased, but I think this is one of big reasons for people to bounce after trying it a bit (at least on macOS).

So this PR adds a `font_thicken` option, along with `font_thicken_strength`, to render glyphs with a slightly heavier stroke.

I saw that Ghostty's implementation was macOS only. But I think it's because they use CoreText/CoreGraphics to do it. This PR only adds it to FreeType rasterizer, and I didn't limit this config option to macOS only since it isn't platform dependent. But happy to hear your thoughts and would be happy to change if you prefer. I `FT_Outline_EmboldenXY` to implement the thickening right now, which works fairly well in my local testing. See also the side by side comparisons below.

One thing that I'm not super happy with is the 36 magic number that I pass to `FT_Outline_EmboldenXY`. Since we don't use CoreText, this is mostly an approximation instead of a native value. But I fine tuned it to look very close and I'm happy with this current state now.

Please let me know what you think!

Before (left is ghostty, right is wezterm):
Screenshot 2026-03-21 at 11 59 14 PM

After (left is ghostty, right is wezterm):
Screenshot 2026-03-22 at 12 37 34 AM

### DIFF
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -288,6 +288,19 @@ pub struct Config {
     pub freetype_render_target: Option<FreeTypeLoadTarget>,
     #[dynamic(default)]
     pub freetype_load_flags: Option<FreeTypeLoadFlags>,
+    /// Draw fonts with a thicker stroke, if supported.
+    #[dynamic(default)]
+    pub font_thicken: bool,
+    /// Strength of thickening when `font_thicken` is enabled.
+    ///
+    /// Valid values are integers between `0` and `255`. `0` does not
+    /// correspond to *no* thickening, rather it corresponds to the
+    /// lightest available thickening.
+    ///
+    /// Has no effect when `font_thicken` is set to `false`.
+    ///
+    #[dynamic(default = "default_font_thicken_strength")]
+    pub font_thicken_strength: u8,
 
     /// Selects the freetype interpret version to use.
     /// Likely values are 35, 38 and 40 which have different
@@ -2192,4 +2205,8 @@ fn default_macos_forward_mods() -> Modifiers {
 
 fn default_colr_rasterizer() -> FontRasterizerSelection {
     FontRasterizerSelection::Harfbuzz
+}
+
+fn default_font_thicken_strength() -> u8 {
+    255
 }
--- a/docs/config/fonts.md
+++ b/docs/config/fonts.md
@@ -58,6 +58,8 @@ Additional options for configuring fonts can be found elsewhere in the docs:
 * [font_locator](lua/config/font_locator.md) - override the system font resolver
 * [font_rules](lua/config/font_rules.md) - advanced control over which fonts are used for italic, bold and other textual styles
 * [font_shaper](lua/config/font_shaper.md) - affects kerning and ligatures
+* [font_thicken](lua/config/font_thicken.md) - draw fonts with a thicker stroke
+* [font_thicken_strength](lua/config/font_thicken_strength.md) - control the amount of thickening
 * [font_size](lua/config/font_size.md) - change the size of the text
 * [freetype_load_flags](lua/config/freetype_load_flags.md) - advanced hinting configuration
 * [freetype_load_target](lua/config/freetype_load_target.md) - configure hinting and anti-aliasing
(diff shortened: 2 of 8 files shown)


akthe-at:
Can't you already do this in your config via lua? 

https://wezterm.org/config/fonts.html
https://wezterm.org/config/lua/wezterm/font.html

canova:
> Can't you already do this in your config via lua?
> 
> https://wezterm.org/config/fonts.html https://wezterm.org/config/lua/wezterm/font.html

I don't think you can, this is different compared to the font weight. This allows both regular and bold text to be thickened at the same time, whereas font `weight = "Bold"` changes the default font to use bold all the time. This doesn't work well with programs that use both the regular and bold text like neovim because both of them will look exactly the same. But with this config, they will look different correctly.

I'm not 100% sure if you meant the font weight though. Please let me know which config specifically you refer if it's not it. I tried a bunch of things there but none of them worked for me.

doulighan:
> > Can't you already do this in your config via lua?
> > https://wezterm.org/config/fonts.html https://wezterm.org/config/lua/wezterm/font.html
> 
> I don't think you can, this is different compared to the font weight. This allows both regular and bold text to be thickened at the same time, whereas font `weight = "Bold"` changes the default font to use bold all the time. This doesn't work well with programs that use both the regular and bold text like neovim because both of them will look exactly the same. But with this config, they will look different correctly.
> 
> I'm not 100% sure if you meant the font weight though. Please let me know which config specifically you refer if it's not it. I tried a bunch of things there but none of them worked for me.

I've also messed around with all the built-in font settings extensively, and have not been able to get the right effect. I've even tried to patch it myself but failed. This seems great, I'm patching my build right now.

PR #7684, opened 2026-03-22
Feature: add support to sync input between all pane in same tab

SakaYK:
this feature will add wezterm to support sync input at all pane in the same tab.

add this toggle shortcut in the .wezterm.lua config as default

-- Toggle synchronized input to all panes in the current tab
    {
      key = 'i',
      mods = 'CTRL|SHIFT',
      action = wezterm.action.TogglePaneInputSynchronization,
    },

you can change the toggle shortcut as you like.


SakaYK:
Need more test in multiple OS

PR #7690, opened 2026-03-25
feat(termwiz): add dynamic color probing

DCjanus:
## Why

- `termwiz` already has probing helpers for terminal properties, but not for the current dynamic colors.
- In a downstream TUI, querying the default foreground/background colors makes it possible to build theme-aware interpolation instead of relying on coarse `DIM`/`BOLD` fallback.

Here's a small downstream demo showing the motivation for this:

[asciicast](https://asciinema.org/a/869075)

## What

- add `ProbeCapabilities::dynamic_color(&mut self, which: DynamicColorNumber) -> Result<SrgbaTuple>`
- parse `OperatingSystemCommand::ChangeDynamicColors` replies in `termwiz/src/caps/probed.rs`
- add a small `dynamic_color_wave` example to show the downstream use case

## Testing

```bash
cargo test --all
cargo run -p termwiz --example dynamic_color_wave
```


termwiz/src/caps/probed.rs
```diff
@@ -84,6 +355,51 @@ impl<'a> ProbeCapabilities<'a> {
...
+        let query = OperatingSystemCommand::ChangeDynamicColors(which, vec![ColorOrQuery::Query]);
+        let dev_attributes = CSI::Device(Box::new(Device::RequestPrimaryDeviceAttributes));
+        write!(self.write, "{query}")?;
+        self.write.flush()?;
+
+        // In tmux and ConPTY, the primary device attributes response can be
+        // reordered ahead of the dynamic color reply if we send it immediately.
+        // Delay a little before using it as a sentinel, matching the style used
+        // by the other probes in this module.
+        std::thread::sleep(std::time::Duration::from_millis(100));
```
bew: Please wrap this unfortunate sleep in a check for tmux or cfg windows, similar to the `screen_size` probe, to avoid negatively impact other OS/environments.

termwiz/src/caps/probed.rs
```diff
@@ -84,6 +355,51 @@ impl<'a> ProbeCapabilities<'a> {
         self.xt_version_impl(true)
     }
 
+    /// Probe the terminal for the current value of a dynamic color.
+    pub fn dynamic_color(&mut self, which: DynamicColorNumber) -> Result<SrgbaTuple> {
```
bew: in term of code organization, please move this function below `xt_version_impl`, right now this new function is placed between public xt-version-related function and its actual implementation.

termwiz/src/caps/probed.rs
```diff
@@ -84,6 +355,51 @@ impl<'a> ProbeCapabilities<'a> {
...
+                    saw_dev_attributes = true;
+                    return;
+                }
+
+                if color.is_none() {
+                    color = parse_dynamic_color_response(action, which);
+                }
+            });
+
+            if saw_dev_attributes {
```
bew: So if we never get the DA response this will loop forever?
I'm not super familiar with how probing capabilities is supposed to work but this seems odd 🤔

termwiz/src/caps/probed.rs
```diff
@@ -84,6 +355,51 @@ impl<'a> ProbeCapabilities<'a> {
...
+        // Delay a little before using it as a sentinel, matching the style used
+        // by the other probes in this module.
+        std::thread::sleep(std::time::Duration::from_millis(100));
+
+        write!(self.write, "{dev_attributes}")?;
+        self.write.flush()?;
+
+        let mut parser = Parser::new();
+        let mut color = None;
+        loop {
```
bew: instead of a `loop` block, could you keep the same kind of looping as the other probes, with a `done` flag, and matching only the expected responses, setting `done = true` if we get something unexpected

termwiz/src/caps/probed.rs
```diff
@@ -84,6 +355,51 @@ impl<'a> ProbeCapabilities<'a> {
         self.xt_version_impl(true)
     }
 
+    /// Probe the terminal for the current value of a dynamic color.
+    pub fn dynamic_color(&mut self, which: DynamicColorNumber) -> Result<SrgbaTuple> {
+        let query = OperatingSystemCommand::ChangeDynamicColors(which, vec![ColorOrQuery::Query]);
+        let dev_attributes = CSI::Device(Box::new(Device::RequestPrimaryDeviceAttributes));
```
bew: Why do you query for the device attributes (DA) and only reply with the dynamic color if you get a DA response?
Is that a 'known rule' when querying for a dyn color?

termwiz/src/caps/probed.rs
```diff
@@ -43,6 +48,67 @@ impl XtVersion {
 #[cfg(test)]
 mod test {
     use super::*;
+    use crate::color::SrgbaTuple;
```
bew: In term of code organization, please move all new tests in a `test` module placed below `ProbeCapabilities`.

This `test` module here is about testing `XtVersion` that is defined directly above.

Right now having tests for `ProbeCapabilities` _above_ the def/impl feels weird.

termwiz/src/caps/probed.rs
```diff
@@ -55,6 +121,211 @@ mod test {
...
+        let cursor = probe
+            .dynamic_color(DynamicColorNumber::TextCursorColor)
+            .unwrap();
+
+        assert_eq!(foreground, first);
+        assert_eq!(cursor, second);
+    }
+
+    #[test]
+    fn test_dynamic_color_avoids_primary_device_attributes_reordering() {
```
bew: This test has _substantial_ setup with the custom reader/writer defined above in `test` module.
Is that really necessary?
What is this test testing exactly, that the other tests above are not testing?

termwiz/examples/dynamic_color_wave.rs
```diff
@@ -0,0 +1,271 @@
+use std::f32::consts::PI;
```
bew, line 6: Please add a doc comment for the file explaining what this example is about.
Mentioning that this shows how to guery fg/bg terminal colors, but also how to manipulate rgb colors and handle smooth animation.

bew (commented):
Hello! Thanks for the PR, fancy use-case ;)

Did you use AI for writting this PR and/or the PR description?

I don't know when Wez will be able to review/merge this, but here are a few comments/questions already 🤔

DCjanus:
> Did you use AI for writting this PR and/or the PR description?

Yes. I used AI during development, and the PR description was also written with AI assistance.

Before opening the PR, I reviewed the changes myself and tried to keep them close to the existing style. That said, I am still not very familiar with the wezterm codebase, and I may also be missing some explicit or implicit design constraints. If that is the case, I would really appreciate having them pointed out, as you are doing now.

I’m happy to work through the review feedback and adjust the implementation to better match the existing conventions.

Also, I’m not a native English speaker, and writing longer explanations in English is a bit stressful for me. I worry that awkward grammar or phrasing might come across the wrong way, so this reply was also translated with AI assistance.

> I don't know when Wez will be able to review/merge this, but here are a few comments/questions already 🤔

I also understand that this may or may not be something WezTerm wants to merge, and that is completely fine. My own TUI application already depends on my fork of `termwiz`, so this already covers my personal use case. I opened the PR mainly in the hope that, if this direction makes sense for the project, it might save some repeated work for other `termwiz` users with similar needs.

termwiz/examples/dynamic_color_wave.rs
DCjanus: Added a file-level doc comment in [b771da1ea](https://github.com/DCjanus/wezterm/commit/b771da1ea).

termwiz/src/caps/probed.rs
DCjanus: Moved below `xt_version_impl` in [ff6f01398](https://github.com/DCjanus/wezterm/commit/ff6f01398).

termwiz/src/caps/probed.rs
DCjanus: Wrapped the delay in a tmux/windows check in [7c85dc90c](https://github.com/DCjanus/wezterm/commit/7c85dc90c).

termwiz/src/caps/probed.rs
DCjanus: Part of my earlier implementation came from me reading the existing probe pattern too literally.

I had treated the DA response more like something to wait for directly, but looking at the surrounding code again, it makes more sense to treat it as a sentinel that helps bound the probe window. I reworked the loop in [bf23a6d50](https://github.com/DCjanus/wezterm/commit/bf23a6d50) to follow that style more closely: it now uses a `done` flag, keeps only the expected responses, ignores the string terminator, and stops on unexpected input.

This seems more consistent with the existing design, and also more robust in unexpected cases.

termwiz/src/caps/probed.rs
DCjanus: Moved the `ProbeCapabilities` tests into a separate module below it in [8556fbd09](https://github.com/DCjanus/wezterm/commit/8556fbd09).

termwiz/src/caps/probed.rs
DCjanus: I agree this test was too heavy in its current form.

I removed it in [281da8bc9](https://github.com/DCjanus/wezterm/commit/281da8bc9) and kept the lighter coverage here. I still think the timing/interleaving case is worth protecting, but I think it would be better handled by a more general probe timing test harness.

If this direction makes sense, I'd like to open an issue to track it so there is a place for follow-up discussion and for collecting better ideas for a more general timing-oriented test approach. Ideally that would cover not just `dynamic_color`, but also other probe logic with similar reordering concerns, such as `screen_size`. I’d also like to revisit it myself when I have time, but I don’t want to over-promise on that here.

termwiz/src/caps/probed.rs
DCjanus: Thanks, this question made me realize I may have been thinking about the wrong kind of reordering problem here.

I originally followed the `screen_size()` pattern and treated `DA` as a loose boundary signal. But after thinking about it again, I think I mixed up two different cases: `screen_size()` is gathering multiple replies, while `dynamic_color()` is only waiting for a single `OSC 10/11` response.

So I'm no longer sure `DA` is the right fit here, and I'd really appreciate your view on whether this probe should simply stay as a single-response query.

bew: I don't have a definite answer for that either, I think we'd need Wez input on that one.

bew:
Removing the DA query/response would much simplify the probe' impl, but having the DA as a sentinel is potentially more robust across terminals and eventual multiplexers..

The main downside of the DA is the 100ms sleep.
👉 Did you yourself experience the ordering issue that this helps fix?
(Or did you copy this part from the other probe without really understanding why?)

Could you try this probing code with/without the delay over different terminals/multiplexers to see how they behave ? 🤔

termwiz/src/caps/probed.rs
```diff
@@ -239,3 +300,256 @@ impl<'a> ProbeCapabilities<'a> {
...
+                CSI::Device(Box::new(Device::RequestTerminalNameAndVersion)),
+                CSI::Device(Box::new(Device::RequestPrimaryDeviceAttributes)),
+            )
+        }
+    }
+
+    fn dynamic_color_probe_response(version: &str) -> String {
+        #[cfg(windows)]
+        {
+            let _ = version;
```
bew: Shouldn't there be a xt version response for the windows case as well?

termwiz/src/caps/probed.rs
DCjanus: I wrote up the details in a separate comment here: https://github.com/wezterm/wezterm/pull/7690#issuecomment-4240309695

DCjanus:
> Could you try this probing code with/without the delay over different terminals/multiplexers to see how they behave ? 🤔

I spent a little time trying to reproduce a stable ordering issue for this in tmux, but couldn't get a convincing repro, so I went back to the tmux code.

I think I copied the `DA` sentinel pattern from `screen_size()` too literally. The issue there seems to be specifically that some replies are generated locally by tmux ([`DA`](https://github.com/tmux/tmux/blob/057d169492a9edbf745283b6cab55e35927164e6/input.c#L1555-L1564)), while others are forwarded to the outer terminal ([`DCS tmux; ... ST`](https://github.com/tmux/tmux/blob/057d169492a9edbf745283b6cab55e35927164e6/input.c#L2549-L2610)), so reply order is not guaranteed once those paths are mixed.

`OSC 10/11` does not seem to fit that model: tmux handles those queries [locally](https://github.com/tmux/tmux/blob/057d169492a9edbf745283b6cab55e35927164e6/input.c#L3000-L3063) and appears to [cache tty fg/bg separately](https://github.com/tmux/tmux/blob/057d169492a9edbf745283b6cab55e35927164e6/tty.c#L388-L401), with the cache updated from [tty color replies](https://github.com/tmux/tmux/blob/057d169492a9edbf745283b6cab55e35927164e6/tty-keys.c#L1659-L1729). So I don't think the `DA` sentinel is justified for `dynamic_color()`, and I plan to remove it from this PR.

termwiz/src/caps/probed.rs
DCjanus: On Windows the condition short-circuited on `cfg!(windows)`, so `xt_version()` was not queried there. 

The current `dynamic_color()` path no longer uses `xt_version` here.

ref: https://github.com/wezterm/wezterm/pull/7690/changes/54a418e89f7e30595297309f593f295b83bedb68

DCjanus:
I reproduced a regression in a downstream project after switching to the fully DA-free dynamic color probe implementation. In my case,  still succeeds, but  no longer returns a response on Ghostty, which causes the downstream query to fail.

I want to keep this PR in draft for now while I investigate the probe state machine a bit more before requesting review again.

DCjanus:
> I want to keep this PR in draft for now while I investigate the probe state machine a bit more before requesting review again.

Fixed in [`90b1d62ca`](https://github.com/wezterm/wezterm/commit/90b1d62ca9489a07b9cfd7fa4d94f44372288a65).

After reproducing the downstream regression locally, I tracked it down to a bug in my probe code. This was my mistake: I had missed consuming the trailing byte from an ST-terminated OSC response.

That meant a single dynamic color query could still succeed, but two consecutive queries could break because the leftover terminator byte was then seen by the next probe. In the downstream case, that was exactly what happened: the first query still worked, while the second one no longer returned a usable response.

This commit fixes that trailing-byte handling and adds regression coverage for consecutive dynamic color queries.

DCjanus:
I noticed that this PR had moved to `To Test`, so I manually tested the current head. I found one case where the new probe blocks indefinitely: when the terminal leaves OSC 10/11 unanswered. In that state, the next keypress can also be consumed by the probe instead of reaching normal input handling.

I’m marking the PR as draft until I can add a bounded no-response path.

<details>
<summary>Ghostty reproduction and analysis</summary>

Environment: macOS with Ghostty 1.3.2-main.

1. With Ghostty's default `osc-color-report-format = 16-bit`, the example starts immediately and displays the probed foreground and background colors:

   ```console
   cargo run -p termwiz --example dynamic_color_wave
   ```

2. I set `osc-color-report-format = none`, which makes Ghostty leave OSC 4/10/11 queries unanswered, and ran the same command. The example remained blocked before its first frame and did not reach its fallback path on its own.

3. I pressed `q` once. The fallback frame appeared, but the example did not exit even though `q` is its quit key. Pressing `q` a second time exited normally. The first `q` was therefore consumed during probing rather than reaching the example's normal input handling.

`UnixTerminal` configures its TTY read handle as blocking, while `dynamic_color()` calls `self.read.read(&mut byte)` without a timeout or another finite completion boundary. A return value of `0` represents EOF, not a terminal that remains connected but silently ignores the query.

The current `test_dynamic_color_errors_when_terminal_does_not_respond` uses an empty `Cursor`, so it covers EOF rather than this open-but-silent behavior.

The probe needs to return an error within a finite time when no response arrives, so callers can fall back without hanging or consuming unrelated input.

</details>

Issue #7693, opened 2026-03-28
Failed to decode webp


Currypan1229:
### What Operating System(s) are you seeing this problem on?

Windows

### WezTerm version

wezterm 20260117-154428-05343b38

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

Wezterm cannot display webp files
ex: imgcat, background image

### To Reproduce

wezterm imgcat *.webp

### Configuration

No config

### Expected Behavior

display webp encoded image

### Logs

14:03:27.451  ERROR  wezterm_gui::glyphcache > Error decoding image: Unable to decode image data. Either it is corrupt, or the Image format is not fully supported by https://github.com/image-rs/image/blob/master/README.md#supported-image-formats
14:03:27.451  WARN   wezterm_gui::glyphcache > image decoder thread terminated

### Anything else?

The WebP implementation in the image crate might have some underlying issues.
Migrating to the dedicated WebPDecoder from the image-webp crate (by the same author) should fix this.

Currypan1229:
I performed tests with the attached image.
Since I could reproduce the behavior on multiple Windows 11 devices, it can be considered consistently reproducible.

In addition, it seems to be failing regardless of whether it's lossless or lossy.

[webp.zip](https://github.com/user-attachments/files/26811859/webp.zip)

Issue #6552, opened 2025-01-04
imgcat cannot display webp images


nik-rev:
### What Operating System(s) are you seeing this problem on?

Linux X11

### Which Wayland compositor or X11 Window manager(s) are you using?

i3

### WezTerm version

wezterm 8e9cf91 (main)

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

Wezterm imgcat cannot display webp files

![image](https://github.com/user-attachments/assets/0df72066-148b-4a89-bad3-f404ce6e4ff7)

### To Reproduce

`wezterm imgcat *.webp`

### Configuration

No config

### Expected Behavior

We should see an image

### Anything else?

Wezterm has the capability to display webp files it seems. For example, with yazi i can see the picture just fine

![image](https://github.com/user-attachments/assets/4ca4dbb2-643b-49f7-8026-39e475527aef)

So it seems that `imgcat` is broken here

wez:
Works fine here:

<img width="483" alt="Image" src="https://github.com/user-attachments/assets/49f7bbee-9acb-4cde-8256-acfee54c0c5e" />

I suspect that there is an issue with the webp file you are trying to display.

PR #7694, opened 2026-03-28
Fix error when decode webp images

Currypan1229:
#### What kind of changes does this PR include?

- [ ] Minor content fixes (broken links, typos, exceptions, etc.)
- [x] New or updated content
- [ ] Documentation updates
- [ ] Assets content
- [x] Code refactoring
- [x] Dependency updates
- [ ] Runtime updates
- [ ] Build Logic updates
- [ ] Other

---

#### Description

- Migrated WebP decoding logic from the image crate to the dedicated image-webp crate.

---

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

---

## Additional Notes

- The image-webp crate is provided by the same vendor as the image crate.
- The image crate is supported only lossless encoding

---

## Related Issue or PR

- #7693
- #6552 

---


Currypan1229:
I'm adding a clarification because it turned out that the image crate only supports lossless encoding.
Specifically, this means that lossy compression is unsupported.
However, it seems to be failing across the board, regardless of the encoding type.

PR #7697, opened 2026-03-29
windows: resolve gui socket path from runtime dir

fanis:
## Summary
On Windows, GUI socket discovery publishes only the socket file name into shared memory, but resolution was treating that value as a complete path.

This causes the CLI handoff path to resolve `gui-sock-*` as a bare relative path instead of `config::RUNTIME_DIR.join(...)`, which breaks attaching to an existing GUI instance.

## Changes
- join the published socket file name with `config::RUNTIME_DIR` when resolving on Windows
- add a Windows regression test covering publish/resolve round-trip

## Notes
The Windows publisher already stores only `path.file_name()` in shared memory, so resolving relative to `RUNTIME_DIR` matches the existing contract.

### DIFF
--- a/wezterm-client/src/discovery.rs
+++ b/wezterm-client/src/discovery.rs
@@ -230,7 +230,7 @@ mod windows {
                 let path = std::str::from_utf8(&source_slice[0..len])
                     .context("reading path from shared memory")?;
 
-                let path: PathBuf = path.into();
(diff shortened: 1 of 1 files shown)


fanis:
Reopening with corrected commit metadata: https://github.com/wezterm/wezterm/pull/7698

Issue #3906, opened 2023-06-28
Add a block list to regex matches to exclude matches from quick_select_patterns


eugenesvk:
**Is your feature request related to a problem? Please describe.**

Even with your addition of the fancy regex crate (thanks!) that allows negative lookbehinds the regex matches are "too earger", and in general regexes aren't great at NOT matches, so those lookbehinds can only help you out so much

For example, if I take this default path pattern (1) and replace it with (2) trying to exclude prompt starts I would only exclude the beginning of a match path, not the full prompt path since the latter matches would (and should as otherwise they'd fail at other text chunks)
```lua
disable_default_quick_select_patterns = true,
quick_select_patterns = {
  [[(?x: (?:          [.\w\-@~]+)?  #start       	#
         (?:         /[.\w\-@]+)+   #continuation	)]], -- 1 path default
  [[(?x: (?:(?<!^…|/) [.\w\-@~]+)?  #start       	#
         (?:(?<!^…|/)/[.\w\-@]+)+   #continuation	)]], -- 2 path adjusted
}
```

**Describe the solution you'd like**
I think a better approach would be to have a higher priority "block" regex match list so that I could match arbitrary path (even with spaces) in the prompt and EXCLUDE it, so that further `quick_select_patterns` that __match__ any text elements __within__ this block list are filtered out and not passed further, so they wouldn't be highlighted by the quick select
```lua
quick_select_patterns_BLOCK = { -- block path displayed at prompt between line starting with / or … ending with ❯
  [[(?x: (?: ^…|/)    	# match prompt start	#
         (?: [^\n❯]*❯)	# match prompt end  	#
         )]],
}
```

**Describe alternatives you've considered**
The other alternative I already mentioned in the other issue is excluding semantic zones, but this block filter could also help in excluding maches regexes in non-semantic areas (for example, you could have a couple of predictable `wordA/wordB` patterns in a prompt welcome message that you are certain would never need to "quick select"), this blocking approach would help with that

**Additional context**
https://github.com/wez/wezterm/issues/3247

wez:
FWIW, this feels too complex to me, both in terms of user understanding of the feature and the implementation itself

eugenesvk:
Can't say much about the implementation, but re. user understanding - aren't exclude/include block/allow lists of various kinds a very common pattern?

Destroy666x:
I'm not too sure why it's complicated in terms of both implementation and user perspective. You just find all matches first and the filter out the blacklist.

PR #7699, opened 2026-03-29
feat: allow matching windows paths in Quick Select Mode

Casper9474:
## Summary
Adds abilitiy to match windows paths in Quick Select Mode (currently it doesn't match them at all).

### Demo
paths

Before
before

After
after

### Disclaimer
I couldn't get it to capture backticks and quotes around the path correctly without freezing, so it matches the closing `'` or `` ` ``, if found at the end. It does that because `` C:\Users\xyz\test\` `` is a valid Windows path.

```powershell
C:\Users\xyz\test> ls


    Directory: C:\Users\xyz\test


Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
d-----         3/29/2026   8:56 PM                '
d-----         3/29/2026   8:56 PM                `


C:\Users\xyz\test> cd C:\Users\xyz\test\`
C:\Users\xyz\test\`>
```

### DIFF
--- a/wezterm-gui/src/overlay/quickselect.rs
+++ b/wezterm-gui/src/overlay/quickselect.rs
@@ -23,7 +23,7 @@ use wezterm_term::{
 };
 use window::WindowOps;
 
-const PATTERNS: [&str; 14] = [
+const PATTERNS: [&str; 15] = [
     // markdown_url
     r"\[[^]]*\]\(([^)]+)\)",
     // url
@@ -36,6 +36,8 @@ const PATTERNS: [&str; 14] = [
     r"sha256:([0-9a-f]{64})",
     // path
     r"(?:[.\w\-@~]+)?(?:/+[.\w\-@]+)+",
+    // windows path
+    r#"(?:[a-zA-Z]:|.)\\(?:(?:(?![<>:"/\\|?*]).)+(?:(?<![ .])\\)?)*"#,
     // color
     r"#[0-9a-fA-F]{6}",
     // uuid


bew (approved):
This looks pretty good thank you!

eugenesvk:
Re. `'quoted'` paths, ideally you'd need a way to exclude stuff https://github.com/wezterm/wezterm/issues/3906 or a way to exclude some groups (maybe exclude named groups named `×`?) to be able to match/discard those quoted paths 

But also you can duplicate your rule and have a higher-precedence rule matching path in `'quotes'` but matching those `''` as positive look-behind/lookahead to not highlight them
Here is a lua example from config
```lua
  quick_select_patterns = {
    -- [[(?x: (… [^❯]+ ❯)(?=1)  )]],
    -- [[(?x: (?!:…) [^❯]+ ❯  )]],

    [[(?x:(?<=')                     	# Exclude quoted paths
      (?:         [a-zA-Z]:|\.    )  	# C:\ Drive   or .\ short path
      (?:   [\\/]    )               	#
      (?:         [^\\/:*?"<>|\r\n]+ 	#¹
            [\\/]                 )* 	# Folder
      (?:         [^\\/:*?"<>|\r\n]*)	# File
          (?=')                      	)]],-- Win path: excluded quoted
    [[(?x:
      (?:         [a-zA-Z]:|\.    )  	# C:\ Drive   or .\ short path
      (?:   [\\/]    )               	#
      (?:         [^\\/:*?"<>|\r\n]+ 	#¹
            [\\/]                 )* 	# Folder
      (?:         [^\\/:*?"<>|\r\n]*)	# File
                                     	)]],-- Win path
    -- ¹ originally non-capturing, but in WT it's the opposite since it uses capturing groups itself to combine user rules
  },
```
wt path

@bew by the way, do you know of any tests for these patters? They are rather tricky to get right, with footguns like A-f, so maybe the text above could at least be a seed test
```
C:\Users\xyz> cd '.\AppData\Roaming\Arduino IDE\'
C:/Users/xyz> cd './AppData/Roaming/Arduino IDE/'
C:\Users\xyz\AppData\Roaming\Arduino IDE> cd 'C:\mingw64\'
C:/Users/xyz/AppData/Roaming/Arduino IDE> cd 'C:/mingw64/'
C:\mingw64> cd ~
C:/mingw64> cd ~
C:\Users\xyz> cd 'C:\$Windows.~WS\'
C:/Users/xyz> cd 'C:/$Windows.~WS/'
C:\$Windows.~WS\ cd .\Sources\Panther\
C:/$Windows.~WS/ cd ./Sources/Panther/
C:\$Windows.~WS\Sources\Panther>
C:/$Windows.~WS/Sources/Panther>
```


@bew by the way, are there any tests for these regexes as they're pretty tricky to get right and full of footguns?

wezterm-gui/src/overlay/quickselect.rs
```diff
@@ -36,6 +36,8 @@ const PATTERNS: [&str; 14] = [
     r"sha256:([0-9a-f]{64})",
     // path
     r"(?:[.\w\-@~]+)?(?:/+[.\w\-@]+)+",
+    // windows path
+    r#"(?:[a-zA-Z]:|.)\\(?:(?:(?![<>:"/\\|?*]).)+(?:(?<![ .])\\)?)*"#,
```
bew, line 40: For testing, I'd suggest extracting this regex into a dedicated const, used here, and add a bunch of unit tests in this file (anywhere, we can move them later if needed!)

(if you want to take a look at this @Casper9474 or @eugenesvk)

PR #7701, opened 2026-03-31
Prevent "screen scraping", disable DECRQCRA

jquast:
Problem
-------

March 17 2019 ``DECRQCRA`` was added in 6cbb3ba43, for integration and testing of esctest, 9702d1cf5. Days later, March 23, esctest was removed in c93f967bc, but ``DECRQCRA`` remained enabled for all WezTerm releases since.

On March 2023, @j4james mentioned WezTerm has this option enabled by default https://github.com/microsoft/terminal/issues/14974, "some people consider it a security risk" and on September 2023, @dgl writes in article, "[\E\[31m"?! ANSI Terminal security in 2023 and finding 10 CVEs](https://dgl.cx/2023/09/ansi-terminal-security#_)",

> "using DECRQCRA [..] potential attack here is reading what is displayed on the terminal before a user SSHes to a remote system."

Scraping can occur during/after ssh login, telnet, netcat, serial, or other bidirectional socket-to-terminal connection, the attack is best used early at the login banner stage to collect maximum amount of pre-connection screen data.

Example
------------

In this example, a victim uses sensitive files before a "[screen-scrape.py](https://github.com/jquast/blessed/blob/master/bin/screen-scrape.py)" is executed,

https://github.com/user-attachments/assets/e9f77e29-ed03-4a80-b00f-c9b6bba0a34e

It is more surprising that by switching to "alternate screen" it also possible to scrape the display of last-executed TUI's alternate buffer, such as the contents of the file last opened in an editor as depicted in this video.

Solution
--------

Disable https://vt100.net/docs/vt510-rm/DECRQCRA.html by default, may be re-enabled by configuration.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -24,6 +24,7 @@ As features stabilize some brief notes about them will accumulate here.
 #### Changed
 * DECRQCRA is now disabled by default to prevent silent screen scraping.
   Set `enable_checksum_rectangular_area = true` to re-enable it.
+  Thanks to @jquast! #7701
 * Wayland: currently being reimplemented, it maybe more unstable than usual.
   Please file GH issues for any problems you see.
   Many thanks to @tzx and @tmccombs! #4777 #5781
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -22,6 +22,8 @@ usually the best available version.
 As features stabilize some brief notes about them will accumulate here.
 
 #### Changed
+* DECRQCRA is now disabled by default to prevent silent screen scraping.
+  Set `enable_checksum_rectangular_area = true` to re-enable it.
 * Wayland: currently being reimplemented, it maybe more unstable than usual.
   Please file GH issues for any problems you see.
   Many thanks to @tzx and @tmccombs! #4777 #5781
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -260,6 +260,12 @@ pub struct Config {
     #[dynamic(default)]
     pub enable_title_reporting: bool,
 
+    /// Whether the terminal should respond to DECRQCRA checksum requests.
+    /// Disabled by default as it allows programs to read screen contents.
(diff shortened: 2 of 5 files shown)


j4james:
For the record, I personally don't consider it a serious security risk. It's an incredibly unlikely attack vector, and if I had anything on my screen that I was concerned about someone seeing I would've just cleared the screen. But I don't think it's unreasonable at least having an option to cater for the truly paranoid.

dgl:
iTerm2 has possibly the best implementation of paranoid but still flexible, it shows a relatively unobtrusive banner at the top of the terminal giving the user a choice. I must have missed that WezTerm enabled this when I surveyed all the terminals a few years ago... defaulting it off seems quite reasonable to me as a total drive-by review comment.

@j4james I agree this is not a serious issue. It is however possible for a compromised SSH remote to read a previous alternate screen (e.g. an attacker could drop a script in /etc/profile.d); it crosses a security boundary which the user may quite reasonably assume is there (even if reading the alternate screen may be noticeable via a flash of text). Clearing isn't quite enough (depending possibly on what $TERM is and what reset sequence is defined in the termcap).

bew:
Hey @wez, sorry for the direct ping but this seems worthy of your attention 🤔

wez:
Thanks all; yeah, I agree, this should be off by default.  The PR looks good at a glance, just running the build now.

docs/changelog.md
```diff
@@ -22,6 +22,8 @@ usually the best available version.
 As features stabilize some brief notes about them will accumulate here.
 
 #### Changed
+* DECRQCRA is now disabled by default to prevent silent screen scraping.
+  Set `enable_checksum_rectangular_area = true` to re-enable it.
```
wez, line 26:
```suggestion
  Set `enable_checksum_rectangular_area = true` to re-enable it.
  Thanks to @jquast! #7701
```

wez:
Thank you!

PR #7709, opened 2026-04-03
fix(pty): handle TerminateProcess return code correctly

chitao1234:
On Windows, TerminateProcess use nonzero return value for success.

### DIFF
--- a/pty/src/win/mod.rs
+++ b/pty/src/win/mod.rs
@@ -42,7 +42,7 @@ impl WinChild {
         let proc = self.proc.lock().unwrap().try_clone().unwrap();
         let res = unsafe { TerminateProcess(proc.as_raw_handle() as _, 1) };
         let err = IoError::last_os_error();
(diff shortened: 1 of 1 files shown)


bew:
could you link to official documentation for this? (helps the review)

chitao1234:
I found the documentation at [Microsoft learn](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-terminateprocess)

bew (approved):
I agree this is actually correct now!

---
I checked any code using this to find why we didn't see it before, it turns out that none of the `kill` callers in Wezterm check the return value.

.. How did you find this mistake? Are you making a specific application using the `pty` module?

chitao1234:
Yeah, I was using the `portable_pty` crate, and found this problem with kill.

wez:
Thank you!

PR #7721, opened 2026-04-06
window/windows: fix re-entrant borrow_mut panic in wm_size (IME/TSF)

m-morino:
## Problem

Two separate panics observed on Windows nightly builds.

### 1. `window/src/os/windows/window.rs:1570` — re-entrant borrow in `wm_size`

`WM_WINDOWPOSCHANGED` is dispatched **synchronously** while the `RefCell` is already borrowed:

```
TranslateMessage
  ImmTranslateMessage
  TsfOneCreate / CtfImeCreateInputContext   <- TSF/IME initialization
  NtUserSetWindowPos                        <- repositions IME candidate window
  WM_WINDOWPOSCHANGED dispatched synchronously
  wm_windowposchanged -> wm_size
  inner.borrow_mut()                        <- PANIC: already borrowed
```

**Fix:** Replace `borrow_mut()` with `try_borrow_mut()`. Re-entrant calls are silently skipped; the outer borrow processes the resize when it completes.

### 2. `wezterm-ssh/src/auth.rs:151` — unwrap on dropped channel sender

The libssh auth callback calls `.unwrap()` on `tx.try_send(...)`. If the UI receiver has been dropped (e.g. during window restore via the resurrect plugin or shutdown), the send fails and panics.

**Fix:** Check `is_err()` and return `libssh_rs::Error::fatal(...)` instead of panicking.

## Reproduction

1. Open WezTerm nightly on Windows with a Japanese IME (TSF-based) enabled
2. Move or resize the window → panic #1
3. Use the resurrect plugin to restore a workspace containing an SSH session → panic #2

### DIFF
--- a/wezterm-ssh/src/auth.rs
+++ b/wezterm-ssh/src/auth.rs
@@ -136,19 +136,26 @@ impl crate::sessioninner::SessionInner {
         // Set the callback for pubkey auth
         sess.set_auth_callback(move |prompt, echo, _verify, identity| {
             let (reply, answers) = bounded(1);
-            tx.try_send(SessionEvent::Authenticate(AuthenticationEvent {
-                username: "".to_string(),
-                instructions: "".to_string(),
-                prompts: vec![AuthenticationPrompt {
-                    prompt: match identity {
-                        Some(ident) => format!("{} ({}): ", prompt, ident),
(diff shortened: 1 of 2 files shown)


m-morino:
Closing: submitted without author's consent.

PR #7730, opened 2026-04-11
Fix IPv6 rule in quick select parsing

### LINKED ISSUE #7729: Regex issue: A-f ≠ A-Fa-f  (quick_select_patterns)

Windows

_No response_

### WezTerm version

 20260331-040028-577474d8

### Describe the bug

https://github.com/wezterm/wezterm/blob/main/wezterm-gui/src/overlay/quickselect.rs has IPv6 `r"[A-f0-9:]+:+[A-f0-9:]+[%\w\d]+",` patter, but this happens to match windows paths `C:\Users` because apparently A-f is not A-F a-f, but includes a few extra
\ ~ ] ^ _ `

### To Reproduce

add any `\ ~ ] ^ _` chars to printed text and run quick select mode

### Configuration

no config

### Expected Behavior

_No response_

### Logs

_No response_

### Anything else?

_No response_

eugenesvk:
`A-f` ≠ `A-Fa-f`, but includes a few extra symbols like `\~]^`

Closes https://github.com/wezterm/wezterm/issues/7729

### DIFF
--- a/wezterm-gui/src/overlay/quickselect.rs
+++ b/wezterm-gui/src/overlay/quickselect.rs
@@ -47,7 +47,7 @@ const PATTERNS: [&str; 14] = [
     // ip
     r"\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}",
     // ipv6
-    r"[A-f0-9:]+:+[A-f0-9:]+[%\w\d]+",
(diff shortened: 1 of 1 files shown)


bew (approved):
Good catch!

PR #7734, opened 2026-04-11
Upgrade wgpu 25.0.2 to 27.0.1 for FreeBSD support

rudrabhoj:
wgpu v25 introduced target-specific compilation in gfx-rs/wgpu#7076 but left FreeBSD out of the conditional build attributes for the Vulkan and GLES backends. As a result, no GPU adapter could be found on FreeBSD, failing with: "No wgpu backend feature that is implemented for the target platform was enabled."

Fixed upstream in gfx-rs/wgpu#8246, first shipped in wgpu 27.0.0.

API adjustments for wgpu 27:
- BufferViewMut lost its lifetime parameter
- DeviceDescriptor gained an experimental_features field
- RenderPassColorAttachment gained a depth_slice field

Tested on FreeBSD 15 with NVIDIA RTX 3070 Ti.

Refs:
- gfx-rs/wgpu#8175
- gfx-rs/wgpu#8246

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3327,13 +3327,13 @@ dependencies = [
 
 [[package]]
 name = "metal"
-version = "0.31.0"
+version = "0.32.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f569fb946490b5743ad69813cb19629130ce9374034abe31614a36402d18f99e"
+checksum = "00c15a6f673ff72ddcc22394663290f870fb224c1bfce55734a75c414150e605"
 dependencies = [
  "bitflags 2.10.0",
  "block",
- "core-graphics-types 0.1.3",
+ "core-graphics-types 0.2.0",
  "foreign-types 0.5.0",
  "log",
  "objc",
@@ -3521,25 +3521,26 @@ dependencies = [
 
 [[package]]
 name = "naga"
-version = "25.0.1"
+version = "27.0.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2b977c445f26e49757f9aca3631c3b8b836942cb278d69a92e7b80d3b24da632"
+checksum = "066cf25f0e8b11ee0df221219010f213ad429855f57c494f995590c861a9a7d8"
 dependencies = [
  "arrayvec",
  "bit-set",
  "bitflags 2.10.0",
+ "cfg-if",
  "cfg_aliases",
  "codespan-reporting",
  "half",
- "hashbrown 0.15.5",
+ "hashbrown 0.16.0",
  "hexf-parse",
  "indexmap 2.12.0",
+ "libm",
  "log",
  "num-traits",
  "once_cell",
  "rustc-hash 1.1.0",
  "spirv",
- "strum",
  "thiserror 2.0.17",
  "unicode-ident",
 ]
@@ -3581,9 +3582,9 @@ dependencies = [
 
 [[package]]
 name = "ndk-sys"
-version = "0.5.0+25.2.9519653"
+version = "0.6.0+11769913"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8c196769dd60fd4f363e11d948139556a344e79d451aeb2fa2fd040738ef7691"
+checksum = "ee6cda3051665f1fb8d9e08fc35c96d5a244fb1be711a03b71118828afc9a873"
 dependencies = [
  "jni-sys",
 ]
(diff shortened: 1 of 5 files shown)


joshuataylor:
Adding a +1 from the macOS side, and a case for taking this past 27 to 30 while it's open.

I'm happy to help contribute to get this in. I've been keeping notes, so apologies for the brain dump below, but it helps me keep track of everything and how it ties to together. Hopefully this is the full consolidation of my notes! (Also, learnt way more about Apple UI crap I never wanted to learn..).

Anyway, I've been running wezterm with `front_end = "WebGpu"` on macOS 15.7, and have been working on improving the font rendering to match OpenGL, and running with sRGB Colour Profiles. OpenGL works well, but I'd rather use WebGpu, as that seems the recommended approach.

There has been an issue I've had for a while where WebGpu has a "stall" when changing from another application back to wezterm, especially when running in a non-filled (e.g "floating") window. This is not a "Full Screen" or whatever confusing terminonoly mac has, but a maximised window).

When this happens, it can take a second or two to response/repaint.

~Cmd+`~ (how macOS handles changing windows between the same app) sometimes misses these windows, as I think they're also stalled.

FWIW, maximised windows seem fine, or not observed when doing this (99% of how I use wezterm).

I've dug long and hard into Apple documentation, especially around features which should be deprecated, which aren't (lolapple), as I've been working on other UI projects.

What is going on, from the wgpu side, for this specific issue:

wgpu-hal 25 sets [`allowsNextDrawableTimeout = NO`](https://github.com/gfx-rs/wgpu/blob/v25.0.2/wgpu-hal/src/metal/surface.rs#L174) on the `CAMetalLayer` and calls [`nextDrawable` unconditionally](https://github.com/gfx-rs/wgpu/blob/v25.0.2/wgpu-hal/src/metal/surface.rs#L195).

So once a window is fully occluded the WindowServer stops consuming its drawables, so the next paint parks the main thread in `nextDrawable` until the window is revealed.

Any output in a covered window seems to trigger it, and on activation, `windowDidBecomeKey:` and the other windows' paints all queue behind it. This is [wgpu#8309](https://github.com/gfx-rs/wgpu/issues/8309).

This is fixed in [v29.0.0](https://github.com/gfx-rs/wgpu/releases/tag/v29.0.0) by [wgpu#9141](https://github.com/gfx-rs/wgpu/pull/9141): `acquire_texture` checks `NSWindow.occlusionState` and returns [`SurfaceError::Occluded`](https://github.com/gfx-rs/wgpu/blob/v29.0.0/wgpu-hal/src/metal/surface.rs#L145-L151) instead of blocking.

27.0.1 does not have it, so this PR as written would not pick it up. (I have a local workaround that fiddles with `drawRect:` on `occlusionState` in `window/src/os/macos/window.rs`, happy to send it separately, but the wgpu-side fix is better, and means we don't need this.)

Other macOS related items between 25 and 30 from the [changelog](https://github.com/gfx-rs/wgpu/blob/trunk/CHANGELOG.md) that might be of interest:

- v28: `MTLDevice` treated as thread-safe ([wgpu#8168](https://github.com/gfx-rs/wgpu/pull/8168)).
- v29: Metal backend on autogenerated `objc2` bindings, "which should resolve a lot of leaks and unsoundness" ([wgpu#5641](https://github.com/gfx-rs/wgpu/pull/5641)). Given wezterm's history of macOS over-release crashes, welcome.
- v29: `MTLCommandQueue` mutex removed ([wgpu#9217](https://github.com/gfx-rs/wgpu/pull/9217)).
- v29.0.2: fence creation crash under the macOS sandbox ([wgpu#9415](https://github.com/gfx-rs/wgpu/pull/9415)).
- v30: command buffer completion waits on a condition variable instead of polling, and fixes a `Device::poll` hang when a command buffer errors ([wgpu#9328](https://github.com/gfx-rs/wgpu/pull/9328)). wezterm polls after every present, so this is on the hot path.
- v30: `SurfaceConfiguration::color_space` (sRGB / Display P3 / extended) and `Surface::display_hdr_info` ([wgpu#9658](https://github.com/gfx-rs/wgpu/pull/9658)). Removes the hand-tagged `CAMetalLayer` colour space and opens the door to P3/EDR later.

Changes we'll need to make going to 30 on top of what this PR already handles, from what I can see in `wezterm-gui`:

- `Surface::get_current_texture` returns the `CurrentSurfaceTexture` enum and `SurfaceError` is gone (`render/draw.rs`, and the `Lost | Outdated | Timeout` recovery arm in `termwindow/mod.rs`). `Occluded` should just skip the frame and leave the window invalidated.
- `InstanceDescriptor::default()` is replaced by the `new_with_display_handle` / `new_without_display_handle` constructors (`termwindow/webgpu.rs`, `scripting/mod.rs` for `enumerate_gpus`).
- `PipelineLayoutDescriptor::bind_group_layouts` takes `&[Option<&BindGroupLayout>]` (v29), and `VertexState::buffers` takes `&[Option<&VertexBufferLayout>]` (v30). One site each in `termwindow/webgpu.rs`.
- `Instance::enumerate_adapters` is async from v28 (the `webgpu_preferred_adapter` loop and the Lua `enumerate_gpus`).
- v30 adds the `color_space` field to `SurfaceConfiguration` (`SurfaceColorSpace::Auto` keeps today's behaviour). The v30 shader change (integer I/O no longer defaults to `@interpolate(flat)`) does not affect `shader.wgsl`, every varying there is a float.
- MSRV is 1.87 for v29+, which is fine with wezterm.

IMHO go straight to 30 - we get the device polling and colour space changes (hell yeah!).

29 still works as a smaller step if that's preferred, since it carries the occlusion fix and the objc2 backend, but why not 30(?).

I live in my terminal, and have been driving other changes locally on a Macbook Pro 13" (M2 Pro, 120 Hz ProMotion, yes the same year they released 2 laptops, I'm not still salty, I swear.).

I'm happy to help test and do this PR as well to go to 30, as well as any other commits/PRs as needed.

---

"How much effort can fixing a few quirks on macOS be" - naive Josh.

PR #7752, opened 2026-04-21
Improve UX for quick selection: hide non-matching labels

Felixoid:
Implement the logic similar to `Vimium` plugin.

When the key is pressed, all non-matching labels are hidden.

- Screenshot_20260421_160548
- Screenshot_20260421_160618
- Screenshot_20260421_160631
- Screenshot_20260421_160652


Felixoid:
Dear @bew, it would be nice to get it under the merge umbrella too.

I miss this feature so desperately; it significantly improves word selection by hiding unrelated "unselected" parts.

bew:
> it would be nice to get it under the merge umbrella too.

it is already 😉

bew:
Thanks a lot for the PR, I definitely wanted something like this! 🚀 :P 
I've updated it with latest main, and pushed a commit to simplify the tests a little & remove redundant tests, and to re-use the existing `recompute_results` instead of adding a new `mark_all_matches_dirty` function 😉

Will merge once CI checks pass!

Felixoid:
Thanks for merging it, the UX improved significantly!

PR #7761, opened 2026-04-26
fix copy mode selection flicker in terminal multiplexers

### LINKED ISSUE #7740: CopyMode selection disappears immediately when running inside terminal multiplexers (zellij, tmux)
## Problem

When using `ActivateCopyMode` inside a terminal multiplexer (zellij, tmux), visual selection disappears almost immediately after being created. Each time the selection is adjusted (e.g., moving the cursor with h/j/k/l), the highlight briefly flashes and then vanishes. Copying only succeeds if Ctrl+C is pressed within a very short window after selecting.

This does **not** occur when WezTerm is used without a multiplexer, or when using mouse-based selection.

## Root Cause

`CopyRenderable::adjust_selection()` in `wezterm-gui/src/overlay/copy.rs` (line 511-524) does not update `selection.seqno` after modifying the selection range:

```rust
fn adjust_selection(&self, start: SelectionCoordinate, range: SelectionRange) {
    let pane_id = self.delegate.pane_id();
    let window = self.window.clone();
    let mode = self.selection_mode;
    self.window
        .notify(TermWindowNotif::Apply(Box::new(move |term_window| {
            let mut selection = term_window.selection(pane_id);
            selection.origin = Some(start);
            selection.range = Some(range);
            selection.rectangular = mode == SelectionMode::Block;
            window.invalidate();
            // BUG: selection.seqno is never updated here
        })));
    self.adjust_viewport_for_cursor_position();
}
```

Every other selection update path in the codebase correctly sets `selection.seqno = pane.get_current_seqno()`:

- `termwindow/selection.rs:120` — `extend_selection_at_mouse_cursor`
- `termwindow/selection.rs:280` — `select_text_at_mouse_cursor`
- `termwindow/selection.rs:115` — `clear_selection`
- `overlay/quickselect.rs:955` — QuickSelectOverlay (with explicit comment: *"Ensure that selection doesn't get invalidated when the overlay is closed"*)

Because the seqno remains stale (default 0), `check_for_dirty_lines_and_invalidate_selection()` in `termwindow/mod.rs:1678` calls `pane.get_changed_since(visible_range, seqno)` with the stale value. In a multiplexer environment where the underlying pane continuously produces output (status bar updates, cursor repositioning, etc.), this causes every line to be reported as dirty on every render cycle, leading to rendering instability that makes the selection highlight disappear.

`CopyRenderable::clear_selection()` (line 422-430) has the same missing seqno update, though the impact is lower since it sets the range to `None`.

## Steps to Reproduce

1. Open WezTerm
2. Start zellij (or tmux) inside WezTerm
3. Trigger `wezterm.action.ActivateCopyMode`
4. Press `v` to start visual selection
5. Move cursor with h/j/k/l
6. Observe: selection highlight appears briefly then disappears

## Expected Behavior

Selection highlight should persist and update as the cursor moves, just as it does outside of a multiplexer environment.

## Suggested Fix

Add `selection.seqno` update in `adjust_selection()`:

```rust
self.window
    .notify(TermWindowNotif::Apply(Box::new(move |term_window| {
        let mut selection = term_window.selection(pane_id);
        selection.origin = Some(start);
        selection.range = Some(range);
        selection.rectangular = mode == SelectionMode::Block;
        selection.seqno = pane.get_current_seqno();  // ADD THIS
        window.invalidate();
    })));
```

## Related Issues

- zellij-org/zellij#4252 — Same symptom reported from the zellij side
- #6163 — Same symptom in tmux ("Copy mode deselects when copying in tmux")
- #5952 — Copy/Search mode selection state management issues

## Environment

- OS: macOS (Darwin 25.3.0)
- WezTerm: built from source (main branch, commit 577474d89)

ysoftman:
Closes #7740

## Changes

- Update `selection.seqno = pane.get_current_seqno()` inside `CopyRenderable::adjust_selection()` and `CopyRenderable::clear_selection()` so the selection state stays in sync with the pane's current sequence number.
- Without this update, the seqno remained at its default (0). In a multiplexer (zellij/tmux), the underlying pane keeps producing output (status bar refresh, cursor repositioning, etc.), so `check_for_dirty_lines_and_invalidate_selection()` reports every visible line as dirty on every render cycle and the highlight flashes and disappears on each cursor move.
- Mirrors the pattern already used by `extend_selection_at_mouse_cursor` (`termwindow/selection.rs:120`) and `QuickSelectOverlay::select_and_copy_match_number` (`overlay/quickselect.rs:955`).

## Repro / Verification

1. Open WezTerm, start `zellij` (or `tmux`).
2. Trigger `ActivateCopyMode`, press `v`, then move with `h/j/k/l`.
3. Before the fix: highlight flashes and vanishes.
4. After the fix: highlight persists and extends correctly.

Verified locally on macOS (Darwin 25.3.0) with a release build of this branch.

### DIFF
--- a/wezterm-gui/src/overlay/copy.rs
+++ b/wezterm-gui/src/overlay/copy.rs
@@ -423,9 +423,13 @@ impl CopyRenderable {
         let pane_id = self.delegate.pane_id();
         self.window
             .notify(TermWindowNotif::Apply(Box::new(move |term_window| {
+                let mux = mux::Mux::get();
                 let mut selection = term_window.selection(pane_id);
                 selection.origin.take();
                 selection.range.take();
+                if let Some(pane) = mux.get_pane(pane_id) {
+                    selection.seqno = pane.get_current_seqno();
+                }
             })));
     }
 
@@ -514,10 +518,14 @@ impl CopyRenderable {
         let mode = self.selection_mode;
         self.window
             .notify(TermWindowNotif::Apply(Box::new(move |term_window| {
+                let mux = mux::Mux::get();
                 let mut selection = term_window.selection(pane_id);
                 selection.origin = Some(start);
                 selection.range = Some(range);
                 selection.rectangular = mode == SelectionMode::Block;
+                if let Some(pane) = mux.get_pane(pane_id) {
+                    selection.seqno = pane.get_current_seqno();
+                }
                 window.invalidate();
             })));
         self.adjust_viewport_for_cursor_position();


wezterm-gui/src/overlay/copy.rs
```diff
@@ -423,9 +423,13 @@ impl CopyRenderable {
         let pane_id = self.delegate.pane_id();
         self.window
             .notify(TermWindowNotif::Apply(Box::new(move |term_window| {
+                let mux = mux::Mux::get();
                 let mut selection = term_window.selection(pane_id);
                 selection.origin.take();
                 selection.range.take();
+                if let Some(pane) = mux.get_pane(pane_id) {
+                    selection.seqno = pane.get_current_seqno();
```
bew, line 431: Hello! o/

What is the usefulness of setting selection seqno when the selection is getting cleared anyway?

ysoftman: Hello @bew, thank you for reviewing this, and I apologize for the delayed response.

You're right. Setting `seqno` in `clear_selection()` is meaningless. I added it there simply for symmetry with `adjust_selection()` without actually tracing how `selection.seqno` is used.
  That was my mistake.

Upon closer inspection, selection.seqno is only read in check_for_dirty_lines_and_invalidate_selection(), and its value is used only within a block protected by the following condition:
  
  if pane.downcast_ref::<CopyOverlay>().is_none()
      && pane.downcast_ref::<QuickSelectOverlay>().is_none()
  { ... }

Since the rendered panel in copy mode is a `CopyOverlay`, that block is completely skipped, and the selection is never cleared via this path. In other words, since `seqno` has no observable effect in copy mode, neither the change to `clear_selection()` nor the change to `adjust_selection()` actually resolves the flickering issue described in #7740.

Furthermore, since the original flickering issue can no longer be reproduced on the current main branch, it appears this was not the actual cause (it may have been fixed by other changes in the meantime).

Considering these points, I will close this PR as it has no effect. I apologize for the inconvenience and thank you for taking the time to review this.

PR #7762, opened 2026-04-27
fix: update Se terminfo entry to reset cursor to configured default

kylesower:
## Problem

The `Se` terminfo entry defines how to reset the cursor style. Right now it's a hardcoded `\E[2 q`, which always sets a steady block. I think it's more sensible to set it to the user configured cursor style with `\E[0 q` (which WezTerm supports). Alacritty, kitty (https://github.com/kovidgoyal/kitty/pull/9929), and iTerm2 (https://github.com/gnachman/iTerm2/pull/651) all do this, and Ghostty should be doing it soon (https://github.com/ghostty-org/ghostty/issues/12482).

## Solution

Update `Se` terminfo entry to `\E[0 q` so applications that rely on terminfo can reliably reset the cursor to the user configured default.

Related issue in neovim: https://github.com/neovim/neovim/issues/38987.

**Note**: There were two compiled terminfo files, `termwiz/data/wezterm` and `termwiz/data/w/wezterm`. They were slightly different in that `termwiz/data/w/wezterm` had `Su` capability and `%;` terminating `Sync`. I think the files merely got out of sync (lol), but I don't understand why there are two different ones.


bew (approved):
Looks good!

👉 I'm interested in starting a community fork, see https://github.com/wezterm/wezterm/issues/6341#issuecomment-4483047032, so this might gets merged there. (you'll be notified if so)

PR #7766, opened 2026-04-28
docs: example for adjusting window opacity incrementally

### LINKED ISSUE #5792: Add richness to examples of overriding config with key bindings.
There has been an  [Issue on how to bind opacity to keyboard shortcut](https://github.com/wez/wezterm/issues/329#issue-742064923)

And the answer was  a reference to [Docs showing examples of how to override config](https://wezfurlong.org/wezterm/config/lua/window/set_config_overrides.html)

However, both examples given perform toggling. Can we please have the second example of opacity work on adjusting the value (instead of simply on-off), this will shed new light not provided by example 1.

Thanks.

PS: tried LLM to do it, but config is extremely finiky and 1 character can break it all, so that didn't work.

tautik:
adds a third example to `set_config_overrides` showing how to adjust `window_background_opacity` by a delta with two key bindings, instead of just toggling.

closes #5792

### DIFF
--- a/docs/config/lua/window/set_config_overrides.md
+++ b/docs/config/lua/window/set_config_overrides.md
@@ -73,3 +73,49 @@ return {
 }
 ```
 
+In this example, two key assignments (`CTRL-SHIFT-PageUp` and `CTRL-SHIFT-PageDown`)
+adjust the window opacity incrementally instead of toggling. Each press reads
+the current opacity (falling back to the configured value if no override is
+set yet), nudges it by 0.1, and clamps the result to a sensible range so the
+window stays visible:
+
+```lua
+local wezterm = require 'wezterm'
+
+local function adjust_opacity(window, delta)
+  local overrides = window:get_config_overrides() or {}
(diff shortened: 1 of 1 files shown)


docs/config/lua/window/set_config_overrides.md
```diff
@@ -73,3 +73,49 @@ return {
...
+wezterm.on('decrease-opacity', function(window, _)
+  adjust_opacity(window, -0.1)
+end)
+
+return {
+  keys = {
+    {
+      key = 'PageUp',
+      mods = 'CTRL|SHIFT',
+      action = wezterm.action.EmitEvent 'increase-opacity',
```
bew: note that you can skip the need for explicit event definition/emission by using `wezterm.action_callback`!
For example:
```suggestion
      action = wezterm.action_callback(function(window, _)
          adjust_opacity(window, -0.1)
      end,
```

and similar change for the decrease action below

bew (approved):
Thank you, it looks good 👍

PR #7768, opened 2026-04-29
docs: fix linuxbrew install command

tautik:
on linux, `brew install wezterm` resolves to the macOS-only cask in homebrew-core and fails with "macOS is required for this software". use the fully-qualified tap name `wezterm/wezterm-linuxbrew/wezterm` instead, and add a short note explaining why.

closes #7755

PR #7772, opened 2026-05-01
docs: fix Linuxbrew installation path

puzzleproper:
On Linux, the generic 'wezterm' formula name in Homebrew/Linuxbrew attempts to install the macOS cask, leading to a failure. This PR updates the documentation to use the fully qualified tap name 'wezterm/wezterm-linuxbrew/wezterm' for Linux installation commands.


bew:
I've merged a similar PR with this change https://github.com/wezterm/wezterm/pull/7768 :)

PR #7778, opened 2026-05-06
feat(ssh): support Match exec, GlobalKnownHostsFile, and two-phase parsing

ismell:
This commit enhances the embedded SSH client's configuration parser and host verification logic to properly handle complex `ssh_config` setups.

Key improvements:
- **Match exec support**: `Match exec` directives are now properly parsed, tokens like `%h`, `%r`, `%u`, and `%p` are expanded, and the command is evaluated via `sh -c` (or `cmd.exe /c` on Windows) to determine if the block matches. Previously, this was unimplemented and unconditionally evaluated to `true`, inadvertently injecting incorrect configuration values.
- **Two-phase config parsing**: Ensure `Context::Canonical` and `Context::Final` matching blocks (e.g., `Match canonical all`) are properly evaluated during the `for_host` resolution phase. Previously, WezTerm silently ignored these blocks.
- **GlobalKnownHostsFile support**: The `GlobalKnownHostsFile` option is now parsed, supports token and environment expansion, and is mapped to the underlying `libssh-rs` session. Host key verification logic falls back to checking the global hosts file if `UserKnownHostsFile` does not yield a match, allowing WezTerm to verify SSH Certificate Authorities deployed system-wide.
- **Unit tests**: Added tests to verify `Match exec` evaluation and ensure `GlobalKnownHostsFile` is properly resolved out of `Match canonical all` blocks with token expansion.

### DIFF
--- a/wezterm-ssh/src/config.rs
+++ b/wezterm-ssh/src/config.rs
@@ -98,7 +98,14 @@ struct MatchGroup {
 }
 
 impl MatchGroup {
-    fn is_match(&self, hostname: &str, user: &str, local_user: &str, context: Context) -> bool {
+    fn is_match(
+        &self,
+        hostname: &str,
+        user: &str,
+        local_user: &str,
+        target: &ConfigMap,
+        context: Context,
+    ) -> bool {
         if self.context != context {
             return false;
         }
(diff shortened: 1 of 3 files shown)


wezterm-ssh/src/host.rs
```diff
@@ -70,6 +70,14 @@ impl crate::sessioninner::SessionInner {
                         break;
                     }
                 }
+                if file.is_none() {
+                    if let Some(kh) = self.config.get("globalknownhostsfile") {
+                        for candidate in kh.split_whitespace() {
+                            file.replace(candidate.into());
+                            break;
```
bew, line 77: NOTE: Would prefer to see `.first()` instead of a look/break, to make the intent more explicit

wezterm-ssh/src/sessioninner.rs
```diff
@@ -195,6 +195,14 @@ impl SessionInner {
                 break;
             }
         }
+        if let Some(gkh) = self.config.get("globalknownhostsfile") {
+            for file in gkh.split_whitespace() {
+                sess.set_option(libssh_rs::SshOption::GlobalKnownHosts(Some(
+                    file.to_string(),
+                )))?;
+                break;
```
bew, line 203: NOTE: Would prefer to see `.first()` instead of a look/break, to make the intent more explicit

wezterm-ssh/src/config.rs
```diff
@@ -1633,4 +1703,41 @@ Config {
...
     }
+
+    #[test]
+    fn test_match_exec() {
+        let mut config = Config::new();
+
+        // Set up a config with an Exec that passes and an Exec that fails.
+        config.add_config_string(
+            r#"
+            Match exec "exit 0"
```
bew, line 1714: I think it'd be more interesting test to have 3 Matches, the first and last one fails, the middle wins 🤔

wezterm-ssh/src/config.rs
```diff
@@ -98,7 +98,14 @@ struct MatchGroup {
...
 
 impl MatchGroup {
-    fn is_match(&self, hostname: &str, user: &str, local_user: &str, context: Context) -> bool {
+    fn is_match(
+        &self,
+        hostname: &str,
+        user: &str,
+        local_user: &str,
+        target: &ConfigMap,
+        context: Context,
```
bew, line 107: please move context _before_ target, like `apply_matches` 🤔

bew (commented):
Looks pretty good! Here are a few nit comments.

👉 I'm interested in starting a community fork, see https://github.com/wezterm/wezterm/issues/6341#issuecomment-4483047032, so this might gets merged there once ready.

PR #7787, opened 2026-05-13
Fix kitty spec for ESC key

Felixoid:
As it was found in https://github.com/earendil-works/pi/issues/4323, the `Esc` pressing event is sent wrongly from wezterm when `config.enable_kitty_keyboard = true`.

This patch addresses it by excluding the `\x1b` from `Char(c)` explicitly.

### DIFF
--- a/wezterm-input-types/src/lib.rs
+++ b/wezterm-input-types/src/lib.rs
@@ -1723,6 +1723,12 @@ impl KeyEvent {
             match &self.key {
                 Char('\x08') => return '\x7f'.to_string(),
                 Char('\x7f') => return '\x08'.to_string(),
+                // With DISAMBIGUATE_ESCAPE_CODES, ESC must not be sent as a
+                // raw \x1b byte — the entire point of the flag is to eliminate
+                // that ambiguity.  Let it fall through to the CSI-u path below
+                // which will produce \x1b[27;1u as required by the spec.
+                // https://sw.kovidgoyal.net/kitty/keyboard-protocol/#disambiguate
+                Char('\x1b') if flags.contains(KittyKeyboardFlags::DISAMBIGUATE_ESCAPE_CODES) => {}
                 Char(c) => return c.to_string(),
                 _ => {}
             }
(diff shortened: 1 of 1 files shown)


Rahmyeon:
Thanks for the fix, (have you tested on windows)

Felixoid:
No, I don't have Win hosts to check it.

Rahmyeon:
Tested on windows so far no issues

aumuell:
This seems to do the same as #6872

Felixoid:
Right, thanks a lot for pointing it 

I'll close this PR if favor if the existing

wezterm-input-types/src/lib.rs
```diff
@@ -3208,4 +3214,62 @@ mod test {
...
+                modifiers: Modifiers::NONE,
+                leds: KeyboardLedStatus::empty(),
+                repeat_count: 1,
+                key_is_down: true,
+                raw: None,
+                #[cfg(windows)]
+                win32_uni_char: None,
+            }
+            .encode_kitty(flags),
+            "\x1b[27;1u".to_string()
```
the-mikedavis, line 3236: (Also discussed in #6872.) It's equivalent in the spec but in my testing, Kitty itself sends `\x1b[27;u`.

the-mikedavis (approved):
I haven't tested this branch on Windows but the test cases look good to me, it should solve the same issues I saw using the kitty keyboard protocol on Windows (i.e. needing to hit Esc twice for the code to be sent).

wezterm-input-types/src/lib.rs
Felixoid: `CSI number; modifier u` here the modifier `1` can be explicit or omitted.

From the spec: `If the modifier field is not present in the escape code, its default value is 1 which means no modifiers`.

So `\x1b[27;u` and `\x1b[27;1u` are equal from the interpretation perspective.

Felixoid: Even `\u{1b}[27u` from your PR is a correct sequence. It's just a bit inconsistent with other codes in wezterm.

bew:
Thank you @Felixoid & @the-mikedavis for the work & tests on this fix! 🚀

PR #7789, opened 2026-05-13
filedescriptor: Add `into_stdio` conversion

xtqqczze:
- Add `FileDescriptor::into_stdio` to create a `std::process::Stdio` by transferring ownership of the underlying handle without duplicating it.
- Add `From<FileDescriptor>` for `std::process::Stdio` for convenience.

### DIFF
--- a/filedescriptor/src/windows.rs
+++ b/filedescriptor/src/windows.rs
@@ -234,6 +234,12 @@ impl FileDescriptor {
     }
 
     #[inline]
+    pub(crate) fn into_stdio_impl(self) -> std::process::Stdio {
+        let handle = self.into_raw_handle();
+        unsafe { std::process::Stdio::from_raw_handle(handle) }
+    }
+
+    #[inline]
     pub(crate) fn as_file_impl(&self) -> Result<std::fs::File> {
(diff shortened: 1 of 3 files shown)


xtqqczze:
@bew PTAL

filedescriptor/src/windows.rs
```diff
@@ -233,6 +233,12 @@ impl FileDescriptor {
         Ok(stdio)
     }
 
+    #[inline]
+    pub(crate) fn into_stdio_impl(&self) -> std::process::Stdio {
```
bew: hmm why does this take a reference to self instead of passing self as an owned value?
(like the other impls)

bew (commented):
Thanks for the ping, however I'm not confident enough to judge the potential ramifications of this and if it makes sense as is for the crate 🤔

I'd prefer if @wez did the review instead.

filedescriptor/src/windows.rs
xtqqczze: right, good catch!

wez (approved):
This looks good, thanks!

bew:
Thanks for the review, all checks are ✅ 
Let's merge!

Issue #7265, opened 2025-09-29
segfault w/macos Sequoia


bdols:
### What Operating System(s) are you seeing this problem on?

macOS

### WezTerm version

20240203-110809-5046fc22

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

No, and I'll explain why below

### Describe the bug

the release has been stable and have not felt the need to try nightly

not sure what happened, i had three windows open with multiple tabs, i was doing ^D to exit tabs and fully close one window, and then wezterm crashed. attaching crash report excerpt below

### Configuration

local wezterm = require("wezterm")
local act = wezterm.action

local config = wezterm.config_builder()

config.color_scheme = "Atelier Dune Light (base16)"
config.color_scheme = "Atelier Dune (base16)"
config.color_schemes = {
        ["Light Scheme"] = {
                background = "white",
                foreground = "black",
        },
        ["Red Scheme"] = {
                background = "red",
        },
        ["Blue Scheme"] = {
                background = "blue",
        },
}
config.font = wezterm.font("FiraCode Nerd Font Mono")

config.mouse_bindings = {
        {
                event = { Up = { streak = 1, button = "Left" } },
                mods = "NONE",
                action = act.CompleteSelection("ClipboardAndPrimarySelection"),
        },
        {
                event = { Up = { streak = 1, button = "Left" } },
                mods = "CTRL",
                action = act.OpenLinkAtMouseCursor,
        },
        -- Disable the 'Down' event of CTRL-Click to avoid weird program behaviors
        {
                event = { Down = { streak = 1, button = "Left" } },
                mods = "CTRL",
                action = act.Nop,
        },
}
config.ssh_domains = {
        {
                -- This name identifies the domain
                name = "linuxnuc",
                -- The hostname or address to connect to. Will be used to match settings
                -- from your ssh config file
                remote_address = "linuxnuc.local",
                -- The username to use on the remote host
                username = "bdols",
        },
}
config.window_background_opacity = 0.97
config.macos_window_background_blur = 20
-- config.unicode_version = 14

config.keys = {
  {key="Enter", mods="SHIFT", action=wezterm.action{SendString="\x1b\r"}},
}

return config

### Logs

Date/Time:             2025-09-29 11:07:54.1961 -0500
OS Version:            macOS 15.6.1 (24G90)
Report Version:        12
Anonymous UUID:        FC459E51-C5CA-CC6A-9CCB-356A3766273C

[...]

Issue #1396, opened 2021-12-13
portable-pty example not working properly on Windows


CyriacBr:
(I omittedthe bug report form as this issue concerns `portable-pty`)

I noticed that `portable-pty` behaves very differently on Windows.
When running the whoami example, no outpud is read, and the child exits with an error code:

```rs
use portable_pty::{CommandBuilder, NativePtySystem, PtySize, PtySystem};

fn main() {
    let pty_system = NativePtySystem::default();

    let pair = pty_system
        .openpty(PtySize {
            rows: 24,
            cols: 80,
            pixel_width: 0,
            pixel_height: 0,
        })
        .unwrap();

    let cmd = CommandBuilder::new("whoami");
    let mut child = pair.slave.spawn_command(cmd).unwrap();
    drop(pair.slave);

    let mut reader = pair.master.try_clone_reader().unwrap();
    drop(pair.master);

    // Consume the output from the child
    let mut s = String::new();
    reader.read_to_string(&mut s).unwrap();

    print!("output: ");
    for c in s.escape_debug() {
        print!("{}", c);
    }

    println!("child status: {:?}", child.wait().unwrap());
}
```

I found that it only works when waiting for the child before reading:
```rs

    let cmd = CommandBuilder::new("whoami");
    let mut child = pair.slave.spawn_command(cmd).unwrap();
    child.wait().unwrap();
    drop(pair.slave);
```
But this is problematic because this makes it directly impossible to handle very long-running commands (e.g a python or node script that infinitely runs something on interval). It's easier to handle such a case through the reader directly with either a timeout or a limit of characters/lines to read.

So first of all it seems either there's an issue with `portable_pty` or with the windows computer I have.
Hopefully you have an idea of what's happening, thanks :)

bbigras:
I think I have the same problem on Windows server 2019.

wez:
I've made a couple of changes in master and updated the example. I haven't published this version to crates.io yet.

https://github.com/wez/wezterm/blob/main/pty/examples/whoami.rs

MBanucu:
See [https://github.com/MBanucu/portable-pty-fixed](https://github.com/MBanucu/portable-pty-fixed).

PR #7799, opened 2026-05-18
Fix SIGABRT on window close caused by OpenGL teardown ordering (MacOS)

joshuataylor:
When a window closes on MacOS (15.x), the GPU drawable/IOSurface gets torn down before NSView dealloc fires. RenderState is still alive at that point, so [glium](https://github.com/glium/glium)'s Drop for RawProgram calls make_current, which walks into NSOpenGLContext update -> CGLUpdateContext -> IOSurface lookup on a dead Mach port. Process aborts. wompwomp.

Thanks to the Wezterm logging and for RustRover for helping debug this, made it muuuuuuuch simpler! (Not paid).

I BELIEVE the fix is dropping `render_state` in the `WindowEvent::Destroyed` handler (which runs from windowWillClose, before dealloc) while the surface is still valid. The field is already Option<RenderState> so .take() does the job.

Crash looks like this:
```
_kernelrpc_mach_port_mod_refs_trap
IOGPUGLDrawableLookupIOSurfaceFromIndex
GLRDrawable::attachDrawable
CGLUpdateContext / glUseProgramObjectARB_Exec
<RawProgram as Drop>::drop
drop_in_place<Option<RenderState>>
WindowView::dealloc
```

Fixes(?): https://github.com/wezterm/wezterm/issues/7265

I've tested this for a week now and haven't crashed once since!!

### DIFF
--- a/wezterm-gui/src/termwindow/mod.rs
+++ b/wezterm-gui/src/termwindow/mod.rs
@@ -909,6 +909,17 @@ impl TermWindow {
                 // the window is gone and we'll linger forever.
                 // <https://github.com/wezterm/wezterm/issues/3522>
                 self.clear_all_overlays();
+                // Drop OpenGL/render resources while the window surface is still
+                // alive, before NSView dealloc invalidates the GPU drawable.
+                // If deferred until TermWindow::drop, glium's Drop impls
+                // (RawProgram, Context, VertexBuffer, etc.) call make_current
+                // which triggers NSOpenGLContext update on a stale IOSurface,
+                // causing SIGABRT.
+                // Order matters: render_state first (its Drop deletes programs,
+                // buffers, textures via the context), then gl (drops the
+                // context itself, which does FBO/VAO/sampler cleanup).
+                self.render_state.take();
+                self.gl.take();
                 Ok(false)
             }
             WindowEvent::CloseRequested => {
--- a/wezterm-gui/src/termwindow/mod.rs
+++ b/wezterm-gui/src/termwindow/mod.rs
@@ -914,6 +914,17 @@ impl TermWindow {
                 // the window is gone and we'll linger forever.
                 // <https://github.com/wezterm/wezterm/issues/3522>
                 self.clear_all_overlays();
+                // Drop OpenGL/render resources while the window surface is still
+                // alive, before NSView dealloc invalidates the GPU drawable.
+                // If deferred until TermWindow::drop, glium's Drop impls
+                // (RawProgram, Context, VertexBuffer, etc.) call make_current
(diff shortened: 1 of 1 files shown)


bew (approved):
This looks good, thank you!

I don't know when this will be merged though, as development has ~stopped (Wez isn't responding.. and he's the only one who can merge..)

joshuataylor:
All good, I run a local copy of WezTerm anyway, so just wanted to help others who might be crashing

joshuataylor:
Hey @bew, saw you did a commit - feel free to let me know if you need anything else for this.

Also would love https://github.com/wezterm/wezterm/pull/7734 -- I got a little nerdsniped and did this today. I did find a couple of regressions I'm testing now.

I also have a series of patches that I've been daily-driving for the past few months.

Quick list of some of the work I've done:

1. I believe I have solved WebGpu font rendering on macOS(!!). It should now render the same as OpenGL.
2. I've fixed all the main crashes Wezterm can have on macOS. These were fun to figure out, thanks to RustRover debug and the logging wezterm provides! I haven't had a single crash across three machines for months.
3. I've worked around macOS colour issues on sRGB, due to wgpu quirks.
4. I added a bunch of caching work around to cache the macOS keyboard layout instead of copying it per keypress.
5. I've updated Symbols Nerd Font Mono to 3.5.1.
6. I added a new config option `send_eof_on_pane_close` to stop wezterm closing tmux. #1396 #2392

I've got way more as well. I've been doing them as branches/commits.

I have just been running these locally, fixing the quirks I've found, instead of creating these PRs, as I know wez has been away from the project (no dramas with this btw!).

Also wanted to test my changes.

Given the work now being done again, I'd be happy to upstream, share, and contribute all these (and more).

Let me know!

wezterm-gui/src/termwindow/mod.rs
```diff
@@ -914,6 +914,17 @@ impl TermWindow {
...
+                // alive, before NSView dealloc invalidates the GPU drawable.
+                // If deferred until TermWindow::drop, glium's Drop impls
+                // (RawProgram, Context, VertexBuffer, etc.) call make_current
+                // which triggers NSOpenGLContext update on a stale IOSurface,
+                // causing SIGABRT.
+                // Order matters: render_state first (its Drop deletes programs,
+                // buffers, textures via the context), then gl (drops the
+                // context itself, which does FBO/VAO/sampler cleanup).
+                self.render_state.take();
+                self.gl.take();
```
bew: TODO: Check where window is destroyed, check where its drop impl is / where gl & render_state used / dropped.

PR #7809, opened 2026-05-23
tab bar: add tab_bar_width config to stretch tabs across the bar

### LINKED ISSUE #1914: iTerm/Safari-style tabs

znzryb:
## Summary

Adds a new `tab_bar_width` opt-in config option (default `"Content"`) that, when set to `"Stretch"`, distributes tab widths equally across the tab bar so the bar is fully covered, eliminating the trailing empty area after the new-tab button.

This is a lightweight alternative to #7591's taffy-based rewrite — ~60 lines, no new dependencies, default behavior unchanged.

## Motivation

Closes #1914 (4-year-old open issue: *"Add 'stretch tabs to fill tab bar' option"*).

The tab bar currently lays out tabs by their natural content width, leaving an empty unusable area between the `+` button and the right edge of the bar. With many tabs the gap grows because the per-tab `1.5 * cell_width` budget subtraction accumulates linearly with N.

## Before (default `tab_bar_width = "Content"`)

The trailing strip on the right is reserved for [`update-right-status`](https://wezterm.org/config/lua/window-events/update-right-status.html) content (date / git branch / pwd / etc) — useful if you surface anything there, wasted space otherwise:

before

## With `tab_bar_width = "Stretch"`

Stretch mode budgets each tab `(pixel_width − new_tab_button_width) / N_tabs`, so the bar is always exactly covered regardless of how many tabs are open:

<p align="center">
  
</p>

<p align="center">
  
</p>

<p align="center">
  
</p>

## Implementation

Addresses @bew's review feedback:

* New `tab_bar_width` enum with `Content` (default) / `Stretch` variants, following the `tab_bar_` prefix convention of existing tab bar config
* The Stretch arm subtracts the new-tab button's footprint up front via a `tabs_available_pixel_width` variable, so the remainder divides evenly across real tabs and no dead zone is left at the right edge (this also resolves the v1 known limitation about a residual `+` button gap)
* The Content arm preserves the historical `1.5 * cell_width` subtraction; the comment now explains the rationale (room for the new-tab button and any `update-right-status` content like date / git branch / pwd)
* Each tab gets `min_width = max_tab_width` in Stretch mode so tabs fill their budgeted slot rather than just being capped by it

`#[dynamic(default)]` keeps the new field at `Content` by default; existing user configs see zero behavioral change.

## Test plan

- [x] Built locally on macOS (M-series, nightly baseline `20260117-154428-05343b38`)
- [x] Default (`Content`): existing behavior unchanged
- [x] Enabled (`Stretch`): tabs fill bar with 2 / 3 / 4 / 12+ open tabs (see screenshots above)
- [x] `cargo fmt --all` clean
- [x] `cargo test --all --exclude wezterm-ssh --no-fail-fast`: **129 suites OK, 335 tests passed, 0 failures** (the 3 SFTP symlink failures in `wezterm-ssh` are caused by macOS `sftp-server` permissions, orthogonal to this change)

## Out of scope (per @bew's review)

* **Classic (non-fancy) tab bar**: @bew has offered to add the same behavior in a follow-up after this lands, so this PR focuses on the default tab bar
* **`tab_title_alignment` config**: orthogonal to stretch mode (alignment matters in `Content` mode too); better tracked as a separate issue

### DIFF
--- a/docs/config/lua/config/tab_bar_width.md
+++ b/docs/config/lua/config/tab_bar_width.md
@@ -8,7 +8,8 @@ tags:
 
 {{since('nightly')}}
 
-Controls how tab widths are distributed across the tab bar.
+Controls how tab widths are distributed across the **fancy** tab bar.
+The classic (non-fancy) tab bar is not affected for now.
 
 Accepted values:
 
@@ -17,7 +18,12 @@ Accepted values:
   fill the bar. This preserves the historical behavior.
 * `"Stretch"` — tabs share the bar width equally so the bar is fully
   covered, eliminating the empty trailing space that otherwise appears
-  after the new-tab button.
+  after the new-tab button. Each tab's rendered width is forced to
+  `(pixel_width − new_tab_button_width) / N_tabs`. The glyph-based
+  [`tab_max_width`](tab_max_width.md) limit still caps the title
+  string upstream, but in Stretch mode the pixel slot is typically
+  the tighter constraint and titles longer than the slot are
+  truncated with the standard `…` ellipsis.
 
 ```lua
 config.tab_bar_width = "Stretch"
--- a/docs/config/lua/config/tab_bar_width.md
+++ b/docs/config/lua/config/tab_bar_width.md
@@ -1,0 +1,26 @@
+---
+tags:
+  - appearance
+  - tab_bar
+---
+
+# `tab_bar_width = "Content"`
+
+{{since('nightly')}}
+
+Controls how tab widths are distributed across the tab bar.
+
+Accepted values:
+
+* `"Content"` (default) — each tab is sized to fit its title, leaving
+  empty space after the last tab when the combined tab widths do not
+  fill the bar. This preserves the historical behavior.
+* `"Stretch"` — tabs share the bar width equally so the bar is fully
+  covered, eliminating the empty trailing space that otherwise appears
+  after the new-tab button.
+
+```lua
+config.tab_bar_width = "Stretch"
+```
+
+See also [`tab_max_width`](tab_max_width.md).
--- a/wezterm-gui/src/termwindow/render/fancy_tab_bar.rs
+++ b/wezterm-gui/src/termwindow/render/fancy_tab_bar.rs
@@ -6,7 +6,7 @@ use crate::termwindow::render::corners::*;
 use crate::termwindow::render::window_buttons::window_button_element;
 use crate::termwindow::{UIItem, UIItemType};
 use crate::utilsprites::RenderMetrics;
-use config::{Dimension, DimensionContext, TabBarColors};
+use config::{Dimension, DimensionContext, TabBarColors, TabBarWidth};
 use std::rc::Rc;
 use wezterm_font::LoadedFont;
 use wezterm_term::color::{ColorAttribute, ColorPalette};
@@ -303,9 +303,27 @@ impl crate::TermWindow {
                 _ => 0.,
             })
             .sum();
-        let max_tab_width = ((self.dimensions.pixel_width as f32 / num_tabs)
-            - (1.5 * metrics.cell_size.width as f32))
-            .max(0.);
+        let max_tab_width = match self.config.tab_bar_width {
+            TabBarWidth::Stretch => {
+                // Subtract the new-tab button's footprint up front so the
+                // remaining space divides evenly across real tabs and no
+                // dead zone is left at the right edge of the bar
+                let new_tab_button_width = 2.0 * metrics.cell_size.width as f32;
+                let tabs_available_pixel_width =
+                    (self.dimensions.pixel_width as f32 - new_tab_button_width).max(0.);
+                let tab_count = (num_tabs - 1.0).max(1.0);
+                (tabs_available_pixel_width / tab_count).max(0.)
+            }
+            TabBarWidth::Content => {
+                // Reserve room at the right edge of the bar for the new-tab
+                // button and any update-right-status content (date / git
+                // branch / pwd / etc) so left-floated tabs don't push them
+                // off-screen
+                ((self.dimensions.pixel_width as f32 / num_tabs)
+                    - (1.5 * metrics.cell_size.width as f32))
+                    .max(0.)
+            }
+        };
 
         // Reserve space for the native titlebar buttons
         if self
@@ -341,6 +359,10 @@ impl crate::TermWindow {
                 TabBarItem::Tab { tab_idx, active } => {
                     let mut elem = item_to_elem(item);
                     elem.max_width = Some(Dimension::Pixels(max_tab_width));
+                    if self.config.tab_bar_width == TabBarWidth::Stretch {
+                        // Stretch the tab to fill the full slot budgeted to it
+                        elem.min_width = Some(Dimension::Pixels(max_tab_width));
+                    }
                     elem.content = match elem.content {
                         ElementContent::Text(_) => unreachable!(),
                         ElementContent::Poly { .. } => unreachable!(),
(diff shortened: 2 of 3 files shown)


wezterm-gui/src/termwindow/render/fancy_tab_bar.rs
```diff
@@ -341,6 +349,9 @@ impl crate::TermWindow {
                 TabBarItem::Tab { tab_idx, active } => {
                     let mut elem = item_to_elem(item);
                     elem.max_width = Some(Dimension::Pixels(max_tab_width));
+                    if self.config.stretch_tabs_to_fill_bar {
+                        elem.min_width = Some(Dimension::Pixels(max_tab_width));
```
bew, line 364: Please add a comment explaining that this ensures that tabs use the max available space to visually stretch across the whole width.

 🤔 I don't see any rule for tab content alignment, so I guess it sits left-aligned?

What do you think about adding a `tab_title_alignement = "Left/Center/Right"` to control this explicitly?
Let's say that it applies only when the content is < `tab_max_width`.

wezterm-gui/src/termwindow/render/fancy_tab_bar.rs
```diff
@@ -303,9 +303,17 @@ impl crate::TermWindow {
...
             })
             .sum();
-        let max_tab_width = ((self.dimensions.pixel_width as f32 / num_tabs)
-            - (1.5 * metrics.cell_size.width as f32))
-            .max(0.);
+        let max_tab_width = if self.config.stretch_tabs_to_fill_bar {
+            // Stretch mode: budget the entire bar width equally across tabs +
+            // new-tab button. Skip the 1.5*cell padding subtraction used in the
+            // default case — accumulating it across N items is exactly what
+            // produces the trailing dead zone we want to eliminate here.
```
bew: please don't explain why we don't do something, instead prefer to explain below why we do the 1.5*cell padding subtraction.
(is there a good reason? 👀)

bew (commented):
Hello, thanks for the PR!

I'm thinking it would be nice to use the same `tab_bar_` prefix as most other tabbar-related config options.

Also I think we can do something that can evolve a bit more in the future than just a boolean, using an enum instead 🤔

Maybe something like:
```lua
-- Tab bar takes all the space, individual tabs have equal size
-- Each tab title is centered if smaller, or cut with `…` char if too long
-- (overrides `tab_max_width` and enforces it)
tab_bar_width = "Stretch"

-- (default)
-- Tab bar width is derived from each tab sizes
-- Each tab title respects `tab_max_width` (with `…` cut-off if needed)
tab_bar_width = "Content"
```

What do you think?

---

For the NewTab button, it should be removed from the full width used to compute tab max width, do a `tabs_available_pixel_width` variable first (I think?) to make this explicit.

We should also be able to have the same behavior available for the classic (non-fancy) tab bar, but I can do it after your PR if you want to focus on fancy-stuff 🤔

wezterm-gui/src/termwindow/render/fancy_tab_bar.rs
znzryb: Updated the comment! 🙂 We keep this padding to reserve room at the right edge for the new-tab button and any `update-right-status` content (date / git branch / pwd / etc) — without it, left-floated tabs would push those off-screen.

docs/config/lua/config/tab_bar_width.md
```diff
@@ -0,0 +1,26 @@
...
+Controls how tab widths are distributed across the tab bar.
+
+Accepted values:
+
+* `"Content"` (default) — each tab is sized to fit its title, leaving
+  empty space after the last tab when the combined tab widths do not
+  fill the bar. This preserves the historical behavior.
+* `"Stretch"` — tabs share the bar width equally so the bar is fully
+  covered, eliminating the empty trailing space that otherwise appears
+  after the new-tab button.
```
bew: Written as is it looks like it's implemented for both fancy & classic tab bar, could you add a mention for 'fancy only' for now?

Also please mention how tab_max_width is enforced for stretch 🤔

znzryb: Updated 🙂 The page now mentions fancy-only up front, and the `"Stretch"` bullet explains that `tab_max_width` still caps title glyphs upstream while the pixel slot is typically the tighter constraint in Stretch mode.

bew:
> tab_title_alignment config: ... (alignment matters in Content mode too)

Hmm I don't agree, it's only when the tabs are stretched that the tab area is potentially larger than the title and title alignment is useful 🤔

But yes it can come as a later PR

wezterm-gui/src/termwindow/render/fancy_tab_bar.rs
```diff
@@ -303,9 +303,27 @@ impl crate::TermWindow {
                 _ => 0.,
             })
             .sum();
-        let max_tab_width = ((self.dimensions.pixel_width as f32 / num_tabs)
-            - (1.5 * metrics.cell_size.width as f32))
-            .max(0.);
+        let max_tab_width = match self.config.tab_bar_width {
```
bew, line 306: I think the whole computation of the tab max width should be reworked with better input vars, same as `num_tabs` which should NOT include the NewTab button. (`tab_count` below shouldn't be necessary)

I'd want to have proper vars in the `build_fancy_tab_bar` function:
- `new_tab_width`
- `left_zone_width` for buttons and left status area
- `right_zone_width` for buttons and left status area
- `tabs_available_width` which is focused on tabs, uses the above vars to compute the area for tabs

Feel free to move these vars definition around to have the proper inputs available.

.. This should make the computations more straight-forward and more readable.

znzryb: Makes sense 👍 Will rework the width computation along those lines — `num_tabs` filtered to real tabs only, the four input vars defined up front, and both arms expressed in terms of `tabs_available_width`.

bew: Technically I think there won't be 2 arms anymore for `tab_max_width`, the only check will be for setting the element's `min_width` below 🤔

znzryb:
@bew I tried to make `left_zone_width` / `right_zone_width` exact locally, but that ended up touching much more of the fancy tab bar layout than this PR is meant to change.

The integrated title buttons do not all participate in layout the same way: `MacOsNative` returns an empty `Element` here and uses the native window buttons plus the existing placeholder spacing, while the Windows/Gnome styles produce real `Element`s with their own padding/border/margin and then flow into `left_eles` or `right_eles` depending on `integrated_title_button_alignment`. On top of that, left/right status widths are only really known after layout.

So I would rather keep this PR scoped to the original issue: preserve the existing Content behavior, subtract the new-tab button budget for Stretch, and set `min_width` only in Stretch so tabs fill their slots.

A more complete zone-width/layout cleanup feels better as a separate PR.

bew:
Alright, let's keep it more focused.

(you'll be notified if so)

znzryb:
> Alright, let's keep it more focused.
> 
> Note that WezTerm is currently on hold/unmaintained, Wez hasn't been responding to PRs for a while. 👉 I'm interested in starting a community fork, see [#6341 (comment)](https://github.com/wezterm/wezterm/issues/6341#issuecomment-4483047032), so this might gets merged there. (you'll be notified if so)

That's okay. Thanks for the heads-up, and thanks for reviewing this.

PR #7812, opened 2026-05-26
fix(wezterm-ssh): support tilde expansion and multiple files for include directive

erics118:
this PR fixes a few issues with wezterm-ssh:

1. adds support for multiple paths in the `Include` directive
2. expand leading ~/ to user's home directory
3. reads files in lexical order

Previously, the config below would fail because (1) there are multiple paths and (2) it uses `~`. The following (valid) ssh config now works with wezterm.
```
Include ~/.ssh/config.local ~/.orbstack/ssh/config ~/.ssh/1Password/config
```

[reference](https://man7.org/linux/man-pages/man5/ssh_config.5.html#:~:text=appear%20before%20it.-,Include,-Include%20the%20specified)

non-features:
- `~user` substitutions

### DIFF
--- a/wezterm-ssh/src/config.rs
+++ b/wezterm-ssh/src/config.rs
@@ -169,32 +169,47 @@ impl ParsedConfigFile {
     }
 
     fn do_include(
-        pattern: &str,
+        value: &str,
         cwd: Option<&Path>,
         options: &mut ConfigMap,
         groups: &mut Vec<MatchGroup>,
         loaded_files: &mut Vec<PathBuf>,
     ) {
-        match filenamegen::Glob::new(&pattern) {
-            Ok(g) => {
-                match cwd
-                    .as_ref()
-                    .map(|p| p.to_path_buf())
-                    .or_else(|| std::env::current_dir().ok())
-                {
-                    Some(cwd) => {
-                        for path in g.walk(&cwd) {
+        // the Include directive can list multiple whitespace-separated paths,
+        // and a leading `~/` should be expanded to the user's home directory
+        for pattern in value.split_ascii_whitespace() {
+            let (expanded, base): (String, Option<PathBuf>) =
+                if let Some(rest) = pattern.strip_prefix("~/") {
+                    match dirs_next::home_dir() {
+                        Some(home) => (rest.to_string(), Some(home)),
+                        None => (pattern.to_string(), None),
+                    }
+                } else {
+                    (pattern.to_string(), None)
+                };
+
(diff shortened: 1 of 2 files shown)


bew (changes requested):
Hello, thanks for the PR, could you focus on the multiple include support?
#7745 is already fully re-working how `IdentitiesOnly=yes` is handled 🤔 Feel free to review his PR if you have some time!

wezterm-ssh/src/config.rs
```diff
@@ -169,32 +169,47 @@ impl ParsedConfigFile {
...
-                    .or_else(|| std::env::current_dir().ok())
-                {
-                    Some(cwd) => {
-                        for path in g.walk(&cwd) {
+        // the Include directive can list multiple whitespace-separated paths,
+        // and a leading `~/` should be expanded to the user's home directory
+        for pattern in value.split_ascii_whitespace() {
+            let (expanded, base): (String, Option<PathBuf>) =
+                if let Some(rest) = pattern.strip_prefix("~/") {
+                    match dirs_next::home_dir() {
```
wez: Use `resolve_home` to resolve the home directory so that it is consistent with the rest of the code, and so that it can be tested; please also add a test case for this.

wez (changes requested):
Thanks for this! Need to make it testable and add a unit test, as well as update the metadata here in the PR to reflect that it is now refocused on just multiple includes.

wezterm-ssh/src/config.rs
```diff
@@ -5,6 +5,50 @@ use std::path::{Path, PathBuf};
...
+            c => current.push(c),
+        }
+    }
+
+    // remaining tokens
+    if !current.is_empty() {
+        tokens.push(current);
+    }
+
+    tokens
```
erics118, line 50: We need a special split function for `Include` because it supports quoting, and backslash to escape. The tests go through all the most common cases that would occur

wezterm-ssh/src/config.rs
```diff
@@ -5,6 +5,50 @@ use std::path::{Path, PathBuf};
...
+        match c {
+            '"' => in_quotes = !in_quotes,
+            '\\' => {
+                // handle basic escaping of the next character
+                if let Some(next) = chars.next() {
+                    current.push(next);
+                } else {
+                    current.push('\\');
+                }
+            }
```
erics118, line 34: This allows a lot more cases than normally accepted, but satisfies all cases regarding `'\'`, `'"'`, `' '`

Issue #6534, opened 2025-01-01
hovering the mouse focuses options on the launcher


PrayagS:
The title may be a bit confusing. Please find a GIF below,

![2025-01-01 at 21 02 59 - Amber Smelt](https://github.com/user-attachments/assets/40f13aca-94a3-42ac-a8c0-736b237adcc0)

Upon opening the launcher, hovering the mouse is tracked and the options are focused.

Even when I don't hover, before opening the launcher, if the mouse happened to be at a height that's adjacent to the launcher, it focuses that option automatically.

It leads to a bad experience for me because when I type to find, it focuses the intended option but then it focuses back on the option that's being focused by the cursor. It keeps on flapping until I manually move the cursor out of the way.

My setup,
- OS/compositor: macOS 15.2
- Wezterm version: Happening on both stable (`20240203-110809-5046fc22`) and nightly (`20241205-083533-6f375e29`).

I looked at the docs but didn't find any relevant options that would affect this behavior. `disable_default_mouse_bindings` also doesn't affect this.

I'm not sure if this is a bug or not so I didn't use the bug report template. Please let me know if any additional information is needed from my end.

sudo-tee:
I have the same issue with the `InputSelector`.

The issue seems less frequent when  using the option `hide_mouse_cursor_when_typing = false`  but still happen from time to time.

PrayagS:
> The issue seems less frequent when using the option hide_mouse_cursor_when_typing = false but still happen from time to time.

Yes, that's the band aid fix here.

sudo-tee:
@PrayagS  I'm using wezterm on windows with WSL.

sudo-tee:
@usrlocalben I was not clear when I said it's less frequent. It's less frequent because I see the cursor and move it out of the way because it annoys me.

sudo-tee:
I am still facing this issue everyday.. has anybody found a workaround ?

mattcargile:
Yes I quite agree. This is quite jarring and frustrating.

So it just appears the launcher is designed with the mouse in mind. I guess I'm going to switch to pure keyboard short cuts.

sudo-tee:
I ended up doing a hack to move the cursor at the bottom of the screen before opening the InputSelector.

Since my work machine is on windows I'm using this snippet

```lua
  -- HACK: Move the mouse out of the way to prevent accidental hover-selection in InputSelector
  --https://github.com/wezterm/wezterm/issues/6534
  wezterm.background_child_process({
[...]

PR #7816, opened 2026-05-30
Fix stale mouse move restoring hidden cursor and URL hover

### LINKED ISSUE #4249: hide_mouse_cursor_when_typing doesn't work on Windows

Windows

_No response_

### WezTerm version

20230904-133103-914f18b1

### Describe the bug

On Windows, mouse cursor isn't being hidden properly while typing.  The cursor does appear to be hidden after each key press, but after a short time (a couple hundred ms or so) it returns again even though the mouse cursor isn't moving.

This is depicted below:

wezterm-cursor

### To Reproduce

Open Wezterm on Windows, move the mouse within the window region, and start typing. :)

### Configuration

No config.

### Expected Behavior

The cursors should stay hidden until the mouse is moved again.

### Logs

wezterm version: 20230904-133103-914f18b1 x86_64-pc-windows-msvc
Window Environment: Windows
WebGPU: name=AMD Radeon RX 7900 XTX, device_type=DiscreteGpu, backend=Vulkan, driver=AMD proprietary driver, driver_info=23.8.1 (LLPC), vendor=4098, device=29772

### Anything else?

Also doesn't work with the latest official release.

Works fine on my Linux systems, however.

yamam:
Fixes #4249.

When `hide_mouse_cursor_when_typing` is enabled, WezTerm hides the mouse cursor
on key input. On Windows, however, a subsequent `MouseEventKind::Move` can be
delivered without the mouse position actually changing.

That stale same-position move event can cause two visible issues:

- it can call `set_cursor(Some(MouseCursor::Text))`, making the mouse cursor
  visible again immediately;
- it can re-evaluate the hyperlink under the stale mouse position and restore
  `current_highlight`, causing a clickable URL underline to reappear even though
  the mouse cursor is still hidden.

This change tracks the mouse position when the cursor is hidden due to typing.
If a later mouse `Move` event reports the same position, WezTerm keeps the cursor
hidden and suppresses restoring the hyperlink hover state.

When typing hides the cursor while it is over a clickable URL, the existing
`current_highlight` is cleared immediately. This removes the URL underline at
the same time as the mouse cursor is hidden.

Once the mouse position actually changes, or another mouse event is processed,
the hidden-cursor state is cleared and normal mouse cursor / hyperlink hover
behavior resumes.

### DIFF
--- a/window/src/os/windows/window.rs
+++ b/window/src/os/windows/window.rs
@@ -127,6 +127,9 @@ pub(crate) struct WindowInner {
     config: ConfigHandle,
     paint_throttled: bool,
     invalidated: bool,
+    /// Screen position at which the cursor was hidden. Windows may emit a
+    /// stale mouse move at this position after keyboard input.
+    cursor_hidden_at: Option<(i32, i32)>,
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash, Ord, PartialOrd)]
@@ -547,6 +550,7 @@ impl Window {
             config: config.clone(),
             paint_throttled: false,
             invalidated: true,
+            cursor_hidden_at: None,
         }));
 
         // Careful: `raw` owns a ref to inner, but there is no Drop impl
@@ -625,6 +629,10 @@ impl WindowInner {
     }
 
     fn set_cursor(&mut self, cursor: Option<CursorIcon>) {
+        self.cursor_hidden_at = match cursor {
+            None => current_cursor_screen_position(),
+            Some(_) => None,
+        };
         apply_mouse_cursor(cursor);
     }
 
@@ -1710,6 +1718,19 @@ fn client_to_screen(hwnd: HWND, point: Point) -> ScreenPoint {
     ScreenPoint::new(point.x.try_into().unwrap(), point.y.try_into().unwrap())
 }
 
+fn current_cursor_screen_position() -> Option<(i32, i32)> {
+    let mut point = POINT { x: 0, y: 0 };
+    if unsafe { GetCursorPos(&mut point) } != 0 {
+        Some((point.x, point.y))
+    } else {
+        None
+    }
+}
+
+fn screen_position_tuple(point: ScreenPoint) -> (i32, i32) {
+    (point.x as i32, point.y as i32)
+}
+
 fn apply_mouse_cursor(cursor: Option<CursorIcon>) {
     match cursor {
         None => unsafe {
@@ -1754,6 +1775,7 @@ fn cursor_icons_use_windows_cursors() {
 
 unsafe fn mouse_button(hwnd: HWND, msg: UINT, wparam: WPARAM, lparam: LPARAM) -> Option<LRESULT> {
     let inner = rc_from_hwnd(hwnd)?;
+    inner.borrow_mut().cursor_hidden_at = None;
     // To support dragging the window, capture when the left
     // button goes down and release when it goes up.
     // Without this, the drag state can be confused when dragging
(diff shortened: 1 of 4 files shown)


sudo-tee:
This should help with this bug 

https://github.com/wezterm/wezterm/issues/6534

yamam:
I pushed an additional update for this PR.

This PR now handles two related pieces of state when
`hide_mouse_cursor_when_typing` hides the mouse cursor while it is positioned
over a clickable hyperlink:

1. Clear the existing hyperlink hover state when typing hides the cursor.
2. Prevent a later same-coordinate synthetic `Move` event from restoring that
   hyperlink hover state.

The first part is done in `keyevent.rs`:

```rust
self.current_highlight.take();
````

`current_highlight` stores the hyperlink currently being hovered by the mouse.
Calling `take()` clears the `Option`, effectively setting it to `None`, so the
URL underline disappears immediately when the mouse cursor is hidden by typing.

The second part is done in `mouseevent.rs`. The existing same-coordinate Move
check, which was already used to avoid restoring `MouseCursor::Text`, is now
also reused when deciding whether to restore `current_highlight`.

Previously, the cursor itself could remain hidden, but `mouse_event_terminal`
could still re-evaluate the hyperlink under the stale mouse position and restore
`current_highlight`. That caused the clickable URL underline to reappear even
though the mouse cursor was still hidden.

With this change:

* typing hides the mouse cursor
* the current hyperlink hover state is cleared immediately
* same-coordinate synthetic `Move` events do not restore `current_highlight`
* real mouse movement still restores normal hover behavior

**Comment by bew** [comment]
Hello, thanks for tracking this down!

One question: this stale `Move` event sounds Windows-specific.
But your current fix is in shared code (`keyevent.rs`/`mouseevent.rs`), so it affects X11, Wayland, and macOS too.

Could we scope this to the Windows input path instead?
For example, filter the stale `Move` where Windows messages get translated into WezTerm's event.
That way other platforms keep their current behavior which already works well AFAIK.

Also for transparency, how much AI was used for this contribution? What for? (research/implementation/testing/PRwriting)

**Comment by yamam** [comment]
Thanks for the review. I agree that the stale Move event is Windows-specific, and that the fix should not change the shared mouse-event behavior on other platforms.

   I reworked the implementation accordingly:

   - Reverted the shared `wezterm-gui` changes.
   - Moved the handling to `window/src/os/windows/window.rs`.
   - When the Windows cursor is hidden, the current screen position is recorded.
   - A subsequent `WM_MOUSEMOVE` or `WM_NCMOUSEMOVE` at that same position is filtered before it is dispatched as a `WindowEvent::MouseEvent`.
   - A move to a different position is dispatched normally, so the existing cursor and hyperlink-hover behavior resumes.
   - Mouse leave, button, and wheel events clear the temporary state.

GitHub Copilot wrote all of the code in this PR. I reviewed the generated code and tested it on Windows to confirm that it works as expected.

File: window/src/os/windows/window.rs [hunk 1]
```diff
@@ -1854,12 +1881,16 @@ unsafe fn mouse_move(hwnd: HWND, _msg: UINT, wparam: WPARAM, lparam: LPARAM) ->
         inner.track_mouse_leave = TrackMouseEvent(&mut trk) == winapi::shared::minwindef::TRUE;
     }
 
+    if same_position_as_hidden_cursor {
```
bew: please add a comment explaining what this check is about here

And mention `#4249` at the end for reference.

window/src/os/windows/window.rs
```diff
@@ -1754,6 +1775,7 @@ fn cursor_icons_use_windows_cursors() {
 
 unsafe fn mouse_button(hwnd: HWND, msg: UINT, wparam: WPARAM, lparam: LPARAM) -> Option<LRESULT> {
     let inner = rc_from_hwnd(hwnd)?;
+    inner.borrow_mut().cursor_hidden_at = None;
```
bew: I'd suggest defining a method on inner with explicit naming so the code stays readable 🙏

PR #7820, opened 2026-06-01
Add Wine build instructions for wezterm-gui

hktonylee:
Summary
- Provide step‑by‑step setup of MSVC and Strawberry Perl under Wine
- Explain how to build wezterm‑gui with Cargo using rust‑lld and explicit /LIBPATH arguments

Why need to build with Wine?
- Because we can develop the Wezterm inside WSL, without installing separated libraries in Windows

### DIFF
--- a/docs/wine.md
+++ b/docs/wine.md
@@ -1,0 +1,135 @@
+# Building Windows with Wine
+
+This is a local WSL/Linux setup for building `wezterm-gui.exe` with the
+`x86_64-pc-windows-msvc` Rust target, using MSVC through Wine. It keeps the
+Wine prefix and Windows Perl under `./wine/` so the setup does not affect the
+rest of the system.
+
+## Install MSVC for Wine
+
+Install [msvc-wine](https://github.com/mstorsjo/msvc-wine) into this checkout:
+
+```console
+$ git clone https://github.com/mstorsjo/msvc-wine.git msvc-wine
+$ cd msvc-wine
+$ ./vsdownload.py --dest msvc
+$ ./install.sh msvc
+$ cd ..
+```
+
+The examples below assume the MSVC tools are available at
+`./msvc-wine/msvc/bin/x64`.
+
+## Install Portable Strawberry Perl
+
+OpenSSL's Windows/MSVC build requires a Windows-style Perl. The system
+`/usr/bin/perl` is not enough; OpenSSL will reject it because it does not
+produce Windows paths.
+
+Download the latest 64-bit portable Strawberry Perl ZIP from the
+[Strawberry Perl releases repo](https://github.com/StrawberryPerl/Perl-Dist-Strawberry/releases)
+into `./wine/`. Rename it to `strawberry-perl.zip`:
+
+```console
+$ mkdir -p wine/downloads wine/strawberry wine/bin wine/prefix
+$ unzip -q wine/downloads/strawberry-perl.zip -d wine/strawberry
+```
+
+Create a repo-local `perl` wrapper that runs Strawberry Perl through Wine:
+
+```console
+$ cat > wine/bin/perl <<'EOF'
+#!/usr/bin/env bash
+set -euo pipefail
+
+ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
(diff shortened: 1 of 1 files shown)


bew:
I have no way to test these instructions myself, so I'll let Wez review this at some point

sdrik:
I gave these instructions an end-to-end test on WSL2 (Ubuntu 24.04, Wine 9.0) and they do successfully cross-build a working `wezterm-gui.exe` — nice work, and thanks for writing this up.

**Result**

| | |
|---|---|
| Host | WSL2, Ubuntu 24.04, Wine 9.0, rustc stable |
| MSVC toolchain | msvc-wine, MSVC 14.51, Windows SDK 10.0.26100 |
| Build | `Finished release profile in 18m06s`, 0 errors |
| Output | `wezterm-gui.exe` → `PE32+ executable (GUI) x86-64, for MS Windows` (~67 MB) |
| Runs | `wine wezterm-gui.exe` executes and exits cleanly |

OpenSSL built fine through the Strawberry-Perl-via-Wine wrapper (`libcrypto.lib`/`libssl.lib` produced), which confirms the rationale for the Windows Perl.

A few things I hit or noticed while following the doc:

**1. Missing prerequisite — `msitools` (required)**

On a clean machine the very first step (`vsdownload.py`) fails with:

```
FileNotFoundError: [Errno 2] No such file or directory: 'msiextract'
```

`msiextract` comes from `msitools`, which msvc-wine needs to unpack the CRT/SDK MSI payloads. Worth calling out explicitly alongside Wine, e.g.:

```console
$ sudo apt install wine64 msitools
```

**2. The page isn't reachable from the site navigation**

`docs/wine.md` isn't referenced in the nav (the `TOC` in `ci/generate-docs.py`). To be clear, this does **not** break the docs build — with mkdocs 1.6.1 / mkdocs-material 9.7.6 the "not included in the nav" message is INFO-level (not fatal under `strict: true`), and the page is still rendered — but it's orphaned: it doesn't appear anywhere in the navigation and is only reachable by direct URL or search. Registering it in the Install section fixes that:

```diff
             Page("Build from source", "install/source.md"),
+            Page("Build for Windows with Wine", "install/wine.md"),
```

Since it's a build/install topic, it also reads more naturally at `docs/install/wine.md` next to `source.md`/`windows.md`, with a pointer from `install/source.md`'s "Building on Windows" section:

````diff
 $ set PATH=c:\Strawberry\perl\bin;%PATH%
 ```
+
+If you'd rather cross-build the Windows `wezterm-gui.exe` from a WSL/Linux
+environment instead of building natively on Windows, see
+[Building for Windows with Wine](wine.md).
````

**3. On the `rust-lld` vs `link.exe` note**

Using `rust-lld` for the final link is the right call — but in my testing the reason isn't quite that `link.exe` "gets stuck". `link.exe` under Wine works fine for normal links (a trivial `cl` + `link` produces a valid exe, and small links neither fail nor hang). What breaks the *wezterm-gui* link specifically is command-line length: rustc passes the final link **inline as 562 arguments totalling ~58 KB**, which exceeds the Windows/Wine `CreateProcess` limit of 32,767 characters, so `link.exe` fails immediately (silently, exit 1 — even with `/DEBUG:NONE`). `rust-lld` runs natively on Linux (`ARG_MAX` ~2 MB), so the same command links fine — which is exactly why it's the right choice here. Might be worth rewording "can get stuck at the final `wezterm-gui` link step" to something like "fails the final link because the linker command line exceeds Windows' ~32 KB limit under Wine".

**4. Minor**

- `-Clink-arg=/LIBPATH:$SDK_VER/km/x64` points at a directory the documented `vsdownload.py` steps don't create — `km` (kernel-mode driver libs) isn't part of the base Windows SDK. It's harmless (the linker ignores a missing `/LIBPATH:`) and unnecessary for a user-mode binary, so it can just be dropped.
- Strawberry Perl: the prose says to download the zip "into `./wine/`", but the `unzip` command reads `wine/downloads/strawberry-perl.zip` — should be `./wine/downloads/`.
- The page starts with a single `#`; other pages here start their body at `##` (mkdocs-material renders the nav title as the H1).
- Optional: `vsdownload.py --architecture x64` restricts the download to the x64 target used here and is a good deal smaller than the all-architecture default.

I've applied 1, 2 and 4 locally and confirmed a clean `mkdocs build` with the page showing up in the nav — happy to open a follow-up PR with these tweaks if that's easier.

PR #7823, opened 2026-06-02
Add opt-in forwarding for macOS marked text

alexzhang1030:
## Summary
- add `macos_forward_marked_text_to_pane` for macOS
- forward AppKit marked-text updates as incremental pane input when the setting is enabled
- reconcile the final `insertText` callback with forwarded marked text and document shell/readline-oriented behavior

## Motivation
Some speech IMEs stream partial transcripts through `setMarkedText`. WezTerm keeps those updates as composition state and sends the final `insertText` to the pane, so shells receive one final chunk. This opt-in setting lets users forward the live marked-text updates into the active pane.

## Testing
- `cargo test -p window marked_text_update`
- `cargo check -p window`
- `cargo fmt --all --check` (stable rustfmt reports the existing nightly-only `imports_granularity` warnings)

### DIFF
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -721,6 +721,8 @@ pub struct Config {
     pub xim_im_name: Option<String>,
     #[dynamic(default)]
     pub ime_preedit_rendering: ImePreeditRendering,
+    #[dynamic(default)]
+    pub macos_forward_marked_text_to_pane: bool,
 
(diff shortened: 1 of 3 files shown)


bew:
Hello, could you open a dedicated feature request first so this can be discussed properly with context?
Please describe in full your use-case and why you'd want something like this

PR #7827, opened 2026-06-04
feat(wayland): allow hiding fallback frame

wezzle:
## Summary

- add `window_frame.hidden` to let Wayland users hide WezTerm's fallback client-side decoration frame
- keep the default unchanged so client-side decorations remain visible unless explicitly disabled
- document the option together with `window_decorations = "RESIZE"`, including the resize-border tradeoff

## Context

On compositors that support server-side decorations, `window_decorations = "NONE"` can negotiate compositor-provided decorations. Using `window_decorations = "RESIZE"` selects client-side decorations, but currently always displays Smithay's fallback frame. This option lets users keep native Wayland while opting out of that fallback frame.

## Testing

- `nix develop ./nix -c cargo test --all`
- `nix develop ./nix -c cargo fmt --all`

### DIFF
--- a/config/src/color.rs
+++ b/config/src/color.rs
@@ -567,6 +567,17 @@ fn default_window_close() -> String {
 
 #[derive(Debug, Clone, FromDynamic, ToDynamic)]
 pub struct WindowFrameConfig {
+    /// Hide the Wayland fallback client-side decoration frame.
+    ///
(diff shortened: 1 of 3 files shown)


bew:
Hello! Thanks for the PR but I don't think we need a separate config for this, it should get fixed once #7095 lands 🤔

PR #7828, opened 2026-06-05
Fix hyperlink hover underline for identical text with different links

### LINKED ISSUE #7835: Hyperlink hover underline can be incorrect for identical hyperlinks

Windows

_No response_

### WezTerm version

dab1816d0

### Describe the bug

When the same hyperlink is printed multiple times with identical displayed
text, hyperlink hover underlining can be inconsistent.

For example, hovering one occurrence may underline another identical occurrence,
or hovering an occurrence may change the mouse cursor to `Hand` while the
expected underline does not appear on that occurrence.

This appears to be caused by the line-to-element shape cache reusing shaped
results for identical rendered text. Hyperlink hover underlining depends on the
`Hyperlink` identity for the specific line being rendered, but the shape cache is
keyed by rendered shape.

This is fixed by PR #7828.

### To Reproduce

Print the same OSC 8 hyperlink twice:

```sh
printf '\033]8;;file:///tmp/same\033\\same-display-text\033]8;;\033\\\n'
printf '\033]8;;file:///tmp/same\033\\same-display-text\033]8;;\033\\\n'
````

Then hover each `same-display-text` occurrence.

### Expected behavior

Only the hovered hyperlink occurrence should be underlined.

Moving the mouse away from the hyperlinks should clear the underline.

### Actual behavior

The hover underline can be applied inconsistently between the identical
occurrences.

For example:

* Hovering one occurrence may underline another identical occurrence.
Image

* Hovering another occurrence may show the `Hand` cursor, but the expected
  underline may not appear on the hovered occurrence.
Image

### Additional context

The displayed text is identical, so the line-to-element shape cache can reuse the
same shaped result. However, hover underlining is line-specific because it depends
on the `Hyperlink` identity for the line being rendered.

If a shaped result containing hover underline state is reused for another
identical-looking occurrence, the underline state can be applied to the wrong
line. Conversely, a shaped result without hover underline state can be reused for
the hovered occurrence, preventing the underline from appearing.

PR #7828 changes the rendering logic so that the shape cache is bypassed for a
line that contains the currently highlighted hyperlink, and shaped results for
such lines are not stored in the line-to-element shape cache.

### Configuration

no config

### Logs

_No response_

### Anything else?

_No response_

yamam:
Fixes #7835.

## Summary

Fix inconsistent hyperlink hover underlines when identical text appears on
multiple lines with different hyperlink identities.

The line-to-element shape cache is keyed by rendered shape. This means identical
text on different lines can share the same cache entry. However, hyperlink hover
highlighting depends on the `Hyperlink` identity for the specific line being
rendered, not just on the rendered text.

When one line is hovered, the shaped result may include hover underline state.
If that shaped result is reused for another line with the same rendered text but
a different `Hyperlink` Arc, the underline can appear on the wrong line. In the
opposite direction, a non-highlighted shaped result can be reused for the hovered
line, preventing the expected underline from appearing.

This change checks whether the current line contains the current highlighted
hyperlink before consulting the line-to-element shape cache.

- If the current line contains the highlighted hyperlink, the shape cache is
  bypassed.
- Shaped results that contain the highlighted hyperlink are not stored in the
  shape cache.
- Non-highlighted hyperlink lines remain cacheable.

This keeps hyperlink hover underline state line-specific while preserving shape
cache reuse for non-highlighted hyperlink lines.

## Testing

Manual testing:

- Display identical hyperlink text on multiple lines where each occurrence has a
  different `Hyperlink` Arc.
- Hover the upper occurrence.
  - Confirm that only the upper occurrence is underlined.
- Hover the lower occurrence.
  - Confirm that only the lower occurrence is underlined.
- Move the mouse away from the hyperlink.
  - Confirm that the underline disappears.

### DIFF
--- a/wezterm-gui/src/termwindow/render/mod.rs
+++ b/wezterm-gui/src/termwindow/render/mod.rs
@@ -102,10 +102,6 @@ pub struct LineToEleShapeCacheKey {
 pub struct LineToElementShapeItem {
     pub expires: Option<Instant>,
     pub shaped: Rc<Vec<LineToElementShape>>,
-    // Only set if the line contains any hyperlinks, so
-    // that we can invalidate when it changes
-    pub current_highlight: Option<Arc<Hyperlink>>,
-    pub invalidate_on_hover_change: bool,
 }
 
 pub struct LineToElementShape {
--- a/wezterm-gui/src/termwindow/render/screen_line.rs
+++ b/wezterm-gui/src/termwindow/render/screen_line.rs
@@ -113,28 +113,50 @@ impl crate::TermWindow {
         let cursor_range_pixels = params.left_pixel_x + cursor_range.start as f32 * cell_width
             ..params.left_pixel_x + cursor_range.end as f32 * cell_width;
 
+        let current_highlight = self.current_highlight.as_ref();
+        let mut line_contains_any_hyperlink = false;
+        let mut line_contains_current_highlight = false;
+
+        // The line-to-element shape cache is keyed by rendered shape.  Identical
+        // text on different lines can therefore share a cache key even when the
+        // lines have different hyperlink attributes, for example when OSC 8
+        // hyperlinks are involved.  Check the current line before consulting the
+        // cache so that we don't reuse a shaped result whose hover underline
+        // state was computed for a different Hyperlink Arc.
+        for cell_idx in 0..num_cols {
+            let Some(cell) = params.line.get_cell(cell_idx) else {
+                continue;
+            };
+
+            let Some(hyperlink) = cell.attrs().hyperlink() else {
+                continue;
+            };
+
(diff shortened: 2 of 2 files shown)


bew:
Hello, could you create an issue first?
With information about your environment & minimal steps to reproduce the issue 🙏

Also, could you mention how much AI was used and how much review of the code you did manually?

yamam:
Thanks for the feedback!

I created an issue with environment information and minimal reproduction steps:
#7835

I also updated the PR description to reference the issue. This PR fixes the issue
described there.

Regarding AI usage: I used M365 Copilot(GPT-5.5) as an assistant during
the investigation and drafting. I manually reviewed the suggestions, adjusted
the code changes, built the code locally, inspected the behavior with logs,
and verified the final behavior manually.

PR #7841, opened 2026-06-07
ci/cd: Update nix-related workflows

bew:
Review pass on the Nix-related workflows.
- lock actions to a major branch, at least
- various tweaks and readability improvements
- only push built result to Cachix if workflow triggered by `schedule` or manual `workflow_dispatch` event

Note @happenslol @gabyx @emiller88: I've removed you from auto-assigned role on flake.lock update PRs, I don't think you recently participated there. But if you're still interested we can talk about it!

### DIFF
--- a/.github/workflows/nix-build.yml
+++ b/.github/workflows/nix-build.yml
@@ -22,15 +22,21 @@ jobs:
     name: Build
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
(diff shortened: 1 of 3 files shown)


bew:
Is there a way to test continuous workflow without merging? 🤔

Also, what's the intended difference between `build` & `continuous` workflows?

bew:
Replaced with #7854

Issue #6979, opened 2025-05-21
Window uses transparency when unfocused


kkanden:
### What Operating System(s) are you seeing this problem on?

Windows

### WezTerm version

20250518-060218-5663e749

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

Having the transparency and acrylic effect on, the window loses transparency when it is unfocused.

### Configuration

```lua
config.window_background_opacity = 0.75
config.win32_system_backdrop = "Acrylic"
```

### Expected Behavior

The window retains transparency even if unfocused.

### Anything else?

This was already once reported in #5895 but was not responded to.

wez:
Duplicate of #5895

Issue #1099, opened 2021-09-02
When wezterm launched "Minimize/maximize all windows" doesn't work.


shaddu38:
### What Operating System(s) are you seeing this problem on?

Windows

### WezTerm version

20210901-211314-3d227d27

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

When wezterm launched "Minimize/maximize all windows"-function doesn't work.

### To Reproduce

OS: Windows 10
Launch any app (e.g. notepad), launch wezterm, press Win+D or button at right bottom corner and you see that windows minimization does not occur.

### Configuration

No config.

### Expected Behavior

"Minimize/maximize all windows" should work with lauched wezterm.

### Logs

10:53:20.050 INFO wezterm_mux_server_impl::local > setting up C:\Users\admin\.local/share/wezterm\gui-sock-3584
10:53:20.099 WARNING window::os::windows::window > EGL init failed Config says to avoid EGL, fall back to WGL
10:53:20.217 WARNING wezterm_term::terminalstate > unhandled DecPrivateMode SetDecPrivateMode(Unspecified(9001))
10:53:20.226 INFO wezterm_gui::termwindow > OpenGL initialized! GeForceGTX 1060 6GB/PCIe/SSE2 4.5.0 NVIDIA 456.71 is_context_loss_possible=true wezterm version: 20210901-211314-3d227d27

### Anything else?

In version 20210502-154244-3f7122cb and previous "Minimize/maximize all windows" works fine.

wez:
Sorry it's taken a while to get to this. I can't reproduce what you described; for me, it is sometimes a bit slow to minimize everything when wezterm is open. I don't know why this is.  Even with wezterm closed, it is sometimes slow to minimize all, so I'm not really sure that this is attributable to wezterm :-/

wez:
Possibly fixed by https://github.com/wez/wezterm/commit/67b8c453cad275c7ea78fd505049e66016d17dba

wez:
I'm pretty confident that this is fixed

Issue #593, opened 2021-03-31
running local build can take up to a minute to start


JeanMertz:
While working on #588 I noticed that fairly often, doing `cargo run --bin wezterm`, or `cargo build && ./target/debug/wezterm` would not start the application as expected. It seemed as if the process hung, and so I always aborted, tried a couple of more builds, and _somehow_ at some point it would start again.

**update** it also appears to happen when I build using `cargo build --release`.

I did check to see if any servers were running in the background causing issues, but couldn't find any in the list of running processes.

Eventually, through chance, I left the hanging process running, only to find the terminal become visible/available after waiting for about a minute. This seems to be reproducible (as in, it always shows up _when_ the hang happens, but the hanging itself doesn't always happen, sometimes the window shows up straight away).

I'm not sure what's happening here, but I figured it's worth a bug report.

Here's part of the log output I took, I added a couple of newlines to the location in the logs where nothing happens for a minute, before the terminal GUI shows up.

<details>
<summary>Details</summary>

```log
env RUSTC_LOG=trace WEZTERM_LOG=trace ./target/debug/wezterm

 2021-03-31T05:16:19.548Z TRACE config > consider config: /Users/jean/.config/wezterm/wezterm.lua
 2021-03-31T05:16:19.548Z TRACE async_io::driver > block_on()
 2021-03-31T05:16:19.549Z TRACE async_io::driver > block_on: completed
 2021-03-31T05:16:19.549Z TRACE polling::kqueue  > new: kqueue_fd=4, read_stream=UnixStream { fd: 5, local: (unnamed), peer: (unnamed) }
 2021-03-31T05:16:19.549Z TRACE async_io::driver > main_loop: waiting on I/O
 2021-03-31T05:16:19.549Z TRACE async_io::reactor > process_timers: 0 ready wakers
 2021-03-31T05:16:19.549Z TRACE polling           > Poller::wait(_, None)
 2021-03-31T05:16:19.549Z TRACE polling::kqueue   > wait: kqueue_fd=4, timeout=None
 2021-03-31T05:16:19.656Z DEBUG config            > Reloaded configuration! generation=1
 2021-03-31T05:16:19.657Z TRACE async_io::reactor > react: 0 ready wakers
 2021-03-31T05:16:19.657Z TRACE async_io::driver  > main_loop: waiting on I/O
 2021-03-31T05:16:19.657Z TRACE async_io::reactor > process_timers: 0 ready wakers
[...]

PR #7848, opened 2026-06-11
windows: add win32_keep_system_backdrop_when_inactive option

### LINKED ISSUE #5895: Acrylic / Mica for unfocused windows

Windows

_No response_

### WezTerm version

20240730-082727-56a27e93

### Describe the bug

When setting `win32_system_backdrop` to either Mica or Acrylic, the transparency only works when the window is currently focused. When focusing another application, the wezterm window goes opaque.

This seems to happen with every combination of Mica/Acrylic and OpenGL/WebGPU/Software

I thought this might be a general Windows issue, but I tried the same thing with Terminal and that seems to work fine:

- top left: wezterm focused
- bottom left: wezterm unfocused
- top right: terminal unfocused

image

### To Reproduce

_No response_

### Configuration

	front_end = "OpenGL",
	win32_system_backdrop = "Mica",

### Expected Behavior

_No response_

### Logs

_No response_

### Anything else?

_No response_

ChristianOrrala:
## Summary

On Windows, when a window uses a `win32_system_backdrop` material
(`Acrylic`, `Mica` or `Tabbed`), DWM stops rendering the material and
fades it to a solid fallback color as soon as the window becomes
inactive, so the window looks opaque until it regains focus. This is a
long-standing pain point reported in #5895 (and its duplicates #6979,
#1099, #593).

This PR adds an opt-in option, `win32_keep_system_backdrop_when_inactive`,
that keeps the configured backdrop material rendered while the window is
inactive.

## Root cause

The fade is not a hard OS limitation — Windows Terminal keeps its
backdrop while inactive, so an app *can* opt out of it. For the DWM
system backdrop (`DWMWA_SYSTEMBACKDROP_TYPE`, which is what wezterm
uses), the inactive fade is driven by the non-client activation state
that `DefWindowProc` applies when handling `WM_NCACTIVATE`.

## Change

When the option is enabled, `wm_ncactivate` reports the window as always
active for non-client purposes: it forwards `WM_NCACTIVATE` to
`DefWindowProc` with `wParam = TRUE` and `lParam = -1` (the latter tells
`DefWindowProc` not to repaint the non-client area for the faked state
change). DWM then keeps drawing the backdrop material while the window is
inactive.

- **Opt-in**, defaults to `false` — no behavior change for existing
  configs.
- **No-op** when `win32_system_backdrop` is `Auto` or `Disable`.
- **Side effect** (documented): with a native title bar
  (`window_decorations` including `TITLE`), the title bar always renders
  in its active colors. This is inherent to the `WM_NCACTIVATE` approach.

I kept it opt-in rather than changing the default because it alters
long-standing non-client behavior and has the title-bar side effect
above; this way existing setups are untouched and users who want the
effect can enable it.

## Testing

The behavior depends on DWM compositing a real desktop session and on
focus changes, which the test suite doesn't exercise (there are no
automated tests around `win32_system_backdrop` for the same reason). I
validated it manually on Windows 11 (build 28120):

- Option **off**: Acrylic/Mica fades to solid on focus loss (current
  behavior, reproduced).
- Option **on**: the material persists across focus changes.
- `Auto`/`Disable`: no effect, as expected.
- `cargo check`, `cargo fmt --all --check` and `cargo test -p config`
  pass.

Happy to add an automated test if there's an established pattern for
exercising win32 backdrop/focus behavior that I've missed.

## Docs & changelog

- New page:
  `docs/config/lua/config/win32_keep_system_backdrop_when_inactive.md`
- Changelog entry under *Continuous/Nightly → New*.

## Related issue

Refs #5895 — this provides the capability requested there (and in the
duplicates #6979, #1099, #593). I've left it as a reference rather than
an automatic close since the option is opt-in; please close #5895 if you
consider it resolved.

### DIFF
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -581,6 +581,13 @@ pub struct Config {
     #[dynamic(default = "default_win32_acrylic_accent_color")]
     pub win32_acrylic_accent_color: RgbaColor,
 
+    /// Only works on Windows: when enabled, the window keeps rendering
+    /// its configured `win32_system_backdrop` material (Acrylic/Mica/Tabbed)
+    /// while the window is inactive, instead of letting DWM fade the
+    /// backdrop to its solid fallback color.
+    #[dynamic(default)]
+    pub win32_keep_system_backdrop_when_inactive: bool,
+
     /// Specifies the alpha value to use when rendering the background
     /// of the window.  The background is taken either from the
     /// window_background_image, or if there is none, the background
--- a/docs/config/lua/config/win32_keep_system_backdrop_when_inactive.md
+++ b/docs/config/lua/config/win32_keep_system_backdrop_when_inactive.md
@@ -1,0 +1,32 @@
+---
+tags:
+  - appearance
+---
+
+# `win32_keep_system_backdrop_when_inactive = false`
+
+{{since('nightly')}}
+
+By default, when a window using a
+[win32_system_backdrop](win32_system_backdrop.md) material (*Acrylic*, *Mica*
+or *Tabbed*) becomes inactive, DWM stops rendering the backdrop material and
+fades it to a solid fallback color, making the window look opaque until it is
+focused again.
+
+Setting this option to `true` keeps the configured backdrop material rendered
+while the window is inactive, so the window stays translucent with the same
+effect regardless of focus:
+
+```lua
+config.window_background_opacity = 0
+config.win32_system_backdrop = 'Acrylic'
+config.win32_keep_system_backdrop_when_inactive = true
+```
+
+This works by reporting the window to DWM as always active for non-client
+rendering purposes (the `WM_NCACTIVATE` state). A side effect is that, when
+using a `window_decorations` style that shows the native title bar, the title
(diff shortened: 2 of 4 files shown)


config/src/config.rs
```diff
@@ -581,6 +581,13 @@ pub struct Config {
     #[dynamic(default = "default_win32_acrylic_accent_color")]
     pub win32_acrylic_accent_color: RgbaColor,
 
+    /// Only works on Windows: when enabled, the window keeps rendering
+    /// its configured `win32_system_backdrop` material (Acrylic/Mica/Tabbed)
+    /// while the window is inactive, instead of letting DWM fade the
+    /// backdrop to its solid fallback color.
```
bew:
```suggestion
    /// backdrop to its default solid fallback color.
```

docs/config/lua/config/win32_keep_system_backdrop_when_inactive.md
bew: also add a cross link to this page in the doc page for `win32_system_backdrop`

Issue #7847, opened 2026-06-11
Crash when using pane:move_to_new_window when single pane in window


eliasreid:
### What Operating System(s) are you seeing this problem on?

Linux Wayland

### Which Wayland compositor or X11 Window manager(s) are you using?

KDE Plasma 6.6.4

more system info:
```
Operating System: Nobara Linux 43
KDE Plasma Version: 6.6.4
KDE Frameworks Version: 6.25.0
Qt Version: 6.10.3
Kernel Version: 7.0.9-200.nobara.fc43.x86_64 (64-bit)
Graphics Platform: Wayland
Processors: 24 × AMD Ryzen 9 5900X 12-Core Processor
Memory: 16 GiB of RAM (15.5 GiB usable)
Graphics Processor: NVIDIA GeForce RTX 3080
Manufacturer: Micro-Star International Co., Ltd.
Product Name: MS-7C91
System Version: 2.0
```

### WezTerm version

20260607_235149_6c16469e and main branch @ 891bed31b75f7a71b78e8f42ad07ae89bf99a7de

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

I copied the config sample shown in: https://wezterm.org/config/lua/pane/move_to_new_window.html

When triggering this action_callback in a window that only has one pane, I get a crash.

### To Reproduce

Launch wezterm with the below config, press shortcut to trigger action (CTRL+M)

### Configuration

```lua
local wezterm = require 'wezterm'

local config = wezterm.config_builder()

config.keys = {
  { key = 'm', mods = 'CTRL', action = wezterm.action_callback(function (win, pane)
    local tab, window = pane:move_to_new_window()
  end)}
}

return config
```

### Expected Behavior

I expect nothing to happen when triggering this shortcut in a new window

### Logs

```
00:30:49.186  ERROR  env_bootstrap > panic at window/src/os/wayland/window.rs:1117:14 - !?
   0: env_bootstrap::register_panic_hook::{{closure}}
   1: std::panicking::panic_with_hook
   2: std::panicking::panic_handler::{closure#0}
   3: std::sys::backtrace::__rust_end_short_backtrace::<std::panicking::panic_handler::{closure#0}, !>
   4: __rustc::rust_begin_unwind
   5: core::panicking::panic_fmt
   6: core::option::expect_failed
[...]

PR #7849, opened 2026-06-12
don't move the pane to a new window if it's already the only pane in the window

eliasreid:
Addresses this crash: https://github.com/wezterm/wezterm/issues/7847

I am new to the wezterm codebase (and rust), so I'm happy to accept any feedback if this is not the right fix!

### DIFF
--- a/mux/src/lib.rs
+++ b/mux/src/lib.rs
@@ -1251,30 +1251,9 @@ impl Mux {
         window_id: Option<WindowId>,
         workspace_for_new_window: Option<String>,
     ) -> anyhow::Result<(Arc<Tab>, WindowId)> {
-        let (domain_id, src_window_id, src_tab) = self
+        let (domain_id, src_window_id, src_tab_id) = self
             .resolve_pane_id(pane_id)
-            .ok_or_else(|| anyhow::anyhow!("pane {} not found", pane_id))?;
-
-        if self
-            .get_window(src_window_id)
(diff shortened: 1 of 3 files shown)


mux/src/lib.rs
```diff
@@ -1251,10 +1251,30 @@ impl Mux {
...
+                .get_tab(src_tab)
+                .ok_or_else(|| anyhow::anyhow!("Invalid tab id {}", src_tab))?;
+
+            //if there's only one tab and one pane in the current window, it doesn't make
+            //sense to move it to a new tab or window
+            if src_tab
+                .count_panes()
+                .map_or(true, |num_panes| num_panes <= 1)
+            {
+                return Ok((src_tab, src_window_id));
```
bew: Please add a small log here mentioning what's happening in this case.

bew (commented):
Hello, thank you for the PR!

We'll need a log to ensures we can properly debug a potential issue about this code without silent automatic behavior!

mux/src/lib.rs
eliasreid: done, let me know if that looks alright

bew:
Hello! I did some testing this WE and pushed a commit to change the direction a bit:
- Instead of simply returning `Ok`, better to return `Err` to show the caller that the action wasn't done.
  This makes your original Lua config `pane:move_to_new_window()` fail with an error in the console.
- I updated the docs for the Lua functions accordingly, with config examples

I did more testing and I'm confident this is good to go if Wez approves the approach.

PR #7850, opened 2026-06-12
build(deps): bump `harfbuzz` to 14.2.1

RoloEdits:
Previous update (`11.2.1`): f412c817c6aa6a16f2d41583ac6284821e5b3eb8


RoloEdits:
Not sure if there is an automated way that updates the `flake.lock`. I don't use nix and am not very familiar with it.

bew:
Hello, don't worry about Nix, its files can be updated later on.

I'm sorry but I'm not sure these kind of update PRs are welcome from people outside the core maintainers to ensure the updates come from trusted source 🤔

Do you need something specific from the new harfbuzz version?

RoloEdits:
Fixes for some glyphs. I build from source anyways, so already run this, so if this needs to be from wez than thats fine too. 

As far as the process:
- I checked out the 14.2.1 tag in the harfbuzz submodule
- git add the changes so that the rev doesnt change
-  ran the `regenerate.sh` in `deps/harfbuzz`, which used my install of bindgen.
- git add .
- Changed the nix file version
- updated the changelog

PR #7854, opened 2026-06-13
nix: Simplify flake.nix; update Nix-related workflows

bew:
With recent enough Nix we can now remove the explicit submodules dependencies and let Nix do the git actions.

Includes #4986 with comments, for nicer package version And a few other improvements to match modern Nix standards (`finalAttrs`).

Also includes #7841 to update Nix-related workflows:
- lock actions to a major branch, at least
- various tweaks and readability improvements
- only push built result to Cachix if workflow triggered by schedule or manual workflow_dispatch event

### DIFF
--- a/.github/workflows/nix-build.yml
+++ b/.github/workflows/nix-build.yml
@@ -22,15 +22,21 @@ jobs:
     name: Build
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v6
+
       - name: Install nix
-        uses: DeterminateSystems/nix-installer-action@main
+        uses: DeterminateSystems/nix-installer-action@v22
+
       - name: Cache build artifacts
-        uses: DeterminateSystems/flakehub-cache-action@main
+        uses: DeterminateSystems/flakehub-cache-action@v3
+
       - name: Check flake inputs
+        # note: waiting for a release with node24 upgrade
+        # see: https://github.com/DeterminateSystems/flake-checker-action/pull/110
         uses: DeterminateSystems/flake-checker-action@main
         with:
           flake-lock-path: ./nix/flake.lock
           ignore-missing-flake-lock: false
+
       - name: Build default package
         run: nix build ./nix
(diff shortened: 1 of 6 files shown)


bew:
The Determinate Nix version 3.21.1 seems to be broken with that branch 👀
Opened: https://github.com/DeterminateSystems/nix-src/issues/500

ruiiiijiiiiang:
I started getting this error: `error: input attribute 'submodules' not supported by scheme 'github'` after this merge. What should I do?

bew:
Hello, sorry about that @ruiiiijiiiiang which version of Nix are you using?

Support for `inputs.self.submodules = true` was added on Nix 2.27 AFAIK:
https://nix.dev/manual/nix/2.34/release-notes/rl-2.27.html

I'm using Nix 2.34.7
Try to update your Nix version, using this option is very useful for this repo as there are a few submodules that are relatively annoying to keep in sync otherwise 🤔

vtumuluri:
@bew  Its not due to the Nix version. I had 2.34.7 as well and I got the same error as @ruiiiijiiiiang 
```
❯ nix --version                                                                                                                                                      12:54
nix (Nix) 2.34.7
```
I guess `inputs.self.submodules` only works for git urls (`git+https://github.com/wezterm/wezterm?dir=nix`)and not github urls ( `github:wezterm/wezterm?dir=nix`) because this fetches tarballs without .git info like submodules.
Changing my wezterm input url accordingly fixed the issue after git fetching 2 screenfuls of tags and branches.  
However, I dont like this solution. I dont want to fetch refs and tags every time I want to update wezterm. I hope there is a better way. 
The relevant issue with github urls is : [13571](https://github.com/NixOS/nix/issues/13571)

bew:
Thanks for the feedback, well that's annoying... ><
👉 I've opened #7867 to revert only this part, can you test it?

EDIT: merged!

vtumuluri:
I commented in the new PR as well but for completeness repeating here. I tested #7867 and its working with that. Thank you very much.

ruiiiijiiiiang:
It is working for me now. Thank you for the fast response!

PR #7859, opened 2026-06-15
fix(term): loop defang_paste until no embedded bracketed markers remain

### LINKED ISSUE #7784: Bracketed Paste “defanged” transformation is not thorough

Linux Wayland

Plasma

### WezTerm version

20240203-110809-5046fc22

### Describe the bug

Bracketed paste attempts to fix a security issue by removing the brackets from the pasted text:
https://github.com/wezterm/wezterm/blob/577474d89ee61aef4a48145cdec82a638d874751/term/src/terminalstate/mod.rs#L836
However, this transformation is only done *once* so can worked around easily.

(I did not test on the nightly because I can see the problem in the code right there)

### To Reproduce

With `xclip` installed, run
```sh
printf 'echo foo; echo malicious command\e\e[201~[201~\n\e\e[200~[200~' | xclip -i -selection CLIPBOARD
```
to get the malicious clipboard contents.  Then with bracketed paste ON you can hit Ctrl+Shift+V and it will still run both `echo` commands.

The actual output `Wezterm` sends, as shown by `socat -x -v`, is
```
> 2026/05/11 21:04:46.884619  length=57 from=0 to=56
 1b 5b 32 30 30 7e 65 63 68 6f 20 66 6f 6f 3b 20  .[200~echo foo;
 65 63 68 6f 20 6d 61 6c 69 63 69 6f 75 73 20 63  echo malicious c
 6f 6d 6d 61 6e 64 1b 5b 32 30 31 7e 0a           ommand.[201~.
 1b 5b 32 30 30 7e 1b 5b 32 30 31 7e              .[200~.[201~
--
```

### Configuration

No relevant config; setting the font, setting `$TERM` correctly, and working around a mouse wheel issue are all irrelevant.

### Expected Behavior

When sending bracketed paste, the end-bracketed-paste escape sequence `\e[201~` is never sent until all the actual pasting is done.

### Logs

No log messages from the times I was testing, at least not at INFO level or higher (I donʼt see *any* log messages lower than INFO so they might just be skipped)

### Anything else?

Most terminals I test donʼt do *any* sanitizing so this is still ahead of the curve, but itʼd be nice for it to be consistent.  One (I forgot which) got rid of every `\e` when in bracketed paste mode, which is certainly easy but sometimes you actually want to paste an escape character deliberately.

Fixing this wouldnʼt be as safe as explicitly prompting the user like #2900 suggests, but is mostly already implemented and people might want to turn explicit prompts off.

molloyzak13:
The bracketed paste defanging in `send_paste` strips embedded `\x1b[200~` and `\x1b[201~` sequences from pasted text so an attacker can't inject their own paste boundaries. The current code does a single pass of `str::replace`, which finds non-overlapping matches. That means a crafted input like `\x1b\x1b[200~[200~` gets the inner pair stripped on pass one, leaving `\x1b[200~[200~` — the outer sequences recombine into a valid marker that hits the terminal.

The fix extracts a `defang_paste` helper that loops until the output stabilises. On each iteration it clones the current string, runs both replaces, and breaks if nothing changed. The logic is the same, it just runs it to a fixed point.

Added six unit tests: plain text, single start marker, single end marker, both markers, the reporter's double-nested exploit case, triple nesting, and mixed double-nested start+end. The double-nested case fails on the old single-pass code (asserts `"[200~injected"` against `"injected"`) and passes with the loop. Ran the full `term` test suite — all 231 existing tests pass plus the 6 new ones.

Closes #7784.

### DIFF
--- a/term/src/terminalstate/mod.rs
+++ b/term/src/terminalstate/mod.rs
@@ -817,9 +817,24 @@ impl TerminalState {
 
     /// Send text to the terminal that is the result of pasting.
     /// If bracketed paste mode is enabled, the paste is enclosed
-    /// in the bracketing, otherwise it is fed to the writer as-is.
     /// De-fang the text by removing any embedded bracketed paste
-    /// sequence that may be present.
+    /// sequence that may be present.  Loops until no more occurrences
+    /// remain, because a single pass of str::replace is not idempotent:
+    /// nested sequences like `\x1b\x1b[200~[200~` can leave a valid
+    /// marker behind after the first sweep.
+    fn defang_paste(text: &str) -> String {
+        let mut result = text.to_string();
+        loop {
+            let prev = result.clone();
+            result = result.replace("\x1b[200~", "").replace("\x1b[201~", "");
+            if result == prev {
+                break;
+            }
+        }
+        result
+    }
+
+    /// in the bracketing, otherwise it is fed to the writer as-is.
     pub fn send_paste(&mut self, text: &str) -> Result<(), Error> {
         let mut buf = String::new();
         if self.bracketed_paste {
@@ -833,7 +848,7 @@ impl TerminalState {
         };
 
         let canon = canon.canonicalize(text);
-        let de_fanged = canon.replace("\x1b[200~", "").replace("\x1b[201~", "");
+        let de_fanged = Self::defang_paste(&canon);
         buf.push_str(&de_fanged);
 
         if self.bracketed_paste {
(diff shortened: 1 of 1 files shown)


bew:
> [!WARNING]
> This looks a lot like a bot account, it was created yesterday and opened 15 security issues on 15 repos, as well as ~30 PRs (most closed by the account itself..)..

The changes might be good though, worth checking out (except the doc comments ^^')

molloyzak13:
Fair observation. The account's new and I've been shipping a lot of small fixes, I get that the volume looks odd from the outside. Happy to address any review notes.

PR #7863, opened 2026-06-15
fix: Save wayland selection offer to every surface

### LINKED ISSUE #6685: Clipboard (most times) not working between windows

bew:
Copied from #7034 from @XeroOl, updated with current `main` with explaining comment.

Original PR description:
In Wayland, when something is copied to the clipboard, an "offer" is given to every "device".
WezTerm holds these offers in a "CopyAndPaste" struct, and reads from the offer when the user presses paste.

When WezTerm receives a data offer, it previously only saved it to one CopyAndPaste based on the active surface id.

After this change, it writes it to every CopyAndPaste.

Fixes #6685
Closes #7034 (replaces it)

---
Tested on KDE 6.6.5, all scenarios on #6685 now work.


bew:
All tests are green, let's merge this and solve copy/paste on Wayland 🚀

XeroOl:
Ideally when someone who understand's wayland's copy/paste protocol better has time, there could be a "real" fix.

It doesn't make sense to create multiple CopyAndPastes just to keep them all in sync with eachother. I wrote my PR trying to find a minimal change that works; It probably makes sense to clean up at some point.

bew:
Thanks, I agree that saving a separate `CopyAndPaste` per surface doesn't seem like a great idea now and could probably be simplified to a single store 🤔

PR #7869, opened 2026-06-18
refactor(wayland): share one `CopyPasteOffer` across all windows

bew:
Previously each window held its own CopyAndPaste (one per surface), causing the data_device selection handler to broadcast the incoming offer to every window on each clipboard change.

Since the compositor sends one selection event per client (not per surface), a single shared CopyPasteOffer on WaylandState is sufficient. The broadcast loop is replaced by a single one confirm_selection() call.

Also add doc comments on all clipboard-related fields in WaylandState.

Assisted-by: Claude Sonnet 4.6; iterated, reviewed & tested manually

refs: #6685 & #7863

### DIFF
--- a/window/src/os/wayland/copy_and_paste.rs
+++ b/window/src/os/wayland/copy_and_paste.rs
@@ -15,25 +15,29 @@ use crate::{Clipboard, ConnectionOps};
 use super::data_device::TEXT_MIME_TYPE;
 use super::state::WaylandState;
 
+/// Holds the most recent clipboard offer received from another Wayland client.
+/// Updated on every `wl_data_device::selection` event.
 #[derive(Default)]
-pub struct CopyAndPaste {
+pub struct CopyPasteOffer {
     data_offer: Option<SelectionOffer>,
 }
 
-impl std::fmt::Debug for CopyAndPaste {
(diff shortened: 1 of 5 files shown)


XeroOl:
I really don't like the spawn_into_main_thread.
I thought we were wrapping it in a Arc<Mutex<CopyPasteOffer>> specifically to deal with thread safety, so we don't need to be on the main thread to edit it.

bew:
> I really don't like the spawn_into_main_thread. I thought we were wrapping it in a Arc<Mutex> specifically to deal with thread safety, so we don't need to be on the main thread to edit it.

I initially looked into removing the additional "thread, but the issue is that the Wayland connection is stored in `thread_local!` storage on the main thread, so all GUI operations must be as well to be able to access the Wayland connection.

And since we're supposed to be on the same thread I tried to remove the `Arc<Mutex>` but had some nasty lifetime issues.. I'm not super familiar with this or how to better fix it so if you want to try an alternative path I'm interested to see it!
The AI suggested an alternative to move the `{get,set}_clipboard_data` on the `WaylandState` directly to avoid all issues with cross borrows, but I did want to keep the methods separate..

bew:
I went ahead & merged it, we can align other PRs later on if needed 🚀

PR #7871, opened 2026-06-20
Fix PaneFocused notification feedback loop - take 3 

### LINKED ISSUE #4390: Switching panes in multiple directions at once causes the focused pane to rapidly switch when any key is pressed

ankitson:
Following up from #7763 - see that PR for the explanation.

Addressed the review comments on that PR - no real changes, just comments and variable renames.
Tested rapidly switching panes manually, and with cli - did not see the issues from #4390 (also #4737) with these changes.
Some janky diagrams explaining it [here](https://github.com/ankitson/wezterm/blob/main/investigations/sshmux-panefocused-storm/DIAGRAMS.md)

Fixes #4390
Closes #7763

### DIFF
--- a/mux/src/tab.rs
+++ b/mux/src/tab.rs
@@ -2226,6 +2226,13 @@ mod test {
     use wezterm_term::color::ColorPalette;
     use wezterm_term::{KeyCode, KeyModifiers, Line, MouseEvent, StableRowIndex};
 
+    // The mux is a process-global singleton (`Mux::set_mux`), and focus
+    // notifications dispatch through `Mux::get()`. Tests that install a mux must
+    // therefore run serially; cargo's default parallel runner would otherwise let
+    // them clobber each other's global and cross-deliver notifications. Hold this
+    // guard for the lifetime of any such test.
+    static MUX_TEST_GUARD: Mutex<()> = Mutex::new(());
+
     struct FakePane {
         id: PaneId,
         size: Mutex<TerminalSize>,
@@ -2538,6 +2545,7 @@ mod test {
 
     #[test]
     fn set_active_pane_can_suppress_mux_notification() {
+        let _guard = MUX_TEST_GUARD.lock();
         let size = TerminalSize {
             rows: 24,
             cols: 80,
@@ -2583,6 +2591,60 @@ mod test {
 
         // Assert: the active pane changed, but not a single PaneFocused was emitted.
         assert_eq!(Vec::<PaneId>::new(), *notified_panes.lock());
+        assert_eq!(1, tab.get_active_pane().unwrap().pane_id());
+
+        Mux::shutdown();
+    }
+
+    #[test]
+    fn set_active_pane_notifies_mux_by_default() {
+        let _guard = MUX_TEST_GUARD.lock();
+        let size = TerminalSize {
+            rows: 24,
+            cols: 80,
+            pixel_width: 800,
+            pixel_height: 600,
+            dpi: 96,
+        };
+
+        let mux = Arc::new(Mux::new(None));
+        Mux::set_mux(&mux);
+
+        // Record every PaneFocused the mux broadcasts. `recorded` is the clone
+        // moved into the subscriber; `notified_panes` is read by the assertion.
+        let notified_panes = Arc::new(Mutex::new(Vec::new()));
+        let recorded = Arc::clone(&notified_panes);
+        mux.subscribe(move |notification| {
+            if let MuxNotification::PaneFocused(pane_id) = notification {
+                recorded.lock().push(pane_id);
+            }
+            true
+        });
+
+        // Build a tab with two side-by-side panes (pane 2 becomes active on split).
+        let tab = Tab::new(&size);
+        let pane_1 = FakePane::new(1, size);
+        tab.assign_pane(&pane_1);
+        let pane_2 = FakePane::new(2, size);
+        let split = tab
+            .compute_split_size(
+                0,
+                SplitRequest {
+                    direction: SplitDirection::Horizontal,
+                    ..Default::default()
+                },
+            )
+            .unwrap();
+        tab.split_and_insert(0, SplitRequest::default(), Arc::clone(&pane_2))
+            .unwrap();
+        pane_1.resize(split.first).unwrap();
+
+        // Act: switch the active pane back to pane 1 via the default helper,
+        // which requests a mux notification.
+        tab.set_active_pane(&pane_1);
+
+        // Assert: the active pane changed and exactly one PaneFocused was emitted.
+        assert_eq!(vec![1], *notified_panes.lock());
         assert_eq!(1, tab.get_active_pane().unwrap().pane_id());
 
         Mux::shutdown();
--- a/mux/src/tab.rs
+++ b/mux/src/tab.rs
@@ -20,6 +20,12 @@ pub type Cursor = bintree::Cursor<Arc<dyn Pane>, SplitDirectionAndSize>;
 static TAB_ID: ::std::sync::atomic::AtomicUsize = ::std::sync::atomic::AtomicUsize::new(0);
 pub type TabId = usize;
 
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+pub enum NotifyMux {
+    Yes,
+    No,
+}
+
 #[derive(Default)]
 struct Recency {
     count: usize,
@@ -699,7 +705,11 @@ impl Tab {
     }
 
     pub fn set_active_pane(&self, pane: &Arc<dyn Pane>) {
-        self.inner.lock().set_active_pane(pane)
+        self.set_active_pane_with_notify(pane, NotifyMux::Yes)
+    }
+
+    pub fn set_active_pane_with_notify(&self, pane: &Arc<dyn Pane>, notify: NotifyMux) {
+        self.inner.lock().set_active_pane(pane, notify)
     }
 
     pub fn set_active_idx(&self, pane_index: usize) {
@@ -1747,7 +1757,7 @@ impl TabInner {
         self.active
     }
 
-    fn set_active_pane(&mut self, pane: &Arc<dyn Pane>) {
+    fn set_active_pane(&mut self, pane: &Arc<dyn Pane>, notify: NotifyMux) {
         let prior = self.get_active_pane();
 
         if is_pane(pane, &prior.as_ref()) {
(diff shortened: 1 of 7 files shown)


mux/src/tab.rs
```diff
@@ -2517,6 +2536,58 @@ mod test {
...
+                    ..Default::default()
+                },
+            )
+            .unwrap();
+        tab.split_and_insert(0, SplitRequest::default(), Arc::clone(&pane_2))
+            .unwrap();
+        pane_1.resize(split.first).unwrap();
+
+        // Act: switch the active pane back to pane 1, requesting NO mux notification.
+        tab.set_active_pane_with_notify(&pane_1, NotifyMux::No);
```
bew, line 2590: I think we can add a second 'act' to this test to test that we can pass Yes and mux gets notified

mux/src/tmux_commands.rs
```diff
@@ -341,7 +341,10 @@ impl TmuxDomainState {
 
                     match mux.get_tab(local_tab.tab_id) {
                         Some(tab) => {
-                            tab.set_active_pane(&local_pane);
+                            // We fire our own PaneFocused on the next line;
+                            // suppress the duplicate advise_focus_change would emit.
+                            // See <https://github.com/wezterm/wezterm/issues/4390>
+                            tab.set_active_pane_with_notify(&local_pane, NotifyMux::No);
                             mux.notify(MuxNotification::PaneFocused(local_pane.pane_id()));
```
bew, line 350: for this case we could instead remove the manual notify & use `NotifyMux::Yes` 🤔
WDYT?

mux/src/tab.rs
ankitson: Added a new test for this

mux/src/tmux_commands.rs
ankitson: I've been digging into this, but so far I don't have a full picture. The `set_active_pane_with_notify` path calls `advise_focus_change` which skips the `mux.notify` in certain cases - I don't know if that might cause clients to go out of sync in some cases. Like for example, the active_pane path does not sync the active tab - only the pane within a tab.

Keeping the `mux.notify` should be correct and at worst just creates a few extra notifications we don't need - but no feedback loops. I added a small comment about that. I would recommend we keep it as is for safety, unless you're sure we don't need it, and we can revisit when we better understand the full picture?

bew: Alright 👍

mux/src/tab.rs
```diff
@@ -698,8 +704,17 @@ impl Tab {
         self.inner.lock().get_active_idx()
     }
 
+    /// Activate `pane` and broadcast a `PaneFocused` mux notification.
+    /// Use this for genuine, user-initiated activations. Call sites that emit
+    /// their own `PaneFocused` should use [`Tab::set_active_pane_with_notify`]
+    /// with [`NotifyMux::No`] to avoid doubling it.
+    /// See <https://github.com/wezterm/wezterm/issues/4390>
     pub fn set_active_pane(&self, pane: &Arc<dyn Pane>) {
```
bew, line 712: _from my [original review](https://github.com/wezterm/wezterm/pull/7763#pullrequestreview-4333659248) on previous PR_

what do you think about completely removing `set_active_pane` (or add `notify_mux` to that function) and always force caller to think about mux notification or not (and document when not) 🤔

Effectively:
- removing the new function
- adding a `notify_mux` param to existing one (we don't need the long function name I think)
- make sure to update the doc, mention that usage of `NotifyMux::No` should be explained

---
NOTE: in doc / comments I like to have each idea/sentence start on its own line (not in the middle of an existing one).

pro-vi:
The pre-set of `focused_remote_pane_id` before reconcile breaks the echo loop. That part of the fix is correct. Side effect: it also suppresses the `SetFocusedPane` round-trip on the receiving client, which is what used to update the server's per-client identity tracking (`ClientInfo.focused_pane_id`).

Concrete symptom: in a multi-client attach, client B activates remote pane X. Server broadcasts `PaneFocused(X)`. Client A receives it, pre-sets cache to X, reconciles, displays X locally. Because the cache already matches, `advise_focus` on A suppresses `SetFocusedPane(X)`. Server's `ClientInfo` for A still says Y (A's previous focus). `wezterm cli list-clients` now prints A's FOCUS as Y while A visibly has X.

Pre-PR this was incidentally tracked via the echo (the same echo this PR correctly breaks). Now both behaviors share one field, so the fix that breaks the echo also breaks the tracking.

One option that covers both: split the cache into two states. A persistent "last advised pane" (current semantic) plus a transient "currently reconciling this inbound focus" guard set and cleared around the PDU handler. The transient guard breaks the echo without affecting `advise_focus` for later local activations.

Non-blocking. The hang fix matters more than `list-clients` accuracy, and this can be a follow-up. Flagging because @bew is already weighing the API shape.

Hit by the underlying freeze on macOS 26.4 Tahoe (M5 Max, wezterm nightly, ~30 panes mux-attached). Thanks @ankitson and @bew for the work.

bew:
@pro-vi this looks very much AI written, it's pretty hard to follow what this is saying with all the short sentences, low context and many technical words 'mushed' together.. 😬
👉 Do you understand everything in this message you posted?

Could you rewrite it yourself in a more understandable way?
E.g. what "echo loop" are you talking about?

If there is a concrete issue with this PR, please provide minimal reproduction steps + analysis if you can

ankitson:
@pro-vi I think this is the issue you're reporting -

1. 2 wezterm clients are connected to the same remote session
2. Clients are both on pane X
3. Client A switches to pane Y
4. Client B's UI also switches to pane Y, however `wezterm cli list-clients` will show client B is on pane X

Is that right? I was not able to reproduce this issue on my setup running this branch. If i switch panes/tabs on client A and then run `wezterm cli list-clients` on the server, it reports both clients having the correct focused pane Y.

pro-vi:
@bew @ankitson I must apologize for letting the agent which was exploring the same issue that I ran into (while also hunting down related PRs) gone a bit awry. The intention was good but the output was mushy indeed. Sorry about the time wasted trying to reproduce and reading slop.

That said, the underlying finding does seem to hold up empirically. I went back and tested the PR branch vs. the baseline. Sharing minimal repro and observations below (edited with agent, but I made sure it's intentional and more clear):

What "echo loop" was supposed to mean by the agent: pre-PR, when client A received a `PaneFocused(X)` broadcast from the server, A's `advise_focus` saw the cache mismatch and sent `SetFocusedPane(X)` back to the server. The server then re-broadcast. That round-trip is what this PR fixes. The PR pre-sets `focused_remote_pane_id = X` BEFORE the reconcile so the round-trip is suppressed.

The side effect: that round-trip was ALSO what kept the server's per-client `ClientInfo.focused_pane_id` accurate. Without it, the per-client tracking on the server goes stale for non-activating clients.

The asymmetry comes from `sessionhandler.rs:337,366`: when A sends `SetFocusedPane` via its own connection, the server scopes `with_identity(A)` and `record_focus_for_current_identity` updates A's `ClientInfo`. The broadcast path in `dispatch.rs:169` just encodes and sends; it never calls `record_focus_*` for any receiver. So per-client tracking only updates via the receiver's own echo, which is exactly what the PR's dedup suppresses.

Minimal repro on the PR branch (head `eb5b952f`, built with `cargo build --release`):

1. Isolated mux config:
```lua
config.unix_domains = { { name = 'spike', socket_path = '/tmp/spike.sock' } }
config.default_domain = 'spike'
```
2. Launch two `wezterm-gui` instances pointing at this config.
3. Bring window A to focus briefly so its paint loop populates its `ClientInfo` on the server.
4. Verify: `wezterm cli list-clients` shows `A.FOCUS = <active pane id>`.
5. From a third terminal (a CLI client, separate identity), activate the OTHER pane: `wezterm cli activate-pane --pane-id <other>`.
6. Run `wezterm cli list-clients` again. A.FOCUS stays at the OLD pane id.

Actual output from my run:

```
BASELINE:
  panes: pane_id=0 is_active=False, pane_id=1 is_active=True
  A.FOCUS = 1

(activate-pane 0 via CLI)

AFTER:
  panes: pane_id=0 is_active=True, pane_id=1 is_active=False
  A.FOCUS = 1   <-- stale, expected 0
```

Same test on the merge-base (`c4b1d89f`, pre-PR) shows `A.FOCUS = 0` after the activate, correctly. So this is a regression introduced by the PR, not preexisting. @ankitson, if you switched pane via A's own keybind, it goes through A's own connection and DOES correctly update A's `ClientInfo`. The bug only appears when the activator is a different client (see the asymmetry from earlier).

User-visible impact: scripts/automation that read `wezterm cli list-clients` to figure out which pane a given client is looking at get stale data when the activator is a different client (CLI, another GUI window, another remote actor). Visible only in multi-client mux setups. So it is quite an edge case.

ankitson:
@bew @pro-vi Okay.. i think this is the hack to fix the bug that you recommended earlier - 
https://github.com/ankitson/wezterm/pull/1/changes

We update the servers view of client-focused panes before sending the message to client (discussed at the end about this too - can cause a problem). We have to hack a thing there to make sure client doesn't send a SetFocusedPane in response to having received a PaneFocused from server. So the deeper problem seems to be this loop

```
Pdu::PaneFocused(X)
    -> mux.focus_pane_and_containing_tab(local_pane)
    -> tab.set_active_pane_with_notify(..., NotifyMux::No)
    -> current.focus_changed(true)
    -> ClientPane::advise_focus()
    -> sends SetFocusedPane(X)
```

Somewhere along this loop - in `focus_changed` I think, client loses track of whether we actually initiated this pane focus change, or if we are responding to a `PaneFocused` from the server. If we are responding to a `PaneFocused` from the server, it should just not echo a `SetFocusedPane` back to the server..

Maybe we need something like:

```
focus_changed(focused, FocusChangeSource::Local | FocusChangeSource::ServerReconcile)
```
which i think is sort of pushing the current `NotifyMux::No` a layer deeper maybe? Like we need to understand throughout the system when we are responding to an event from the server, which doesn't need to be echo'd back to the server, vs. when we are responding to an event on the client itself, which does need to be sent to server.

But the other view is - with this PR/method, the server assumes the client has received the PaneFocused and updated its state, without actually receiving an ACK back from the client that this actually happened.. so that can create adifferent problem as well. So maybe we actually need to wait for an ACK from client before server marks focus - but all this should apply to general architecture of Wezterm as well, for any server-client reconciliation - i'm not sure how these problems are currently handled.

PR #7881, opened 2026-06-25
termhost: Add Windows Default Terminal support

XTheocharis:
OK


bew:
Hello,

This looks 100% vibe-coded, this PR is way too big and changes/adds too many precise things to be reviewed as is.

This kind of changes needs to be planned & discussed in issues first, so we can check that this is the right direction for the project, and the right way to do it before investing in drafting an implementation. And such implementation would have to be split into different incremental PRs to be human reviewable / testable / ...

I'll have to close this.

bew:
> I think it is a feature that many Windows users would enjoy.

Yeah I'm sure they would love to use Wezterm for everything!

----

I concede that it might have been a bit too harsh for me to immediately close the PR..
But I personally have no way (nor the knowledge to review in detail at the moment) to test these changes. Other people might chime in though.
.. I'd appreciate if you can make an issue about this features, with reference material you might be aware of, existing implementation you might know about, ..

:point_right: Why did you put all this new code in `wezterm-gui` (which has almost no platform-specific code) and not as a kindof `windows-termhost` backend in `window` crate, which has all the OS-specific handling 🤔
You could re-use some parts of the `windows` backend for non-`termhost` -specific Windows things.

Also: How much AI was used here, and how much manual review, testing, debugging did you do?
What's your level of knowledge on Windows programming?
_Do you 'trust-the-AI' or do you have practical knowledge to explain the changes without AI?_

NOTE: I see that you've pushed a few binary files, those shouldn't be included there, only project maintainers are allowed to get & push binary files from trusted sources since those are sensitive.
The PR will have to take this into account somehow :thinking:

PR #7890, opened 2026-07-01
Optimize case conversion performance

HuaGu-Dragon:
# Objective

* Make ASCII-only, case-insensitive matching clearer and more direct (use eq_ignore_ascii_case / to_ascii_lowercase instead of to_lowercase).

### DIFF
--- a/wezterm-gui/src/main.rs
+++ b/wezterm-gui/src/main.rs
@@ -139,7 +139,7 @@ async fn async_run_ssh(opts: SshCommand) -> anyhow::Result<()> {
         ssh_option.insert("wezterm_ssh_verbose".to_string(), "true".to_string());
     }
     for (k, v) in opts.config_override {
-        ssh_option.insert(k.to_lowercase().to_string(), v);
+        ssh_option.insert(k.to_lowercase(), v);
     }
 
     let dom = SshDomain {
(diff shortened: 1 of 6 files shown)


wezterm-gui/src/overlay/quickselect.rs
```diff
@@ -76,7 +76,7 @@ fn compute_labels_for_alphabet_impl(
     let alphabet = if make_lowercase {
         alphabet
             .chars()
-            .map(|c| c.to_lowercase())
+            .map(|c| c.to_lowercase().to_string())
```
HuaGu-Dragon, line 79: Sorry for previous I misunderstand the return type here

HuaGu-Dragon:
I'm a bit confused about the macOS build failure. Could it be a random failure?

bew:
> I'm a bit confused about the macOS build failure. Could it be a random failure?

Yes the macOS build is kinda broken at the moment, I don't hold you to it ;)

wezterm-gui/src/main.rs
```diff
@@ -139,7 +139,7 @@ async fn async_run_ssh(opts: SshCommand) -> anyhow::Result<()> {
         ssh_option.insert("wezterm_ssh_verbose".to_string(), "true".to_string());
     }
     for (k, v) in opts.config_override {
-        ssh_option.insert(k.to_lowercase().to_string(), v);
+        ssh_option.insert(k.to_lowercase(), v);
```
bew: why not use `to_ascii_lowercase` here? AFAIK ssh option keys are known and always ascii 🤔

HuaGu-Dragon: Good point—I didn’t fully consider that, sorry about that.

bew:
Thanks for the PR ;)

PR #7892, opened 2026-07-01
nix: fix binary locations in macos app

erics118:
previously, the nix build script would incorrectly place some of the files:
- `wezterm`,`wezterm-mux-server`,`wezterm-gui`,`strip-ansi-escapes` were incorrectly placed into `$OUT_APP/` rather than `$OUT_APP/Contents/MacOS/`
- `shell-integration` was placed in `$OUT_APP/` rather than `$OUT_APP/Contents/Resources/`

This PR fixes the behavior so it now matches the correct behavior of macOS app packaging (and the `ci/deploy.sh` script).

### DIFF
--- a/nix/flake.nix
+++ b/nix/flake.nix
@@ -146,7 +146,7 @@
               OUT_APP="$out/Applications/WezTerm.app"
               cp -r assets/macos/WezTerm.app "$OUT_APP"
               rm $OUT_APP/*.dylib
-              cp -r assets/shell-integration/* "$OUT_APP"
+              cp -r assets/shell-integration/* "$OUT_APP/Contents/Resources"
               # macOS will only recognize our application bundle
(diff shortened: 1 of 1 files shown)


bew:
Hello, is this related to an issue you're having?

erics118:
I implemented this fix a few months ago, so I can't rmr if it was broken before. IIRC, the wezterm app (built via nix) would still launch properly, even though the executable was incorrectly placed.

This issue only manifests when using nix to build the app, with `nix build .#default`, or from a flake overlay (which I use)

But this is definitely a correctness issue that should be addressed

erics118:
I just tested, and the wezterm app from before this change will still launch, so yeah, its mostly for correctness

bew (approved):
I don't see why not, thanks for the PR!

PR #7894, opened 2026-07-02
fix(deps): enable chrono "clock" feature to restore Utc/Local::now()

sdrik:
## Problem

Building an individual crate such as `cargo build -p mux` fails:

```
error[E0599]: no associated function or constant named `now` found for struct `Utc`
  --> mux/src/client.rs:61:32
   |
61 |             connected_at: Utc::now(),
   |                                ^^^ associated function or constant not found in `Utc`
```

## Root cause

The workspace declares chrono with `default-features=false` and only `["unstable-locales", "serde"]`. Starting with chrono **0.4.43** (the lockfile recently moved 0.4.42 → 0.4.44), `Utc::now()` and `Local::now()` were moved behind a new `now` cargo feature (`#[cfg(feature = "now")]`), which is enabled by the `clock` feature. Neither was enabled here, so those constructors no longer exist.

## Why a full build still worked

Cargo unifies the features of a shared dependency across the whole build graph. Other crates in the tree — `spa` (pulled in by `time-funcs`) and `serde_with`'s `std` feature — request `chrono/clock`, which transitively enabled `now` for every crate and masked the missing feature. Building a subtree that excludes those crates (e.g. `-p mux`) drops the accidental activation and exposes the latent breakage.

## Affected crates

`mux` and `frecency` (`Utc::now()`), `env-bootstrap` and `time-funcs` (`Local::now()`), plus the `wezterm` CLI helpers.

## Fix

Declare `clock` explicitly on the workspace chrono dependency so the feature is available regardless of build scope, removing the fragile implicit activation. `clock` (rather than bare `now`) is required because `Local::now()` needs `iana-time-zone` for local timezone detection.

## Verification

`cargo build -p mux -p env-bootstrap -p time-funcs -p frecency` now compiles cleanly (previously failed on `-p mux`).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -58,7 +58,7 @@ camino = "1.0"
 cassowary = "0.3"
 cc = {version="1.0", features = ["parallel"]}
 cgl = "0.3"
-chrono = {version="0.4", default-features=false, features=["unstable-locales", "serde"]}
+chrono = {version="0.4", default-features=false, features=["clock", "unstable-locales", "serde"]}
 clap = {version="4.0", features=["derive"]}
 clap_complete = "4.4"
 clap_complete_fig = "4.0"


bew:
Hello, I'm curious why are you trying to check/build only the `mux`  crate?

sdrik:
I'm working on various tmux control-mode enhancements, and my iteration loop builds individual crates (e.g. `cargo build -p mux`) rather than the whole workspace — it's much faster for tight feedback. Part of my workflow uses Claude Code for automated verification, and it keeps tripping over this. It can work around it, but that lengthens and adds cost to the loop.

More importantly, I don't think it's purely a workflow quirk: these crates call `Utc::now()`/`Local::now()` without declaring the chrono feature that provides them. They only compile today because `spa` happens to pull in `chrono/clock` and Cargo unifies it across the graph — any package-scoped build that excludes `spa` breaks. So the fix removes a latent fragility rather than just papering over my setup. I suspect #7897's author hit the same thing.

bew (approved):
Alright, thanks for the PR and the detailed build-features debugging 👍

Issue #4371, opened 2023-09-30
Alternative actions for hyperlinks (e.g., copy hyperlink on ctrl+mouse left)


fabian-thomas:
**Is your feature request related to a problem? Please describe.**
I'm currently using alacritty and would like to switch to wezterm for the multiplexing functionality. In alacritty one has the option to specify alernative actions for clicking on hyperlinks or links defined by regex and I would love to have that in wezterm as well. It's easiest to explain that with my current alacritty config (yaml):
```yaml
  # NOTE: this is just below but the following removed:
  # - ipv6
  # - ipv4
  # - hashs
  # - versions
  # - all words
  define: &link_regex "((ipfs:|ipns:|magnet:|mailto:|gemini:|gopher:|https:|http:|news:|file:|git:|ssh:|ftp:|vnc:)\
            [^\u0000-\u001F\u007F-\u009F<>\"\\s{-}\\^⟨⟩`]+)\
            |[a-zA-Z0-9-_~./!?&@+]+/[a-zA-Z0-9-_~./!?&@]+(:[0-9]+)?(:[0-9]+)?\
            |[a-zA-Z0-9-_~./!?&@+]+(:[0-9]+)?:[0-9]+"

  define: &copy_regex "((ipfs:|ipns:|magnet:|mailto:|gemini:|gopher:|https:|http:|news:|file:|git:|ssh:|ftp:|vnc:)\
            [^\u0000-\u001F\u007F-\u009F<>\"\\s{-}\\^⟨⟩`]+)\
            |([0-9a-fA-F]{1,4}:){1,7}([0-9a-fA-F]{1,4}){0,1}(:[0-9a-fA-F]{1,4}){1,7}\
            |[0-9]{1,3}\\.[0-9]{1,3}\\.[0-9]{1,3}\\.[0-9]{1,3}\
            |[0-9a-f]{32,64}\
            |([0-9]+\\.)+[0-9]+(-[0-9]+)?\
            |[a-zA-Z0-9-_~./!?&@+]+/[a-zA-Z0-9-_~./!?&@]+(:[0-9]+)?(:[0-9]+)?\
            |[a-zA-Z0-9-_~./!?&@+]+(:[0-9]+)?:[0-9]+\
            |[a-zA-Z0-9-_~./!?&@+:]+"
            # description:
            # protocol links
            # ipv6/mac
            # ipv4
            # hashes/git commits (git log)
            # versions
            # files with colon + line
            # files with colon + line
[...]

Issue #3791, opened 2023-05-28
Run any command when clicking on a link matching a regexp


nojhan:
I want to be able to open a file at a specific line in my IDE, just by clicking on an error message in my terminal.

For instance, for opening a python script indicated in a python error, within Vim, I would use configure the following option:

    regexp = \B(/\S+?\.py)\S{2}\sline\s(\d+)
    command = gvim --servername IDE --remote +$1 $0

For command such as "open in a [web|file] browser", the hyperlink_rules feature already does the job. But if one want to trigger just any command, it cannot work, as it is limited to a predefined set of protocols.

At least URxvt and Terminator have this feature. I [implemented it in Terminator](https://github.com/gnome-terminator/terminator/pull/399), by using the existing link handler, and allow adding a `terminator://` protocol, that bypass xdg-open and run the configured command. This had the minimum footprint and reused the existing link handling code.

# Example

Run a script that display the info to catch, here a Python script displaying THE ABSOLUTE PATH.

```shell
echo "print('What the F)" >  test.py
python3 $(pwd)/test.py
```

Note that if you just do ./test.py, then the path information is contextual and it's not going to work (or else we would need to pass the current directory from the shell to the URL manager). In that case, the observed behaviour would be to open an empty file.

Then, you have to pass the mouse over the path segment to see it underlined, and [Ctrl-] click to open it.

Example (the mouse is over the link):
![](https://user-images.githubusercontent.com/285009/114521947-2b695f80-9c43-11eb-9ede-f23ebb423a19.png)

wez:
See:
* https://github.com/wez/wezterm/discussions/529
* https://github.com/wez/wezterm/discussions/3261

or generally search for `open-uri` here on github.

docs:

https://wezfurlong.org/wezterm/config/lua/window-events/open-uri.html

nojhan:
Thanks. That's actually 118 lines of code with several config entry points for the user to maintain. Is there any chance that such a useful feature would be included in the codebase at some point?

PR #7903, opened 2026-07-05
feat(lua): expose link/mouse position under cursor; OpenLinkAtMouseCursor fallback

### LINKED ISSUE #7899: Lua mouse bindings: branch on hyperlink presence — expose link/position under cursor, or add a fallback to OpenLinkAtMouseCursor

In a `mouse_bindings` callback there's no way to ask "is there a hyperlink (OSC 8 or a matched `hyperlink_rules` entry) under the cursor right now, and if so which one?" `window:current_event()` returns only `{ Down = { button, streak } }` (or the wheel delta) — no mouse cell, no link. `pane:get_cursor_position()` is the *text* cursor, not the mouse. So any binding that wants to branch on link presence can't do it directly in Lua.

**Concrete use cases** (these keep coming up across separate issues; listing them together since one root capability covers all of them):

1. **Ctrl+click: open the hyperlink if one is under the cursor, otherwise resolve the clicked word as a path / `file:line` / issue-ref and open that** — one chord that "opens whatever's here." Currently needs a fragile workaround (below).
2. **Open the semantic zone under the click in `$EDITOR`** (kitty's `mouse_map ctrl+right press ungrabbed mouse_show_command_output`). Needs the mouse cell to feed `pane:get_semantic_zone_at`, which already exists but takes a position. ([discussion #5836](https://github.com/wezterm/wezterm/discussions/5836))
3. **Copy the link under the cursor** instead of opening it. (#527)
4. **Run a custom command for a link matching a regexp** (e.g. open `file:line` from a traceback in an IDE). (#3791)
5. **Alternative actions on a hyperlink** (e.g. ctrl+click a link to copy it). (#4371)

**Current workaround (use case 1) and why it's fragile**

```lua
wezterm.action_callback(function(window, pane)
  link_opened = false
  window:perform_action(wezterm.action.OpenLinkAtMouseCursor, pane)
  if link_opened then return end            -- open-uri set the flag synchronously
  window:perform_action(fallback_action)    -- no link → resolve clicked word as a path
end)
```
where `wezterm.on('open-uri', ...)` sets `link_opened = true` when it sees a real URI. This works only because `open-uri` appears to fire *synchronously* inside `perform_action(OpenLinkAtMouseCursor)` — that ordering isn't documented, so it could silently break. It also can't distinguish "link present but I want to override it" from "no link," which is exactly what #4371 needs.

Two tiers; either would help, the first is minimal:

- **Minimal — a fallback on `OpenLinkAtMouseCursor`:** generalize the existing `CompleteSelectionOrOpenLinkAtMouseCursor` composition so `OpenLinkAtMouseCursor` accepts an optional fallback action, e.g. `wezterm.action.OpenLinkAtMouseCursor { fallback = <action> }`. Mouse bindings could then branch on link presence without the `open-uri` side channel. Covers 1, 3, and part of 5.
- **General — expose what's under the cursor to Lua:**
  - a `position` / `mouse_cell` field on the table returned by `window:current_event()` during mouse callbacks (the existing `pane:get_semantic_zone_at` then becomes usable from a click → use case 2 / [discussion #5836](https://github.com/wezterm/wezterm/discussions/5836)); and/or
  - `pane:get_hyperlink_at_mouse_cursor()` (or `get_hyperlink_at(x, y)`) returning the URI + which rule matched → covers 4 and 5.

**Related / how this differs from #4371**

#4371 asks for alternative actions *when a link is present*; this issue is the underlying capability that it (and the cases above) depend on — letting Lua know whether a link is under the cursor at all, and where the mouse is. Filed as a root primitive to tie the cluster together rather than another per-action request.

Also related: #7491 / #4536 (hyperlink open bindings under mouse reporting).

timotheecour:
Closes #7899.

## What

Lets `mouse_bindings` callbacks branch on what is under the cursor, and adds an optional fallback to `OpenLinkAtMouseCursor`. Implements the "general" + "minimal" tiers from the issue.

## Why

Today a mouse-binding callback can't ask "is there a link under the cursor, and where is the mouse?" `window:current_event()` returns only `{Down/Drag/Up={streak,button}}`, `pane:get_cursor_position()` is the *text* cursor, and the only signal about a hovered link is the synchronous-`open-uri` side channel inside `OpenLinkAtMouseCursor` — undocumented, and unable to distinguish "link present but I want to override" from "no link". This blocks ctrl+click "open whatever's here", semantic-zone-at-click, copy-the-link, custom regexp-link actions, and alternative link actions (use cases 1–5 in #7899).

## Changes

**Two new window methods** — the state already lives on `TermWindow`, so this only exposes it:

- `window:get_mouse_position()` → `{ x = column, y = stable_row } | nil`.
  `x`/`y` use the same convention as `pane:get_semantic_zone_at(x, y)`, so they compose directly:
  ```lua
  local pos = window:get_mouse_position()
  local zone = pane:get_semantic_zone_at(pos.x, pos.y)  -- use case 2
  ```
- `window:get_hyperlink_at_mouse_cursor()` → `uri | nil`.
  Reads the existing `current_highlight`, so it covers **both** OSC-8 hyperlinks and `hyperlink_rules` matches — exactly what `OpenLinkAtMouseCursor` opens. Use cases 1/3/4/5.

**Fallback action** (minimal tier):

- `OpenLinkAtMouseCursor` is now `OpenLinkAtMouseCursor { fallback = <action> }` (fallback optional). When no link is under the cursor, the fallback runs:
  ```lua
  action = wezterm.action.OpenLinkAtMouseCursor {
    fallback = wezterm.action_callback(function(window, pane)
      -- resolve the clicked word as a path / file:line / issue-ref
    end),
  }
  ```
- **Backward compatible**: `wezterm.action.OpenLinkAtMouseCursor` (no args) still works. The `Enum::__index` metatable in `luahelper/src/enumctor.rs` default-constructs the now-struct variant via the empty-table form `{OpenLinkAtMouseCursor={}}`; verified by a dedicated round-trip test. The default key/mouse bindings and the menu item are unchanged in behavior.

## Implementation notes

- The link-resolution block in `mouse_event_impl` runs unconditionally on every mouse event, so `current_highlight` and the tracked mouse cell are fresh at click time.
- A new global `TermWindow::current_mouse_position` field is set alongside the existing per-pane `mouse_terminal_coords`; the window methods read it via the existing `TermWindowNotif::Apply` channel (same pattern as `current_event`).

## Verification

- `cargo test -p config` — 11/11 pass, including 3 new round-trip tests for the variant (`OpenLinkAtMouseCursor{fallback:None}` round-trips; the empty-table no-arg form decodes to `fallback:None`; the fallback form decodes correctly).
- `cargo check` for `config`, `wezterm-gui`, and `wezterm` — clean.
- End-to-end: built `wezterm-gui` and launched it with a scratch config that force-evaluates both action forms under `pcall` + shape assertions — both `wezterm.action.OpenLinkAtMouseCursor` and `wezterm.action.OpenLinkAtMouseCursor{fallback=...}` construct correctly in the real Lua environment.

## Docs

Adds `docs/config/lua/window/get_mouse_position.md`, `docs/config/lua/window/get_hyperlink_at_mouse_cursor.md`, updates `docs/config/lua/keyassignment/OpenLinkAtMouseCursor.md`, and adds a changelog entry.

## Related issues

This provides the Lua primitives that unblock these (linked for cross-reference, not auto-closed — each closes only if a maintainer agrees it's resolved):

- Refs #527 — copy the link under the cursor (now: `get_hyperlink_at_mouse_cursor()` + `window:copy_to_clipboard()`).
- Refs #4371 — alternative actions on a present link (link presence is now queryable in Lua).
- Refs #3791 — custom command for a regexp-matched link (the URI is exposed, but not *which rule matched*).
- Refs discussion #5836 — open the semantic zone under the click (now: `get_mouse_position()` + existing `pane:get_semantic_zone_at(x, y)`).

Out of scope: #7491 / #4536 (hyperlink-open bindings being intercepted by mouse-reporting mode) — a different problem this PR doesn't address.

### DIFF
--- a/config/src/keyassignment.rs
+++ b/config/src/keyassignment.rs
@@ -587,7 +587,12 @@ pub enum KeyAssignment {
 
     SelectTextAtMouseCursor(SelectionMode),
     ExtendSelectionToMouseCursor(SelectionMode),
-    OpenLinkAtMouseCursor,
+    /// Open the hyperlink under the mouse cursor, if any.
+    /// When no link is present, optionally perform `fallback` instead.
+    OpenLinkAtMouseCursor {
+        #[dynamic(default)]
+        fallback: Option<Box<KeyAssignment>>,
+    },
     ClearSelection,
     CompleteSelection(ClipboardCopyDestination),
     CompleteSelectionOrOpenLinkAtMouseCursor(ClipboardCopyDestination),
@@ -733,4 +738,61 @@ pub struct KeyTables {
 #[derive(Debug, Clone, PartialEq)]
 pub struct KeyTableEntry {
     pub action: KeyAssignment,
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use wezterm_dynamic::Value;
+
+    fn obj(pairs: impl IntoIterator<Item = (Value, Value)>) -> Value {
+        Value::Object(pairs.into_iter().collect())
+    }
+
+    #[test]
+    fn open_link_at_mouse_cursor_round_trip_no_fallback() {
+        let original = KeyAssignment::OpenLinkAtMouseCursor { fallback: None };
+        let encoded = original.to_dynamic();
+        let decoded =
+            KeyAssignment::from_dynamic(&encoded, FromDynamicOptions::default()).unwrap();
+        assert_eq!(original, decoded);
+    }
+
+    /// `wezterm.action.OpenLinkAtMouseCursor` (no args) is produced by the
+    /// `Enum::__index` metatable as the empty-table form; it must decode back
+    /// to the variant with no fallback. This is the backward-compatibility
+    /// guarantee for existing configs.
+    #[test]
+    fn open_link_at_mouse_cursor_decodes_empty_table_form() {
+        let empty_table_form = obj(std::iter::once((
+            Value::String("OpenLinkAtMouseCursor".to_string()),
+            obj(std::iter::empty()),
+        )));
+        let decoded =
+            KeyAssignment::from_dynamic(&empty_table_form, FromDynamicOptions::default()).unwrap();
+        assert_eq!(
+            decoded,
+            KeyAssignment::OpenLinkAtMouseCursor { fallback: None }
+        );
+    }
+
+    #[test]
+    fn open_link_at_mouse_cursor_decodes_fallback() {
+        let with_fallback = obj(std::iter::once((
+            Value::String("OpenLinkAtMouseCursor".to_string()),
+            obj(std::iter::once((
+                Value::String("fallback".to_string()),
+                // A unit-variant KeyAssignment encodes as a bare string.
+                Value::String("ClearSelection".to_string()),
+            ))),
+        )));
+        let decoded =
+            KeyAssignment::from_dynamic(&with_fallback, FromDynamicOptions::default()).unwrap();
+        assert_eq!(
+            decoded,
+            KeyAssignment::OpenLinkAtMouseCursor {
+                fallback: Some(Box::new(KeyAssignment::ClearSelection)),
+            }
+        );
+    }
 }
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -75,6 +75,14 @@ As features stabilize some brief notes about them will accumulate here.
   easier to spot the remaining candidates. Thanks to @mr-felixoid and @bew! #7752
 
 #### New
+* Mouse bindings can now branch on what's under the cursor from Lua:
+  [window:get_mouse_position()](config/lua/window/get_mouse_position.md)
+  returns the cell column and stable row under the mouse, and
+  [window:get_hyperlink_at_mouse_cursor()](config/lua/window/get_hyperlink_at_mouse_cursor.md)
+  returns the URI of the hyperlink under the mouse (or `nil`).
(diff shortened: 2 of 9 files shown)


timotheecour:
friendly ping @wez

bew:
Hello! Thanks for your interest in improving this and for the PR!

How much AI was used for this contribution?

---

I'd be willing to accept the Lua part to get mouse coords in a pane, but I think the method should be on the Pane object, not on the Window object, to be more in-line with other pane related query methods like `get_semantic_zone_at` or `get_cursor_position`, and avoids confusions with potential window-level pixel coordinates.

For the `window:get_hyperlink_at_mouse_cursor` the same logic would apply, although I feel like this function is too narrow and could be generalized to a way to query information about a pane cell.
Doing this would need some careful design to avoid coupling too hard with how cells are implemented in Wezterm internally (only expose a stable subset, may be extended later on as more use-cases emerge).
👉 I think this should be split into a separate PR that focuses on this method only.

For the fallback on OpenLinkAtMouseCursor, I feel like this is too narrow and doesn't compose well..
I'd much prefer people to use custom the Lua API instead for this (for now at least).

---

👉 So please keep this PR about `pane:get_mouse_position` only, feel free to open a separate (draft to start?) PR for a method like `pane:get_cell_metadata_at`

PR #7915, opened 2026-07-10
fix: encode Composed keys properly in Kitty keyboard protocol

2bitbit:
# fix: encode Composed keys properly in Kitty keyboard protocol

## Summary

This PR fixes a bug where single-character IME inputs (e.g., CJK punctuation like `，` or single characters like `你`) are encoded as empty strings `""` when the Kitty keyboard protocol is active on Windows, making CJK typing impossible in terminal applications (such as [pi](https://github.com/earendil-works/pi)) that request this protocol.

## Cause of the Bug

When Windows IME commits a single character, WezTerm triggers a `KeyEvent` with `key` set to `KeyCode::Composed(String)`. 
Under the Kitty keyboard protocol, WezTerm routes this event to `encode_kitty(flags)`. However, `encode_kitty` fell back to the default `_` match branch, which returned an empty string `""` because the event had no physical raw key code (`self.raw` is `None`). 

This empty string `""` was then wrapped as `Some("")` by `encode_kitty_input` and returned to the caller in `keyevent.rs` (simplified for illustration):
```rust
} else if let Some(encoded) = self.encode_kitty_input(&pane, &window_key) {
    // We get here because `encoded` is `Some("")`.
    // It writes the empty string to the PTY, and halts processing.
    pane.writer().write_all(encoded.as_bytes()).ok();
} else {
    // The default fallback (which writes raw characters to PTY) is bypassed/short-circuited.
    pane.key_down(key, modifiers)
}
```
This short-circuits the default text-fallback path (`pane.key_down`), causing the committed character to be entirely swallowed.

## Proposed Fix

We add a dedicated `KeyCode::Composed(s)` branch inside the main `match &self.key` loop in `encode_kitty`:
1. **Legacy Fallback**: If no modifiers are present, `REPORT_ALL_KEYS_AS_ESCAPE_CODES` is not active, and the event does not require event-type formatting (e.g., not a key-release event), we directly return the raw Unicode string `s` as-is. This aligns perfectly with how simple `Char` inputs are encoded.
2. **CSI-u Formatting**: Otherwise, we format it into a proper CSI-u sequence `\x1b[{first};{modifiers}{event_type}{generated_text}u`.
   - The primary key code is the Unicode code point of the first character in the composed text.
   - The associated text (`generated_text`) contains the code points for all characters in the string, formatted as `;code1:code2`.

## How to Reproduce

You can reproduce this behavior using a clean setup on Windows 11 with Microsoft Pinyin IME:

1. Create a minimal WezTerm configuration `wezterm-debug-kitty.lua`:
   ```lua
   config.enable_kitty_keyboard = true
   config.debug_key_events = true
   return config
   ```
2. Start WezTerm using this config:
   ```powershell
   wezterm --config-file wezterm-debug-kitty.lua start
   ```
3. Inside the new WezTerm window, run the following standalone Node.js script to simulate a Kitty protocol handshake and capture inputs:
   ```javascript
   const stdin = process.stdin, stdout = process.stdout;
   if (!stdin.setRawMode) { console.error("raw mode not supported"); process.exit(1); }
   
   // Send Kitty keyboard protocol handshake sequences
   stdout.write("\x1b[>7u\x1b[?u\x1b[c");
   stdin.setRawMode(true); stdin.setEncoding("utf8"); stdin.resume();
   
   let gotHandshake = false;
   stdin.on("data", (data) => {
       const s = String(data);
       if (s === "\x03") { stdin.setRawMode(false); process.exit(0); }
       if (!gotHandshake) { gotHandshake = true; return; } // swallow handshake response
       const hex = [...Buffer.from(s, "utf8")].map(b => "\\x" + b.toString(16).padStart(2, "0")).join("");
       stdout.write(`[Received] ${hex}\r\n`);
   });
   ```
4. Switch to CJK IME, and press `,` once (committing `，`).
   - **Before this fix**: The script receives absolutely nothing for the key press (and WezTerm logs `kitty: Encoded input as ""`).
   - **After this fix**: The script correctly receives the UTF-8 bytes `\xef\xbc\x8c` of `，`.

## Testing & Verification

We added a robust test suite `encode_composed_kitty` in `wezterm-input-types/src/lib.rs` verifying 7 distinct Kitty protocol encoding scenarios for composed keys:
1. **Plain Input** (`flags = NONE`, `key_is_down = true`) → returns `"，"` (raw text).
2. **Plain Release** (`flags = NONE`, `key_is_down = false`) → returns `""`.
3. **All Keys as Escapes** (`flags = REPORT_ALL_KEYS_AS_ESCAPE_CODES`) → returns `\x1b[65292;1u`.
4. **Associated Text** (`flags = REPORT_ALL_KEYS_AS_ESCAPE_CODES | REPORT_ASSOCIATED_TEXT`) → returns `\x1b[65292;1;65292u`.
5. **Event-Type Release** (`flags = REPORT_EVENT_TYPES`, `key_is_down = false`) → returns `\x1b[65292;1:3u`.
6. **With Modifier** (`modifiers = CTRL`, `flags = NONE`) → returns `\x1b[65292;5u`.
7. **Typical Real-world Combo** (`flags = DISAMBIGUATE | REPORT_EVENT_TYPES | REPORT_ALTERNATE_KEYS`) → returns `"，"` (raw text).

All 14 tests in the `wezterm-input-types` package pass successfully.

---

## Proposed Changelog Entry

```markdown
- **Fixed**: Windows: Fixed single-character CJK/IME inputs (e.g., punctuation `，`) being swallowed when the Kitty keyboard protocol is active.
```

### DIFF
--- a/wezterm-input-types/src/lib.rs
+++ b/wezterm-input-types/src/lib.rs
@@ -2002,6 +2002,23 @@ impl KeyEvent {
                 format!("{intro};{modifiers}{event_type}{end_char}")
             }
 
+            Composed(s) => {
+                let use_legacy = self.modifiers.is_empty()
+                    && !flags.contains(KittyKeyboardFlags::REPORT_ALL_KEYS_AS_ESCAPE_CODES)
+                    && event_type.is_empty();
+
+                if use_legacy {
+                    s.clone()
+                } else {
+                    // Note: Multi-character IME composed strings are handled by a separate
+                    // text path in keyevent.rs (Key::Composed) and do not reach here.
+                    // This fallback handles single-character Composed keys that were mapped
+                    // to Key::Code, or any potential multichar keys if the layout changes.
+                    let first = s.chars().next().map(|c| c as u32).unwrap_or(0);
+                    format!("\x1b[{first};{modifiers}{event_type}{generated_text}u")
+                }
+            }
+
             _ => {
                 let code = self.raw.as_ref().and_then(|raw| raw.kitty_function_code());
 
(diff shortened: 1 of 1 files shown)


wezterm-input-types/src/lib.rs
```diff
@@ -2002,6 +2002,23 @@ impl KeyEvent {
...
+
+                if use_legacy {
+                    s.clone()
+                } else {
+                    // Note: Multi-character IME composed strings are handled by a separate
+                    // text path in keyevent.rs (Key::Composed) and do not reach here.
+                    // This fallback handles single-character Composed keys that were mapped
+                    // to Key::Code, or any potential multichar keys if the layout changes.
+                    let first = s.chars().next().map(|c| c as u32).unwrap_or(0);
+                    format!("\x1b[{first};{modifiers}{event_type}{generated_text}u")
```
bew, line 2018: I don't understand how this can be an expected behavior for a multi-chars IME composed string 🤔
This would only send the first char and dismiss the remaining chars of the composed text..

Do you have any reference for this?

2bitbit: Hi @bew,

Thanks for looking into this logic. Using `chars().next()` to extract the first character here is completely safe, and we don't need to worry about multi-character IME inputs being truncated.

This is based on the following two reasons:

### 1. Pre-routing and Event Dispatching at Code Level

In `win_key_code_to_termwiz_key_code` within [`wezterm-gui/src/termwindow/keyevent.rs`](https://github.com/wezterm/wezterm/blob/e22070a9b8bdfb1e8484e56598505c2196fb10b2/wezterm-gui/src/termwindow/keyevent.rs#L775-L784):

```rust
            WK::Composed(ref s) => {
                let mut chars = s.chars();
                if let Some(first_char) = chars.next() {
                    if chars.next().is_none() {
                        // Only single characters (e.g. "你", "，") are converted to a standard Key::Code(KeyCode::Char)
                        return self.win_key_code_to_termwiz_key_code(&WK::Char(first_char));
                    }
                }
                return Key::Composed(s.to_owned());
            }
```

Here is how the event routing details work:
* **Multi-character strings bypass Kitty encoder**: When the IME submits a multi-character string (e.g. `"你好"`), it matches the `Key::Composed(s)` variant, which writes the text directly to the PTY and completely bypasses the `encode_kitty` key event encoder.
* **Single-character strings route to key event cycle**: Only when the `Composed` string has a length of 1, it gets converted to `Key::Code(KeyCode::Char)` to route it into the standard key event handler (so that it can be captured and encoded under the Kitty keyboard protocol).
* **Original event data remains intact**: Because this routing translation only yields a local key routing variable `key`, the original `window_key` reference passed to `encode_kitty` remains unmodified. Thus, when `encode_kitty` matches on `self.key`, it still correctly hits the `KeyCode::Composed(s)` arm, where `s` is guaranteed to contain only a single character.

---

### 2. Kitty Keyboard Protocol (CSI u) Constraints

According to the official [Kitty Keyboard Protocol Specification](https://sw.kovidgoyal.net/kitty/keyboard-protocol/):
* **CSI u key_code only represents a single character**: In the protocol definition, the primary `key_code` in a `CSI u` sequence is a single decimal value representing a single Unicode code point. It cannot carry multiple Unicode code points.
* **Multi-character input relies on Associated Text**: When a key press or IME input yields multiple characters, the [Report associated text](https://sw.kovidgoyal.net/kitty/keyboard-protocol/#report-associated-text) specification states that the primary `key_code` must be set to the code point of the first character (or `0` if there is no mapping physical key), and the remaining characters must be appended to the end of the sequence separated by colons (e.g., `;code1:code2` if the `REPORT_ASSOCIATED_TEXT` flag/value 16 is enabled, which is exactly how `generated_text` is formatted for `Composed(s)` in our implementation).

Therefore, even if future architectural refactorings or layout changes cause a multi-character `Composed` string to flow into this branch, extracting the first character via `chars().next()` remains the only conformant and safe fallback strategy under the Kitty protocol, preventing invalid escape sequences.

---

### 3. Event Flow Chart

```mermaid
graph TD
    A["IME Submits Text"] -->|Multi-char e.g. '你好'| B["win_key_code_to_termwiz_key_code"]
    A -->|Single-char e.g. '你'| B
    
    B -->|Multi-char| C["Return Key::Composed"]
    B -->|Single-char| D["Return Key::Code"]
    
    C --> E["Key Event Loop"]
    D --> E
    
    E -->|Key::Composed Branch| F["Direct Write: Writes full text to PTY"]
    E -->|Key::Code Branch| G["Call encode_kitty_input"]
    
    G -->|Passed window_key is still KeyCode::Composed| H{"encode_kitty match arms"}
    H -->|Old code: matched '_' due to raw=None| I["Return empty string: swallows character"]
    H -->|New code: matches 'Composed' using chars.next| J["Return single character or CSI u sequence"]
```

### Conclusion
Multi-character inputs are written directly to PTY via the routing dispatch and bypass the encoder. Only single-character `Composed` events will ever reach `encode_kitty`.

Thus, using `s.chars().next()` here is purely for extracting the single generated character and will not result in any data truncation. In the meantime, it acts as a protocol-compliant fallback for the CSI u format.

---
*\*Note: This reply was co-authored with AI and manually reviewed by the author.\**

PR #7922, opened 2026-07-11
fix: emit text for Composed keys under kitty keyboard protocol

Zhou-jw:
When the kitty keyboard protocol is active, KeyCode::Composed values (eg. IME composition results such as full-width punctuation) were encoded to an empty string and silently dropped. This made IME input disappear in applications that enable the kitty keyboard protocol (eg. fish), while working fine in shells that don't (eg. bash).

Only substitute the composed text in the fallback branch that would otherwise return an empty string, so keys that have a real kitty encoding remain unaffected.


bew:
#7915 is a better impl for this, checking the Compose type at the `match self.key` branch above.

PR #7924, opened 2026-07-11
feat(term): support kitty graphics protocol unicode placeholders

### LINKED ISSUE #4531: Kitty Image Protocol on tmux

macOS

_No response_

### WezTerm version

20230712-072601-f4abf8fd

### Describe the bug

Recently Kitty image protocol can be used inside tmux and within Kitty terminal it seems to work properly (I just tried a handful of times). However in Wezterm it is not working, it prints a bunch of symbols instead of the image.

### To Reproduce

kitty +kitten icat image.jpg

### Configuration

no config

### Expected Behavior

Image shown in terminal window

### Logs

_No response_

### Anything else?

_No response_

### LINKED ISSUE #7807: Kitty Unicode Rendering Weird Behaviour

Linux Wayland

Cosmic

### WezTerm version

wezterm 0-unstable-2026-03-31

### Describe the bug

I am using <https://github.com/PartyWumpus/typst-concealer>, which works perfectly with kitty, but under wezterm the characters and laying out break.

Left Wezterm, Right kitty

Image

### To Reproduce

Use <https://github.com/PartyWumpus/typst-concealer>

And use this example for testing:
```typ
In fact $5+4=9$
and
$
  integral 0^5 5 = 25
$

but $5+5!= 25$

longer stuff:
$
  vec(a, b, c)
  +
  vec(d, e, f)
  =
  vec(a+d, b+e, c+f)
$
```

### Configuration

no config

### Expected Behavior

The Rendering should behave the same as Kitty.

### Logs

Debug Overlay
wezterm version: 0-unstable-2026-03-31 x86_64-unknown-linux-gnu
Window Environment: Wayland
Lua Version: Lua 5.4
OpenGL: AMD Radeon RX 6650 XT (radeonsi, navi23, ACO, DRM 3.64, 6.18.31) 4.6 (Compatibility Profile) Mesa 26.1.0
Press ESC or CTRL-D to exit
06:31:58.999 WARN wezterm_font > No fonts contain glyphs for these codepoints: \u{10eeee}.
Placeholder glyphs are being displayed instead.
You may wish to install additional fonts, or adjust your
configuration so that it can find them.
https://wezterm.org/config/fonts.html has more information about configuring fonts.
Set warn_about_missing_glyphs=false to suppress this message.
06:40:54.304 ERROR wezterm_mux_server_impl::local > writing pdu data buffer: Broken pipe (os error 32)

### Anything else?

-

sjuxax:
## Summary

Implements the [Unicode placeholders](https://sw.kovidgoyal.net/kitty/graphics-protocol/#unicode-placeholders) feature of the kitty graphics protocol: virtual placements (`U=1`) and `U+10EEEE` placeholder cells.

Client libraries that position images through the text grid rely on this scheme — [textual-image](https://github.com/lnqs/textual-image) (used by Textual TUI apps), [ratatui-image](https://github.com/benjajaja/ratatui-image), and [typst-concealer](https://github.com/PartyWumpus/typst-concealer) among others. They transmit the image, create a virtual placement, and then "draw" it by printing placeholder characters whose combining diacritics encode the row/column within the placement and whose foreground color carries the image id.

### Behavior before this change

wezterm ignored the `U` key on placements, with two visible failure modes:

* The virtual placement was treated as a regular placement and drawn **at the cursor position**, which under a full-screen TUI is essentially arbitrary — images appear scattered/staggered around the pane instead of where the application put the placeholder cells.
* The `U+10EEEE` placeholder text fell through to font rendering. No font covers Plane-16 PUA, so users get the "Font problem: No fonts contain glyphs for these codepoints: \u{10eeee}" toast and placeholder boxes.

### Behavior after this change

* `wezterm-escape-parser`: `KittyImagePlacement` gains `unicode_placeholder` (the `U` key), parsed and serialized like the other placement keys.
* `term`: a placement with `U=1` is recorded as a *virtual placement* (grid size and z-index); it does not touch the cell model and does not move the cursor. If `c=`/`r=` are omitted, the grid is derived from the image and cell pixel sizes, as for regular placements.
* At print time, a grapheme based on `U+10EEEE` is decoded — row and column from the first two diacritics (kitty's `rowcolumn-diacritics` table), the most significant image-id byte from the optional third, the image id low bits from the foreground color, and the optional placement id from the underline color. Omitted diacritics are inferred from the placeholder cell to the left, as kitty specifies. The referenced tile of the image is attached to the cell and the cell text is blanked, so the placeholder never reaches font rendering/fallback.
* Placeholder cells that reference an unknown image or placement pass through unchanged as regular text.
* Kitty delete operations (`a=d` by id, delete-all) and soft reset clean up virtual placements, detaching any placeholder cells' images from the model.

Placeholder resolution happens at print time rather than render time. This relies on the placement being created before the placeholder text is printed, which is how the client libraries above behave (they must, since the image data has to be transmitted first); cells printed before their placement exists remain text until the application redraws them.

## Tests

Four new tests in `term/src/test/image.rs` (alongside the existing kitty image tests):

* a virtual placement displays nothing at the cursor and does not move it
* placeholder cells resolve to the correct image tiles (texture coordinates) with blanked text
* omitted diacritics continue from the cell to the left
* a placeholder with no matching virtual placement passes through as text

`cargo test -p wezterm-term -p wezterm-escape-parser` passes (113 tests).

Also verified interactively against a real client: textual-image's TGP renderable in a live wezterm-gui build renders the image as a single correctly-positioned block, with no font warning and nothing drawn at the cursor.

Closes #7807


sjuxax:
Claude took it upon itself to push this PR. Nonetheless, this fixes the issue I was having and should at least serve as a reasonable point-of-reference. Let me know what else may be needed to push it through. Thanks!

cremarch:
Tested this PR with neovim + [folke/snacks.nvim](https://github.com/folke/snacks.nvim) image preview on macOS (Apple Silicon), both running directly and through tmux passthrough.

First of all — thank you for working on this; unicode placeholder support is exactly what nvim image plugins need to work reliably in wezterm.

**As written, the PR doesn't yet work for snacks.image**: the placeholder cells fall through to font rendering and show up as glyph boxes instead of image tiles.

### Root cause

- snacks.image creates the virtual placement **with** an explicit placement id: `a=p,U=1,i=<id>,p=<id>,C=1,c=…,r=…`, so it is stored under the key `(image_id, Some(p))`.
- However, nvim's TUI never emits SGR 58 (underline color) for cells without an underline attribute, so the placeholder cells encode **no** placement id (verified with `WEZTERM_LOG`: zero `UnderlineColor` sequences in the whole session).
- The lookup in `kitty_print_unicode_placeholder` only tries `(image_id, decoded_placement_id)` and then `(image_id, None)`, so in this combination it always misses and returns `false`.

Per the [kitty docs on unicode placeholders](https://sw.kovidgoyal.net/kitty/graphics-protocol/#unicode-placeholders), a placeholder that doesn't specify a placement id (or specifies id 0) may resolve to *any* virtual placement of that image — kitty implements this fallback, which is why the same client code works there. The existing tests in this PR only exercise `a=T,U=1` without `p=`, and clients like textual-image set the underline color explicitly, so this path wasn't covered.

### Patch

With the following patch on top of this branch, snacks.image works correctly — switching between images leaves no stale frames, both directly and under tmux. It also treats an underline color of 0 as "unspecified" per the spec, and adds a regression test mirroring the nvim/snacks flow (fails before, passes after; all 57 `wezterm-term` tests pass):

```diff
diff --git a/term/src/terminalstate/kitty.rs b/term/src/terminalstate/kitty.rs
index 89e2df65..209b6b55 100644
--- a/term/src/terminalstate/kitty.rs
+++ b/term/src/terminalstate/kitty.rs
@@ -326,14 +326,28 @@ impl TerminalState {
         if let Some(msb) = values[2] {
             image_id |= msb << 24;
         }
-        let placement_id = color_encoded_id(pen.underline_color());
+        // An underline color of zero means the same as not specifying one
+        let placement_id = color_encoded_id(pen.underline_color()).filter(|&id| id != 0);
 
         let vp = match self
             .kitty_img
             .virtual_placements
             .get(&(image_id, placement_id))
             .or_else(|| self.kitty_img.virtual_placements.get(&(image_id, None)))
-        {
+            .or_else(|| {
+                // A placeholder that doesn't encode a placement id refers
+                // to any virtual placement of the image, as in kitty
+                if placement_id.is_none() {
+                    self.kitty_img
+                        .virtual_placements
+                        .iter()
+                        .filter(|((id, _), _)| *id == image_id)
+                        .min_by_key(|((_, pid), _)| *pid)
+                        .map(|(_, vp)| vp)
+                } else {
+                    None
+                }
+            }) {
             Some(vp) => *vp,
             None => return false,
         };
diff --git a/term/src/test/image.rs b/term/src/test/image.rs
index 9b1e9b9f..32ee5406 100644
--- a/term/src/test/image.rs
+++ b/term/src/test/image.rs
@@ -172,3 +172,25 @@ fn kitty_unicode_placeholder_without_placement_is_text() {
     assert_eq!(cell.str(), "\u{10eeee}\u{305}\u{305}");
     assert_eq!(cell.attrs().images().is_none(), true);
 }
+
+/// Clients like nvim's snacks.image create the virtual placement with an
+/// explicit placement id (a=p,p=NN,U=1), but never emit the underline
+/// color, so their placeholder cells don't encode a placement id. As in
+/// kitty, such placeholders resolve to any virtual placement of the image.
+#[test]
+fn kitty_unicode_placeholder_without_placement_id_matches_any_placement() {
+    let mut term = TestTerm::new(3, 10, 0);
+
+    let seq = format!("\x1b_Ga=t,i=42,t=d,f=100;{}\x1b\\", TINY_PNG_BASE64);
+    term.print(seq.as_bytes());
+    term.print("\x1b_Ga=p,U=1,i=42,p=7,C=1,c=2,r=2\x1b\\");
+
+    term.print("\x1b[38;2;0;0;42m");
+    term.print("\u{10eeee}\u{305}\u{305}");
+
+    let lines = term.screen().visible_lines();
+    let cells: Vec<_> = lines[0].visible_cells().collect();
+    assert_eq!(cells[0].str(), " ");
+    let images = cells[0].attrs().images().expect("image tile attached");
+    assert_eq!(images[0].image_id(), Some(42));
+}
```

Feel free to fold this into the PR directly — or if it's easier, I'm happy to open it as a PR against your branch. Thanks again!

caoer:
Also tested this branch with neovim + snacks.image on macOS (Apple Silicon) through tmux passthrough, on top of @cremarch's lookup patch above (which I can confirm fixes the glyph-box fallthrough — thanks!).

With that in place, images render and position correctly, but two intermittent failure modes remain, both rooted in the same design point: **this PR resolves placeholder cells at print time** (the tile attaches to the cell when the placeholder grapheme is printed, using whatever virtual placement exists at that moment), whereas **kitty resolves placeholders at render time**. Under tmux passthrough the graphics APC and the placeholder text travel on independent paths, so ordering isn't guaranteed:

1. **Late arrival (persistent breakage):** when the placeholder cells reach the terminal before the image transmission / virtual placement (easy to hit with slow-to-render images, e.g. snacks.image's mermaid pipeline), the cells fall through to text and are never revisited — the image area stays blank/garbled until something forces a full repaint.
2. **Replaced placement (transient shift):** when the client re-creates a virtual placement while its placeholder cells are on screen (snacks does this when re-rendering at a new size), already-printed cells keep tiles cut from the *old* grid — the image renders shifted/mis-scaled for a frame or two until the next full repaint.

### Patch: retro-resolution

The patch below (on top of this branch + cremarch's patch) approximates kitty's render-time semantics without touching the render path: whenever an image's data arrives or one of its virtual placements is created/replaced, re-scan the visible screen for that image — raw placeholder text resolves as if just printed (the cell still carries the full grapheme + encoding attributes), and tiles bound to a replaced placement re-bind to the new grid (recovering each tile's row/col from its texture coords against the old placement).

Two regression tests cover the late-arrival and replaced-placement cases (both fail before, pass after); all 59 `wezterm-term` tests pass. Verified interactively with snacks.image (plain images + mermaid) under tmux: no more stale frames or transient offsets.

<details>
<summary>Diff (kitty.rs + tests, also folds in cremarch's lookup patch)</summary>

```diff
term/src/terminalstate/kitty.rs --- 1/5 --- Rust
 266             rows
 267         );
 268 
 269         let old_vp = self.kitty_img.virtual_placements.insert(
 270             (image_id, placement.placement_id),
 271             KittyVirtualPlacement {
 272                 cols,

term/src/terminalstate/kitty.rs --- 2/5 --- Rust
 281  281             cols,
 282  282             rows
 283  283         );
 ...  284         self.kitty_retro_resolve_placeholders(image_id, old_vp);
 284  285         Ok(())
 285  286     }
 286  287 

term/src/terminalstate/kitty.rs --- 3/5 --- Rust
 326  327         if let Some(msb) = values[2] {
 327  328             image_id |= msb << 24;
 328  329         }
 ...  330         // An underline color of zero means the same as not specifying one
 329  331         let placement_id = color_encoded_id(pen.underline_color()).filter(|&id| id != 0);
 330  332 
 331  333         let vp = match self
 332  334             .kitty_img
 333  335             .virtual_placements
 334  336             .get(&(image_id, placement_id))
 335  337             .or_else(|| self.kitty_img.virtual_placements.get(&(image_id, None)))
 ...  338             .or_else(|| {
 ...  339                 // A placeholder that doesn't encode a placement id refers
 ...  340                 // to any virtual placement of the image, as in kitty
 ...  341                 if placement_id.is_none() {
 ...  342                     self.kitty_img
 ...  343                         .virtual_placements
 ...  344                         .iter()
 ...  345                         .filter(|((id, _), _)| *id == image_id)
 ...  346                         .min_by_key(|((_, pid), _)| *pid)
 ...  347                         .map(|(_, vp)| vp)
 ...  348                 } else {
 ...  349                     None
 ...  350                 }
 336  351             }) {
 337  352             Some(vp) => *vp,
 338  353             None => return false,
 339  354         };

term/src/terminalstate/kitty.rs --- 4/5 --- Rust
 405  420         Some((row, col))
 406  421     }
 ...  422 
 ...  423     /// Re-resolve placeholder cells for `image_id` on the visible screen.
 ...  424     ///
 ...  425     /// Placeholder cells resolve at print time, but the graphics APC and
 ...  426     /// the placeholder text can arrive in either order (tmux passthrough
 ...  427     /// delivers them on independent paths), and a client may replace a
 ...  428     /// virtual placement while its placeholder cells are on screen. Kitty
 ...  429     /// resolves placeholders at render time so neither case has a visible
 ...  430     /// window; approximate that by re-scanning the screen when the
 ...  431     /// placement side of an image changes: raw placeholder text resolves
 ...  432     /// as if just printed, and tiles attached against `old_vp` re-attach
 ...  433     /// against the current placement.
 ...  434     fn kitty_retro_resolve_placeholders(
 ...  435         &mut self,
 ...  436         image_id: u32,
 ...  437         old_vp: Option<KittyVirtualPlacement>,
 ...  438     ) {
 ...  439         if !self.config.enable_kitty_graphics() {
 ...  440             return;
 ...  441         }
 ...  442         enum Retry {
 ...  443             Raw(String, CellAttributes),
 ...  444             Stale {
 ...  445                 row: u32,
 ...  446                 col: u32,
 ...  447                 placement_id: Option<u32>,
 ...  448                 pen: CellAttributes,
 ...  449             },
 ...  450         }
 ...  451         let seqno = self.seqno;
 ...  452         let rows = self.screen().physical_rows;
 ...  453         let cols = self.screen().physical_cols;
 ...  454         for y in 0..rows {
 ...  455             let y = y as VisibleRowIndex;
 ...  456             for x in 0..cols {
 ...  457                 let retry = {
 ...  458                     let cell = match self.screen_mut().get_cell(x, y) {
 ...  459                         Some(cell) => cell,
 ...  460                         None => continue,
 ...  461                     };
 ...  462                     let g = cell.str();
 ...  463                     if g.starts_with(KITTY_UNICODE_PLACEHOLDER) {
 ...  464                         Retry::Raw(g.to_string(), cell.attrs().clone())
 ...  465                     } else if let (Some(old), Some(images)) = (old_vp, cell.attrs().images()) {
 ...  466                         match images.iter().find(|i| i.image_id() == Some(image_id)) {
 ...  467                             Some(tile) => {
 ...  468                                 let row = (tile.top_left().y.into_inner() * old.rows as f32)
 ...  469                                     .round() as u32;
 ...  470                                 let col = (tile.top_left().x.into_inner() * old.cols as f32)
 ...  471                                     .round() as u32;
 ...  472                                 let mut pen = cell.attrs().clone();
 ...  473                                 pen.clear_images();
 ...  474                                 Retry::Stale {
 ...  475                                     row,
 ...  476                                     col,
 ...  477                                     placement_id: tile.placement_id(),
 ...  478                                     pen,
 ...  479                                 }
 ...  480                             }
 ...  481                             None => continue,
 ...  482                         }
 ...  483                     } else {
 ...  484                         continue;
 ...  485                     }
 ...  486                 };
 ...  487                 match retry {
 ...  488                     Retry::Raw(g, pen) => {
 ...  489                         self.kitty_print_unicode_placeholder(x, y, &g, &pen, seqno);
 ...  490                     }
 ...  491                     Retry::Stale {
 ...  492                         row,
 ...  493                         col,
 ...  494                         placement_id,
 ...  495                         pen,
 ...  496                     } => {
 ...  497                         let vp = match self
 ...  498                             .kitty_img
 ...  499                             .virtual_placements
 ...  500                             .get(&(image_id, placement_id))
 ...  501                             .or_else(|| self.kitty_img.virtual_placements.get(&(image_id, None)))
 ...  502                         {
 ...  503                             Some(vp) => *vp,
 ...  504                             None => continue,
 ...  505                         };
 ...  506                         let img = match self.kitty_img.id_to_data.get(&image_id) {
 ...  507                             Some(img) => Arc::clone(img),
 ...  508                             None => continue,
 ...  509                         };
 ...  510                         let mut cell = Cell::new(' ', pen);
 ...  511                         if (row as usize) < vp.rows && (col as usize) < vp.cols {
 ...  512                             let cols_f = vp.cols as f32;
 ...  513                             let rows_f = vp.rows as f32;
 ...  514                             cell.attrs_mut().attach_image(Box::new(ImageCell::with_z_index(
 ...  515                                 TextureCoordinate::new_f32(
 ...  516                                     col as f32 / cols_f,
 ...  517                                     row as f32 / rows_f,
 ...  518                                 ),
 ...  519                                 TextureCoordinate::new_f32(
 ...  520                                     (col + 1) as f32 / cols_f,
 ...  521                                     (row + 1) as f32 / rows_f,
 ...  522                                 ),
 ...  523                                 img,
 ...  524                                 vp.z_index,
 ...  525                                 0,
 ...  526                                 0,
 ...  527                                 0,
 ...  528                                 0,
 ...  529                                 Some(image_id),
 ...  530                                 placement_id,
 ...  531                             )));
 ...  532                         }
 ...  533                         self.screen_mut().set_cell(x, y, &cell, seqno);
 ...  534                     }
 ...  535                 }
 ...  536             }
 ...  537         }
 ...  538     }
 407  539 
 408  540     fn kitty_img_inner(&mut self, img: KittyImage) -> anyhow::Result<()> {
 409  541         match self

term/src/terminalstate/kitty.rs --- 5/5 --- Rust
1133 1265             .raw_image_to_image_data(img)
1134 1266             .context("storing image data")?;
1135 1267         self.kitty_img.record_id_to_data(image_id, img);
.... 1268         self.kitty_retro_resolve_placeholders(image_id, None);
1136 1269 
1137 1270         if image_number.is_some() {
1138 1271             self.kitty_send_response(

term/src/test/image.rs --- Rust
173 173     assert_eq!(cell.attrs().images().is_none(), true);
174 174 }
    175 
    176 /// Clients like nvim's snacks.image create the virtual placement with an
    177 /// explicit placement id (a=p,p=NN,U=1), but never emit the underline
    178 /// color, so their placeholder cells don't encode a placement id. As in
    179 /// kitty, such placeholders resolve to any virtual placement of the image.
    180 #[test]
    181 fn kitty_unicode_placeholder_without_placement_id_matches_any_placement() {
    182     let mut term = TestTerm::new(3, 10, 0);
    183 
    184     let seq = format!("\x1b_Ga=t,i=42,t=d,f=100;{}\x1b\\", TINY_PNG_BASE64);
    185     term.print(seq.as_bytes());
    186     term.print("\x1b_Ga=p,U=1,i=42,p=7,C=1,c=2,r=2\x1b\\");
    187 
    188     term.print("\x1b[38;2;0;0;42m");
    189     term.print("\u{10eeee}\u{305}\u{305}");
    190 
    191     let lines = term.screen().visible_lines();
    192     let cells: Vec<_> = lines[0].visible_cells().collect();
    193     assert_eq!(cells[0].str(), " ");
    194     let images = cells[0].attrs().images().expect("image tile attached");
    195     assert_eq!(images[0].image_id(), Some(42));
    196 }
    197 
    198 /// Under tmux passthrough the graphics APC and the placeholder text
    199 /// travel on independent paths and can arrive in either order. Cells
    200 /// printed before the image/placement exist must resolve retroactively
    201 /// once both arrive.
    202 #[test]
    203 fn kitty_unicode_placeholder_late_placement_retro_resolves() {
    204     let mut term = TestTerm::new(3, 10, 0);
    205 
    206     // placeholder text lands first: stays raw text
    207     term.print("\x1b[38;2;0;0;42m");
    208     term.print("\u{10eeee}\u{305}\u{305}");
    209     {
    210         let lines = term.screen().visible_lines();
    211         let cell = lines[0].visible_cells().next().unwrap();
    212         assert_eq!(cell.attrs().images().is_none(), true);
    213     }
    214 
    215     // image data + virtual placement arrive afterwards
    216     let seq = format!("\x1b_Ga=t,i=42,t=d,f=100;{}\x1b\\", TINY_PNG_BASE64);
    217     term.print(seq.as_bytes());
    218     term.print("\x1b_Ga=p,U=1,i=42,C=1,c=2,r=2\x1b\\");
    219 
    220     let lines = term.screen().visible_lines();
    221     let cells: Vec<_> = lines[0].visible_cells().collect();
    222     let images = cells[0].attrs().images().expect("retro-resolved tile");
    223     assert_eq!(images[0].image_id(), Some(42));
    224 }
    225 
    226 /// A client that replaces a virtual placement (e.g. re-rendering at a new
    227 /// size) must see existing placeholder cells re-bind to the new grid, not
    228 /// keep tiles cut from the old one.
    229 #[test]
    230 fn kitty_unicode_placeholder_replaced_placement_rebinds_tiles() {
    231     let mut term = TestTerm::new(3, 10, 0);
    232 
    233     let seq = format!("\x1b_Ga=t,i=42,t=d,f=100;{}\x1b\\", TINY_PNG_BASE64);
    234     term.print(seq.as_bytes());
    235     term.print("\x1b_Ga=p,U=1,i=42,C=1,c=2,r=2\x1b\\");
    236     term.print("\x1b[38;2;0;0;42m");
    237     term.print("\u{10eeee}\u{305}\u{305}");
    238     term.print("\u{10eeee}\u{305}\u{30d}");
    239     {
    240         let lines = term.screen().visible_lines();
    241         let cells: Vec<_> = lines[0].visible_cells().collect();
    242         let tile = &cells[1].attrs().images().unwrap()[0];
    243         assert_eq!(tile.top_left().x.into_inner(), 0.5);
    244     }
    245 
    246     // same placement key, wider grid: tile (0,1) is now a quarter column
    247     term.print("\x1b_Ga=p,U=1,i=42,C=1,c=4,r=4\x1b\\");
    248     let lines = term.screen().visible_lines();
    249     let cells: Vec<_> = lines[0].visible_cells().collect();
    250     let tile = &cells[1].attrs().images().unwrap()[0];
    251     assert_eq!(tile.top_left().x.into_inner(), 0.25);
    252     assert_eq!(tile.top_left().y.into_inner(), 0.0);
    253 }

```

</details>

Happy to open this as a PR against the branch if that's easier.

cremarch:
Thanks @caoer for the sharp analysis and the retro-resolution patch — and for confirming my lookup fix along the way!

Verified the patch on top of this branch + my earlier lookup patch, same setup as before (macOS Apple Silicon, neovim + snacks.image, under tmux):

- All 59 `wezterm-term` tests pass, including the two new regression tests.
- Interactively: image switching works as before, and resizing the preview while an image is displayed (`:vsplit` / `:only`, which makes snacks re-create the virtual placement at the new size) now re-binds cleanly — no shifted or mis-scaled tiles.

The print-time vs render-time analysis makes sense to me, and retro-resolving on the placement/data side keeps the render path untouched — seems like the right trade-off for this PR.

@sjuxax — with these two patches folded in, the branch works reliably for nvim + snacks.image under tmux. We're ready to open a PR against your branch whenever that's helpful — or feel free to fold the diffs in directly.

sjuxax:
Thanks @cremarch , I've added your patch to this PR.

bew:
Hello!

Did you take inspiration from https://github.com/sergei-grechanik/wezterm/commit/d3df8e5e608df7bb9261f95e8a8f01b06a7a32b4 a prototype by @sergei-grechanik, the original author of the unicode placeholder feature on Kitty?
I'd be interested in how your impl compares to his 🤔

I don't have much review time at the moment for non-trivial features (though quite important I concede!) at the moment, but other users are encouraged to try and test it and report issues so it get better and better until this can get reviewed 🙃

---

Also, could you add an AI-disclosure mentioning:
How much AI was used for this contribution?
How much domain knowledge you have on the topic, and what / how much manual review/iteration/testing did you do? 🤔

paranoidi:
I have been running LLM adapted implementation from https://github.com/sergei-grechanik/wezterm/commit/d3df8e5e608df7bb9261f95e8a8f01b06a7a32b4 for a week now and it seems to be working quite well with my graphics enabled file manager. Most notably it completely solves the intended issue of placing images at correct location within tmux.

That code is available at https://github.com/paranoidi/wezterm/tree/kitty-img-placeholder

AI Disclaimer: I have zero rust knowledge but since LLM is used only to port sergeis POF code to current main branch and fix issues identified while doing so it should not be taken too negatively.

Should I consider opening PR for easier reviewing / discovery? I did ask sergei if he's ok with this and got a green light.

Alternatively, anyone can freely pick commits from my repo.

bew:
> Should I consider opening PR for easier reviewing / discovery?

If you cannot explain the changes yourself / answer potentially tricky questions about it, it's best to not open a PR or you'd become accountable for the change.
Posting a comment about it is a good move.

Now if @sergei-grechanik is interested into taking some of your changes (or re-making them) & opening the PR, he'd be more than welcome as he's the actual author 😀

paranoidi:
I don't think we need two competing PRs. I would have tested this one first if I had found it sooner. I have switched to running this PR and I'm happy to report there are zero issues. 

In fact this seems to work better than the one I ported. It rarely failed to render image but I could not reliably determine where exactly the fault was (my app, tmux or wezterm). The issue existed even with different terminals and file managers making pinpointing the issue extremely tricky. Terminal graphics are quite glitchy across the board.

This PR gets big thumbs up from here.

lzrd:
Verified on Fedora 43.

Rebased onto main at 4fbd6b8e9.
Conflicts were both adding to `docs/changelog.md` and new tests in `term/src/test/image.rs`. Kept both.

`cargo test -p wezterm-term` passes 58 tests.

My case of running mermaid rendering in nvim under tmux and just nvim both work now.

Previously, mermaid renderings are placed at the outer terminal's origin and can persist through tmux window changes.
Applying this PR, everything works properly.

`snacks.nvim` marks [WezTerm as missing placeholder support](https://github.com/folke/snacks.nvim/blob/882c996cf28183f4d63640de0b4c02ec886d01f2/docs/image.md?plain=1#L19) and falls back to cursor positioning. That makes it unusable under tmux.

I did not do code review apart from taking a last look after the rebasing as a sanity check.
Superficially, it looks good to me.

theeternalsw0rd:
Images have worked great for me on macOS, Windows, and wayland within superfile which has mistakenly hardcoded WezTerm as already supporting this feature of the protocol. Has also worked fine when run in ssh both with and without herdr as well as herdr remote.

swebber-upngo:
I think there is a bug here. I'm working on an app (toy/self learning project) that outputs kitty images. In this case, the images i'm transmitting are squares. In this screenshot you can see this PR on the left vs Ghostty on the right.  This PR stretches the image to fit the cell space, whereas Ghostty centers the scaled image to maintain the aspect ratio.

To my understanding, the Ghostty approach is more correct. The kitty docs say:
>The image will eventually be fit to the specified rectangle, its aspect ratio preserved. 

image

BTW, thank you for working on this. I'm very grateful to see WezTerm progress again.

PR #7929, opened 2026-07-13
mux: answer capability queries and bound the hold while a synchronized update is open

### LINKED ISSUE #7918: Queries are not answered while a synchronized update (mode 2026) is open

macOS

_No response_

### WezTerm version

wezterm 20260707-093716-fff02ca5

### Describe the bug

I was trying to figure out why a TUI app takes 2 seconds to show its first frame in WezTerm when the same binary starts instantly under tmux. I put a [pty relay](https://gist.github.com/luizribeiro/daac63943e0d87fa2737a0a84477fcdb) between the app and WezTerm to log both directions, and it turns out WezTerm sits on query responses while a synchronized update is open (timestamps in seconds):

```
0.0119  APP>TERM  ESC[?25l ESC[?2026h            ; app begins a synchronized update
0.0123  APP>TERM  ESC[?u ESC[c                   ; kitty keyboard probe + DA1
2.0137  APP>TERM  <frame> ... ESC[?2026l         ; app gives up after a 2s timeout, closes the update
2.0139  TERM>APP  ESC[?1u ESC[?65;4;6;18;22;52c  ; response arrives 0.2ms after ESU
```

The app opens a synchronized update, probes for kitty keyboard support (crossterm's `supports_keyboard_enhancement()`, a blocking round-trip with a 2s timeout), and waits for the answer. WezTerm waits for the update to close before answering. Nobody moves for 2 seconds until crossterm gives up. The answer then arrives 0.2ms after the ESU, which is also too late in a second way: the app has already concluded the kitty keyboard protocol is unsupported, on a terminal that supports it. The same two queries sent from a quiet prompt come back in about 4ms.

I went looking for where this happens. In `mux/src/lib.rs`, `parse_buffered_data` (around line 140), `?2026h` sets `hold = true`, and from then on every parsed action piles up in a buffer that the `!hold` gate keeps away from the terminal model. That includes `Device::RequestPrimaryDeviceAttributes` and `Keyboard::QueryKittySupport`. Responses are only written when the performer applies those actions, so nothing can be answered until `?2026l` or DECSTR flushes the buffer.

One more thing I noticed while in there: `hold` has no deadline. It's only cleared by ESU, soft reset, or pane EOF, so a client that opens an update and never closes it wedges the pane's model application for as long as it lives. In my trace it was the client's own 2s timeout doing the un-sticking, not WezTerm.

### To Reproduce

Run any iocraft 0.8.3 app in WezTerm and watch it pause for 2 seconds before the first frame (that framework happens to probe inside its first synchronized update). Or do it by hand: with the tty in raw mode, write `ESC[?2026h ESC[?u ESC[c` and wait. No response comes until you close the update.

### Configuration

The relevant part (the rest is fonts, colors, and keybindings):

```lua
enable_kitty_keyboard = true,
front_end = "WebGpu",
```

`enable_kitty_keyboard = true` is why the terminal answers `CSI ?u` at all here, but the hold happens in the mux action dispatch, so I don't think it's config dependent.

### Expected Behavior

I'd expect queries to be answered while the update is open. The de facto mode 2026 spec ([Terminal Spec: Synchronized Output](https://gist.github.com/christianparpart/d8a62cc1ab659194337d73e399004036)) only talks about freezing rendering: "When the synchronization mode is enabled following render calls will keep rendering the last rendered state," while "the terminal Emulator keeps processing incoming text and sequences." Query responses aren't screen content, so answering them mid-update can't cause tearing. #882 describes the feature the same way, batching rendering until reset flushes the queued screen data.

Holding the answers also trips up the [kitty keyboard protocol's detection procedure](https://sw.kovidgoyal.net/kitty/keyboard-protocol/#detection-of-support-for-this-protocol): "If an answer for the device attributes is received without getting back an answer for the progressive enhancement the terminal does not support this protocol." A terminal that has the protocol but holds its answer until ESU reads as "unsupported" to any client that probes inside a frame.

I got curious what other emulators do here:

- kitty: the parser keeps interpreting everything; 2026 only snapshots render state (`screen_pause_rendering` in screen.c), and the query handlers write immediately with no check of the paused state. Stale updates expire after 2s.
- Ghostty: stream processing is untouched; the renderer just skips frames while the mode is set, and a 1s timer resets abandoned updates. Queries answered immediately.
- Alacritty: buffers raw bytes during an update like WezTerm holds actions, but vte caps it with a 150ms timeout and a 2MiB buffer limit, so the worst case is an answer 150ms late.
- tmux: answers queries regardless of synchronized-update state.

### Logs

The relay trace above is the interesting part; I can attach the full log if useful.

### Anything else?

I believe two changes could fix this. The first one is letting queries whose answers don't depend on pending screen state (DA1/DA2/DA3, XTVERSION, kitty `?u`, DECRQM) through the hold gate as their own single-action batch instead of parking them in the buffer. CPR/DSR-6 would need to stay held, since their answers do depend on the buffered updates being applied. The second one is giving `hold` a deadline like the other emulators have, so a stalled client can't freeze a pane forever. Happy to attempt a PR for either if the approach sounds reasonable.

For context, this has been confusing people downstream: ccbrown/iocraft#125 and ccbrown/iocraft#74 concluded that WezTerm doesn't support probing for keyboard enhancement. It does answer the probe, just not while an update is open.

luizribeiro:
Fixes #7918.

This implements the two changes discussed in the issue:

1. **Capability queries are answered while a synchronized update is open.** In `parse_buffered_data`, queries whose answers reflect terminal capabilities rather than screen state (DA1/DA2/DA3, XTVERSION, DSR `CSI 5n`, and the kitty keyboard query `CSI ?u`) are forwarded to the terminal as their own batch instead of being parked in the hold buffer. Everything whose meaning depends on the held actions stays held: CPR (the answer is the cursor position), DECRQM (the answer is mode state a held action may change; the issue originally proposed passing it through, but the terminal-side handler for querying mode 2026 itself relies on the hold intercepting it, so blocking on DECRQM now resolves via the timeout below instead), and the kitty state mutations (the keyboard stacks live on the screens themselves, so a held screen switch could route an eagerly-applied mutation to the wrong screen's stack).

2. **The hold now has a deadline.** A synchronized update that stays open longer than the new `synchronized_output_timeout_ms` option (default 1000ms) is released: the buffered output is applied and the eventual `CSI ?2026l` becomes a no-op. This is the same protection kitty (2s) and alacritty (150ms) have, and it means a client that crashes or stalls mid-update can no longer freeze its pane forever. The deadline is anchored to each update's opening sequence and cleared when it closes, so back-to-back updates can't inherit a stale deadline; re-sending BSU while an update is open is idempotent (it previously flushed the partially-held frame). While waiting out the deadline, an interrupted poll is retried and other poll errors fail open by releasing the hold. The timeout only kicks in while a hold is actually open, so it costs nothing on the normal path.

I went with a config option rather than a hardcoded value since the terminals that bound this disagree by an order of magnitude on what the bound should be; 1000ms sits in the middle and is generous enough that a legitimate frame should never hit it.

To make this testable I added a small test suite around `parse_buffered_data` (the first one, as far as I can tell): a new `mux/src/test/` module mirroring the layout of `term/src/test/`, with a recording `Pane` impl fed through a socketpair in `mod.rs` and the tests in `sync_update.rs`. It covers: queries passing through an open update as their own single-action batch while `Print` output stays held; CPR, DECRQM, and kitty mutations staying held until the update closes; repeated BSU not leaking the held frame; a back-to-back update not inheriting the previous update's deadline (using a passed-through DA1 as a synchronization barrier so the test can't race the parser); the timeout expiring a stalled update; a late ESU after expiry being a harmless no-op; and the timeout clamp against extreme configured values.

Verified end to end by running an iocraft 0.8.3 binary against a build of this branch, with the pty relay from the issue in between. That version of iocraft probes kitty keyboard support inside its first synchronized update, which is the pathological case from the issue. Before, the probe's answers were deferred until the update closed and the app ate crossterm's full 2 second query timeout before its first frame. Now the probe is answered while the update is still open and the first frame goes out a few ms after spawn:

```
0.0235 APP>TERM b'\x1b[?25l\x1b[?1049h\x1b[?2026h'
0.0378 APP>TERM b'\x1b[?u\x1b[c'
0.0482 TERM>APP b'\x1b[?65;4;6;18;22;52c'
0.0483 APP>TERM ... (first frame follows immediately)
```

(That run was against wezterm-mux-server, which shares the code path under change; the `?u` query itself gets no reply because `enable_kitty_keyboard` defaults to off, which is why crossterm pairs it with DA1 as the terminator.)

### DIFF
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -411,6 +411,13 @@ pub struct Config {
     #[dynamic(default = "default_mux_output_parser_coalesce_delay_ms")]
     pub mux_output_parser_coalesce_delay_ms: u64,
 
+    /// How many ms a synchronized update (DEC private mode 2026) may
+    /// hold back output before wezterm stops waiting for the closing
+    /// sequence and applies the buffered output anyway, so that a
+    /// stalled application cannot freeze the pane indefinitely
+    #[dynamic(default = "default_mux_synchronized_output_timeout_ms")]
+    pub mux_synchronized_output_timeout_ms: u64,
+
     #[dynamic(default = "default_mux_env_remove")]
     pub mux_env_remove: Vec<String>,
 
@@ -1661,6 +1668,10 @@ fn default_swallow_mouse_click_on_window_focus() -> bool {
 
 fn default_mux_output_parser_coalesce_delay_ms() -> u64 {
     3
+}
+
+fn default_mux_synchronized_output_timeout_ms() -> u64 {
+    1000
 }
 
 fn default_mux_output_parser_buffer_size() -> usize {
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -276,6 +276,11 @@ As features stabilize some brief notes about them will accumulate here.
   a zero-sized placement (e.g. `w=0`/`h=0`), or displaying a cell-sized image on a pane
   whose pty reported no pixel dimensions (e.g. in `tmux -CC` domain).
   Such images are now refused instead of taking down the pane. Thanks to @zakrad! #6344
+* Capability queries (DA1, the kitty keyboard probe, etc.) are now answered
+  while a synchronized update is open, and an update held open for too long
+  now times out; see
+  [mux_synchronized_output_timeout_ms](config/lua/config/mux_synchronized_output_timeout_ms.md).
+  Thanks to @luizribeiro! #7918
 
 #### Updated
 * Bundled conpty.dll and OpenConsole.exe to build 1.22.250204002.nupkg
--- a/mux/src/lib.rs
+++ b/mux/src/lib.rs
@@ -7,7 +7,9 @@ use anyhow::{anyhow, Context, Error};
 use config::keyassignment::SpawnTabDomain;
 use config::{configuration, ExitBehavior, GuiPosition};
 use domain::{Domain, DomainId, DomainState, SplitSource};
-use filedescriptor::{poll, pollfd, socketpair, AsRawSocketDescriptor, FileDescriptor, POLLIN};
+use filedescriptor::{
+    poll, pollfd, socketpair, AsRawSocketDescriptor, Error as FdError, FileDescriptor, POLLIN,
+};
 #[cfg(unix)]
 use libc::{c_int, SOL_SOCKET, SO_RCVBUF, SO_SNDBUF};
 use log::error;
@@ -167,6 +169,18 @@ fn is_passthrough_query(action: &Action) -> bool {
     }
 }
 
+/// The poll timeout is a c_int of milliseconds on the platforms this
+/// runs on, and the deadline is Instant arithmetic; clamp the configured
+/// value (to ~24.8 days) so that an absurdly large setting can't wrap
+/// into a negative poll timeout or overflow the deadline
+fn hold_timeout_from(config: &config::ConfigHandle) -> Duration {
+    Duration::from_millis(
+        config
+            .mux_synchronized_output_timeout_ms
+            .min(i32::MAX as u64),
+    )
+}
+
 fn parse_buffered_data(pane: Weak<dyn Pane>, dead: &Arc<AtomicBool>, mut rx: FileDescriptor) {
     let mut buf = vec![0; configuration().mux_output_parser_buffer_size];
     let mut parser = termwiz::escape::parser::Parser::new();
@@ -175,8 +189,62 @@ fn parse_buffered_data(pane: Weak<dyn Pane>, dead: &Arc<AtomicBool>, mut rx: Fil
     let mut action_size = 0;
     let mut delay = Duration::from_millis(configuration().mux_output_parser_coalesce_delay_ms);
     let mut deadline = None;
+    let mut hold_timeout = hold_timeout_from(&configuration());
+    let mut hold_deadline: Option<Instant> = None;
 
     loop {
+        if hold {
+            let target = *hold_deadline.get_or_insert_with(|| Instant::now() + hold_timeout);
+            let expired = match target.checked_duration_since(Instant::now()) {
+                Some(remaining) => {
+                    let mut pfd = [pollfd {
+                        fd: rx.as_socket_descriptor(),
+                        events: POLLIN,
+                        revents: 0,
+                    }];
+                    match poll(&mut pfd, Some(remaining)) {
+                        Ok(0) => true,
+                        Ok(_) => false,
+                        // Interrupted before the deadline; recompute the
+                        // remaining time and poll again. The error variant
+                        // depends on the backend: poll(2) reports
+                        // FdError::Poll, while the macOS select(2) shim and
+                        // Windows WSAPoll report FdError::Io
+                        Err(FdError::Poll(err) | FdError::Io(err))
+                            if err.kind() == std::io::ErrorKind::Interrupted =>
+                        {
+                            continue;
+                        }
+                        Err(err) => {
+                            // Polling is what enforces the deadline, so if
+                            // it is broken the only way to keep the promise
+                            // that a stalled client cannot freeze the pane
+                            // is to fail open and release the hold
+                            log::error!(
+                                "error polling for pane output while a \
+                                 synchronized update is open: {err:#}; \
+                                 releasing the hold"
+                            );
+                            true
+                        }
+                    }
+                }
+                None => true,
+            };
+            if expired {
+                // The synchronized update has been open for longer than
+                // the configured timeout; release the hold so that a
+                // stalled application cannot freeze the pane indefinitely
+                hold = false;
+                hold_deadline = None;
+                if !actions.is_empty() {
+                    send_actions_to_mux(&pane, &dead, std::mem::take(&mut actions));
+                    action_size = 0;
+                }
+                continue;
+            }
+        }
+
         match rx.read(&mut buf) {
             Ok(size) if size == 0 => {
                 dead.store(true, Ordering::Relaxed);
@@ -197,22 +265,30 @@ fn parse_buffered_data(pane: Weak<dyn Pane>, dead: &Arc<AtomicBool>, mut rx: Fil
                         Action::CSI(CSI::Mode(Mode::SetDecPrivateMode(DecPrivateMode::Code(
                             DecPrivateModeCode::SynchronizedOutput,
                         )))) => {
-                            hold = true;
+                            // Setting the mode while it is already active is
+                            // idempotent: the update keeps its original
+                            // deadline and the frame held so far stays held
+                            if !hold {
+                                hold = true;
+                                hold_deadline = Some(Instant::now() + hold_timeout);
 
-                            // Flush prior actions
-                            if !actions.is_empty() {
-                                send_actions_to_mux(&pane, &dead, std::mem::take(&mut actions));
-                                action_size = 0;
+                                // Flush prior actions
+                                if !actions.is_empty() {
+                                    send_actions_to_mux(&pane, &dead, std::mem::take(&mut actions));
+                                    action_size = 0;
+                                }
                             }
                         }
                         Action::CSI(CSI::Mode(Mode::ResetDecPrivateMode(
                             DecPrivateMode::Code(DecPrivateModeCode::SynchronizedOutput),
                         ))) => {
                             hold = false;
+                            hold_deadline = None;
                             flush = true;
                         }
                         Action::CSI(CSI::Device(dev)) if matches!(**dev, Device::SoftReset) => {
                             hold = false;
+                            hold_deadline = None;
                             flush = true;
                         }
                         _ => {}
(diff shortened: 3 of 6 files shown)


docs/changelog.md
```diff
@@ -145,6 +145,13 @@ As features stabilize some brief notes about them will accumulate here.
   Thanks to @j4james! #7046
 
 #### Fixed
+* Queries such as the primary device attributes request or the kitty keyboard
```
bew: move this at the end of the section, also please reduce this paragraph, keep it short and link to related documentation.

mux/src/lib.rs
```diff
@@ -1464,3 +1572,6 @@ impl wezterm_term::DownloadHandler for MuxDownloader {
         }
     }
 }
+
+#[cfg(test)]
+mod test;
```
bew: move this to the top, near the other mod declarations.

mux/src/lib.rs
```diff
@@ -137,6 +139,45 @@ fn send_actions_to_mux(pane: &Weak<dyn Pane>, dead: &Arc<AtomicBool>, actions: V
     histogram!("send_actions_to_mux.rate").record(1.);
 }
 
+/// Returns true for queries that are safe to answer while a synchronized
+/// update (DEC private mode 2026) is holding back the output stream:
+/// their responses reflect terminal capabilities rather than screen state,
+/// so they don't need to wait for the held actions to be applied.
+/// Applications commonly block waiting for these responses, and would
+/// otherwise stall until the update closes.
+/// The kitty keyboard query is answered eagerly, but the kitty state
```
bew: I think you can stop the fn doc before this line, the rest can be moved as comments above each branch in the function.

mux/src/lib.rs
```diff
@@ -145,8 +186,62 @@ fn parse_buffered_data(pane: Weak<dyn Pane>, dead: &Arc<AtomicBool>, mut rx: Fil
...
+                        revents: 0,
+                    }];
+                    match poll(&mut pfd, Some(remaining)) {
+                        Ok(0) => true,
+                        Ok(_) => false,
+                        // Interrupted before the deadline; recompute the
+                        // remaining time and poll again. The error variant
+                        // depends on the backend: poll(2) reports
+                        // FdError::Poll, while the macOS select(2) shim and
+                        // Windows WSAPoll report FdError::Io
```
bew, line 212: Do you have refs for this? has you tried this yourself on those platforms?

mux/src/lib.rs
```diff
@@ -137,6 +139,45 @@ fn send_actions_to_mux(pane: &Weak<dyn Pane>, dead: &Arc<AtomicBool>, actions: V
...
+                | Device::StatusReport
+        ),
+        Action::CSI(CSI::Keyboard(Keyboard::QueryKittySupport)) => true,
+        _ => false,
+    }
+}
+
+/// The poll timeout is a c_int of milliseconds on the platforms this
+/// runs on, and the deadline is Instant arithmetic; clamp the configured
+/// value (to ~24.8 days) so that an absurdly large setting can't wrap
```
bew, line 174: I think we can clamp to a much lower value.. 10sec maybe? 60sec?
Can you think about any use-case for a longer timeout?

mux/src/lib.rs
```diff
@@ -145,8 +186,62 @@ fn parse_buffered_data(pane: Weak<dyn Pane>, dead: &Arc<AtomicBool>, mut rx: Fil
     let mut action_size = 0;
     let mut delay = Duration::from_millis(configuration().mux_output_parser_coalesce_delay_ms);
     let mut deadline = None;
+    let mut hold_timeout = hold_timeout_from(&configuration());
+    let mut hold_deadline: Option<Instant> = None;
```
bew, line 193: What's the benefit from having a separate delay/deadline and poll loop from the existing one for when there are no actions & no hold?

Could we try to merge both of them instead, to avoid duplicating code?

I'd be interested as well if we can isolate the fd polling code in a separate function, with an intentional enum result for the different cases after the poll 🤔

config/src/config.rs
```diff
@@ -411,6 +411,13 @@ pub struct Config {
     #[dynamic(default = "default_mux_output_parser_coalesce_delay_ms")]
     pub mux_output_parser_coalesce_delay_ms: u64,
 
+    /// How many ms a synchronized update (DEC private mode 2026) may
+    /// hold back output before wezterm stops waiting for the closing
+    /// sequence and applies the buffered output anyway, so that a
+    /// stalled application cannot freeze the pane indefinitely
+    #[dynamic(default = "default_synchronized_output_timeout_ms")]
+    pub synchronized_output_timeout_ms: u64,
```
bew: what do you think about renaming this to have `mux_` prefix like other related mux options 🤔

mux/src/lib.rs
```diff
@@ -137,6 +139,45 @@ fn send_actions_to_mux(pane: &Weak<dyn Pane>, dead: &Arc<AtomicBool>, actions: V
...
+        Action::CSI(CSI::Device(dev)) => matches!(
+            **dev,
+            Device::RequestPrimaryDeviceAttributes
+                | Device::RequestSecondaryDeviceAttributes
+                | Device::RequestTertiaryDeviceAttributes
+                | Device::RequestTerminalNameAndVersion
+                | Device::StatusReport
+        ),
+        Action::CSI(CSI::Keyboard(Keyboard::QueryKittySupport)) => true,
+        _ => false,
```
bew, line 168: I think you can include `CSI::Window`' `Report*` requests as well 🤔

See https://github.com/wezterm/wezterm/blob/main/termwiz/src/caps/probed.rs#L128 impl & sources in csi.rs

mux/src/lib.rs
```diff
@@ -158,27 +253,39 @@ fn parse_buffered_data(pane: Weak<dyn Pane>, dead: &Arc<AtomicBool>, mut rx: Fil
             }
             Ok(size) => {
                 parser.parse(&buf[0..size], |action| {
+                    if hold && is_passthrough_query(&action) {
+                        send_actions_to_mux(&pane, &dead, vec![action]);
+                        return;
```
bew, line 261: should also set `action_size = 0` like all other calls to `send_actions_to_mux` ?

bew (commented):
Thanks for the PR!

I didn't read the test harness yet, but I have a few questions/comments on the rest already 🤔

mux/src/lib.rs
bew: hum, comments are usually placed _above_ what the comment is about, but you placed them _below_ for some reason in your updated code...

luizribeiro: hmm they seem right to me? The first comment is about the kitty arm and the second about the _ fallback. But they do read a bit vague which maybe is where the confusion comes from? I'll reword them a bit.

Or maybe we don't even need these comments altogether since the method one feels pretty clear?

mux/src/lib.rs
luizribeiro: I have only tested on macOS.

These come from the `filedescriptor` crate: the non-macOS unix `poll_impl` wraps failures in `Error::Poll`:

https://github.com/wezterm/wezterm/blob/fff02ca501c3b457f99b467a86061d2b150c51f2/filedescriptor/src/unix.rs#L456-L461

while the macOS select shim and the Windows `WSAPoll` impl both return `std::io::Error::last_os_error().into()`, which becomes `Error::Io`:

https://github.com/wezterm/wezterm/blob/fff02ca501c3b457f99b467a86061d2b150c51f2/filedescriptor/src/unix.rs#L562-L564

and

https://github.com/wezterm/wezterm/blob/fff02ca501c3b457f99b467a86061d2b150c51f2/filedescriptor/src/windows.rs#L565-L567

Honestly, this comment was Claude's touch and I'm not super convinced it's worth keeping, since it's basically duplicating information that's already in the code. If this happens to change the type checker would likely complain anyway. What do you think?

mux/src/lib.rs
luizribeiro: I don't think there's any benefit tbh. I'll give a shot at extracting the polling into a `poll_for_input(fd, deadline)` helper returning an enum (`Ready`, `TimedOut`, `Failed`) tomorrow and see if that makes it better. I'll likely do this as its own commit on top of the ones we already have so the refactor is easier to review (so the duplicated poll loop would be in the history for a commit or two).

mux/src/lib.rs
bew: Ah yes sorry I read too quickly, it's just that the `Action::CSI(CSI::Device(dev))` branch doesn't actually have a comment and I though it had one below instead ><

mux/src/lib.rs
bew: Hmm good catch, honestly it looks like an oversight in `filedescriptor`, all platforms should return a `Error::Poll` out of `poll` fn :thinking: 
If this gets 'fixed' later, I don't think you'd get a type error since the 2 Error variants are likely to stay. Just that the Io one wouldn't be used anymore for poll errors..

:point_right: I think you can remove/simplify the overly detailed comment and leave the checks on Poll + Io error.

mux/src/lib.rs
bew: don't worry about the commits, you can add as many as you want here in the PR, they'll be squashed at the end anyway
(but good commit messages are still useful to know what's going on & for writing the final one!)

bew:
Hello @luizribeiro, someone raised a similar issue at #8146 so I'm coming back to your PR 🤔
👉 Are you willing to continue working on this PR to address my comments above?

PR #7933, opened 2026-07-14
Add win32_system_backdrop_keep_inactive: keep the active backdrop material on unfocused windows

### LINKED ISSUE #5895: Acrylic / Mica for unfocused windows

KhangHLe:
## Summary

DWM swaps a window's system-backdrop material (Acrylic/Mica/Tabbed) for a desaturated/solid fallback whenever the window is deactivated — a power-saving behavior keyed off the non-client activation state that `DefWindowProc` records while processing `WM_NCACTIVATE`. There is no system-wide way to disable it, and windows using `win32_system_backdrop` lose their transparency/material whenever they lose focus.

This PR adds an opt-in config option:

```lua
config.window_background_opacity = 0.75
config.win32_system_backdrop = "Acrylic"
config.win32_system_backdrop_keep_inactive = true
```

When enabled (and a system backdrop is in use), WezTerm handles `WM_NCACTIVATE` by passing `wParam=TRUE` to `DefWindowProc` (with `lParam=-1` to suppress the frame repaint), so DWM keeps rendering the vivid active material while the window is unfocused. Real focus handling (`WM_SETFOCUS`/`WM_KILLFOCUS`) is untouched, and deactivation (`wParam == 0`) is the only case intercepted.

Fixes #5895 (also reported as #6979).

## Notes

- Off by default — default behavior is unchanged.
- No effect when `win32_system_backdrop = "Disable"`.
- This is the same approach Win32 apps use in their own wndprocs to keep active-state visuals; Windows Terminal achieves the equivalent by composing its own in-app acrylic brush (microsoft/terminal#593).
- Tested on Windows 11 25H2 with Acrylic: unfocused windows keep the vivid material; without the option they gray out as before.

## Docs

- New page: `docs/config/lua/config/win32_system_backdrop_keep_inactive.md`
- Changelog entry under Nightly → New

### DIFF
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -585,6 +585,15 @@ pub struct Config {
     #[dynamic(default)]
     pub win32_system_backdrop: SystemBackdrop,
 
+    /// Only works on Windows: when a `win32_system_backdrop` material is
+    /// active, keep rendering the *active* material while the window is
+    /// unfocused. By default DWM swaps the backdrop for a desaturated /
+    /// solid fallback on inactive windows (a power-saving behavior); this
+    /// option suppresses that by reporting the window as visually active.
+    /// See <https://github.com/wezterm/wezterm/issues/5895>.
+    #[dynamic(default)]
+    pub win32_system_backdrop_keep_inactive: bool,
+
     #[dynamic(default = "default_win32_acrylic_accent_color")]
     pub win32_acrylic_accent_color: RgbaColor,
 
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -22,6 +22,7 @@ usually the best available version.
 As features stabilize some brief notes about them will accumulate here.
 
 #### Changed
+
 * DECRQCRA is now disabled by default to prevent silent screen scraping.
   Set `enable_checksum_rectangular_area = true` to re-enable it.
   Thanks to @jquast! #7701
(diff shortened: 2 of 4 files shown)


bew:
Hello, is this an ~exact duplicate of #7848 ?

KhangHLe:
> Hello, is this an ~exact duplicate of #7848 ?

Yes! I did not do the due diligence haha, I can close this PR

KhangHLe:
Closing in favor of #7848, which got there first — same mechanism, same fix. (Said I'd do this back in July and let it ride; tidying up now.)

PR #7942, opened 2026-07-16
Guard shell integration against unset ZSH_NAME

### LINKED ISSUE #7920: /etc/profile.d/wezterm.sh fails with "unbound variable" when nounset (set -u) is enabled
## Problem
`/etc/profile.d/wezterm.sh` references `ZSH_NAME` without a default value. When the calling shell has `nounset` (`set -u`) enabled, this causes the shell to exit with an "unbound variable" error.

## Affected lines
- Line 18: `if [ -z "${BASH_VERSION}" -a -z "${ZSH_NAME}" ] ; then`
- Line 471: `if [[ -n "$ZSH_NAME" ]] ; then`

## Steps to reproduce
1. Run a bash script that starts with `set -euo pipefail`
2. After the script exits, the shell terminates with a `ZSH_NAME: unbound variable` error

## Expected behavior
The script should use a default value (e.g. `${ZSH_NAME:-}`) so it doesn't error under `nounset`.

## Environment
- WezTerm: 20240203-110809-5046fc22
- OS: Ubuntu 24.04
- Installed via: apt (apt.fury.io/wez repository)

tianrking:
## What changed

Guard the remaining `ZSH_NAME` check in the shell-integration prompt hook with a default expansion.

## Why

When a Bash user enables `set -u`, `__wezterm_semantic_precmd` previously expanded an unset `ZSH_NAME` directly and aborted before rendering the prompt's semantic markers.

## User impact

Bash users who enable nounset can load WezTerm shell integration without an unset `ZSH_NAME` breaking their prompt hook.

Fixes #7920.

## Validation

- Reproduced before the change with interactive Bash and `set -u`: `ZSH_NAME: unbound variable`.
- Re-ran the same source-and-prompt-hook command after the change; it exits 0.
- `git diff --check`

ShellCheck is not installed in the local environment.

### DIFF
--- a/assets/shell-integration/wezterm.sh
+++ b/assets/shell-integration/wezterm.sh
@@ -468,7 +468,7 @@ __wezterm_semantic_precmd() {
     __wezterm_save_ps2="$PS2"
     # Markup the left and right prompts so that the terminal
     # knows that they are semantically prompt output.
-    if [[ -n "$ZSH_NAME" ]] ; then
+    if [[ -n "${ZSH_NAME-}" ]] ; then
       PS1=$'%{\e]133;P;k=i\a%}'$PS1$'%{\e]133;B\a%}'
       PS2=$'%{\e]133;P;k=s\a%}'$PS2$'%{\e]133;B\a%}'
     else


bew (approved):
Makes sense, thanks for the fix !

PR #7955, opened 2026-07-20
Fix duplicate blob storage on Windows

Tomauskasz:
## Summary

- reuse an existing live content-addressed blob instead of replacing its file
- increment the blob reference count for each duplicate store
- add a regression test that keeps the first reader open while storing identical content again

## Problem

`SimpleTempDir::store` persisted a new temporary file over the content-hash destination on every store. When a reader still had that destination open, Windows rejected the replacement with error 5 (`Access is denied`). This caused later consumers of the same image data, such as animated backgrounds in additional windows, to fail to load it.

The storage contract already requires duplicate content IDs to be deduplicated. A positive reference count means the content-addressed file is live, so this change reuses it and records the additional reference.

## Testing

- Before the fix, the new regression test failed with:
  `Io(Os { code: 5, kind: PermissionDenied, message: "Access is denied." })`
- `cargo test -p wezterm-blob-leases --features simple_tempdir`
- `cargo test --all --no-fail-fast`
- `cargo +nightly fmt --all -- --check`
- Manual Windows validation with one source window and two newly spawned windows: all three animated backgrounds showed motion, and the GUI log contained no blob/access-denied error.

### DIFF
--- a/wezterm-blob-leases/src/simple_tempdir.rs
+++ b/wezterm-blob-leases/src/simple_tempdir.rs
@@ -74,6 +74,15 @@ impl BlobStorage for SimpleTempDir {
     fn store(&self, content_id: ContentId, data: &[u8], _lease_id: LeaseId) -> Result<(), Error> {
         let mut refs = self.refs.lock().unwrap();
 
+        // A live reference means that this content-addressed file is already
+        // stored, so reuse it rather than trying to replace it.
+        if let Some(count) = refs.get_mut(&content_id) {
+            if *count > 0 {
+                *count += 1;
+                return Ok(());
+            }
+        }
+
         let path = self.path_for_content(content_id)?;
         let mut file = tempfile::Builder::new()
             .prefix("new-")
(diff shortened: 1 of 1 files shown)


bew (commented):
Hello, thanks for the PR!
The change makes sense 👍 

How did you find about this issue?
Are you using the `wezterm-blob-leases` crate directly or did you find this through WezTerm usage?

Tomauskasz:
> Hello, thanks for the PR! The change makes sense 👍
> 
> How did you find about this issue? Are you using the `wezterm-blob-leases` crate directly or did you find this through WezTerm usage?

I've made this https://github.com/Tomauskasz/electric-control-room.wez on Mac first and it was working just fine there, but then once I've set it up on Windows - it would only work for 1 window. If I opened new windows, then this custom visual of mine wouldn't show, so I investigated and found out that it's due to duplicate blob storage. That visual is working just fine now after the fix

PR #7956, opened 2026-07-20
Install desktop entry and icons in Linuxbrew formula

pir0c0pter0:
The Linuxbrew formula only installs the AppImage binary. This leaves no desktop entry or icon, so desktop launchers may show a missing icon or fail to start WezTerm.

This change installs the desktop entry and icons already included in the AppImage and points Exec and TryExec to the stable Homebrew opt path. It also fixes the existing brew style issues.

Tested on Bazzite 44 Kinoite, an atomic Fedora-based Linux distribution, using brew style, desktop-file-validate and gtk-launch.

### DIFF
--- a/ci/wezterm-linuxbrew.rb.template
+++ b/ci/wezterm-linuxbrew.rb.template
@@ -1,9 +1,9 @@
-# Note: if you are viewing this from the tap repo, this file is automatically
+# NOTE: If you are viewing this from the tap repo, this file is automatically
 # updated from:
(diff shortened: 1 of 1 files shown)


bew (commented):
Thanks for the PR!

Just one comment: why did you change the description? 🤔 
_this has nothing to do with the scope of this PR_

pir0c0pter0:
Only to pass brew style check.

PR #7958, opened 2026-07-21
fix: macOS crash (SIGABRT) from RefCell re-entrancy panic in WindowView::draw_rect

### LINKED ISSUE #7427: Crash (segfault) with no apparent reason on mac

macOS

N/A

### WezTerm version

20240203-110809,5046fc22

### Describe the bug

Wezterm just crashed. If I remember correctly I wasn't even using it at that time.
This is a one-time crash, and Wezterm was running for weeks without any issue before this, and since this crash.
This is on macOS 15.6.1 (24G90).

### To Reproduce

I don't have any actionable STR to reproduce, this is a one-time crash.

### Configuration

```
config.hide_tab_bar_if_only_one_tab = true
config.native_macos_fullscreen_mode = false
```
Besides this, only a binding and some themes/font configs.

### Expected Behavior

No crash

### Logs

This is logs as seen on this crash window:
<details>
<summary>
Screenshot
</summary>

Image

</details>

It doesn't seems to contain a stack trace/coredump as I'm used to on Linux, but maybe it's just because it looks very different on mac. I tried asking on Matrix how useful this report'd be, but had no answer so here's the report and the related logs.

Logs as attachment as GitHub found out them to be too big to post here.
[wezterm-crash-logs.log](https://github.com/user-attachments/files/24082061/wezterm-crash-logs.log)

### Anything else?

_No response_

mvanhorn:
`draw_rect` in `window/src/os/macos/window.rs` (around line 3047) holds a live. `this.inner.borrow_mut()` guard while calling. Reporter Porkepix sees wezterm-gui abort/segfault on macOS with no reliable. repro ("one-time crash", recurring roughly weekly).

Happy path: resizing/repainting a wezterm window on macOS repeatedly; (including rapid successive repaints, e.g. resizing while content is.

Fixes #7427

### DIFF
--- a/window/src/os/macos/window.rs
+++ b/window/src/os/macos/window.rs
@@ -3063,12 +3063,17 @@ impl WindowView {
             if inner.paint_throttled {
                 inner.invalidated = true;
             } else {
-                inner.events.dispatch(WindowEvent::NeedRepaint);
                 inner.invalidated = false;
                 inner.paint_throttled = true;
 
                 let window_id = inner.window_id;
                 let max_fps = inner.config.max_fps;
+                let mut events =
+                    std::mem::replace(&mut inner.events, WindowEventSender::new(|_, _| {}));
+                drop(inner);
+                events.dispatch(WindowEvent::NeedRepaint);
+                this.inner.borrow_mut().events = events;
+
                 promise::spawn::spawn(async move {
                     async_io::Timer::after(std::time::Duration::from_millis(1000 / max_fps as u64))
                         .await;
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -154,6 +154,7 @@ As features stabilize some brief notes about them will accumulate here.
 * macOS: wacky initial window size with external monitors or certain font
   sizes. #4966 #4250
 * macOS: dragging non-filename data over wezterm could cause it to crash. #4771
+* macOS: repainting a window could cause wezterm to crash. #7427
 * New tabs spawned by the gui could spawn into the wrong domain when using
   multiplexing together `default_domain`. Thanks to @bogdan2412! #4994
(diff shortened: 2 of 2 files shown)


window/src/os/macos/window.rs
```diff
@@ -3063,12 +3063,17 @@ impl WindowView {
...
                 inner.invalidated = false;
                 inner.paint_throttled = true;
 
                 let window_id = inner.window_id;
                 let max_fps = inner.config.max_fps;
+                let mut events =
+                    std::mem::replace(&mut inner.events, WindowEventSender::new(|_, _| {}));
+                drop(inner);
+                events.dispatch(WindowEvent::NeedRepaint);
+                this.inner.borrow_mut().events = events;
```
bew, line 3075: This new code feels weird, it's not clear why you need to do this..
could you explain what's going on here?

mvanhorn: Sure. The crash in #7427 is a RefCell double-borrow abort: draw_rect held the this.inner.borrow_mut() guard while calling events.dispatch(NeedRepaint). If the dispatch re-enters anything that borrows inner again (the handler runs synchronously), the second borrow_mut() panics with BorrowMutError and the process aborts, which matches the no-repro weekly crash in the report.

The mem::replace dance makes the dispatch run with no borrow held: move the real sender out of inner (leaving a no-op sender in its place), drop(inner) to release the guard, dispatch, then re-borrow and put the real sender back. Re-entrant draw_rect calls during the dispatch take the paint_throttled branch and just mark invalidated, so no repaint is lost.

Happy to restructure if you would prefer a different shape, e.g. ending the borrow scope before a plain dispatch instead of the swap.

Issue #5158, opened 2024-03-13
thin line when opacity and shadow enabled


Faetu:
### What Operating System(s) are you seeing this problem on?

macOS

### WezTerm version

20240203-110809-5046fc22

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

When using opacity < 1.0 with "MACOS_FORCE_ENABLE_SHADOW" flag, there is a black thin frame:
<img width="147" alt="image" src="https://github.com/wez/wezterm/assets/55803495/e3675059-1859-4495-b83c-12db789c57f8">
This goes around the whole window.

### To Reproduce

Set opacity below 1.0 and add the "MACOS_FORCE_ENABLE_SHADOW" decoration flag.

### Configuration

```lua
-- Pull in the wezterm API
local wezterm = require("wezterm")

local config = wezterm.config_builder()

config.color_scheme = "Catppuccin Mocha"
config.front_end = "OpenGL"
config.bold_brightens_ansi_colors = true
config.hide_tab_bar_if_only_one_tab = true
config.tab_bar_at_bottom = true
config.default_prog = { "/opt/homebrew/bin/fish", "-l" }
config.window_decorations = "NONE | MACOS_FORCE_ENABLE_SHADOW"
config.window_background_opacity = 0.8
config.macos_window_background_blur = 90
config.max_fps = 120
config.use_fancy_tab_bar = false
config.notification_handling = "AlwaysShow"
config.animation_fps = 60
config.initial_rows = 45
config.initial_cols = 175
config.send_composed_key_when_left_alt_is_pressed = true
config.native_macos_fullscreen_mode = true
config.keys = {
	{
		key = "d",
		mods = "SUPER|SHIFT",
		action = wezterm.action.SplitHorizontal({ domain = "CurrentPaneDomain" }),
	},
	-- Send "CTRL-A" to the terminal when pressing CTRL-A, CTRL-A
	{
		key = "d",
		mods = "SUPER",
		action = wezterm.action.SplitVertical({ domain = "CurrentPaneDomain" }),
	},
	{
		key = "f",
		mods = "SUPER|CTRL",
		action = wezterm.action.ToggleFullScreen,
	},
	{
		key = "LeftArrow",
		mods = "SUPER|OPT",
		action = wezterm.action.AdjustPaneSize({ "Left", 5 }),
	},
	{
		key = "RightArrow",
[...]

PR #7962, opened 2026-07-22
Fix window_frame border color missing at rounded macOS corners

raemr:
On macOS, the WindowServer rounds a titleless window's outer silhouette, but the border rectangles drawn for window_frame only overlap in border_width-sized squares at each corner. Since the OS's real corner radius is bigger than that, the whole colored overlap sits inside the region the OS clips away, leaving the corner bare.

Rebuild each corner as a solid square of border color with a background-colored quarter-disc carved out of it, inset by the configured border width and centered on the same point as the OS's own corner arc. The visible ring between the OS clip and the disc then reads as a constant-thickness border that follows the rounded corner, without needing to know the OS's exact radius.

### DIFF
--- a/wezterm-gui/src/termwindow/render/borders.rs
+++ b/wezterm-gui/src/termwindow/render/borders.rs
@@ -1,6 +1,10 @@
 use crate::quad::TripleLayerQuadAllocator;
+use crate::termwindow::render::corners::{
+    BOTTOM_LEFT_ROUNDED_CORNER, BOTTOM_RIGHT_ROUNDED_CORNER, TOP_LEFT_ROUNDED_CORNER,
+    TOP_RIGHT_ROUNDED_CORNER,
+};
 use crate::utilsprites::RenderMetrics;
-use ::window::ULength;
+use ::window::{PointF, SizeF, ULength};
 use config::{ConfigHandle, DimensionContext};
 
 impl crate::TermWindow {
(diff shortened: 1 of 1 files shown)


raemr:
Ah snap I forgot about full screen mode, this does not work while the terminal is full screened.

(Clarification): The rounded corners are present in full screen for the inner line, even though the outer line is straight and a square.

bew:
Hello, thanks for the PR!
Since this is a visual change, could you show some screenshots of different cases & before/after?
Thank you!

raemr:
Hi, yes of course! Note: Since the previews are rounded, both fullscreens look off for me, but if you view them in full they are in fact working as intended.

## Before Fix 

### Windowed:
image

### Fullscreen:

image

## After fix

### Windowed:
image

### Fullscreen:
(I fixed this in the latest commit)

image

raemr:
Also, slight issue (?). The method I'm using to draw borders is pretty much just guessing how MacOS draws borders. I’m working on a commit that queries `NSThemeFrame`, a native AppKit class *currently* used by macOS to render the window frame and corner radius.

There is no public API that exposes the actual corner radius of a standard `NSWindow`, so the only way to retrieve the live value is through this private AppKit method. The implementation will use runtime lookup and fall back to the current estimate if the method is unavailable.

The main tradeoff is compatibility, so because the API is undocumented, Apple could rename or remove it in a future macOS release. It does not introduce any third-party dependency or external access. 

Not sure if this is an issue if I were to commit it, but the alternative is a guess, since MacOS' corner radius is dynamic based on factors like if there's a navbar (per my research, not a lot I could reliably find on this).

raemr:
> Also, slight issue (?). The method I'm using to draw borders is pretty much just guessing how MacOS draws borders. I’m working on a commit that queries `NSThemeFrame`, a native AppKit class _currently_ used by macOS to render the window frame and corner radius.
> 
> There is no public API that exposes the actual corner radius of a standard `NSWindow`, so the only way to retrieve the live value is through this private AppKit method. The implementation will use runtime lookup and fall back to the current estimate if the method is unavailable.
> 
> The main tradeoff is compatibility, so because the API is undocumented, Apple could rename or remove it in a future macOS release. It does not introduce any third-party dependency or external access.
> 
> Not sure if this is an issue if I were to commit it, but the alternative is a guess, since MacOS' corner radius is dynamic based on factors like if there's a navbar (per my research, not a lot I could reliably find on this).

Actually, after looking into this and implementing it, it doesn't even look as good as this guess (maybe I'm just incompetent), but I think this is as far as I go for this!

raemr:
TL;DR, I think everything is in order. I looked at contributing.md, and my code passes the tests. Additionally, the issue that the PR was made to address has been (in my view) sufficiently resolved. Sorry about the long-winded-ness, first time contributor here!

bew:
Hello! what minimal Wezterm config are you using?

I'm thinking your issue could be related to #8038 (issue: #5158) which is much shorter 🤔

bew:
I merged https://github.com/wezterm/wezterm/pull/8038, could you test whether you still have these issues with latest nightly?

And if so how could I reproduce them?

PR #7967, opened 2026-07-23
macos: Fix top row dragging window when decorations are hidden

### LINKED ISSUE #4522: Selecting topmost line drags the window instead.

macOS

_No response_

### WezTerm version

wezterm 20231030-074559-75909682

### Describe the bug

When trying to select text on the topmost line it won't work and the dragging of the window kicks in instead.
So far I could not find a good workaround for this issue.
It would be nice to be able to disable dragging without modifier. It starts dragging on CMD+Click anywhere AND on NONE+Click on the topmost line.

### To Reproduce

Try selecting the topmost line of text.

### Configuration

```local wezterm = require 'wezterm'

local config = {}

default_gui_startup_args = {"start"}

if wezterm.config_builder then
    config = wezterm.config_builder()
end


config.window_decorations = 'RESIZE'
config.hide_tab_bar_if_only_one_tab = true

return config

### Expected Behavior

Can select text on the topmost line and only drag window on modifier (eg. CMD+Click by default).

### Logs

_No response_

### Anything else?

_No response_

### PR DESCRIPTION
**Pull request by draconivis**
## Summary

Fixes #4522. On macOS, clicking the topmost row of the window (e.g. to select text) started a native window drag instead, even with `window_decorations = "RESIZE"` / `"NONE"`.

## Root cause

WezTerm keeps `NSWindowStyleMaskTitled` in the window's style mask even when the titlebar is fully hidden, to preserve rounded window corners (#1034). `NSWindow.isMovable` defaults to `true`, and AppKit ties a native "click the title-bar area to move the window" behavior to that style mask — independent of whether the titlebar UI is visually hidden/transparent. That native drag zone was intercepting clicks on the topmost row before they reached wezterm's content view.

## Fix

In `apply_decorations_to_window` (`window/src/os/macos/window.rs`), set `window.isMovable = false` whenever the titlebar is fully hidden (no `TITLE`, no `INTEGRATED_BUTTONS`). WezTerm's own dragging mechanisms (tab-bar drag, `SUPER`+drag, `StartWindowDrag`) reposition the window manually and don't depend on native movability, so this doesn't regress those. Also hides the now-inert `NSTitlebarContainerView` so it's out of the responder chain.

Extracted the "is titlebar hidden" condition into a small pure function (`titlebar_is_hidden`) and added unit tests covering all `WindowDecorations` combinations, since the actual AppKit drag behavior isn't unit-testable without a real windowing session.

macOS-only change — X11/Wayland/Windows don't exhibit this bug.

## Test plan

- [x] `cargo build --release -p wezterm-gui`
- [x] `cargo test -p window titlebar`
- [x] `cargo +nightly fmt --package window -- --check`
- [x] Manually reproduced with `window_decorations = "RESIZE"`, `hide_tab_bar_if_only_one_tab = true`: before the fix, clicking/dragging the topmost line moved the window; after the fix, it selects text normally.

### DIFF
--- a/window/src/os/macos/window.rs
+++ b/window/src/os/macos/window.rs
@@ -1397,6 +1397,13 @@ fn apply_decorations_to_window(
         } else {
             window.setTitlebarAppearsTransparent_(to_yes_no(!has_decorations));
         }
+
+        if let Some(titlebar_view_container) = get_titlebar_view_container(window) {
+            // hiding the subview prevents it from participating in normal visible/hit-tested view behavior
+            let _: () = msg_send![*titlebar_view_container.load(), setHidden: to_yes_no(!has_decorations)];
+        }
+
+        let _: () = msg_send![**window, setMovable: to_yes_no(has_decorations)];
     }
 }
 
--- a/window/src/os/macos/window.rs
+++ b/window/src/os/macos/window.rs
@@ -1400,7 +1400,8 @@ fn apply_decorations_to_window(
 
         if let Some(titlebar_view_container) = get_titlebar_view_container(window) {
             // hiding the subview prevents it from participating in normal visible/hit-tested view behavior
-            let _: () = msg_send![*titlebar_view_container.load(), setHidden: to_yes_no(!has_decorations)];
+            let _: () =
+                msg_send![*titlebar_view_container.load(), setHidden: to_yes_no(!has_decorations)];
         }
 
         let _: () = msg_send![**window, setMovable: to_yes_no(has_decorations)];

### CODE REVIEW

**Comment by bew** [comment]
Thanks for the PR!

I tested it locally (before/after), it seems to work correctly!
I tried with different combination of:
```lua
return {
  -- hide_tab_bar_if_only_one_tab = true,
  window_decorations = "RESIZE",
  -- window_decorations = "TITLE",
  -- window_decorations = "INTEGRATED_BUTTONS|RESIZE",
  -- window_decorations = "TITLE|INTEGRATED_BUTTONS",
}
```

Although I'm not sure making a function is worth it, the tests you added are trivial and the function won't change any time soon, honestly I'd just fix the condition instead which gives a much simpler diff for the PR 🤔
But keep a `titlebar_is_hidden` var, so it's clear what 'is hidden' in the checks!

---
_I noticed something that predates this PR but that definitely looks fun / like a bug, using `TITLE|INTEGRATED_BUTTONS` makes the titlebar fully transparent ^^ it's probably not meant to be used together 🤷_
_You don't need to fix it, just wanted to mention it 😀_

window/src/os/macos/window.rs
```diff
@@ -1405,6 +1405,36 @@ fn apply_decorations_to_window(
         } else {
             window.setTitlebarAppearsTransparent_(hidden);
         }
+
+        if let Some(titlebar_view_container) = get_titlebar_view_container(window) {
+            let _: () = msg_send![*titlebar_view_container.load(), setHidden: hidden];
```
bew:
```suggestion
            // hiding the subview prevents it from participating in normal visible/hit-tested view behavior
            let _: () = msg_send![*titlebar_view_container.load(), setHidden: hidden];
```

bew:
Hello! I've rebased this on top of https://github.com/wezterm/wezterm/commit/2582d8b04a33cffad2e342efbd66a67646c6a802 to get a helper I made recently to simplify how bools are handled around MacOS-specific window code.
I also simplified it to focus on its core changes.

Thanks a lot for the PR! 💪

PR #7973, opened 2026-07-24
chore(deps): upgrade thiserror to 2.0

xtqqczze:
https://github.com/dtolnay/thiserror/releases/tag/2.0.0

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -401,7 +401,7 @@ dependencies = [
  "num-traits",
  "pastey",
  "rayon",
- "thiserror 2.0.19",
+ "thiserror 2.0.20",
  "v_frame",
  "y4m",
 ]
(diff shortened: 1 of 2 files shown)


bew:
Not sure why I missed this PR 😵‍💫

Thanks a lot!

⚠️ Please always add a description for the PRs, even short.
In this case it's worth linking to the changes in v2: https://github.com/dtolnay/thiserror/releases/tag/2.0.0

xtqqczze:
@bew WDYT?

bew:
What does this help with exactly?

Also, it would be interesting to see if we can leverage the new features in our `Error`s 🤔

xtqqczze:
`thiserror` v2 was released nearly two years ago, and most crates have since migrated. In particular, `filedescriptor` depends on `thiserror`, so upgrading here would help avoid pulling both v1 and v2 into downstream dependency trees.

PR #7976, opened 2026-07-24
cargo: revert edition to 2021

xtqqczze:
```log
$ cargo +1.71.0 check -p filedescriptor                
error: failed to load manifest for workspace member `~/wezterm/strip-ansi-escapes`

Caused by:
  failed to load manifest for dependency `termwiz`

Caused by:
  failed to load manifest for dependency `wezterm-cell`

Caused by:
  failed to parse manifest at `~/wezterm/wezterm-cell/Cargo.toml`

Caused by:
  failed to parse the `edition` key

Caused by:
  this version of Cargo is older than the `2024` edition, and only supports `2015`, `2018`, and `2021` editions.
  ```

### DIFF
--- a/wezterm-cell/Cargo.toml
+++ b/wezterm-cell/Cargo.toml
@@ -2,7 +2,7 @@
 authors = ["Wez Furlong"]
 name = "wezterm-cell"
 version = "0.1.0"
-edition = "2024"
+edition = "2021"
 repository = "https://github.com/wezterm/wezterm"
 description = "Model a Cell in a Terminal display"
 license = "MIT"
--- a/wezterm-escape-parser/Cargo.toml
+++ b/wezterm-escape-parser/Cargo.toml
@@ -2,7 +2,7 @@
 authors = ["Wez Furlong"]
 name = "wezterm-escape-parser"
 version = "0.1.0"
-edition = "2024"
+edition = "2021"
 repository = "https://github.com/wezterm/wezterm"
 description = "Escape Sequence Parser"
 license = "MIT"


bew:
Erm.. Why though?

xtqqczze:
So that the `filedescriptor` crate can be built with a compiler prior to 1.85.

protoz:
It would be better to bring filedescriptor up to 2024 than keep using an old edition

xtqqczze:
Most of the crates in the workspace are still using the 2021 or 2018 edition, so it’s simpler to avoid adopting the 2024 edition for now. This also avoids implicitly increasing the MSRV for many crates to 1.85.

bew:
Yeah we won't be doing that, 2021 is so old now!
Wezterm & we maintainers usually use the latest Rust version anyway.

To get Wezterm on your platform, prefer to get pre-built packages from your distributions, or the Flatpack package, or a compatible AppImage from github releases.

If you need the individual crates, pin it to a compatible version for you or feel free to fork it!

xtqqczze:
> Yeah we won't be doing that, 2021 is so old now! Wezterm & we maintainers usually use the latest Rust version anyway.

@wez Neither `wezterm-cell` nor `wezterm-escape-parser` use any Edition 2024 features, so reverting them to Edition 2021 has no downside for `wezterm` while avoiding an unnecessary MSRV bump for downstream users.

If the plan is to move the entire workspace to Edition 2024 to use its new features, I'm happy with that. I just don't see a benefit to upgrading these crates in isolation.

wez:
There's a difference between a crate using an older edition and the overall project requiring a newer compiler version.

In general, the work we're doing in the main branch should be unconstrained by legacy compiler and environment versions.  There's not usually a good reason for an environment to use an old version of the rust compiler; it's freely available after all.

We should probably update `filedescriptor` to declare the newer edition for the sake of consistency, but it seems to me that that is on a different axis from continuing to support older versions of the Rust compiler.

Issue #2490, opened 2022-09-06
a new function like `get_lines_as_text()` which includes empty lines


naikrovek:
**Is your feature request related to a problem? Please describe.**
`get_lines_as_text()` and `get_logical_lines_as_text()` do not get empty lines.

Sometimes I want to capture the window including empty lines, such as when I want to capture a TUI as-is.  `get_lines_as_text()` and `get_logical_lines_as_text()` omit empty lines, and according to the documentation, this is by design.

**Describe the solution you'd like**
A new function which does include empty lines.

**Describe alternatives you've considered**
The only alternative I have is to manually edit the text captured by these methods to look like what I see on screen.

**Additional context**
An option for the new method/function which optionally also includes textual formatting information (as inline formatting sequences exactly as the shell would receive them.)

Basically I want a "textual screenshot" of the window, without copying with the mouse.

jj51726-jd:
oh, by the way, if this feels like I'm making a demand, then I have misspoken.  If this doesn't make sense for the application, don't do it.  if you don't want to do this, don't do it.  If I annoy you, don't do it.  etc.

naikrovek:
oh, by the way, if this feels like I'm making a demand, then I have misspoken. If this doesn't make sense for the application, don't do it. if you don't want to do this, don't do it. If I annoy you, don't do it. etc.

wez:
https://wezfurlong.org/wezterm/config/lua/pane/get_text_from_region.html was recently added

PR #7985, opened 2026-07-28
Keep blank lines in pane:get_lines_as_text()

delphinus:
## Problem

`pane:get_lines_as_text()` and `pane:get_logical_lines_as_text()` document that

> Trailing blank lines are stripped, which may result in fewer lines being returned than you might expect if the pane only had a couple of lines of output.

but in practice they strip blank lines from *anywhere* in the pane. Capturing a TUI, or any output containing a blank line, returns fewer lines than are on screen and the text no longer corresponds row for row with what the user sees.

```lua
-- pane showing:  hello / (blank) / there
pane:get_lines_as_text()  --> "hello\nthere"
```

Runs of consecutive blank lines collapse entirely. On a pane I hit this with, 72 grid rows came back as 63, with 10 blank rows reduced to 1.

## Cause

Both methods build the result by appending each line to a single buffer and then calling `trim_end()` on the *whole* buffer:

```rust
for line in lines {
    for cell in line.visible_cells() {
        text.push_str(cell.str());
    }
    let trimmed = text.trim_end().len();   // trims the entire buffer
    text.truncate(trimmed);
    text.push('\n');
}
```

`'\n'` is whitespace, so when a line contributes no characters the trim also removes the newline that was written for the preceding line, and that blank line vanishes.

## Fix

Scope the per-line trim to the line just written, and keep the final trim of the whole buffer — that final trim is what implements the documented stripping of *trailing* blank lines, so the documented behaviour is unchanged.

The shared logic is lifted into one `lines_to_text()` helper so the two entry points cannot drift apart, and it is unit tested. Reverting just the trim line makes exactly the three interior-blank-line tests fail, while `strips_trailing_blank_lines` keeps passing.

`get_lines_as_escapes()` and `get_text_from_region()` were never affected; only the two plain-text methods.

## Notes

This was previously reported as #2490 and closed as by design. I think that was based on the documented trailing-blank-line behaviour — the loss of *interior* blank lines looks unintended, since nothing documents it and the two behaviours share the same buggy line.

The docs for both methods and a changelog entry are updated.

- [x] `cargo test -p mux-lua`
- [x] `cargo fmt --all`

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -149,6 +149,10 @@ As features stabilize some brief notes about them will accumulate here.
   OSC 9 escapes set the progress state to "Indeterminate".
 
 #### Fixed
+* [pane:get_lines_as_text()](config/lua/pane/get_lines_as_text.md) and
+  [pane:get_logical_lines_as_text()](config/lua/pane/get_logical_lines_as_text.md)
+  removed blank lines from anywhere in the pane, rather than only the trailing
+  ones they document stripping. #2490
 * Race condition when very quickly adjusting font scale, and other improvements
   around resizing. Thanks to @jknockel! #4876 #5032 #5033
 * macOS: wacky initial window size with external monitors or certain font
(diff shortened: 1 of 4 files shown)


delphinus:
Sorry, I've opened this mistakingly. I'll check code again.

PR #7988, opened 2026-07-28
Update SshDomain.md

hagaigold:
identity file should be the private key


bew (approved):
Good catch, thank you!

PR #7992, opened 2026-07-30
Add llms.txt for LLM-friendly documentation index

abyworkings-coder:
Adds an `llms.txt` file following the emerging [llms.txt convention](https://llmstxt.org/) — a curated, structured Markdown index that helps AI assistants and LLM-based tools navigate the site's documentation more effectively than crawling raw HTML.

The file covers Getting Started, Configuration, Usage, and Project sections — linking to the most-used pages out of the ~600 generated doc pages (mostly per-Lua-function API reference pages) rather than an exhaustive dump, per the convention's intent of a concise, high-signal index.

Placed at `docs/llms.txt` so mkdocs' static-passthrough (no `exclude_docs` entry matches it) copies it into `gh_pages/llms.txt`, serving it at `https://wezterm.org/llms.txt` alongside the rest of the built site. Verified all linked URLs resolve with 200 on the live site before submitting.

### DIFF
--- a/docs/llms.txt
+++ b/docs/llms.txt
@@ -1,0 +1,42 @@
+# WezTerm
+
+> WezTerm is a GPU-accelerated cross-platform terminal emulator and multiplexer, configured and scriptable in Lua, supporting Linux, macOS, Windows, and FreeBSD/NetBSD.
+
+## Getting Started
+
+- [Home](https://wezterm.org/): Overview and feature highlights
+- [Installation](https://wezterm.org/installation.html): Platform installation index
+- [Install on Linux](https://wezterm.org/install/linux.html)
+- [Install on macOS](https://wezterm.org/install/macos.html)
+- [Install on Windows](https://wezterm.org/install/windows.html)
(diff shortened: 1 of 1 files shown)


bew:
Hello, I don't think it's a good idea to add a manually written file for this, this should be generated when we build the project documentation.

Maybe we could use the https://github.com/pawamoy/mkdocs-llmstxt plugin for mkdocs ?

abyworkings-coder:
That makes sense — a generated file from the actual mkdocs build would stay in sync automatically, which a hand-maintained one can't guarantee. I don't have a strong attachment to the manual version; happy to close this in favor of the mkdocs-llmstxt plugin approach if that's the direction you'd rather take. Thanks for the pointer.

bew:
Yes, if you're willing to do the change to a more automated solution, feel free to change this PR or we can close it (we won't merge a manual solution)

abyworkings-coder:
Makes sense — closing in favor of the mkdocs-llmstxt plugin route, which is the better long-term fit for this project. Thanks for the guidance!

PR #7993, opened 2026-07-31
fix(open-url): percent-encode spaces in URLs before opening

### LINKED ISSUE #7983: [Linux/KDE] Opening a file hyperlink fails with argument splitting if path or file name contains spaces Environment:

Linux Wayland

Operating System: EndeavourOS
KDE Plasma Version: 6.7.3
KDE Frameworks Version: 6.28.0
Qt Version: 6.11.1
Kernel Version: 7.1.5-xanmod1-1-x64v2 (64-bit)
Graphics Platform: Wayland

### WezTerm version

### Describe the bug

Description:
When clicking on a file link inside the terminal, WezTerm passes the current working directory (cwd) and file name to the system opener. If there are spaces anywhere in the absolute path or the file name itself, the path gets incorrectly split. KDE's KIO Client receives a truncated string, causing it to fail with a "file or folder does not exist" error.

Actual Behavior:
KDE pops up a KIO Client error dialog saying:

The file or folder /home/user/My Projects/Invoice Archive/Q1 does not exist.

The terminal application passes the arguments without strict encapsulation, resulting in the path breaking at the space character right before Financial Report.pdf.

### To Reproduce

Steps to Reproduce:
   1. Create a directory structure containing spaces: mkdir -p "/home/user/My Projects/Invoice Archive"
   2. Navigate into that directory: cd "/home/user/My Projects/Invoice Archive"
   3. Create a dummy PDF file that also contains spaces: touch "Q1 Financial Report.pdf"
   4. Expose the file path or list it to generate a clickable hyperlink: ls -l
   5. Hold the appropriate modifier key and click the hyperlinked filename Q1 Financial Report.pdf to open it.

### Configuration

no config

### Expected Behavior

The hyperlink handler must safely escape or encapsulate the entire absolute path string so xdg-open processes it as a single literal argument.

### Logs

_No response_

### Anything else?

Image

MeGaurav4:
Fixes #7983. When a file hyperlink contains spaces (e.g. OSC 8 link to `/home/user/My Projects/Q1 Financial Report.pdf`), the URL is passed to `xdg-open` with the raw spaces. KDE's KIO re-parses the argument and splits at the whitespace, so the file fails to open with a "does not exist" error.

Raw spaces are never valid in a URI, so encoding them as %20 is safe and never double-encodes an existing escape sequence. Applies to both `open_url` and `open_with`.

### DIFF
--- a/wezterm-open-url/src/lib.rs
+++ b/wezterm-open-url/src/lib.rs
@@ -2,9 +2,18 @@
 // Copyright © 2015 Sebastian Thiel
 // <https://github.com/Byron/open-rs>
 
+/// Some desktop openers (notably KDE's KIO) re-parse the URL argument
+/// and split it on whitespace, so a file path containing spaces is
+/// truncated and fails to open. A raw space is never valid in a URI,
+/// so encoding it as %20 is safe and cannot double-encode an existing
+/// escape sequence.
+fn encode_spaces(url: &str) -> String {
+    url.replace(' ', "%20")
+}
+
 #[cfg(not(windows))]
 pub fn open_url(url: &str) {
-    let url = url.to_string();
+    let url = encode_spaces(url);
     std::thread::spawn(move || {
         #[cfg(target_os = "macos")]
         let candidates: &[&[&str]] = &[&["/usr/bin/open", &url]];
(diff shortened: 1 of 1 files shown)


bew:
(note: waiting for repro steps on #7983 first)

PR #8006, opened 2026-08-04
tmuxcc: fix TmuxPty* Write impl to return length of consumed data

### LINKED ISSUE #8003: tmux -CC TmuxPty Write implementations enqueue input but incorrectly return Ok(0)

Linux X11

**Desktop environment:** MATE
**Windowing system:** X11
**Window manager:** Marco, reported by X11 as Metacity (Marco)
**Compositor:** Marco’s built-in compositor, if compositing is enabled; no separate Picom/Compton process was detected.

### WezTerm version

wezterm 20260731-083216-d69264df
Commit `d69264df66fdcc928c7a30c673df108984fda821`.

No, mine is a very recent version and recent examination tells me this was not addressed. 

### Describe the bug

Both `TmuxPty::write()` and `TmuxPtyWriter::write()` copy the complete non-empty input buffer into
a queued `SendKeys` command, but return `Ok(0)`.

Rust's `std::io::Write` contract requires `write()` to return the number of bytes consumed.
Returning `Ok(0)` for a non-empty buffer means that no progress was made. In particular,
`write_all()` converts that result into `ErrorKind::WriteZero`, even though the bytes were already
queued.

This affects keyboard and paste paths that use `write_all()` and can produce false write failures
after the input has already been accepted.

### Actual behavior
The complete input is queued, but the caller is told that zero bytes were consumed.

### Expected behavior
- A non-empty accepted buffer returns `Ok(buf.len())`.
- An empty buffer returns `Ok(0)` and does not queue an empty tmux command.

### Existing reports
No matching WezTerm issue was found for the `Ok(0)`/`WriteZero` contract violation.

### Minimal possible code change

The smallest correction is two lines:

```diff
diff --git a/mux/src/tmux_pty.rs b/mux/src/tmux_pty.rs
--- a/mux/src/tmux_pty.rs
+++ b/mux/src/tmux_pty.rs
@@
-        Ok(0)
+        Ok(buf.len())
@@
-        Ok(0)
+        Ok(buf.len())
```

### More complete possible code change

The following locally tested change was intentionally larger than the minimal two-line correction. It:

- deduplicates the two queueing implementations;
- avoids queueing an empty `send-keys` command;
- schedules command processing only when bytes were accepted;
- returns the exact number of consumed bytes; and
- adds a focused regression test.

```rust
fn enqueue_send_keys(
    cmd_queue: &Arc<Mutex<TmuxCmdQueue>>,
    pane_id: u64,
    buf: &[u8],
) -> usize {
    if buf.is_empty() {
        return 0;
    }

    cmd_queue.lock().push_back(Box::new(SendKeys {
        pane: pane_id,
        keys: buf.to_vec(),
    }));
    buf.len()
}

impl Write for TmuxPtyWriter {
    fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
        let pane_id = self.master_pane.lock().pane_id;
        log::trace!("pane:{}, content:{:?}", &pane_id, buf);

        let written = enqueue_send_keys(&self.cmd_queue, pane_id, buf);
        if written != 0 {
            TmuxDomainState::schedule_send_next_command(self.domain_id);
        }
        Ok(written)
    }

    fn flush(&mut self) -> std::io::Result<()> {
        Ok(())
    }
}

impl Write for TmuxPty {
    fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
        let pane_id = self.master_pane.lock().pane_id;
        log::trace!("pane:{}, content:{:?}", &pane_id, buf);

        let written = enqueue_send_keys(&self.cmd_queue, pane_id, buf);
        if written != 0 {
            TmuxDomainState::schedule_send_next_command(self.domain_id);
        }
        Ok(written)
    }

    fn flush(&mut self) -> std::io::Result<()> {
        Ok(())
    }
}
```

The corresponding regression test verifies both empty and non-empty writes:

```rust
#[test]
fn enqueue_send_keys_reports_consumed_bytes() {
    let cmd_queue = Arc::new(Mutex::new(TmuxCmdQueue::new()));

    assert_eq!(enqueue_send_keys(&cmd_queue, 7, b""), 0);
    assert!(cmd_queue.lock().is_empty());

    assert_eq!(enqueue_send_keys(&cmd_queue, 7, b"abc"), 3);
    let command = cmd_queue
        .lock()
        .pop_front()
        .expect("missing send-keys command")
        .get_command(0);
    assert_eq!(command, "send-keys -H -t %7 0x61 0x62 0x63 \n");
}
```

### Related upstream reports

These reports are related context but reflect the issue describe above above:

- [#336 — Support tmux control mode](https://github.com/wezterm/wezterm/issues/336)
- [#6806 — tmux -CC causing latency with remote VMs](https://github.com/wezterm/wezterm/issues/6806)
- [#7085 — tmux control mode does not detect a broken SSH pipe](https://github.com/wezterm/wezterm/issues/7085)
- [#7144 — Cannot input Chinese characters in control mode](https://github.com/wezterm/wezterm/issues/7144)
- [#7162 — UTF8 problem with tmux control mode](https://github.com/wezterm/wezterm/issues/7162)
- [PR #7154 — fix: tmux -CC Unicode input](https://github.com/wezterm/wezterm/pull/7154)
- [#8001 — tmux -CC keyboard input stalls with tmux 3.7b because SendKeys uses CR instead of LF](https://github.com/wezterm/wezterm/issues/8001)

### To Reproduce

This was discovered while addressing:
- [#8001 — tmux -CC keyboard input stalls with tmux 3.7b because SendKeys uses CR instead of LF](https://github.com/wezterm/wezterm/issues/8001)

### Configuration

no config

### Expected Behavior
- A non-empty accepted buffer returns `Ok(buf.len())`.
- An empty buffer returns `Ok(0)` and does not queue an empty tmux command.

### Logs
_No response_

### Anything else?

_No response_

bew:
Fixes #8003 

Although I'm not 100% sure this is right here, as technically we didn't send the data to tmux yet, it's scheduled to send but it's not sent yet 🤔

@joexue (I cannot add you as a reviewer for some reason..) and maybe @wez
👉 could you confirm this is correct?

---

#8003 bug report also suggested thess small additional behavior changes to:
* avoid queueing an empty send-keys command (note @bew: not sure this is actually needed, but why not)
* schedule command processing _only_ when non-empty input was given

👉 WDYT about that? 🤔

### DIFF
--- a/mux/src/tmux_pty.rs
+++ b/mux/src/tmux_pty.rs
@@ -35,7 +35,7 @@ impl Write for TmuxPtyWriter {
             keys: buf.to_vec(),
         }));
         TmuxDomainState::schedule_send_next_command(self.domain_id);
-        Ok(0)
+        Ok(buf.len())
     }
 
     fn flush(&mut self) -> std::io::Result<()> {
@@ -56,7 +56,7 @@ impl Write for TmuxPty {
             keys: buf.to_vec(),
         }));
         TmuxDomainState::schedule_send_next_command(self.domain_id);
-        Ok(0)
+        Ok(buf.len())
     }
 
     fn flush(&mut self) -> std::io::Result<()> {


wez:
> Although I'm not 100% sure this is right here, as technically we didn't send the data to tmux yet, it's scheduled to send but it's not sent yet 🤔

The way to consider it is about transfer of ownership: the return value signals to the caller how many bytes have been transferred to the next most responsible part of the chain, and since we've queued them, we should report the number that were queued.

For the Write contract that sort of buffering is fine: the `flush` function is the part of the contract that is intended to represent whether the data got flushed through the buffer.

So this looks good me: definitely better than returning 0 which is the wrong signal for the upstream caller; any value we return that is smaller than the number of bytes we captured is effectively a signal to the caller that they need to retry writing the remainder of unwritten bytes.

wez (approved):

PR #8008, opened 2026-08-04
Fix READMEs (MDs) typos

jessp01:
Hi Wez ,

Nice project:)

If you're open to it, I'd also recommend using the GH Actions/pre-commit hook for flagging future typos; see:
https://github.com/crate-ci/typos/blob/master/docs/github-action.md
https://github.com/crate-ci/typos/blob/master/docs/pre-commit.md


bew (approved):
Nice, thanks for the PR!

I agree we could add a GHA workflow with `typos`, that opens a PR with changes it found, we'll consider it!

PR #8010, opened 2026-08-04
Check for superuser privileges and exit if false and make script POSIX compliant

jessp01:
On the superuser check, please see https://github.com/wezterm/wezterm/pull/8010#issuecomment-5180477103

On POSIX compatibility:
```
  ^-- SC3014 (warning): In POSIX sh, == in place of = is undefined.

  In get-deps line 337:
       ^-----^ SC3028 (warning): In POSIX sh, OSTYPE is undefined.
```

For more information:
  https://www.shellcheck.net/wiki/SC3014 -- In POSIX sh, == in place of = is ...
  https://www.shellcheck.net/wiki/SC3028 -- In POSIX sh, OSTYPE is undefined.

I've replaced `OSTYPE` with `uname -s`. I do not have a Darwin or .*BSD handy for testing, but I've used
https://git.kernel.org/pub/scm/git/git.git/tree/config.mak.uname as reference, which I reckon is safe enough:) Still, would be good for you to confirm, if you have access to machines running these.

### DIFF
--- a/get-deps
+++ b/get-deps
@@ -1,5 +1,10 @@
 #!/bin/sh
 
+if [ "$(id -u)" -ne 0 ];then
+	echo "$0 requires superuser privileges. Exiting."
+	exit 1
+fi
+
 TESTING=0
 DOCS=0
 for arg in "$@"; do
@@ -22,7 +27,7 @@ done
 NOTFOUND=0
 
 test_cmd() {
-  command -v $1 >/dev/null
+  command -v "$1" >/dev/null
 }
 
 test_flag() {
@@ -110,7 +115,7 @@ fedora_deps() {
 
   # Fedora 41 and above split a part of openssl-devel into openssl-devel-engine.
   # CentOS/RHEL 10 are doing the same thing.
-  if { [ "$VERSION_ID" -ge "41" ] && [ "$ID" == "fedora" ]; } \
+  if { [ "$VERSION_ID" -ge "41" ] && [ "$ID" = "fedora" ]; } \
      || { [ "$VERSION_ID" -ge "10" ] && [ "$ID" != "fedora" ]; }; then
     $YUM install -y 'openssl-devel-engine'
   fi
@@ -119,7 +124,7 @@ fedora_deps() {
 suse_deps() {
   ZYPPER="$SUDO zypper"
   RESOLVE=""
-  if [ "${CI}" == "yes" ] ; then
+  if [ "${CI}" = "yes" ] ; then
     RESOLVE="--allow-downgrade"
   fi
   $ZYPPER install $RESOLVE -yl \
@@ -333,19 +338,20 @@ fallback_method() {
     suse_deps
   fi
 
+  OSTYPE=$(uname -s)
   # OSTYPE is set by bash
   case $OSTYPE in
-    darwin*|msys)
+    Darwin*|msys)
       echo "skipping darwin*/msys"
     ;;
-    freebsd*)
+    FreeBSD*)
       freebsd_deps
     ;;
-    ''|linux-gnu)
+    ''|Linux)
       # catch and known OSTYPE
       echo "\$OSTYPE is '$OSTYPE'"
     ;;
-    netbsd*)
+    NetBSD*)
       netbsd_deps
     ;;
     *)
--- a/get-deps
+++ b/get-deps
@@ -1,10 +1,5 @@
 #!/bin/sh
 
-if [ "$(id -u)" -ne 0 ];then
-	echo "$0 requires superuser privileges. Exiting."
-	exit 1
-fi
-
 TESTING=0
 DOCS=0
 for arg in "$@"; do
@@ -38,16 +33,8 @@ docs_flag() {
   test $DOCS -eq 1
 }
 
-if test -z "$SUDO"; then
-  if test_cmd 'sudo'; then
-    SUDO="sudo"
-  elif test_cmd 'doas'; then
-    SUDO="doas"
-  fi
-fi
-
 alpine_deps() {
-  APK="$SUDO apk"
+  APK="apk"
   $APK add \
     'alpine-sdk' \
     'bash' \
@@ -79,9 +66,9 @@ alpine_deps() {
 
 fedora_deps() {
   if test_cmd 'dnf'; then
-    YUM="$SUDO dnf"
+    YUM="dnf"
   elif test_cmd 'yum'; then
-    YUM="$SUDO yum"
+    YUM="yum"
   else
     echo "No idea what package manager to use, sorry! (perhaps 'dnf' or 'yum' is not in \$PATH?)"
     return 1
@@ -122,7 +109,7 @@ fedora_deps() {
 }
 
 suse_deps() {
-  ZYPPER="$SUDO zypper"
+  ZYPPER="zypper"
   RESOLVE=""
   if [ "${CI}" = "yes" ] ; then
     RESOLVE="--allow-downgrade"
@@ -152,7 +139,7 @@ suse_deps() {
 }
 
 debian_deps() {
-  APT="$SUDO apt-get"
+  APT="apt-get"
   apt-cache show 'libxcb-util-dev' > /dev/null 2>&1 && XCBUTIL="libxcb-util-dev" || XCBUTIL="libxcb-util0-dev"
   $APT install -y \
     'bsdutils' \
@@ -190,7 +177,7 @@ debian_deps() {
 }
 
 arch_deps() {
-  PACMAN="$SUDO pacman"
+  PACMAN="pacman"
   $PACMAN -S --noconfirm --needed \
     'base-devel' \
     'cargo' \
@@ -215,7 +202,7 @@ arch_deps() {
 }
 
 freebsd_deps() {
-  PKG="$SUDO pkg"
+  PKG="pkg"
   $PKG install -y \
     'cmake' \
     'curl' \
@@ -246,7 +233,7 @@ freebsd_deps() {
 }
 
 netbsd_deps() {
-  PKG="$SUDO pkgin"
+  PKG="pkgin"
   $PKG -y install \
     'bash' \
     'fontconfig' \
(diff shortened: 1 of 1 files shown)


jessp01:
This now makes the CI fail with:
```
Run env CI=yes PATH=$PATH ./get-deps
./get-deps requires superuser privileges. Exiting.
Error: Process completed with exit code 1.
```

Because the CI action does not run this with superuser privileges. I do see this in `get-deps`:
```sh
 41 if test -z "$SUDO"; then
 42   if test_cmd 'sudo'; then
 43     SUDO="sudo"
 44   elif test_cmd 'doas'; then
 45     SUDO="doas"
 46   fi
 47 fi 
 ```
 
 But personally, I don't think we should assume every system is configured with `sudo` or `doas` (I know of enough that are not) and that it's better to add this check instead as it's less limiting (users that have sudo/doas configured will just re-run with them, users that do not will `su`). Of course, *if* you decide to go with that, `.github/workflows/gen_*yml` will need to be revised accordingly, and, we should remove explicit invocations of `$SUDO` from the script as well. If you'd rather keep it as is, I can revise the PR accordingly and remove the check.

**Comment by bew** [comment]
I think we can change the shebang to explicitly mention bash (like ~all the other scripts in the repo), and avoid the shellcheck warnings.

For the other changes, could you explain why you want to do them?

**Comment by jessp01** [comment]
Hi @bew ,

> I think we can change the shebang to explicitly mention bash (like ~all the other scripts in the repo), and avoid the shellcheck warnings.

You could, sure, but some OSes (I'm thinking of the .*BSD variety in particular) do not install BASH as part of their base system; I think it's best to be POSIX compliant wherever possible (and it definitely is in this case).

 
> For the other changes, could you explain why you want to do them?

The only other change is adding the check for superuser privileges; as I wrote in my earlier comment:

> personally, I don't think we should assume every system is configured with sudo or doas (I know of enough that are not) and that it's better to add this check instead as it's less limiting (users that have sudo/doas configured will just re-run with them, users that do not will su).

Also, speaking for myself, I don't like it when scripts elevate privileges without explicit consent. IMHO, even if `sudo`/`doas` is configured, it's better to explicitly require elevated privileges and exit if we don't have them.

**Comment by bew** [comment]
> You could, sure, but some OSes (I'm thinking of the .*BSD variety in particular) do not install BASH as part of their base system; I think it's best to be POSIX compliant wherever possible (and it definitely is in this case).

ah good call, didn't think of that
then we should be explicit about this, and add a small comment below shebang about why we use sh instead of bash (to support *BSD variants)

> Also, speaking for myself, I don't like it when scripts elevate privileges without explicit consent. IMHO, even if sudo/doas is configured, it's better to explicitly require elevated privileges and exit if we don't have them.

Ah, I thought it already worked like that (I never used this script, I'm on NixOS 😇)...
👉 Fully agree then, I think that instead of bailing on script start, we could have a new function to exit if not root, and call it at the beginning of each `somedistro_deps` functions when _somedistro_ need root to install its deps, and remove the $SUDO stuff 🤔

This takes care of the platforms that don't require root, and actually bail only when it's necessary with clear error message 🤔

WDYT?

---

Could you actually split this PR in 2 ?
1. posix compliance + small comment under shebang
2. change how root is requested

**Comment by jessp01** [comment]
Hi @bew ,


> Could you actually split this PR in 2 ?
>    posix compliance + small comment under shebang
>    change how root is requested

Sure, happy to do so. Some thoughts, however:

> then we should be explicit about this, and add a small comment below shebang about why we use sh instead of bash (to support *BSD variants)

I don't think we need to explain why we're using the default shell (`/bin/sh`), it's the natural thing to do after all :)

> I think that instead of bailing on script start, we could have a new function to exit if not root, and call it at the beginning of each somedistro_deps functions when somedistro need root to install its deps, and remove the $SUDO stuff 🤔

Since the sole purpose of this script is to install the needed deps, which requires superuser privileges on all supported OSes/distros, could I ask why you feel it's better to invoke this function multiple times from OS/dist specific implementations rather than once in the beginning?

```sh
$ grep SUDO -B1 get-deps

if test -z "$SUDO"; then
  if test_cmd 'sudo'; then
    SUDO="sudo"
  elif test_cmd 'doas'; then
    SUDO="doas"
--
alpine_deps() {
  APK="$SUDO apk"
--
  if test_cmd 'dnf'; then
    YUM="$SUDO dnf"
  elif test_cmd 'yum'; then
    YUM="$SUDO yum"
--
suse_deps() {
  ZYPPER="$SUDO zypper"
--
debian_deps() {
  APT="$SUDO apt-get"
--
arch_deps() {
  PACMAN="$SUDO pacman"
--
freebsd_deps() {
  PKG="$SUDO pkg"
--
netbsd_deps() {
  PKG="$SUDO pkgin"
--
  || (echo 'X is not found in USE flags' && exit 1)
  EMERGE="$SUDO emerge"
--
void_deps() {
  XBPS="$SUDO xbps-install"
--
solus_deps() {
  EOPKG="$SUDO eopkg"
````

Also, having reviewed the script more minutely, I'm not sure why we need both:
```sh
343   case $OSTYPE in
```

And
```sh
365 if test -e /etc/os-release; then
366   . /etc/os-release
367 else
368   ID=$(uname -s | tr '[:upper:]' '[:lower:]')
369 fi
370
371 set -e
372
373 case $ID in
```

Any thoughts/insights on that?

jessp01:
Hi @bew ,

Normally, I'd wait much longer before pinging, but as your response time is superb (well done on that), I just want to make sure you received a notification for this https://github.com/wezterm/wezterm/pull/8010#issuecomment-5190496455?

bew:
> Normally, I'd wait much longer before pinging, but as your response time is superb (well done on that)

yeah well, don't mention it to my employer 🤫 👀

I did get the notification and read your comment, but was on my phone and didn't want to type too much at that time, I guess I can answer now 🤔

> > then we should be explicit about this, and add a small comment below shebang about why we use sh instead of bash (to support *BSD variants)
>
> I don't think we need to explain why we're using the default shell (/bin/sh), it's the natural thing to do after all :)

I still think it's important to mention because I don't have that reflex, and I'd have changed it in a heartbeat if I saw that script ^^ Even more so since all ci and other maintainer scripts are using bash shebang



> Since the sole purpose of this script is to install the needed deps, which requires superuser privileges on all supported OSes/distros, could I ask why you feel it's better to invoke this function multiple times from OS/dist specific implementations rather than once in the beginning?

Well I mis-interpreted the failing build on MacOS as "This platform doesn't need sudo to install deps", but the real reason is that on darwin the script just doesn't do anything ^^'
image

Still, I'd prefer to get an error in the paths where we know we're going to need sudo, a no-op like on darwin shouldn't stop early.

> Also, having reviewed the script more minutely, I'm not sure why we need both:

No idea either, maybe some tricky fallback on some systems, I don't know, I'd leave it as is for now, maybe do another PR pinging Wez later.

jessp01:
Hi @bew ,

> yeah well, don't mention it to my employer 👀

You can rely on my discretion :)

 > I did get the notification and read your comment, but was on my phone and didn't want to type too much at that time, I guess I can answer now
 
 Cheers, certainly did not mean to nag...
 
 > I still think it's important to mention because I don't have that reflex, and I'd have changed it in a heartbeat if I saw that script ^^ Even more so since all ci and other maintainer scripts are using bash shebang
 
 Well, actually, `ci/appstreamcli` , `termwiz/update-changelog.sh` and `assets/open-wezterm-here` all use `#!/bin/sh` as well:) I'd keep them that way, too, nothing in any of them that requires BASH specifically (all just run the one command, really).
 
 > Well I mis-interpreted the failing build on MacOS as "This platform doesn't need sudo to install deps", but the real reason is that on darwin the script just doesn't do anything ^^'
 
 Indeed. In fact, you can just remove the invocation of `get-deps` from `.github/workflows/gen_*mac*`, since it does nothing.
 
 > Still, I'd prefer to get an error in the paths where we know we're going to need sudo, a no-op like on darwin shouldn't stop early.
 
 We can place the check for superuser privileges right before `case $ID in` (line 367), since everything that follows will require superuser privs. IMHO, it makes more sense than adding individual invocations of the same check in all these functions (`.*_deps()`). 
 
> No idea either, maybe some tricky fallback on some systems, I don't know, I'd leave it as is for now, maybe do another PR pinging Wez later.

Oh, it's definitely a fallback (I mean, it's inside a function called `fallback_method()` :)), but since the bit that pertains to `OSTYPE`  (line 337-355) does not actually cover individual distros but only `darwin*|msys` (does nothing), `freebsd*`, `linux-gnu` and `netbsd*`  and seeing how these broad detections can be (and *are*) done using `uname -s` (line 362) unless `ID` is set in `/etc/os-release`, there's no added value to that block in my view and I'd recommend its removal.

bew:
> Indeed. In fact, you can just remove the invocation of `get-deps` from `.github/workflows/gen_*mac*`, since it does nothing.

Feel free to open a PR to update the `gen_workflows.py` script :P 
I might merge it before this one so you can assume macos won't break 🤔

> Well, actually, [...] all use #!/bin/sh as well:)

I stand corrected, will try to pay attention to the /bin/sh next times 😀

> but since the bit that pertains to `OSTYPE` (line 337-355) does not actually cover individual distros but only `darwin*|msys` (does nothing), `freebsd*`, `linux-gnu` and `netbsd*` and seeing how these broad detections can be (and _are_) done using `uname -s` (line 362) unless `ID` is set in `/etc/os-release`, there's no added value to that block in my view and I'd recommend its removal.

open another PR after this one if you want 😉

---

Another alternative, if you're willing to follow me there:
I think it would be nice to have a kind of dry-mode when sudo/.. was not used 🤔 with clear message like a `DRY-MODE:` prefix and a line at beginning telling user to use sudo/.. to actually install stuff 🤔

WDYT?

jessp01:
Hi @bew .

> Feel free to open a PR to update the gen_workflows.py script :P

See https://github.com/wezterm/wezterm/pull/8029

I've made the `get-deps` changes I suggested in this commit: https://github.com/wezterm/wezterm/pull/8010/commits/e48ac2be6ab2a1d8758fc62fe312b2c9c2de6e61, I believe it covers everything, let me know what you think?

> Another alternative, if you're willing to follow me there:
> I think it would be nice to have a kind of dry-mode when sudo/.. was not used 🤔 with clear message like a DRY-MODE: prefix and a line at beginning telling user to use sudo/.. to actually install stuff 🤔

Maybe I am misunderstanding something here, but I'm not sure what value such a "dry-run" op would provide? I mean, if one just wants to review the needed packages without actually installing, one can simply read the script, no? :)

PR #8012, opened 2026-08-04
sixel: default omitted `!` repeat count to 1

Osamaali313:
## Summary

In the sixel parser (`wezterm-escape-parser/src/parser/sixel.rs`), the `!` (DECGRI "graphics repeat") handler reads the repeat count as:

```rust
repeat_count: self.params[0] as u32,
```

`params` is initialised to `-1` (the "parameter unspecified" sentinel). When the repeat count is omitted, `params[0]` is still `-1`, and `(-1i64) as u32` wraps to **4294967295**. So a sequence like `ESC P q ! ~ ESC \` yields `Repeat { repeat_count: 4294967295 }`.

The consumer (`term/src/terminalstate/sixel.rs`) then runs `for _ in 0..repeat_count { … }`, i.e. ~4.3 billion iterations for a handful of input bytes — a significant CPU/UI stall on the output-processing path, reachable by anything that writes to the terminal (a `cat`-ed file, or a remote host over SSH). In a debug build the `x` counter also overflow-panics.

The sibling `"` raster-attributes branch already special-cases the `-1` sentinel (e.g. `pan = if self.params[0] == -1 { 2 } else { … }`); the repeat branch just missed it.

## Fix

Per the DEC spec, an omitted `Pn` for `!` defaults to `1`. Handle the sentinel the same way:

```rust
let repeat_count = if self.params[0] < 0 {
    1
} else {
    self.params[0] as u32
};
```

An explicit repeat count is unchanged. Added `sixel_repeat_omitted_count_defaults_to_one` to the existing test module; `cargo test -p wezterm-escape-parser` passes.

### DIFF
--- a/wezterm-escape-parser/src/parser/sixel.rs
+++ b/wezterm-escape-parser/src/parser/sixel.rs
@@ -53,8 +53,18 @@ impl SixelBuilder {
                 self.sixel.data.push(SixelData::NewLine);
             }
             0x3f..=0x7e if self.current_command == b'!' => {
+                // An omitted repeat count leaves params[0] at the -1 sentinel;
+                // per DEC, `!` (DECGRI) defaults to 1 when Pn is absent. Without
+                // this, `-1 as u32` becomes 4294967295 and the consumer loops
+                // billions of times on a tiny input. Mirrors the `"` raster
+                // branch's -1 handling below.
+                let repeat_count = if self.params[0] < 0 {
+                    1
+                } else {
+                    self.params[0] as u32
+                };
                 self.sixel.data.push(SixelData::Repeat {
-                    repeat_count: self.params[0] as u32,
+                    repeat_count,
                     data: data - 0x3f,
                 });
                 self.finish_command();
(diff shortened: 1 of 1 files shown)


wezterm-escape-parser/src/parser/sixel.rs
```diff
@@ -193,6 +203,41 @@ mod test {
...
+        };
+        assert_eq!(
+            sixel.data,
+            vec![SixelData::Repeat {
+                repeat_count: 1,
+                data: 0x3f
+            }]
+        );
+
+        // An explicit repeat count is still honored.
```
bew, line 225: Please split this into a separate test 🙏

wezterm-escape-parser/src/parser/sixel.rs
```diff
@@ -53,8 +53,18 @@ impl SixelBuilder {
                 self.sixel.data.push(SixelData::NewLine);
             }
             0x3f..=0x7e if self.current_command == b'!' => {
+                // An omitted repeat count leaves params[0] at the -1 sentinel;
+                // per DEC, `!` (DECGRI) defaults to 1 when Pn is absent. Without
+                // this, `-1 as u32` becomes 4294967295 and the consumer loops
+                // billions of times on a tiny input. Mirrors the `"` raster
+                // branch's -1 handling below.
```
bew, line 60: Please link to the PR for context 🙏

```suggestion
                // branch's -1 handling below. #8012
```

wezterm-escape-parser/src/parser/sixel.rs
```diff
@@ -193,6 +203,41 @@ mod test {
     use alloc::boxed::Box;
     use k9::assert_equal as assert_eq;
 
+    #[test]
+    fn sixel_repeat_omitted_count_defaults_to_one() {
+        // `!` (DECGRI) with no repeat count must default to 1. Previously the
```
bew, line 208: ⚠️ "Previously" only makes sense in the context of this change, please rephrase this to be more context agnostic.
And link to the pr like mentioned above, near end of comment.

bew (commented):
Hello, thanks for the PR !

How did you find out about this issue?

Also, how much AI was used and how much / what kind of manual testing did you do? 🤔

Osamaali313:
Hey @bew, thanks for taking a look!

I hit it using wezterm day-to-day — some sixel output rendered wrong, and when I traced it, the omitted-count case after the `!` introducer was the culprit. Per the spec an omitted repeat count should default to 1, which is what the PR restores.

On process: I read through the parser and wrote the fix by hand. Where AI came in was verification — I had a Claude Code–based QA agent exercise the change end-to-end (the omitted-count path plus the explicit-count cases around it) so I wasn't just assuming it was correct, and to confirm it doesn't regress the normal sixel path. Happy to add a specific repro sequence or more test coverage if that'd help review.

j4james:
> Per the spec an omitted repeat count should default to 1

Exactly which spec are you're referring to here?

> I hit it using wezterm day-to-day — some sixel output rendered wrong

Do you recall which application you were using that produced the issue? Because this seems like a bug that's worth reporting to them too.

PR #8013, opened 2026-08-05
termwiz(lineedit): Add Ctrl-u for Kill to StartOfLine

bew:
I've been missing that keybind many times before e.g. when renaming my tabs using a `PromptInputLine`!

I initially added this to the prompt line overlay, but since similar Ctrl-k was already implemented to Kill to EndOfLine on the LineEditor in termwiz, adding Ctrl-u there felt like a logical addition.
.. Not sure why it wasn't impl from the beginning 🤔


wez:
No good reason other than it just wasn't in the initial set of things that I put in there!  This looks good to me, thanks!

PR #8014, opened 2026-08-05
termwiz: Add Action::ClearLine for LineEditor

bew:
This was needed to implement #8015 
Tested manually with #8015 

I'm not 100% sure about the clear impl, especially around how the search state works and should be managed 🤔

### DIFF
--- a/termwiz/src/lineedit/mod.rs
+++ b/termwiz/src/lineedit/mod.rs
@@ -702,6 +702,11 @@ impl<'term> LineEditor<'term> {
 
                 self.state = EditorState::Accepted;
             }
+            Action::ClearLine => {
+                self.clear_completion();
+                self.cancel_search_state();
+                self.line.clear();
+            }
             Action::EndOfFile => {
                 return Err(
                     std::io::Error::new(std::io::ErrorKind::UnexpectedEof, "End Of File").into(),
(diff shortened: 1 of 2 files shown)


termwiz/src/lineedit/mod.rs
```diff
@@ -702,6 +702,11 @@ impl<'term> LineEditor<'term> {
 
                 self.state = EditorState::Accepted;
             }
+            Action::ClearLine => {
+                self.clear_completion();
+                self.cancel_search_state();
+                self.line.set_line_and_cursor("", 0);
```
wez: if this works, this reads better:

```suggestion
                self.line.clear();
```

wez (approved):
looks good; one minor suggestion which I think should work and would be slightly clearer

termwiz/src/lineedit/mod.rs
bew: ah yes, missed it at 4am 😅

fixed with a force push 946f3ac5399de4f8590f3937d70fdf477a48d5f6

bew:
Thanks, I definitely miss stuff like that at 4am 😅👀

Tested again with #8015, works well 👍

bew:
Windows is slow, but ci is ✅ otherwise, let's merge!

PR #8015, opened 2026-08-05
overlays(prompt,debug): Clear line before Cancel when pressing Escape

bew:
This helps e.g. in the debug overlay, to avoid exiting it by easy mistake. This is especially for people that are used to modal editors (vim, ..) where pressing Escape becomes a _reflex_ after some text input 😬

note: recent commit 63061d7797fecd37366e605d39daddc105a43698 refactored these `resolve_action` impl to be more easily extensible.

### DIFF
--- a/wezterm-gui/src/overlay/debug.rs
+++ b/wezterm-gui/src/overlay/debug.rs
@@ -106,7 +106,10 @@ impl LineEditorHost for LuaReplHost {
             InputEvent::Key(KeyEvent {
                 key: KeyCode::Escape,
                 ..
-            }) => line.is_empty().then_some(Action::Cancel),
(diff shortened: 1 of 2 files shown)


wez:
Does this mean that if you double tap escape by accident, you'll now lose your line content and pop out of the overlay? That feels a bit more likely to surprise/frustrate?

bew:
> Does this mean that if you double tap escape by accident, you'll now lose your line content and pop out of the overlay? That feels a bit more likely to surprise/frustrate?

Well current Wezterm behavior is "you loose your line on FIRST escape", so this is slightly better 👀

We could think about Nop'ing Escape, but it's kinda nice to be able to escape out of a prompt with Escape... like many other prompts...

Issue #6739, opened 2025-02-28
Bottom tab bar design


sanpii:
**Is your feature request related to a problem? Please describe.**
Tab bar at bottom is just put at the bottom of the window.

**Describe the solution you'd like**
For a better design, the tabs should be reversed (rounded edges at the bottom) and moved up a little.

Actual:
![Image](https://github.com/user-attachments/assets/edb9f39b-10b7-4b85-80da-53ac9764f0c7)

Suggestion:

![Image](https://github.com/user-attachments/assets/a81bd288-5d28-4e4e-9e98-dab17c7172ca)

doctorkotik187:
I will do it.

hphng:
hey can I do this?

bew:
> hey can I do this?

Sure 😉

PR #8018, opened 2026-08-05
Bottom tab bar design

hphng:
hi @bew, I just add the botttom tab bar design (issue #6739 )

image

Also, there a bug in `box_model` that the right border (the R piece) got wrong in calculation for **bottom bar mode** because x-position calculate only with `top_right_width`.

Otherwise, I just check the state of configuration, and add `if..else` where it necessary

This is my first time to make a opensouce contribution, so I will make some rookie mistakes. If there is anything I need to improve, please show me and I am really appreciate it!!


bew:
Hello, welcome to opensource! 🥳 🚀 

I noticed that you didn't fetch latest `main` branch before starting to work on this, you had ~1300 commits missing ^^'

I did a merge on my local side to be able to test your PR with latest changes, but I'll let you do it on your own for the branch ;)
You basically have 2 choice:
- merge main into your branch (easier if there are many commits already in your branch, but gives a bit of a bad history in git log graph)
- or rebase your branch on top of main (easier if there are few commits, nice history in git log graph)

In this repo I tend to merge in 'squash-and-merge' way, so the details of your feature branch is not important, it'll be all _'squashed'_ together at merge-time into a single commit anyway.

So you have the choice here, which one will you choose :P 

image

bew:
Hmm I just tested it on MacOS, and this is what I see 🤔

image
image


(and I do not see the gap when tab bar is at the top)

Do you maybe see the same thing after you merged/rebased with main?
If not it might be a difference between platforms 🤔
.. Which platform did you test this on?

hphng:
Hi @bew, thanks for the welcoming message!

I think when I tried to fork the repo last week I did not regularly update it somehow. Since it's not important, I will choose the easier for me - merge main into this branch.

Also, after I merge and test it's totally fine (no gap)

I tested it on Window 11


image

hphng:
is there anything I need to do in this case?

bew:
NOTE: I can also see this gap on Linux 👀:
Screenshot_20260810_054919

> is there anything I need to do in this case?

Unless you have access to a Linux or MacOS machine to try to understand what might be going on, you could try guessing and asking me ( / other people who might want to review / help test your PR) to test it, or me / other people might do the debugging for you.

Since it seems to be only working on Windows, I'd almost guess this is a Windows bug ^^' you could try to see if it's related to the UI zoom (aka try with bigger zoom / increasing wezterm' dpi config to see if you see the gap or not 🤔

---
NOTE: helpful config file for testing this PR:
```lua
return {
  tab_bar_at_bottom = true,
}
```

hphng:
hi @bew, I already tested both on Mac and Window

157137ec-442c-4798-8b1f-d4ee53e13151

after playing around a bit, I realize that the padding between top and bottom is not equal in the tab bars, hence make a small gap between them. I already fix it :)

bew:
Any idea why the gap didn't show up in your windows tests earlier?

hphng:
I am not 100% sure, but I guess it is about of the way calculate the height of the tab bar. Padding top change from 0.2 -> 0.25. so the tab bar is a little bit bigger, and it filled into the gap.

I know that Window, Mac, and Linux have different DPI and scaling baseline (Mac has Retina/HiDPI I believe), which mean the height is a little bit different ?

Also, for the padding and margin, there are a lot of **magic number**, which is not planned so that maybe why we get a small gap between these twerks.

I think we need to investigate more the root caused of this

bew:
Hey, I tried on Linux, I still get a small gap 😬 (I zoomed my display at 200% here)
image

instead of tweaking .05 stuff, it'd be better to try and understand where the gap is coming from / why we need to tweak these details I think 🤔

Also, I'd recommend we setup some rust constants for the magic numbers used around the tab bar here!

_EDIT: Found it! The tabs were aligned to the bottom of the container in all cases, fixed in https://github.com/wezterm/wezterm/pull/8018/changes/d7ef946466948a844d69ebd97d3dd1b88ea48c82_

Issue #2162, opened 2022-06-22
Weird line wrapping behaviour


Funami580:
### What Operating System(s) are you seeing this problem on?

Linux Wayland

### WezTerm version

wezterm 20220622-091434-80671f29

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

When running bash and you resize the terminal width so that the lines wrap, the lines get duplicated? (See gif below). I couldn't notice this behaviour with Konsole.

With `echo`ing some line and wrapping that, I couldn't notice the same behaviour. The text stayed the same, but the cursor position changed to the next line. (See gif below).

### To Reproduce

For 1st scenario: Run `bash` and resize terminal width.
For 2nd scenario: Run `clear && echo -n some long long text && read` and resize terminal width.

### Configuration

no config

### Expected Behavior

No duplicated lines and no cursor position changes.

### Logs

21:52:13.064  INFO   wezterm_gui::termwindow > OpenGL initialized! Mesa Intel(R) HD Graphics 4600 (HSW GT2) 4.6 (Compatibility Profile) Mesa 22.1.1 is_context_loss_possible=false wezterm version: 20220622-091434-80671f29

### Anything else?

Duplicated lines:
![bash](https://user-images.githubusercontent.com/63090225/175124300-31a149b0-dcad-43b0-9f67-a9dbdbebdd8c.gif)

Cursor position changes:
![cursor](https://user-images.githubusercontent.com/63090225/175124315-9dbee57d-4afc-4095-a1b3-3ea05226626b.gif)

wez:
This should be fixed now in `main`.

It typically takes about an hour before fixes are available as nightly builds for all platforms.  Linux builds are the fastest to build and are often available within about 20 minutes.  Windows and macOS builds take a bit longer.

Please take a few moments to try out the fix and let me know how that works out. You can find the nightly downloads for your system in the [wezterm installation docs](https://wezfurlong.org/wezterm/installation.html).

If you prefer to use packages provided by your distribution or package manager of choice and don't want to replace that with a nightly download, keep in mind that you can download portable packages (eg: a `.dmg` file on macOS, a `.zip` file on Windows and an `.AppImage` file on Linux) that can be run without permanently installing or replacing an existing package, and can then simply be deleted once you no longer need them. 

[...]

PR #8021, opened 2026-08-05
term: preserve cursor at start of line during rewrap

### LINKED ISSUE #7994: [Windows] Width resize corrupts cursor position when it is at column 0 on a non-first line

Windows

_No response_

### WezTerm version

20260731-083216-d69264df

### Describe the bug

On Windows, resizing the WezTerm window while a process is running causes the cursor position to become incorrect. Subsequent output is rendered at the wrong location.

Example: run `ping localhost -t`

Image

The problem reproduces with `wezterm.exe -n` and with cmd.exe, PowerShell, and WSL. Windows Terminal does not reproduce.

### To Reproduce

1. Start WezTerm with clean config:

    `wezterm.exe -n`

2. Run:

    `python -c "import time; print('start'); time.sleep(5); print('end')"`

3. Resize (horizontally stretch) the window during sleep.

    Expected:

    ```
    start
    end
    ```

    Actual:

    ```
    start                                                              end
    ```

### Configuration

no config

### Expected Behavior

Like Windows Terminal:

Image

### Logs

Debug Overlay
wezterm version: 20260731-083216-d69264df x86_64-pc-windows-msvc
Window Environment: Windows
Lua Version: Lua 5.4
OpenGL: AMD Radeon(TM) Graphics 4.5.0 Core Profile Context 22.20.44.221025
Press ESC or CTRL-D to exit

### Anything else?

_No response_

EastMonster:
> [!NOTE]
> This change was developed with assistance from OpenAI Codex. I reviewed and verified the proposed fix, tests, and PR description.

## Summary

- preserve a cursor at the start of a logical line when the terminal width changes
- restrict the wrapped-line cursor adjustment to non-zero logical offsets
- add a regression test for #7994

Fixes #7994.

## Root cause

When the terminal width changes, `Screen::rewrap_lines` converts the cursor position into an offset within its logical line and then calculates its new physical row and column.

The existing special case handles a cursor that lands exactly at a wrapping boundary:

https://github.com/wezterm/wezterm/blob/4b1c3c151eb530e569f867e1461693c56fe89695/term/src/screen.rs#L142-L165

It moves the cursor back to the preceding physical row so that the cursor remains associated with the wrapped logical line.

However, this condition also matches an ordinary cursor at the start of a logical line, where the logical offset `x` is zero. For example, after:

```
start\r\n
```

the cursor is normally at `(x=0, y=1)`. When the terminal is widened, the special case incorrectly treats it as a wrapping boundary, moves it to the previous row, and sets its column to `physical_cols`.

The subsequent output is therefore rendered at the wrong position.

The fix is to apply the special case only when the logical offset is non-zero:

```rust
if x > 0 && adjusted_cursor.0 == 0 && adjusted_cursor.1 > 0 {
```

A positive logical offset that maps to physical column zero represents a real wrapping boundary. A zero logical offset represents the ordinary start of a logical line and should not be adjusted.

Although #7994 was reproduced on Windows, this code is part of the platform-independent terminal reflow implementation.

## Relation to #2162
The special-case adjustment was introduced in commit [4bf89e0c1](https://github.com/wezterm/wezterm/commit/4bf89e0c1062b04b82699f0911d3936c0d9ed0a0) to address cursor drift at exact wrapping boundaries reported in #2162.

This change keeps that behavior for non-zero logical offsets while excluding the distinct `x == 0` case. In other words, it narrows the #2162 workaround rather than removing it.

The existing `test_resize_2162*` tests continue to pass.

## Testing
The new regression test:

1. writes `start\r\n`
2. verifies that the cursor is at `(0, 1)`
3. changes the terminal column count
4. verifies that the cursor remains at `(0, 1)`

Tests run:
```console
cargo test -p wezterm-term test_resize_cursor_issue_7994
cargo test -p wezterm-term test_resize_2162
```

### DIFF
--- a/term/src/screen.rs
+++ b/term/src/screen.rs
@@ -149,7 +149,7 @@ impl Screen {
                 // line and it won't resize with the line correctly.
                 // Put it back on the prior line. The cursor is now
                 // technically outside of the viewport width.
-                if adjusted_cursor.0 == 0 && adjusted_cursor.1 > 0 {
+                if x > 0 && adjusted_cursor.0 == 0 && adjusted_cursor.1 > 0 {
                     if physical_cols < self.physical_cols {
                         // getting smaller: preserve its original position
                         // on the prior line
--- a/term/src/test/mod.rs
+++ b/term/src/test/mod.rs
@@ -916,6 +916,22 @@ fn test_resize_2162() {
     term.assert_cursor_pos(19, 0, None, Some(7));
 }
 
+/// A cursor with logical x=0 is at the start of a logical line,
+/// not a wrapped continuation.
+#[test]
+fn test_resize_cursor_issue_7994() {
+    let mut term = TestTerm::new(4, 20, 0);
+    term.print("start\r\n");
+    term.assert_cursor_pos(0, 1, None, None);
+
+    term.resize(TerminalSize {
+        rows: 4,
+        cols: 21,
+        ..Default::default()
+    });
+    term.assert_cursor_pos(0, 1, None, None);
+}
+
 /// Test the behavior of wrapped lines when we resize the terminal
 /// wider and then narrower.
 #[test]


bew:
Hello again, thank you for the PR, and such a small change!

Quick read through it makes sense, but I'll have to dive a little another time to verify everything here 👍

PR #8022, opened 2026-08-05
cairo: Remove unused features

xtqqczze:
Remove unused `cairo-sys-rs` features that cause `cargo check --all-features` at the workspace root to fail.

```console
$ cargo check -q -p cairo-sys-rs --all-features
error[E0463]: can't find crate for `x11`
  --> deps/cairo/src/lib.rs:12:1
   |
12 | extern crate x11;
   | ^^^^^^^^^^^^^^^^^ can't find crate

For more information about this error, try `rustc --explain E0463`.
error: could not compile `cairo-sys-rs` (lib) due to 1 previous error
```

Compared to commit d7ede9fb04fa59f334986e08dcfa99ba065d9d1a, this approach is cleaner because it removes the unnecessary feature flags instead of working around the issue.

### DIFF
--- a/deps/cairo/Cargo.toml
+++ b/deps/cairo/Cargo.toml
@@ -22,14 +22,13 @@ pdf = []
 svg = []
 ps = []
 script = []
-win32-surface = []
-xlib = []
-xcb = []
-freetype = []
-use_glib = []
 
 [dependencies]
 libc.workspace = true
 
 [build-dependencies]
 cc.workspace = true
+
+[lints.rust]
+# Avoid diagnostics for features that are no longer defined but are still referenced by conditional compilation.
+unexpected_cfgs = { level = "warn", check-cfg = ['cfg(feature, values("freetype","use_glib","win32-surface","x11","xcb","xlib"))'] }


bew:
Hello!

I'm not familiar with the part of the project nor with the issue you're mentioning.
Could you explain a bit more about this?

Also, i'd prefer to keep dependencies-related stuff in cargo.toml, quick chat with an AI pointed me towards the `[lints.rust] unexpected_cfgs` options (https://doc.rust-lang.org/rustc/check-cfg/cargo-specifics.html#check-cfg-in-lintsrust-table), could we use this instead?

note: in build.rs or cargo.toml (wherever we'll put that in the end), could you add a relatively short comment above it explaining why we need to do that?

xtqqczze:
I originally had these changes in `Cargo.toml`, but moved them to `build.rs` because `cairo-sys-rs` is a vendored crate. I thought it would be better to keep `Cargo.toml` as close as possible to the original upstream source.

bew:
I don't really have an opinion of this to debate here.
I'll let Wez decide if he comes around 🤔

wez:
IMO, there is no usecase to test all compilation features for this crate, so it's difficult to find the motivation to dig in and get more opinionated about whether this is better than things are now: it feels like a gateway to more use-cases that are outside of the intended build flow, and it's hard to justify wanting to spend time on either of these concerns.

What's the higher level goal here?  It feels like this might be a result of AI-assisted code exploration to find technicalities more than an actual real issue/concern that helps either our users or maintainers.

xtqqczze:
@wez `cargo check --all-features` is in my command history and I tend to use it a lot when developing. Currently this fails at the repository root which is annoying. The only goal here is convenience.

Issue #7797, opened 2026-05-17
[Tracking Issue] Switch winapi dependency to windows/windows-sys


atouchet:
**Is your feature request related to a problem? Please describe.**
The winapi crate used as a dependency by the wezterm repository is no longer maintained.

**Describe the solution you'd like**
The wezterm repository should transition to using the windows/windows-sys crates instead of winapi.

Crates to update:

- [x] config (#8098)
- [ ] env-bootstrap (#8080)
- [x] filedescriptor (#8026)
- [x] mux (#8072)
- [x] portable-pty (#8073)
- [ ] procinfo
- [x] spawn-funcs (#8069)
- [x] termwiz (#8099)
- [ ] wezterm
- [ ] wezterm-client
- [ ] wezterm-escape-parser (#8107)
- [ ] wezterm-font (#8106)
- [ ] wezterm-gui
- [ ] wezterm-mux-server
- [x] wezterm-mux-server-impl (#8072)
- [x] wezterm-open-url (#8101)
- [ ] window

xtqqczze:
Could we specifically migrate the `filedescriptor` crate?

bew:
> Could we specifically migrate the `filedescriptor` crate?

@xtqqczze  PR welcome :)

bew:
FYI, in 2023 @strottos attempted a one-shot migration to windows-rs in #4635.
Going incrementally as is the plan here should be easier to review and work on module by module.

strottos:
For the record, it was working as far as I could tell but I ran out of steam and time. I agree though oneshotting like this was not the way, incremental would have been better.

bew:
> (from https://github.com/wezterm/wezterm/pull/4635#issuecomment-5354540829 by @xtqqczze )
> For migrating from `winapi`, it seems like `windows-sys` seems like the better fit and might even make it possible to do in one shot.
> The higher-level `windows` crate could then be considered as part of a larger refactoring.

I agree on the 2-staged migration, using the `windows` crate would help Rustify the Windows side a bit more, but I'd still prefer to go incrementally for the first stage to help simplify the reviews 🤔

xtqqczze:
We could also switch the `ntapi` dependency to `windows-sys`, e.g. [`windows_sys::Wdk::System::Threading::NtQueryInformationProcess`](https://docs.rs/windows-sys/latest/windows_sys/Wdk/System/Threading/fn.NtQueryInformationProcess.html)

bew:
[...]

PR #8026, opened 2026-08-06
filedescriptor: Replace winapi with windows-sys

xtqqczze:

### DIFF
--- a/filedescriptor/src/windows.rs
+++ b/filedescriptor/src/windows.rs
@@ -8,21 +8,25 @@ use std::os::windows::prelude::*;
 use std::ptr;
 use std::sync::Once;
 use std::time::Duration;
-use winapi::shared::ws2def::{AF_INET, INADDR_LOOPBACK, SOCKADDR_IN};
-use winapi::um::fileapi::*;
-use winapi::um::handleapi::*;
-use winapi::um::minwinbase::SECURITY_ATTRIBUTES;
-use winapi::um::namedpipeapi::{CreatePipe, GetNamedPipeInfo};
-use winapi::um::processenv::{GetStdHandle, SetStdHandle};
-use winapi::um::processthreadsapi::*;
-use winapi::um::winbase::{FILE_TYPE_CHAR, FILE_TYPE_DISK, FILE_TYPE_PIPE};
-use winapi::um::winnt::HANDLE;
-use winapi::um::winsock2::{
-    accept, bind, closesocket, connect, getsockname, getsockopt, htonl, ioctlsocket, listen, recv,
-    send, WSAGetLastError, WSAPoll, WSASocketW, WSAStartup, INVALID_SOCKET, SOCKET, SOCK_STREAM,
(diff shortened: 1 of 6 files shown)


pty/Cargo.toml
```diff
@@ -33,6 +33,7 @@ winapi = { workspace=true, features = [
     "handleapi",
     "fileapi",
     "namedpipeapi",
+    "processenv",
```
bew, line 36: what is this change about? 🤔 why is it needed _now_ ?

xtqqczze:
https://www.githubstatus.com/incidents/qcvjkzcs7j74

mux/Cargo.toml
```diff
@@ -57,6 +57,7 @@ winapi = { workspace=true, features = [
     "psapi",
     "processthreadsapi",
     "tlhelp32",
+    "winsock2"
```
bew, line 60: Why is this change needed?
The idea is to avoid using winapi stuff anymore, not add enabled features for it 🤔

filedescriptor/src/windows.rs
```diff
@@ -530,7 +526,7 @@ pub fn socketpair_impl() -> Result<(FileDescriptor, FileDescriptor)> {
         }
     }
 
-    let client = socket(AF_INET, SOCK_STREAM, 0)?;
+    let client = socket(AF_INET as _, SOCK_STREAM, 0)?;
```
bew: nit: I think it would be nicer to change `socket`'s first arg type to [`ADDRESS_FAMILY`](https://docs.rs/windows-sys/0.61.2/windows_sys/Win32/Networking/WinSock/type.ADDRESS_FAMILY.html) and do the `as _` conversion in `socket` before calling `WSASocketW` 🤔

bew (commented):
Hello! Changes look good overrall, it's nice that the functions/constant names map the same way to Windows' API so the changes are ~only about imports, not logic!

A few small comments on the cargo stuff though 🤔
(I'm not super familiar with cargo workspaces as they are used a lot in this project, yet, so it might just be some misunderstanding on my part)

pty/Cargo.toml
xtqqczze: Upon removing `winapi` from `filedescriptor`, we now need to specify `winapi` features elsewhere, otherwise there are compile errors.

```
error[E0432]: unresolved import `winapi::um::processenv`
   --> pty/src/cmdbuilder.rs:107:25
    |
107 |         use winapi::um::processenv::ExpandEnvironmentStringsW;
    |                         ^^^^^^^^^^ could not find `processenv` in `um`
    |
```

mux/Cargo.toml
xtqqczze: Same as above

bew:
Alright, thanks for the PR! 🚀

xtqqczze:
@bew Maybe we could release a new version of the `filedescriptor` crate using edition 2021 (https://github.com/wezterm/wezterm/pull/7976). Then, once that’s done, update everything to edition 2024?

xtqqczze:
https://github.com/wezterm/wezterm/pull/8075

bew:
I don't have control over crate releases. I think the crates will all get released in bulk when we do the next Wezterm release 🤔

PR #8027, opened 2026-08-06
feat(selector): add ctrl-u and ctrl-w filter shortcuts

JafarAbdi:
Support the following shortcuts in the `InputSelector` filter:

- `Ctrl-u` clears the filter
- `Ctrl-w` deletes the trailing word, via `LineEditBuffer` so it behaves exactly like `LineEditor`'s Ctrl-w

Also fixes `Ctrl-w`/`Ctrl-u` inserting a literal `W`/`U` into the filter (the catch-all `Char(c)` arm ignores modifiers).

Similar to https://github.com/wezterm/wezterm/pull/8013, but for the `InputSelector` filter.

### DIFF
--- a/wezterm-gui/src/overlay/selector.rs
+++ b/wezterm-gui/src/overlay/selector.rs
@@ -12,6 +12,7 @@ use std::rc::Rc;
 use termwiz::cell::{AttributeChange, CellAttributes};
 use termwiz::color::ColorAttribute;
 use termwiz::input::{InputEvent, KeyCode, KeyEvent, Modifiers, MouseButtons, MouseEvent};
+use termwiz::lineedit::{LineEditBuffer, Movement};
 use termwiz::surface::{Change, Position};
 use termwiz::terminal::Terminal;
 use termwiz_funcs::truncate_right;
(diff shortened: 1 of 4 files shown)


wezterm-gui/src/overlay/selector.rs
```diff
@@ -311,6 +312,22 @@ impl SelectorState {
...
+                    let mut buffer = LineEditBuffer::new(&self.filter_term, self.filter_term.len());
+                    buffer.kill_text(Movement::BackwardWord(1), Movement::BackwardWord(1));
+                    self.filter_term = buffer.get_line().to_string();
+                    self.update_filter();
+                }
+                InputEvent::Key(KeyEvent {
+                    key: KeyCode::Char('U'),
+                    modifiers: Modifiers::CTRL,
+                }) if self.filtering => {
+                    self.filter_term.clear();
```
bew: hmm ctrl-u is not clear, it's clear from cursor to start of line (rest of line after cursor shouldn't be deleted)

bew:
> Also fixes `Ctrl-w`/`Ctrl-u` inserting a literal `W`/`U` into the filter (the catch-all `Char(c)` arm ignores modifiers).

Could you update it to only insert c if no modifiers?

wezterm-gui/src/overlay/selector.rs
JafarAbdi: Addressed in the last commit

bew:
Hey, I tried to test your change a few days ago but didn't see it working at all.
I just realized that I was testing on the TabNavigator (the `launcher` overlay) and assumed that they used the same underlying implementation. TIL: they don't 😬

Could you impl the same change for the launcher overlay as well? 🤔
Thanks!

JafarAbdi:
> Hey, I tried to test your change a few days ago but didn't see it working at all. I just realized that I was testing on the TabNavigator (the `launcher` overlay) and assumed that they used the same underlying implementation. TIL: they don't 😬
> 
> Could you impl the same change for the launcher overlay as well? 🤔 Thanks!

Yes, I'll add that later today. Thanks for the review!

JafarAbdi:
> Hey, I tried to test your change a few days ago but didn't see it working at all. I just realized that I was testing on the TabNavigator (the `launcher` overlay) and assumed that they used the same underlying implementation. TIL: they don't 😬
> 
> Could you impl the same change for the launcher overlay as well? 🤔 Thanks!

Sorry for delay. I pushed the changes! I also added it to the command palette + CharSelect, and unified them to make the workflow hopefully consistent.

Please let me know if I need to change anything

P.S. I used AI to write the tests and part of the changes here. I made sure to review them carefully for AI slop

PR #8028, opened 2026-08-06
fix(async-ossl): vendor OpenSSL for musl targets

JafarAbdi:
Cross-compiling for `*-linux-musl` fails in `openssl-sys`: there is no musl-compatible system OpenSSL to link against. Use the vendored OpenSSL for musl, as already done for Windows and macOS.

Motivation: I'm building a web frontend for the wezterm mux (xterm.js in the browser talking to `wezterm-mux-server`) which I deploy as a static musl binary.

Note: native musl builds (e.g. Alpine #7390) switch from system to vendored OpenSSL with this change.

### DIFF
--- a/async_ossl/Cargo.toml
+++ b/async_ossl/Cargo.toml
@@ -11,9 +11,9 @@ publish = false
 [dependencies]
 async-io.workspace = true
 
-[target.'cfg(not(any(windows, target_os="macos")))'.dependencies]
+[target.'cfg(not(any(windows, target_os="macos", target_env="musl")))'.dependencies]
 openssl.workspace = true
 
-[target.'cfg(any(windows, target_os="macos"))'.dependencies]
+[target.'cfg(any(windows, target_os="macos", target_env="musl"))'.dependencies]
 openssl = { workspace = true, features=["vendored"] }
 


bew:
Looks reasonable 🤔

could you add a mention in some musl docs if you find any? (can't check right now)

> Motivation: I'm building a web frontend for the wezterm mux (xterm.js in the browser talking to wezterm-mux-server) which I deploy as a static musl binary.

Well, that's cool, keep it up! 🚀

JafarAbdi:
> Looks reasonable 🤔
> 
> could you add a mention in some musl docs if you find any? (can't check right now)

Thanks for the review! I couldn't find any musl related docs. Should I create one, or update https://github.com/wezterm/wezterm/blob/main/docs/install/source.md? Cheers!

bew:
Hey, thanks for getting back to me here!
> > Looks reasonable 🤔
> > could you add a mention in some musl docs if you find any? (can't check right now)
> 
> Thanks for the review! I couldn't find any musl related docs. Should I create one, or update [`main`/docs/install/source.md](https://github.com/wezterm/wezterm/blob/main/docs/install/source.md?rgh-link-date=2026-08-25T13%3A22%3A26Z)? Cheers!

Yes, `source.md` sounds good, since we don't have musl docs yet I'll merge this as-is, and I let you make a separate PR with some musl docs if you can 🙏

bew:
Also, could you confirm that #7390 is correct / is still needed?

PR #8029, opened 2026-08-07
`get-deps` does not actually do anything when run on MacOS/Darwin

jessp01:
Remove its invocation for that ENV.

### DIFF
--- a/.github/workflows/gen_macos.yml
+++ b/.github/workflows/gen_macos.yml
@@ -52,9 +52,6 @@ jobs:
         if: steps.cache-cargo-vendor.outputs.cache-hit != 'true'
         shell: bash
         run: "cargo vendor --locked --versioned-dirs >> .cargo/config"
-      - name: "Install System Deps"
-        shell: bash
-        run: "env CI=yes PATH=$PATH ./get-deps"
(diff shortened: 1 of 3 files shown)


bew (changes requested):
Hello! Thanks for the PR!
If you want to remove get-deps from MacOS pipelines this is not enough, this only removes the _trigger_, so changed to that file won't trigger a macos build.
But a change to any other file will and the later step where it tries to run ./get-deps is still there and will run 🤔

.github/workflows/gen_macos.yml
```diff
@@ -53,9 +52,6 @@ jobs:
         if: steps.cache-cargo-vendor.outputs.cache-hit != 'true'
         shell: bash
         run: "cargo vendor --locked --versioned-dirs >> .cargo/config"
-      - name: "Install System Deps"
```
bew, line 56: ⚠️ This file is generated by `ci/generate-workflows.py`, you need to update that script to make any persisted changes to workflows

wez:
It may not do anything today, but it might in the future.  Does it do any harm to leave it in the CI?  If not, I'd suggest just leaving it in place for the sake of consistency

Issue #8020, opened 2026-08-05
tab bar title not changed when switch same panel


LuncyBloont:
### What Operating System(s) are you seeing this problem on?

Windows

### Which Wayland compositor or X11 Window manager(s) are you using?

Windows 11

### WezTerm version

20260805-063407-26270fd8

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

Tab bar title not changed from "Tab Navigator" when switch the same panel current using.

### To Reproduce

Right click tab bar, click the same panel current using.

### Configuration

``` lua
local wezterm = require 'wezterm'

-- This will hold the configuration.
local config = wezterm.config_builder()

-- This is where you actually apply your config choices.

-- For example, changing the initial geometry for new windows:
config.initial_cols = 120
config.initial_rows = 38

-- or, changing the font size and color scheme.
config.font_size = 11
config.color_scheme = 'tango'

-- Spawn a fish shell in login mode
config.default_prog = { 'powershell' }

config.font = wezterm.font 'SauceCodePro NFM'

-- config.hide_tab_bar_if_only_one_tab = true

config.tab_bar_at_bottom = true

-- Finally, return the configuration to wezterm:
return config
```

### Expected Behavior

Tab bar title changed back from "Tab Navigator"

bew:
Yes I noticed this too recently with the Launcher overlay.

I'll try to fix it soon!

(Note: it looks like you're missing a few words on your sentences 🤔)

hphng:
hey @bew can I have this bug? I want to try to fix this

bew:
> hey [@bew](https://github.com/bew) can I have this bug? I want to try to fix this

yeah go ahead ;)

PR #8031, opened 2026-08-08
fix the bug by adding update title to update immediately

hphng:
Hi @bew, 

## The problem ( #8020 )
as I understand, WezTerm has a model on these layered: **Pane, Tab, Window, and Mux**. There is mechanical called **Overlay** that shown a temporary pane. 

So when we activate the **Overlay**, and choose the current active tab, the `update_title()` get called. However, the temporary pane still the current pane, and as a result, `update_title()` recomputed as **Tab Navigator** again.

## The fix
I added `update_title()` for both `cancel_overlay_for_tab()` and `cancel_overlay_for_pane()` so that when an overlay tab/pane removed, its title will compute correctly

What do you think about this?


hphng:
Hi @bew, just a follow up reminder about this pull request :)

Issue #7953, opened 2026-07-19
Kitty image placement reposition + DECSTBM scroll crashes WezTerm (SIGABRT via quad-count blowup)


hisanari-dev:
### What Operating System(s) are you seeing this problem on?

macOS

### Which Wayland compositor or X11 Window manager(s) are you using?

N/A (macOS)

### WezTerm version

20260716-195552-76b606ec

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

WezTerm's process crashes with `SIGABRT` when a kitty graphics protocol placement is repeatedly repositioned (`a=p`) while, concurrently, on-screen terminal text is scrolling **within a DECSTBM scroll region** (`ESC[<top>;<bottom>r`, then writing at the region's bottom row + newline — the standard way full-screen TUI apps like Vim/Neovim scroll a viewport while leaving statusline/tabline chrome outside the region untouched).

Root cause chain, from WezTerm's own log file (`~/.local/share/wezterm/wezterm-gui-log-<pid>.txt` on macOS):

1. The renderer's per-paint quad count grows to **~10.48 million** quads in a single paint pass — many orders of magnitude more than a terminal grid should ever need.
2. GPU vertex buffer allocation for that fails ("not enough memory to create the buffer").
3. `wezterm-gui/src/termwindow/render/draw.rs:258` calls `.unwrap()` on that failed result instead of handling it, so it panics.
4. On macOS that panic fires inside `window::os::macos::window::WindowView::draw_rect`, a Cocoa `drawRect:` callback invoked across an `extern "C"` FFI boundary, so it can't unwind (`panic_cannot_unwind`) and WezTerm calls `abort()`. `SIGABRT`, whole process dies — taking down any application using the terminal (in my case, a Neovim session with unsaved buffers).

I found this while developing a Neovim plugin that displays images via the kitty graphics protocol — scrolling a Neovim buffer with an image placement anchored crashed WezTerm and my Neovim session with it. I've since isolated the trigger to a minimal, Neovim-free shell reproduction (below) and reproduced it 8 times, all with a byte-identical panic/backtrace shape.

This may be the same underlying mechanism as #5892 ("Kitty image protocol visual glitches when updating placement location") — that report's setup (ncurses scrolling + repositioning a placement) looks like it could be hitting the same quad-count growth, just not reaching the crash threshold in that reporter's environment/duration.

[...]

Issue #7400, opened 2025-11-24
Memory leak when using kitty + gif


hasezoey:
### What Operating System(s) are you seeing this problem on?

Linux Wayland

### Which Wayland compositor or X11 Window manager(s) are you using?

Operating System: Manjaro Linux 
KDE Plasma Version: 6.3.6
KDE Frameworks Version: 6.18.0
Qt Version: 6.9.2
Kernel Version: 6.12.48-1-MANJARO (64-bit)
Graphics Platform: Wayland
Processors: 16 × AMD Ryzen 7 5800X 8-Core Processor
Memory: 15.5 GiB of RAM
Graphics Processor: AMD Radeon Graphics
Manufacturer: ASUS

### WezTerm version

20251123-154500-ac08cdf6

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

While using wezterm, outputting kitty gif (animation) content, wezterm just freezes graphically, and observing any process monitor shows 1 CPU being pinned at 100% and memory steadly going up.

### To Reproduce

Setup:
- have wezterm installed
- have [viu](https://github.com/hasezoey/viu/tree/testKittyOnly) ready

1. start wezterm
2. go to the downloaded `viu` project
3. run `cargo run -- img/giphy.gif`
4. observe wezterm being completely stuck and memory leaking

This likely would also happen without viu and some direct command, but `viu` was the easiest to demonstrate this.

### Configuration

no config

### Expected Behavior

No Memory leak, 1 100% pinned cpu and correctly display the animation (or discard the data)

### Logs

wezterm is completely unresponsive, even to CTRL+SHIFT+L.
```sh
$ ls -al $XDG_RUNTIME_DIR/wezterm
total 0
drwx------  2 hasezoey hasezoey 180 Nov 24 19:58 .
drwx------ 16 hasezoey hasezoey 620 Nov 24 19:44 ..
lrwxrwxrwx  1 hasezoey hasezoey  32 Nov 24 19:49 agent.119928 -> /tmp/ssh-XXXXXXcN2oEK/agent.1059
lrwxrwxrwx  1 hasezoey hasezoey  32 Nov 24 19:53 agent.121854 -> /tmp/ssh-XXXXXXcN2oEK/agent.1059
[...]

PR #8035, opened 2026-08-09
term: don't duplicate image attachments when placing kitty images

alanbanks229:
## Problem

Repeated Kitty image placements over cells that already contain image attachments can duplicate those attachments on every update.

`assign_image_to_cells` clones the existing cell with its current attachments, then writes it through `Line::set_cell`, which merges the previous attachments back in again.
Repeating this causes attachment counts (renderer quads) — to grow rapidly causing WezTerm to crash.

This explains the quad-count growth seen in #7953 

## Fix

Use `set_cell_clearing_image_placements` for Kitty placement writes, where the updated cell already contains the complete attachment set.

The existing merging behavior remains unchanged for text, sixel, and iTerm2 images, where it is still required.

## Tests

Added regression coverage for:

- repeated overlapping Kitty placements without accumulating attachments
- preserving Kitty placements when text is written over them

Verified with:

- `cargo nextest run --all --no-fail-fast` — 387/387 passing
- `cargo +nightly fmt --all -- --check` — clean

The overlapping-placement repro previously grew from ~120 MB to >500 MB in 15 updates; with this change it remained around 125 MB over 40 updates.

### DIFF
--- a/term/src/screen.rs
+++ b/term/src/screen.rs
@@ -408,6 +408,18 @@ impl Screen {
         line.set_cell(x, cell.clone(), seqno);
     }
 
+    pub fn set_cell_clearing_image_placements(
+        &mut self,
+        x: usize,
+        y: VisibleRowIndex,
+        cell: &Cell,
+        seqno: SequenceNo,
+    ) {
+        let line_idx = self.phys_row(y);
+        let line = self.line_mut(line_idx);
+        line.set_cell_clearing_image_placements(x, cell.clone(), seqno);
+    }
+
     pub fn set_cell_grapheme(
         &mut self,
         x: usize,
--- a/term/src/terminalstate/image.rs
+++ b/term/src/terminalstate/image.rs
@@ -235,14 +235,21 @@ impl TerminalState {
                     params.placement_id,
                 ));
                 match params.style {
-                    ImageAttachStyle::Kitty => cell.attrs_mut().attach_image(img),
+                    ImageAttachStyle::Kitty => {
+                        cell.attrs_mut().attach_image(img);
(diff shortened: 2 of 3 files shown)


bew:
Hello, for some reason I see that this PR mentioned https://github.com/wezterm/wezterm/issues/7400 at some point:
image
Was this a mistake or would this PR fix that issue as well?

bew:
Hmm I see you've added a fix for #7400, I'd much prefer if you'd make a separate PR instead, so each fix can be discussed independently.

alanbanks229:
> Hmm I see you've added a fix for #7400, I'd much prefer if you'd make a separate PR instead, so each fix can be discussed independently.

Hey @bew that makes sense — sorry about the confusion.

The original #7400 reference was left over from some investigation I had briefly done on this branch before moving it elsewhere.

Looking into it again, I found that the two bugs happen in the same general image-attachment path, but yes they are distinct:

This PR: repeatedly placing/repositioning a Kitty image over cells that already contain image attachments caused the existing attachments to be duplicated on each update.

#7400: viu sends GIF frames as anonymous Kitty images (no explicit image id). Those anonymous re-placements were accumulating attachments instead of replacing the previous frame.

I pushed the #7400 fix here while investigating the relationship between the two, but I agree that they should be reviewed independently. 
- I'll move that second commit to a separate PR and keep this PR focused on the overlapping-placement bug.

alanbanks229:
@bew created the separate PR #8077 for issue #7400 !

They touch adjacent lines in the same `match` arm, so whichever one lands second will need a small rebase.

On a side note I will be away this weekend, so it might be a bit before I can respond again.

PR #8036, opened 2026-08-09
config: cache Config->lua conversion for event dispatch

epsalmond:
`format-tab-title` receives the whole Config on every call, and every call
converted it to a lua value from scratch (`to_lua<Config>` ->
`dynamic_to_lua_value`). When many windows are created at once the cost
multiplies: a client domain re-attach after a network blip re-creates every
window, each `new_window` rebuilds `TabBarState`, and each tab title
re-converts the entire Config. On a loaded host this monopolizes the main
thread for minutes. The UI freezes and mux cli queries starve behind the
busy frontend.

Observed in the wild twice on macOS: GUI wedge during SSHMUX re-attach.
Main-thread samples show >90% of samples inside one `new_window` spawn,
bottoming out in `dynamic_to_lua_value` recursion under
`call_format_tab_title`.

The fix adds `config::config_handle_to_lua`: it caches the converted table
per (lua state, config Arc) pair in a main-thread thread_local, and
`format-tab-title` dispatch uses it. The cache holds strong refs to both,
so the pointers it compares cannot be recycled while an entry is live. A
config reload or a per-window overridden config allocates a fresh Arc and
misses the cache naturally. The lua state is replaced on reload too, and
the pair key covers it.

## Test plan

New unit test in the config crate: same-Arc reuse returns the identical
table, a different config Arc misses, a different lua state misses.

- `cargo test -p config`: green.
- Live check on macOS with the mux host at load 20: three `cli spawn` window
  creations plus a `cli list` complete in 0.34s and the GUI stays
  responsive. Before the fix, samples show a single window creation holding
  the main thread for over a second during re-attach, with cli queries
  timing out behind it.

### DIFF
--- a/config/src/lib.rs
+++ b/config/src/lib.rs
@@ -814,6 +814,82 @@ impl std::ops::Deref for ConfigHandle {
     }
 }
 
+thread_local! {
+    static CONFIG_LUA_CACHE: RefCell<Option<(Rc<Lua>, Arc<Config>, mlua::RegistryKey)>> =
+        RefCell::new(None);
+}
+
+/// Convert a ConfigHandle to a lua value, re-using the cached conversion
+/// when the lua state and the config Arc are both unchanged.
+///
+/// Events such as format-tab-title receive the config on every invocation,
+/// and converting the whole Config each time is expensive: when many windows
+/// are created at once (for example while re-attaching a client domain),
+/// the repeated conversion monopolizes the main thread for minutes.
+///
+/// The cache holds strong refs to the lua state and the config Arc, so the
+/// pointers it compares cannot be recycled while an entry is live.
+///
+/// Must be called from the main thread, same as
+/// run_immediate_with_lua_config.
+pub fn config_handle_to_lua<'lua>(
+    lua: &'lua Rc<Lua>,
+    config: &ConfigHandle,
+) -> mlua::Result<mlua::Value<'lua>> {
+    let cached = CONFIG_LUA_CACHE.with(|cache| -> Option<mlua::Value<'lua>> {
+        let cache = cache.borrow();
+        match &*cache {
+            Some((cached_lua, cached_config, key))
+                if Rc::ptr_eq(cached_lua, lua) && Arc::ptr_eq(cached_config, &config.config) =>
+            {
+                lua.registry_value(key).ok()
+            }
+            _ => None,
+        }
+    });
+    if let Some(value) = cached {
+        return Ok(value);
+    }
+    let value = luahelper::to_lua(lua, (*config.config).clone())?;
+    let key = lua.create_registry_value(&value)?;
+    CONFIG_LUA_CACHE.with(|cache| {
+        cache
+            .borrow_mut()
+            .replace((Rc::clone(lua), Arc::clone(&config.config), key));
+    });
+    Ok(value)
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    fn config_handle_to_lua_caches_per_config_arc() {
+        let lua = Rc::new(Lua::new());
+        let handle = ConfigHandle::default_config();
+
+        let first = config_handle_to_lua(&lua, &handle).unwrap();
+        let second = config_handle_to_lua(&lua, &handle).unwrap();
+        assert!(matches!(first, mlua::Value::Table(_)));
+        assert_eq!(
+            first, second,
+            "same config Arc and lua state must re-use the cached table"
+        );
+
+        // A different config Arc (per-window overrides, or a reload)
+        // must not hit the cache
+        let other = ConfigHandle::default_config();
+        let third = config_handle_to_lua(&lua, &other).unwrap();
+        assert_ne!(first, third);
+
+        // A different lua state must not hit the cache either
+        let other_lua = Rc::new(Lua::new());
+        let fourth = config_handle_to_lua(&other_lua, &handle).unwrap();
+        assert!(matches!(fourth, mlua::Value::Table(_)));
+    }
+}
+
 pub struct LoadedConfig {
     pub config: anyhow::Result<Config>,
     pub file_name: Option<PathBuf>,
--- a/wezterm-gui/src/tabbar.rs
+++ b/wezterm-gui/src/tabbar.rs
@@ -63,19 +63,13 @@ fn call_format_tab_title(
         if let Some(lua) = lua {
             let tabs = lua.create_sequence_from(tab_info.iter().cloned())?;
             let panes = lua.create_sequence_from(pane_info.iter().cloned())?;
+            let config_value = config::config_handle_to_lua(&lua, config)?;
 
             let v = config::lua::emit_sync_callback(
                 &*lua,
                 (
                     "format-tab-title".to_string(),
-                    (
-                        tab.clone(),
-                        tabs,
-                        panes,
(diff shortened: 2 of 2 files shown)


config/src/lib.rs
```diff
@@ -814,6 +814,82 @@ impl std::ops::Deref for ConfigHandle {
...
+) -> mlua::Result<mlua::Value<'lua>> {
+    let cached = CONFIG_LUA_CACHE.with(|cache| -> Option<mlua::Value<'lua>> {
+        let cache = cache.borrow();
+        match &*cache {
+            Some((cached_lua, cached_config, key))
+                if Rc::ptr_eq(cached_lua, lua) && Arc::ptr_eq(cached_config, &config.config) =>
+            {
+                lua.registry_value(key).ok()
+            }
+            _ => None,
```
bew, line 847: Shouldn't we cleanup lua registry in case the cache is stale?

config/src/lib.rs
```diff
@@ -814,6 +814,82 @@ impl std::ops::Deref for ConfigHandle {
...
+        assert!(matches!(first, mlua::Value::Table(_)));
+        assert_eq!(
+            first, second,
+            "same config Arc and lua state must re-use the cached table"
+        );
+
+        // A different config Arc (per-window overrides, or a reload)
+        // must not hit the cache
+        let other = ConfigHandle::default_config();
+        let third = config_handle_to_lua(&lua, &other).unwrap();
```
bew, line 883: I think we should test that next call with this config should get the same cached value?

config/src/lib.rs
```diff
@@ -814,6 +814,82 @@ impl std::ops::Deref for ConfigHandle {
...
+        assert_ne!(first, third);
+
+        // A different lua state must not hit the cache either
+        let other_lua = Rc::new(Lua::new());
+        let fourth = config_handle_to_lua(&other_lua, &handle).unwrap();
+        assert!(matches!(fourth, mlua::Value::Table(_)));
```
bew, line 889: AFAIK this test doesn't test what you say you're testing, since all results from that function are all Lua tables.. It doesn't verify cache hits at all 🤔

config/src/lib.rs
```diff
@@ -814,6 +814,82 @@ impl std::ops::Deref for ConfigHandle {
     }
 }
 
+thread_local! {
+    static CONFIG_LUA_CACHE: RefCell<Option<(Rc<Lua>, Arc<Config>, mlua::RegistryKey)>> =
```
bew, line 818: Please add some documentation about this.
Should mention that it holds the last cached config and any different Lua state or config replace the cache.

Btw, why not hold caches for all (not just last) Lua-Config pairs and invalidate the cache on reload?

wezterm-gui/src/tabbar.rs
```diff
@@ -63,19 +63,13 @@ fn call_format_tab_title(
...
                     "format-tab-title".to_string(),
-                    (
-                        tab.clone(),
-                        tabs,
-                        panes,
-                        (**config).clone(),
-                        hover,
-                        tab_max_width,
-                    ),
+                    (tab.clone(), tabs, panes, config_value, hover, tab_max_width),
```
bew, line 72: Was that a rustfmt change to switchall args to single-line? The multi-line variant is much more readable 🤔

bew (changes requested):
Hello, thanks for the PR 🤔

Could you open a separate bug report issue about the root issue this is trying to fix ? With your environment data as well as how to reproduce the issue.

Could you share exact commands user for testing, so I can reproduce the issue and verify the fix?

And what / how much manual review/iteration/testing did you do?

iliaal:
Independent confirmation from a Windows user, plus a symbolized profile, since this PR
has been sitting without CI or review.

## This fixes a real hard freeze, not just slow rendering

On wezterm nightly `20260828-101755-27d55bef` (commit `27d55bef144f`), Windows 10 build
26200, `wezterm-gui` stops responding entirely with 9 to 13 panes open and a
`format-tab-title` handler registered. The main thread pins one core, the GPU sits idle,
page faults are zero, memory stays flat, and `wezterm cli list` can't reach the GUI
socket. It never recovers, so the only way out is killing the process.

I captured this against a local release build with debug info. Here's the main thread
mid-freeze:

```
wezterm_gui!luaH_finishset / luaH_set / rehash / luaH_newkey / luaH_resize
  mlua::table::Table::set<Value,Value>
  luahelper::dynamic_to_lua_value
  luahelper::to_lua
  config::config::impl$15::into_lua
  mlua::function::Function::call<(TabInformation, Table, Table, Config, bool, usize)>
  config::lua::emit_sync_callback
  config::run_immediate_with_lua_config<tabbar::call_format_tab_title::closure_env$0>
  wezterm_gui::tabbar::call_format_tab_title
  wezterm_gui::tabbar::compute_tab_title
  wezterm_gui::tabbar::TabBarState::new
  wezterm_gui::termwindow::TermWindow::update_title_impl
  wezterm_gui::termwindow::TermWindow::update_title
  wezterm_gui::termwindow::TermWindow::dispatch_notif
  wezterm_gui::termwindow::TermWindow::dispatch_window_event
```

`luaH_resize` and `rehash` dominate, which is the table growth from converting the whole
`Config` on every dispatch. That's exactly what this PR addresses.

The mechanism is congestive collapse rather than a stuck loop. Title change
notifications arrive faster than `update_title` drains them, so the `SpawnQueue` grows
without bound and the message loop never runs again. Anything that animates a pane title
several times a second drives it, and the cost scales with pane count.

## Applying this PR removed the freeze

With the patch applied, the same workload (13 panes, agents writing titles continuously)
has run 35 minutes and counting with no freeze. Before it, the same setup froze
reproducibly within about 8 minutes.

## One caveat: this is necessary but not sufficient

Three separate per-call costs sit on this path. This PR fixes the second one.

1. `format_as_escapes` rebuilds a `TerminfoRenderer` on every call, and
   `TerminfoRenderer::new(CAPS.clone())` deep copies the whole `terminfo::Database`
   because `Capabilities` owns it rather than sharing it. I patched this locally by
   changing `Capabilities::terminfo_db` to `Option<Arc<terminfo::Database>>`, which keeps
   the public `terminfo_db()` signature unchanged. Happy to open a separate PR for it.
2. The `Config` to Lua conversion, which this PR caches.
3. The `tabs` and `panes` sequences built by `create_sequence_from` on every call, which
   this PR leaves uncached.

After applying both 1 and 2, the freeze is gone but the main thread still burns about 49%
of one core at 13 panes (17m19s of CPU across 35 minutes of uptime). Sampling it now
lands here:

```
ntdll!RtlFreeHeap
  core::ptr::drop_glue<mlua::userdata::UserDataCell<TabInformation>>
  mlua::util::userdata_destructor<UserDataCell<TabInformation>>
  precallC / luaD_precall / ccall / luaD_rawrunprotected / luaD_pcall
```

So the remaining cost is Lua GC tearing down the per-call `TabInformation` and
`PaneInformation` userdata. Combined with the double `compute_tab_title` call per tab
noted in #8086, that looks like the next thing worth caching.

## Environment

- wezterm `20260828-101755-27d55bef`, built locally with `[profile.release] debug = true`
- Windows 10 build 26200, 32 logical CPUs
- Reproduced under both `front_end = "WebGpu"` (D3D12) and `front_end = "OpenGL"` (ANGLE)
- Unaffected by `use_fancy_tab_bar` and `tab_max_width`

Issue #118, opened 2020-01-14
 Mouse selection mark still displaying after changing to a new tmux window


rbelem:
## Describe the bug

Mouse selection mark still displaying after changing tmux window

## Environment (please complete the following information):

 - OS: KDE Neon Linux X11
 - Frontend:  OpenGL
 - Version: wezterm 20200113-222147-724ad3a-1-g18bbd2a

## Is it a build problem?

Nope

## To Reproduce

1. Open tmux
2. Select some text with the mouse
3. Create a new tmux window
4. The selection mark will follow the new window

## Configuration

None

## Expected behavior

Keep the selection mark in the original window

## Screenshots

None

## Session Recording

None

## Additional context

None

wez:
Thanks for filing; the same effect could also be produced by running `top`, then double clicking on eg: a pid in one of the rows and then waiting three seconds.  I've been meaning to fix it since I introduced this regression in a recent refactor :)
The fix will show up in the "nightly" (it's really hourly) in about an hour.

PR #8037, opened 2026-08-09
fix: keep selections alive across content-identical redraws

### LINKED ISSUE #7984: Selections are cleared while a terminal application continuously redraws

Windows 11, with the application running in WSL/Arch Linux.

### WezTerm version

Current `main` at `76b606ec597a3c0263fa60321548637451c0a547`.

### Did you try the latest nightly build?

Yes. I can reproduce both behaviors with a local build of current `main` and
with no configuration.

### Describe the bug

I ran into this while using a full-screen CLI that redraws while it is
generating output. It makes selecting text quite difficult:

- a mouse selection can disappear before I release the left button;
- a manual selection in Copy Mode disappears almost immediately.

I originally thought this was specific to that CLI, but the small Python
program below reproduces it without any third-party application or WezTerm
configuration.

### To reproduce

Start WezTerm without loading a config:

```console
wezterm -n
```

Then run:

```python
import sys
import time

sys.stdout.write("\x1b[?1049h\x1b[?25l")

try:
    frame = 0
    while True:
        sys.stdout.write("\x1b[H")
        sys.stdout.write("Try to select this unchanged text with the mouse.\r\n")
        sys.stdout.write("This line is rewritten, but its text never changes.\r\n")
        sys.stdout.write(f"frame: {frame:08d}\x1b[J")
        sys.stdout.flush()
        frame += 1
        time.sleep(0.03)
finally:
    sys.stdout.write("\x1b[?25h\x1b[?1049l")
    sys.stdout.flush()
```

For mouse selection, slowly drag across the first line. The highlight is
cleared repeatedly while the button is still held.

For Copy Mode, activate Copy Mode, start a selection on the `frame:` line, and
move the Copy Mode cursor. The selection is cleared on the next redraw and
copying produces an empty result.

### Expected behavior

An in-progress mouse drag should remain selected until mouse-up. At that point
`CompleteSelection` can copy the current live contents, so this does not
require keeping a stale text snapshot.

A selection made manually in Copy Mode should also survive normal updates from
the underlying pane.

### Additional context

I found two separate paths behind the two symptoms.

For mouse selection,
`TermWindow::check_for_dirty_lines_and_invalidate_selection()` clears any
selection that intersects a dirty row. It does this even while
`MousePress::Left` is still present in `current_mouse_buttons`.

For Copy Mode, `CopyOverlay::get_lines()`/`with_lines_mut()` calls
`CopyRenderable::update_search()` whenever the delegate sequence number
advances. When the search pattern is empty, `update_search()` unconditionally
calls `clear_selection()`. That also clears a selection made manually in Copy
Mode, although there is no active search result.

I have a small local prototype for both paths:

- defer dirty-row invalidation only while the left mouse button is held;
- in the empty-search branch, clear the selection only when an active search
  result is actually being discarded.

I tested the patched build against the reproducer:

- mouse selection stayed highlighted during the drag, copied the current text
  on release, and resumed the normal invalidation behavior afterward;
- Copy Mode kept the selection during continuous redraws and copied it
  successfully;
- search, tabs, pane splitting, and pane zoom still worked;
- `cargo test --all` passed.

Possibly related: #644, #6163, and #7740.

sdrik:
Partially addresses #7984 — see the scope note at the end.

## Background

The core of this is not new work written for #7984. I had been losing selections in
my own setup for a long time; on 2026-07-14 I tracked it down to two independent bugs
coexisting: this one in wezterm, and one in the tmux2k status-bar plugin whose time
segment shelled out to `tmux set-option` on every tick — and *any* `set-option` makes
tmux re-send every row, dirtying the whole viewport once a second. I fixed both but
only upstreamed the tmux2k one, since that alone made my symptom go away, so this
branch sat unused until #7984 turned up and I recognised the wezterm half of my
problem in it. The Copy Mode fix was added then.

Worth stating plainly: I cannot offer real-world mileage on this branch, because what
I have been running is the tmux2k fix, not this. The wezterm bug is independent of
that plugin though — `top`, `watch`, or any application repainting a full screen
reproduce it.

## Problem

An application that repaints the screen destroys a selection even when the text under
it never changed:

- a finished selection goes away at the next refresh — a stable PID or a column title
  in `top`, static content in a tmux pane while only the status bar ticks;
- a selection made by hand in Copy Mode disappears almost at once, so copying it
  yields an empty result.

## Cause

The terminal marks a line dirty on any write, including a rewrite with identical
content (`set_cell_impl` bumps the line seqno unconditionally), and there is no
per-cell dirty signal — a stable PID sharing a line with churning CPU%/TIME cells
cannot be told apart at row level. `check_for_dirty_lines_and_invalidate_selection`
cleared the selection as soon as a selected *row* was dirty, so every redraw wiped it.

Copy Mode reaches the same outcome by another path: `CopyOverlay::get_lines()` /
`with_lines_mut()` call `update_search()` on every seqno advance of the delegate, and
with an empty search pattern `update_search()` called `clear_selection()`
unconditionally.

## Changes

1. **Compare the selected _text_ rather than row dirtiness.** `Selection` gains a
   `text` cache alongside its existing `seqno`, and the selection is cleared only when
   that text actually differs. Selections survive content-identical redraws while
   still going away when their content changes (#9, #118). Nothing here assumes what
   drives the selection.

2. **The cache lives in `Selection` itself**, not beside it: `Selection::set_range()`
   drops it whenever the selection moves, so it can never describe a region the
   selection has since left. `Copy` had to go from the derive list — nothing copied the
   struct by value.

3. **Copy Mode:** in the empty-pattern branch of `update_search()`, only
   `clear_selection()` when an active match is actually being discarded (`result_pos`
   was set). Erasing a search pattern still drops the match's selection; a hand-made
   Copy Mode selection survives the pane's redraws.

## Behaviour note

A redraw that rewrites the selected cells with identical content now keeps the
selection. This intentionally also covers switching between two tmux windows whose
selected region looks the same: there is no terminal-level signal distinguishing such
a switch from an in-place repaint (both are cursor moves plus overwrites, with no full
erase and no alt-screen toggle — I captured the byte stream to check), and it is the
same event as keeping a stable value selected in `top` across a refresh. The code
comment says so, so that it does not get "fixed" later.

## Scope

This does **not** close #7984. The reporter's mouse symptom — a selection cleared while
the button is still held, over cells whose content genuinely changes — is deliberately
left alone for now, and a drag over such content still loses its selection mid-gesture.
The hyperlink-on-drag-release half is going into its own PR.

## Testing

`cargo test` passes. The interactive cases were driven against an isolated GUI with
synthetic input, and each was confirmed to fail on the same build with the patch
reverted:

| Scenario | Before | After |
| --- | --- | --- |
| Drag over a line whose text is stable, then 2s of redraws | kept | kept |
| Copy Mode manual selection, 2s of redraws, then `y` | copies empty | copies the selection |
| Search a pattern, then erase it | match selection dropped | match selection dropped |
| Drag over a line whose text changes, pausing mid-drag | cleared | cleared (out of scope, see above) |

Keyboard injection into the test GUI is not perfectly reliable, so the Copy Mode case
was repeated; it passed 5 runs out of 6, the failure being a dropped keystroke rather
than a lost selection.

### DIFF
--- a/wezterm-gui/src/selection.rs
+++ b/wezterm-gui/src/selection.rs
@@ -8,15 +8,21 @@ use termwiz::surface::line::DoubleClickRange;
 use termwiz::surface::SequenceNo;
 use wezterm_term::{SemanticZone, StableRowIndex};
 
-#[derive(Debug, Default, Copy, Clone, Eq, PartialEq)]
+#[derive(Debug, Default, Clone, Eq, PartialEq)]
 pub struct Selection {
     /// Remembers the starting coordinate of the selection prior to
     /// dragging.
     pub origin: Option<SelectionCoordinate>,
-    /// Holds the not-normalized selection range.
+    /// Holds the not-normalized selection range. Assign it through
+    /// [`Self::set_range`] so that `text` cannot go stale.
     pub range: Option<SelectionRange>,
-    /// When the selection was made wrt. the pane content
+    /// The pane seqno `text` was sampled at.
     pub seqno: SequenceNo,
+    /// The text covered by `range`, as of `seqno`, or None until it has been
+    /// sampled. Redraws are checked against it rather than against the dirty
+    /// flag of the selected rows, because the terminal marks a line dirty on
+    /// any write to it, even a rewrite with identical content.
+    pub text: Option<String>,
     /// Whether the selection is rectangular
     pub rectangular: bool,
 }
@@ -25,13 +31,19 @@ pub use config::keyassignment::SelectionMode;
 
 impl Selection {
     pub fn clear(&mut self) {
-        self.range = None;
+        self.set_range(None);
         self.origin = None;
     }
 
     pub fn begin(&mut self, origin: SelectionCoordinate) {
-        self.range = None;
+        self.set_range(None);
         self.origin = Some(origin);
+    }
+
+    /// Moves the selection, dropping the now meaningless `text`.
+    pub fn set_range(&mut self, range: Option<SelectionRange>) {
+        self.range = range;
+        self.text = None;
     }
 
     pub fn is_empty(&self) -> bool {
--- a/wezterm-gui/src/termwindow/mod.rs
+++ b/wezterm-gui/src/termwindow/mod.rs
@@ -1676,40 +1676,62 @@ impl TermWindow {
     }
 
     fn check_for_dirty_lines_and_invalidate_selection(&mut self, pane: &Arc<dyn Pane>) {
-        let dims = pane.get_dimensions();
-        let viewport = self
-            .get_viewport(pane.pane_id())
-            .unwrap_or(dims.physical_top);
-        let visible_range = viewport..viewport + dims.viewport_rows as StableRowIndex;
-        let seqno = self.selection(pane.pane_id()).seqno;
-        let dirty = pane.get_changed_since(visible_range, seqno);
-
-        if dirty.is_empty() {
+        // The search/copy overlays deliberately mark lines dirty (for match
+        // highlighting) while managing the selection themselves, so leave them
+        // alone.
+        if pane.downcast_ref::<CopyOverlay>().is_some()
+            || pane.downcast_ref::<QuickSelectOverlay>().is_some()
+        {
             return;
         }
-        if pane.downcast_ref::<CopyOverlay>().is_none()
-            && pane.downcast_ref::<QuickSelectOverlay>().is_none()
-        {
-            // If any of the changed lines intersect with the
-            // selection, then we need to clear the selection, but not
-            // when the search overlay is active; the search overlay
-            // marks lines as dirty to force invalidate them for
-            // highlighting purpose but also manipulates the selection
-            // and we want to allow it to retain the selection it made!
 
-            let clear_selection =
-                if let Some(selection_range) = self.selection(pane.pane_id()).range.as_ref() {
-                    let selection_rows = selection_range.rows();
-                    selection_rows.into_iter().any(|row| dirty.contains(row))
-                } else {
-                    false
-                };
-
-            if clear_selection {
-                self.selection(pane.pane_id()).range.take();
-                self.selection(pane.pane_id()).origin.take();
-                self.selection(pane.pane_id()).seqno = pane.get_current_seqno();
+        let pane_id = pane.pane_id();
+        let (range, baseline_seqno, has_text) = {
+            let selection = self.selection(pane_id);
+            match selection.range {
+                Some(range) => (range, selection.seqno, selection.text.is_some()),
+                None => return,
             }
+        };
+
+        // First frame for this selection: sample the text it covers, to compare
+        // later redraws against.
+        if !has_text {
+            let text = self.selection_text(pane);
+            let mut selection = self.selection(pane_id);
+            selection.text = Some(text);
+            selection.seqno = pane.get_current_seqno();
+            return;
+        }
+
+        let dims = pane.get_dimensions();
+        let viewport = self.get_viewport(pane_id).unwrap_or(dims.physical_top);
+        let visible_range = viewport..viewport + dims.viewport_rows as StableRowIndex;
+        let dirty = pane.get_changed_since(visible_range, baseline_seqno);
+        // Rows outside the selection can be rewritten freely, and extracting the
+        // selected text on every frame would be wasteful, so stop here unless a
+        // selected row was touched.
+        if dirty.is_empty() || !range.rows().any(|row| dirty.contains(row)) {
+            return;
+        }
+
+        // A selected row was rewritten, which says nothing about whether the
+        // selected cells changed. Compare the text, and when it is unchanged
+        // move the baseline forward so that the comparison is not redone on
+        // every frame while unrelated cells on those rows keep churning.
+        //
+        // Keeping the selection when the cells are rewritten identically is
+        // intended, and by design also covers switching between two tmux windows
+        // that look the same: a switch emits no full erase and no alt-screen
+        // toggle, so nothing distinguishes it from an in-place repaint.
+        let text = self.selection_text(pane);
+        let seqno = pane.get_current_seqno();
+        let mut selection = self.selection(pane_id);
+        if selection.text.as_deref() == Some(text.as_str()) {
+            selection.seqno = seqno;
+        } else {
+            selection.clear();
+            selection.seqno = seqno;
         }
     }
 }
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -284,6 +284,10 @@ As features stabilize some brief notes about them will accumulate here.
   whose pty reported no pixel dimensions (e.g. in `tmux -CC` domain).
   Such images are now refused instead of taking down the pane. Thanks to @zakrad! #6344
 * Fix render loop freeze when closing workspaces. Thanks to @JafarAbdi! #7444
+* Selections were cleared by applications that repaint the screen on a timer,
+  even when the text under the selection was unchanged. This also affected
+  selections made by hand in Copy Mode, which could vanish before they could be
+  copied. Thanks to @xloudmax! #7984
 
 #### Updated
 * Bundled conpty.dll and OpenConsole.exe to build 1.22.250204002.nupkg
--- a/wezterm-gui/src/overlay/copy.rs
+++ b/wezterm-gui/src/overlay/copy.rs
@@ -307,7 +307,10 @@ impl CopyRenderable {
 
         self.results.clear();
         self.by_line.clear();
-        self.result_pos.take();
+        // Whether we are discarding an active search match, which owns the
+        // selection it highlighted. Captured before clearing, and consulted in
+        // the empty-pattern branch below.
+        let had_active_match = self.result_pos.take().is_some();
 
         SAVED_PATTERN.lock().insert(self.tab_id, self.get_pattern());
 
@@ -352,7 +355,15 @@ impl CopyRenderable {
             .detach();
         } else {
             self.searching.take();
-            self.clear_selection();
+            // No pattern to search for, so there are no matches to highlight.
+            // Only drop the selection if it belonged to a match we just
+            // discarded: update_search() also runs on every sequence number
+            // advance of the delegate (see get_lines/with_lines_mut), so
+            // clearing unconditionally would wipe a selection the user made by
+            // hand in Copy Mode as soon as the underlying pane redraws.
+            if had_active_match {
+                self.clear_selection();
+            }
         }
         self.window.invalidate();
     }
@@ -423,9 +434,7 @@ impl CopyRenderable {
         let pane_id = self.delegate.pane_id();
         self.window
             .notify(TermWindowNotif::Apply(Box::new(move |term_window| {
-                let mut selection = term_window.selection(pane_id);
-                selection.origin.take();
-                selection.range.take();
+                term_window.selection(pane_id).clear();
             })));
     }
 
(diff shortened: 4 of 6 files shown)


wezterm-gui/src/termwindow/mod.rs
```diff
@@ -1676,41 +1684,113 @@ impl TermWindow {
...
-            .unwrap_or(dims.physical_top);
+        let viewport = self.get_viewport(pane_id).unwrap_or(dims.physical_top);
         let visible_range = viewport..viewport + dims.viewport_rows as StableRowIndex;
-        let seqno = self.selection(pane.pane_id()).seqno;
-        let dirty = pane.get_changed_since(visible_range, seqno);
-
-        if dirty.is_empty() {
+        let dirty = pane.get_changed_since(visible_range, baseline_seqno);
+        if dirty.is_empty() || !current_range.rows().any(|row| dirty.contains(row)) {
             return;
```
bew, line 1715: please add a small comment about this case

wezterm-gui/src/termwindow/mod.rs
```diff
@@ -1676,41 +1684,113 @@ impl TermWindow {
...
+                seqno,
+            });
+            return;
+        }
+
+        // A drag gesture is in progress: the pointer is held down and has
+        // already moved. The selection is stationary this frame (otherwise
+        // need_capture above would have taken over), but the user is still in
+        // the middle of making it, so never clear it from under them -- not even
+        // when the selected cells genuinely change, which is the one case the
```
bew: > not even when the selected cells genuinely change

I'd like to avoid this behavior for now, let's try to fix the selection invalidation when the text doesn't change first, and let it get invalidated when the text changes under the selection for now.

We can mention this as a potential change later, but I'd like to get real use without first.

wezterm-gui/src/termwindow/mod.rs
```diff
@@ -1676,41 +1684,113 @@ impl TermWindow {
...
+        let current_range = match current_range {
+            Some(range) => range,
+            None => {
+                self.pane_state(pane_id).selection_snapshot = None;
+                return;
+            }
+        };
+
+        // Whenever the selected region changes -- a drag extending it, a custom
+        // binding, or a Lua-driven ExtendSelectionToMouseCursor -- (re)capture a
```
bew: Remove these detail on when selection region can change, unless the code has a specific  dependency on these specific region change methods, having them listed is redundant with the code, and doesn't gets updated if more are added.

_This is a general documentation guideline, you should teach your AI about this._

wezterm-gui/src/termwindow/mod.rs
```diff
@@ -1676,41 +1684,113 @@ impl TermWindow {
...
+        }
+
+        let pane_id = pane.pane_id();
+
+        // Nothing selected: forget any snapshot and bail.
+        let current_range = self.selection(pane_id).range;
+        let current_range = match current_range {
+            Some(range) => range,
+            None => {
+                self.pane_state(pane_id).selection_snapshot = None;
```
bew: FIXME: why is this here in this function??
Answer: because selection_snapshot tracking is not always in sync with the real selection obj, so this 'syncs' it.
👉 It think it would be easier/preferable to only have ONE selection obj to keep track, to avoid having to sync another one

Also: block comment is not accurate, the goal of this block is not this, but to get the current selection range AFAIK (or it should be moved near the snapshot reset to mention the sync) 🤔

wezterm-gui/src/termwindow/mouseevent.rs
```diff
@@ -137,6 +137,8 @@ impl super::TermWindow {
 
             WMEK::Press(ref press) => {
                 capture_mouse = true;
+                // A new gesture begins; it is a click until proven a drag.
+                self.mouse_gesture_dragged = false;
```
bew: shouldn't there be a similar line in the `WMEK::Release` branch above?

wezterm-gui/src/termwindow/mod.rs
```diff
@@ -2840,7 +2920,9 @@ impl TermWindow {
                     self.copy_to_clipboard(*dest, text);
                     let window = self.window.as_ref().unwrap();
                     window.invalidate();
-                } else {
+                } else if !self.mouse_gesture_dragged {
+                    // Only treat this as a link click if the gesture was a click,
+                    // not a drag whose selection happened to end up empty.
```
bew: Why is this change part of this PR? it's completely unrelated 🤔

wezterm-gui/src/selection.rs
```diff
@@ -39,6 +39,24 @@ impl Selection {
...
+/// A snapshot of a selection's content, used to decide whether the selection
+/// must be invalidated when the underlying pane is redrawn.
+///
+/// We compare the *selected text* rather than merely whether a selected row was
+/// flagged dirty: the terminal marks a line dirty on any write to it — even a
+/// rewrite with identical content (e.g. `top` repainting its whole screen, or
+/// tmux redrawing a pane). Comparing content keeps a selection stable across
+/// such redraws while still clearing it when the selected cells actually change.
+#[derive(Debug, Clone)]
+pub struct SelectionSnapshot {
```
bew: Why didn't you re-use the existing `Selection` struct above?
Did you hit concrete blockers?

wezterm-gui/src/termwindow/mod.rs
```diff
@@ -415,6 +418,10 @@ pub struct TermWindow {
     current_modifier_and_leds: (Modifiers, KeyboardLedStatus),
     current_mouse_buttons: Vec<MousePress>,
     current_mouse_capture: Option<MouseCapture>,
+    /// Set once the current mouse gesture has involved a drag (motion with a
+    /// button held). Used so that the matching button release is not treated as
+    /// a click that opens a hovered hyperlink.
```
bew: I don't think the last sentence is useful here, and would only be hard to maintain + doesn't reflect all usages of this. (LSP references check is much more accurate here)

bew (commented):
Hello!
Thanks for the PR, I like where this is going!

I didn't review everything yet, but here are a few comments already 😉

Also, I saw a few particularly big comments in this PR, I'd recommend trying to refactor them / reduce them to only focus on the important points relative to the code maintenance / explaining _why_ (without being too verbose!) / mentioning constraints / ...

wezterm-gui/src/termwindow/mod.rs
sdrik: Done. The snapshot lives in `Selection` itself now, as `text: Option<String>`,
reusing its existing `seqno` as the sample point — that field had no reader left
otherwise. `PaneState.selection_snapshot` and the syncing block are gone, so the
inaccurate comment goes with them. `range` is assigned through
`Selection::set_range()`, which drops the cached text, so the cache can never
describe a region the selection has since left.

wezterm-gui/src/termwindow/mod.rs
sdrik: Removed.

wezterm-gui/src/termwindow/mod.rs
sdrik: Dropped, along with the flag it relied on. A selection over cells that genuinely
change is cleared again, mid-drag included. Mentioned in the commit message as a
possible follow-up.

wezterm-gui/src/termwindow/mod.rs
sdrik: Added: it now says why we stop when no selected row was touched (rows outside the
selection may be rewritten freely, and extracting the text every frame would be
wasteful).

wezterm-gui/src/termwindow/mouseevent.rs
sdrik: Moot here: the flag is gone from this PR together with the hyperlink change. For
the record, a reset in `Release` would have had to come *after* the action
dispatch of that same release, since that is what reads the flag; `Press` is
where a new gesture starts, so it was the only place that needed it.

wezterm-gui/src/termwindow/mod.rs
sdrik: Less unrelated than it looks, though I will still split it out. The link opening
was the second half of the symptom I originally chased: with the selection cleared
out from under a held button, the left-`Up` handler finds `selection_text` empty
and falls through to `do_open_link_at_mouse_cursor` — so pausing a drag over a URL
opened it. Fixing the invalidation removes that path, but a drag that legitimately
ends on an empty selection still opens the link, which is what the flag was for.
Taken out of here and `mouseevent.rs` is back to matching `main`; I will open it on
its own with that reasoning.

wezterm-gui/src/selection.rs
sdrik: No blocker beyond the `derive(Copy)`: nothing copied the struct by value, so I
dropped `Copy` and moved the field in. Done in the update above.

wezterm-gui/src/termwindow/mod.rs
sdrik: Moot: the field is gone with the hyperlink change.

sdrik:
Some background, since it explains the shape of this PR.

The content comparison is not new work written for #7984. I had been losing
selections in my own setup for a long time; on 2026-07-14 I tracked it down to two
independent bugs coexisting: this one in wezterm, and one in the tmux2k status-bar
plugin, whose time segment shelled out to `tmux set-option` on every tick — and
*any* `set-option` makes tmux re-send every row, dirtying the whole viewport once
a second. I fixed both but only upstreamed the tmux2k one, since that alone made
my symptom go away, and the wezterm branch has been shelved ever since: no issue,
no PR — no time, and I am aware of the project's bandwidth.

That July work covered both halves of what I was hitting: the selection dying, and
the hyperlink firing on release *because* it had died — they were the same hunt,
which is why they arrived here together.

#7984 is what brought the branch back out, yesterday. What I added then is the Copy
Mode path and the mid-drag guard. The branch is a single squashed commit, so the log
does not separate the two.

Which bears on your call: the guard is a day old, has had no use beyond my own
tests, and — unlike the rest — addresses a symptom I never hit myself. Dropping it
costs nothing I can vouch for, so it is dropped. I should also be upfront that I
cannot offer you real-world mileage on any of this, since I have been running the
upstreamed tmux2k fix rather than this branch. The wezterm bug itself is
independent of that plugin, though: `top`, `watch`, or any application repainting
a full screen reproduce it.

Full transparency on the AI question: substantial, and here is the split.

**AI** (Claude Code — Opus 4.8 for the first commit, Opus 5 since, hence the two
`Co-Authored-By` trailers): mapping the relevant code, writing the patch,
capturing tmux's byte stream to check what a `set-option` and a window switch
actually emit, an xdotool-driven GUI harness for the interactive cases, and
essentially all of the prose — commit messages, PR description, this comment
included.

**Me**: hitting the bug, and the differential observations that pinned it (only under
tmux, only for selections wezterm makes locally, absent when the application
requests mouse reporting); manual verification of the behaviour in a real GUI
across five scenarios; isolating the tmux2k trigger by elimination and
re-testing the branch against that config; and pushing back on the framing
several times (an "unavoidable limitation" that was
documented nowhere, and an early draft that implied a per-second full redraw is
normal tmux behaviour — it is not, it was plugin-induced). For the July patch I
also read and assimilated the whole thing by hand; I stand behind the code, not
just behind its output. With one caveat worth knowing where to aim your review: I
read C, C++, C#, PHP, Python and JS/TS comfortably, but Rust is the language I am
least at home in. The logic and the terminal semantics I can vouch for; idiomatic
Rust and ownership subtleties are where my own review is thinnest.

One thing I would rather state than have you guess: I lean on the AI heavily for
prose, this comment included. I can hold a conversation in English, but with far
less precision than a review discussion deserves, and I would rather be
understood than idiomatic. The engineering calls are mine; the wording frequently
is not.

On comment size: agreed, and trimmed — the diff is now ~80 lines smaller than what
you reviewed. I have also dropped the mid-drag behaviour and split the hyperlink
change out, so this PR no longer closes #7984 on its own; the mouse symptom over
genuinely-changing cells is left for a follow-up.

PR #8039, opened 2026-08-10
dynamic: Add fallback="OtherType" container attribute

bew:
Allows a struct or enum to transparently accept an older/simpler format.
The derived from_dynamic first attempts the primary type; on failure it retries with OtherType and converts it to primary type. When both fail, both errors are surfaced to the user.

The fallback attribute is mutually exclusive with try_from.
ToDynamic is unaffected.

(NOTE: I'll merge them separately, not squashed!)
👉 https://github.com/wezterm/wezterm/pull/8039/changes/11dc8b6ca9de2d467f2f485d77c39cbb6ab88845 Restored bool parsing for `NewlineCanon`, which had been _broken_ since https://github.com/wezterm/wezterm/commit/f587cac145868320937a3cd28c426126d83c0257 in 2022 👀
👉 https://github.com/wezterm/wezterm/pull/8039/changes/53ac24b673adc99ee0f3bea883f46f682a31fc4a Simplifies BoldBrightening bool value parsing

> [!NOTE]
> I also added a README for the crate, just before (no PR), to better understand what's going on here: https://github.com/wezterm/wezterm/commit/ad5abbaf21676fb44c6a65f7deb09c979ce34596

---

This is an enabler to be able to transparently migrate old actions/configs from a legacy/simpler format to a more structured one.
For example: https://github.com/wezterm/wezterm/pull/6999#issuecomment-3694491298

Alternatively we could offer simpler ways to set configs, like:
```lua
config.window_alignment = "Center"
-- vs
config.window_alignment = {
  horizontal = "Center",
  vertical = "Center",
}
```

### DIFF
--- a/wezterm-dynamic/derive/src/fromdynamic.rs
+++ b/wezterm-dynamic/derive/src/fromdynamic.rs
@@ -90,7 +90,7 @@ fn derive_struct(input: &DeriveInput, fields: &FieldsNamed) -> Result<TokenStrea
         )
     };
 
-    let from_dynamic = match info.try_from {
+    let primary_body = match info.try_from {
         Some(try_from) => {
             quote!(
                 use core::convert::TryFrom;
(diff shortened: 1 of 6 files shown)


wezterm-dynamic/derive/src/fromdynamic.rs
```diff
@@ -309,6 +330,27 @@ fn derive_enum(input: &DeriveInput, enumeration: &DataEnum) -> Result<TokenStrea
...
+    let from_dynamic = match info.fallback {
+        Some(fallback) => quote!(
+            let primary_result = (|| -> core::result::Result<Self, wezterm_dynamic::Error> {
+                #primary_body
+            })();
+            match primary_result {
+                Ok(v) => Ok(v),
+                Err(primary_err) => {
+                    match <#fallback as wezterm_dynamic::FromDynamic>::from_dynamic(value, options) {
+                        Ok(legacy) => Ok(<#ident as core::convert::From<#fallback>>::from(legacy)),
```
bew, line 342: this here requires `From` on the other type, do you know if there's a ~simple way to get nice compiler error if there is no `impl From<OtherType> for MyType` ? 🤔

Issue #5917, opened 2024-08-09
wezterm with custom config disables OSC52 copy to clipboard


MartyMcFlyInTheSky:
### What Operating System(s) are you seeing this problem on?

Linux X11

### Which Wayland compositor or X11 Window manager(s) are you using?

gnome

### WezTerm version

wezterm 20240730-082727-56a27e93

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

I tested this because nvim OSC 52 clipboard wasn't working when I used it within wezterm. So I retried it without config ( ` wezterm -n`) and it was working fine. As soon as I use a configuration file (with a mere ` return {}` in it) OSC 52 sequences magically stop working.

### To Reproduce

Assert ~/.config/wezterm/wezterm.lua contains:

```
return {}
```

start wezterm normally. Paste `printf $'\e]52;c;%s\a' "$(base64 <<<'Hello world')"` and hit enter. You'd expect Hello world to be pasted from the clipboard buffer. However, the previously copied printf command is pasted instead. So the sequence is NOT copied to clipbord.

Conversely this works:

Start `wezterm -n` and do the same as above - no problems. Everything works as expected. I'm surprised since my config doesn't change anything. Is there an internal trigger maybe?

### Configuration

```
-- ~/.config/wezterm/wezterm.lua
return {}
```

### Expected Behavior

Character sequence is correctly copied to X clipboard

Swetko:
Experiencing the same issue... Very frustrating. The thing is it actually works sometimes even with config file. But then randomly stops until you kill all instances and restart. Fedora 42.

bew:
Have you tried with the latest nightly?

Swetko:
Not yet. Seems like my build is from January 2026. commit 05343b387085842b434d267f91b6b0ec157e4331

bew:
Do you use a clipboard manager on your setup? (it could help 🤔)

heldergg:
Still reproducible on a recent version. Loading any config disables OSC 52 clipboard writes; wezterm -n works fine.

* Linux X11
* wezterm 20260716-195552-76b606ec

bew:
Hello! Changes from #8043 were merged to `main`, could any of you check the latest nightly (available in a few hours) to see if the issue is still reproducible?

@heldergg @Swetko @MartyMcFlyInTheSky

PR #8043, opened 2026-08-10
x11: fix OSC 52 clipboard silently doing nothing

oetiker:
On X11, OSC 52 frequently had no effect at all: the clipboard kept its previous contents, with no error and nothing in the log.

Reproduced with multiple windows open by mouse-copying in one window and then emitting OSC 52 from a pane in another:

```
printf '\033]52;c;%s\033\\' "$(printf 'hello' | base64)"
```

Two independent bugs combine to cause this, so this PR is split into one commit each.

### 1. `SetSelectionOwner` used a stale timestamp

`update_selection_owner` asserted ownership using `self.copy_and_paste.time`. That field only advances on key and button events delivered to *that particular window*, so a window the user has not touched recently carries a timestamp that predates the selection's `lastTimeChanged`. Per ICCCM the X server silently ignores `SetSelectionOwner` in that case — no error, nothing logged, the assertion is simply dropped.

Switched to `CURRENT_TIME`, which is always accepted, for both the assert and the disown path. This mirrors what `focus()` already does, and for the same underlying reason.

This is the bug that makes OSC 52 look intermittent: it works until some other window takes the selection, and appears to "fix itself" after a restart because a fresh window's timestamps win again until the next copy elsewhere.

### 2. The clipboard went to the wrong window

`MuxNotification::AssignClipboard` picked the target window with `fe.known_windows.borrow().keys().next()`. `known_windows` is a `BTreeMap` keyed by the window handle, so this is the lowest handle — the oldest window — regardless of which pane actually emitted the sequence. With more than one window open, OSC 52 from a pane in a newer window set the clipboard by way of an unrelated window.

Now resolves the pane to its mux window and uses the window that owns it, falling back to the previous behaviour if the lookup fails.

### Testing

Built and run on X11 (MATE, single X server, multiple windows and panes). Before: OSC 52 silently no-ops once another window has copied. After: the sequence sets the clipboard reliably from any pane in any window, verified by pasting into other X clients.

I have not tested Wayland or macOS. Change 1 is X11-only; change 2 is in the shared frontend but only alters *which* window receives an already-correct assignment.

Possibly related: #5917 (X11, OSC 52 silently stops working). I can't confirm it's the same root cause — that report attributes the trigger to the presence of a config file, which this doesn't explain — but the symptom matches.

### DIFF
--- a/window/src/os/x11/window.rs
+++ b/window/src/os/x11/window.rs
@@ -923,10 +923,11 @@ impl XWindowInner {
                         but we don't: tell it to clear it"
             );
             // We don't have a selection but X thinks we do; disown it!
+            // See the note about CURRENT_TIME in the assert branch below.
             conn.send_request_no_reply(&xcb::x::SetSelectionOwner {
                 owner: xcb::x::Window::none(),
                 selection,
-                time: self.copy_and_paste.time,
+                time: xcb::x::CURRENT_TIME,
             })?;
         } else if we_own_it {
             log::trace!(
@@ -934,10 +935,19 @@ impl XWindowInner {
                  {current_owner:?}, tell X we now own it"
             );
             // We have the selection but X doesn't think we do; assert it!
+            //
+            // We deliberately use CURRENT_TIME rather than
+            // self.copy_and_paste.time here. That field only advances on key
+            // and button events delivered to *this* window, so for a window
+            // that hasn't been interacted with recently it lags behind the
+            // selection's lastTimeChanged. The X server silently ignores
+            // SetSelectionOwner whose time predates lastTimeChanged, which
+            // made OSC 52 a no-op whenever another window had copied more
+            // recently. CURRENT_TIME is always accepted.
             conn.send_request_no_reply(&xcb::x::SetSelectionOwner {
                 owner: self.window_id,
                 selection,
-                time: self.copy_and_paste.time,
+                time: xcb::x::CURRENT_TIME,
             })?;
         } else {
             log::trace!(


MichalMaruska:
I tested and works

wezterm-gui/src/frontend.rs
```diff
@@ -184,7 +184,24 @@ impl GuiFrontEnd {
                             selection,
                             clipboard
                         );
-                        if let Some(window) = fe.known_windows.borrow().keys().next() {
+                        // Route the assignment to the window that actually
+                        // contains the pane that emitted OSC 52. known_windows
+                        // is a BTreeMap keyed by the window handle, so simply
+                        // taking the first key always landed the clipboard on
+                        // the oldest window, no matter which pane asked for it.
```
bew: the comment is about the current state of the code, not what it was doing before
simplify it!
```suggestion
                        // Route the assignment to the window that actually
                        // contains the pane that emitted OSC 52.
```

wezterm-gui/src/frontend.rs
```diff
@@ -184,7 +184,24 @@ impl GuiFrontEnd {
...
+                            let windows = fe.known_windows.borrow();
+                            Mux::get()
+                                .resolve_pane_id(pane_id)
+                                .and_then(|(_, mux_window_id, _)| {
+                                    windows
+                                        .iter()
+                                        .find(|(_, id)| **id == mux_window_id)
+                                        .map(|(window, _)| window.clone())
+                                })
+                                .or_else(|| windows.keys().next().cloned())
```
bew, line 205: mention why we do that as fallback (e.g. we got an event from a pane that doesn't have a window showing it, but a window usually needs to own the clipboard)

wezterm-gui/src/frontend.rs
```diff
@@ -184,7 +184,24 @@ impl GuiFrontEnd {
...
+                        // Route the assignment to the window that actually
+                        // contains the pane that emitted OSC 52. known_windows
+                        // is a BTreeMap keyed by the window handle, so simply
+                        // taking the first key always landed the clipboard on
+                        // the oldest window, no matter which pane asked for it.
+                        let target = {
+                            let windows = fe.known_windows.borrow();
+                            Mux::get()
+                                .resolve_pane_id(pane_id)
+                                .and_then(|(_, mux_window_id, _)| {
```
bew: be explicit about what you skip!
```suggestion
                                .and_then(|(_domain, mux_window_id, _tab_id)| {
```

same below

bew (commented):
Hello! Thanks for the PR!

oetiker:
hi @bew 
i produced the code with claude opus, and tested by me using it every day since submission. Note though I am using it in my own fork https://github.com/oetiker/wezterm/tree/update-optimization-rebased (which I have no hope of ever getting merged) but I thought this one might be of some use generally. will tend to the comments momentarily.

wezterm-gui/src/frontend.rs
oetiker: Applied your suggestion. The BTreeMap/oldest-window explanation stays in the commit message, where the history belongs.

wezterm-gui/src/frontend.rs
oetiker: Added a comment on the `or_else`: a pane can belong to a mux window that no gui window is attached to, and since the clipboard is owned per window we need some window to assert that ownership, so we fall back to any one of them rather than dropping the request.

wezterm-gui/src/frontend.rs
oetiker: Done, here and in the two closures below: `|(_domain, mux_window_id, _tab_id)|`, `|(_window, id)|`, `|(window, _mux_window_id)|`.

bew:
Merged manually on main as individual commits: d8563fa02802204a5da3f45d5dfbc100ed4337f4 016b96272ba145430312bb73bd088703a5dd466e

Thanks!

PR #8045, opened 2026-08-11
Add mouse and focus accessors to TerminalState

mshamis-meta:
I'm embedding wezterm-term in a terminal where the emulator runs in a separate process from the UI, so the frontend has to reproduce the mode state rather than just react to it locally. Most of that state is parsed but isn't reachable from outside the crate.

### DIFF
--- a/term/src/terminalstate/mod.rs
+++ b/term/src/terminalstate/mod.rs
@@ -264,6 +264,9 @@ pub struct TerminalState {
     /// If true, writing a character inserts a new cell
     insert: bool,
 
+    /// When set, received chars auto-wrap to the next line when the cursor reaches the right border.
+    /// Otherwise received chars continuously replace the last char on the right.
+    /// note: This is set by default, because it's more useful this way.
     /// https://vt100.net/docs/vt510-rm/DECAWM.html
     dec_auto_wrap: bool,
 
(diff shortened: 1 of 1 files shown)


mshamis-meta:
@bew possible to get a review?

term/src/terminalstate/mod.rs
```diff
@@ -752,10 +755,125 @@ impl TerminalState {
...
+
+    pub fn shift_out_enabled(&self) -> bool {
+        self.shift_out
+    }
+
+    /// The SCS final byte designating G0: `b'B'` ASCII, `b'A'` UK, `b'0'` DEC
+    /// line drawing. Emit as `ESC ( <byte>`.
+    ///
+    /// Returns the designator rather than `CharSet` because `CharSet` is
+    /// `pub(crate)`: a getter returning it would widen an existing item, and
```
bew: looks like you let your LLM loose without review here...

you already widen the scope of `MouseEncoding`, why not `CharSet` ? 🤔

bew:
Hey @mshamis-meta, I did some updates & added some docs, what do you think about that?

mshamis-meta:
@bew Looks great! Thanks for the improvement

mshamis-meta:
@bew good to land?

bew:
Yes, thanks for reminding me 😃

Have fun with your own terminal project !
Are you embedding it in another app ? What is it about ?

mshamis-meta:
@bew Thanks for the help. Yes, I am embedding this in an app. I have an Electron app using xterm in the UI, and wezterm on the backend for various pre-processing tasks. Would love to go wezterm both on the front end and back end. Not sure what is the Wasm story for this.

bew:
> Not sure what is the Wasm story for this.

Probably not far ^^'

PR #8059, opened 2026-08-17
mux: run WSL domain discovery off the startup path (#7782)

### LINKED ISSUE #7782: Extremely slow startup on Windows: 15-50 seconds delay in portable_pty::cmdbuilder and local gui-sock setup

Windows

N/A - Windows

### WezTerm version

20240203-110809-5046fc22

### Describe the bug

WezTerm starts extremely slowly on my Windows machine. Opening a new WezTerm window takes around 15 to 50 seconds.

The issue still happens with an empty/minimal config, with `cmd.exe` instead of PowerShell, with software rendering, with auto config reload disabled, and after clearing the local WezTerm mux/socket directory.

Other checks suggest that the executable itself and font scanning are not the bottleneck:

- `wezterm --version` returns quickly, around 40-50 ms.
- `wezterm ls-fonts` returns in around 686 ms after rebuilding the Windows font cache.
- Native Windows terminal / cmd / PowerShell starts in a few hundred milliseconds.
- The slow startup happens with both `cmd.exe` and `pwsh.exe`.
- It also happens when using a minimal config file.

Trace logs show long gaps around `portable_pty::cmdbuilder` and `wezterm_mux_server_impl::local` while setting up the local GUI socket.

### To Reproduce

Steps I used to reproduce:

1. Create a minimal config file:

```lua
local wezterm = require "wezterm"

config.default_prog = { "cmd.exe" }
config.automatically_reload_config = false
config.front_end = "Software"
config.default_domain = "local"
config.ssh_domains = {}
config.unix_domains = {}

return config
```

2. Start WezTerm with:
wezterm --config-file "$env:USERPROFILE\wezterm-test.lua" start --always-new-process --no-auto-connect -- cmd.exe

3. Observe that startup still takes more than 10 seconds.

I also tested:
wezterm start --no-auto-connect -- cmd.exe
wezterm start --no-auto-connect -- pwsh.exe -NoLogo -NoProfile
wezterm --version
wezterm ls-fonts

wezterm --version and wezterm ls-fonts are fast, but opening a new GUI window is slow.

### Configuration

```
local wezterm = require "wezterm"

config.default_prog = { "cmd.exe" }
config.automatically_reload_config = false
config.front_end = "Software"
config.default_domain = "local"
config.ssh_domains = {}
config.unix_domains = {}

return config
```
It also happens after deleting my normal local configuration.

### Expected Behavior

I expected WezTerm to open a new local terminal window within a few hundred milliseconds to around one second, similar to Windows Terminal or the native console.

### Logs

Trace log gap analysis shows the following long delays:

```text
=== GAP: 14,240 ms ===
BEFORE: 10:27:39.646  TRACE  portable_pty::cmdbuilder > adding USER env: "TMP" "C:\\Users\\Administrator\\AppData\\Local\\Temp"
AFTER : 10:27:53.886  TRACE  wezterm_gui              > NoConnectNoPublish

=== GAP: 10,916 ms ===
BEFORE: 10:27:53.944  TRACE  wezterm_mux_server_impl::local > setting up C:\Users\Administrator\.local/share/wezterm\gui-sock-3
AFTER : 10:28:04.860  DEBUG  wezterm_gui::frontend          > workspace is default, fixup windows
```

### Anything else?

wezterm --version: around 40-50 ms
wezterm ls-fonts: around 686 ms
starting cmd.exe / pwsh.exe directly outside WezTerm: a few hundred milliseconds
starting WezTerm GUI: usually 15-50 seconds

Things already tried:

- Reinstalled WezTerm.
- Removed/deleted local config.
- Tested with a minimal config file.
- Disabled `automatically_reload_config`.
- Tested `cmd.exe` instead of PowerShell.
- Tested `pwsh.exe -NoLogo -NoProfile`.
- Tested `front_end = "Software"`.
- Rebuilt Windows font cache.
- Confirmed `wezterm ls-fonts` is fast.
- Removed stale PyCharm-related environment variables and PATH entries.
- Removed HTTP_PROXY / HTTPS_PROXY / NO_PROXY environment variables during testing.
- Cleared `C:\Users\Administrator\.local\share\wezterm`.
- Tried changing `XDG_RUNTIME_DIR` / `XDG_DATA_HOME`.
- Tried changing TEMP / TMP to `C:\Temp`.
- Killed all existing `wezterm.exe`, `wezterm-gui.exe`, and `wezterm-mux-server.exe` processes before testing.
- The issue still persists.

The machine uses Windows with an Administrator user profile. Other terminals start quickly on the same machine.

And more, this happened after I tried some CLI, including Open Code, Oh My Open Code, GSD2, Codex.
At first I used WezTerm with Claude Code and it worked well.

PHPCraftdream:
On Windows, starting wezterm -- and, separately, spawning the first pane --
could take 15-50+ seconds when `wsl_domains` was left unconfigured. Two
separate causes, both of them synchronous `wsl.exe -l -v` invocations:

* `update_mux_domains` built the default WSL domain list on the startup
  path, so the GUI didn't come up until `wsl.exe` did.
* `LocalDomain::resolve_wsl_domain` re-derived that list on *every* pane
  spawn, for every domain type -- including plain local ones that have
  nothing to do with WSL.

`wsl.exe -l -v` is slow for reasons outside wezterm's control (it may have
to start the LxssManager service and/or boot the WSL2 utility VM), and it
can hang outright when WSL is in a bad state, so the fix is to stop
blocking on it rather than to try to make it faster. Fixes #7782.

Discovery now runs on a background thread and publishes into the mux as it
completes; its result is cached for the life of the process rather than
recomputed per spawn. The non-blocking `cached_default_domains()` accessor
is what the spawn path uses, and it distinguishes "discovery hasn't
finished yet" (fall back to the domain's construction-time snapshot, which
carries the real distribution name -- resolving to `None` here would make
`fixup_command` silently run the command on the Windows host instead of
inside WSL) from "the list is known and doesn't contain this name" (a real
"no longer a WSL domain"). Only domains actually constructed as WSL domains
(explicit `wsl_domains` entries or auto-discovery) get this treatment now;
a plain domain whose name happens to collide with a `wsl_domains` entry no
longer does.

The hard part is not the background thread, it's not letting spawns race
it. A `DomainDiscoveryState` state machine (`NotStarted`/`InProgress`/
`Done`) coordinates that:

* Resolving a domain *by name* that isn't registered yet waits for an
  in-flight run, bounded by `DOMAIN_DISCOVERY_TIMEOUT`, instead of
  immediately reporting "invalid domain name". Bounded rather than
  unbounded because such a name is much more likely to be one that will
  never appear -- a typo, or a distribution since removed -- than the
  startup-critical case below, and blocking that spawn for the whole of a
  cold `wsl.exe` before saying so is its own kind of wrong. The cost of
  that choice is real and worth naming: `wsl.exe` can take longer than the
  bound on the very machines this targets, so a by-name spawn issued in the
  first seconds -- most plausibly from a `gui-startup` handler -- can be
  told the name is invalid even though discovery finds it shortly after.
  Configs that need a specific WSL domain available that early should set
  `wsl_domains` explicitly; the constant's own doc comment and the
  `wsl_domains`/`mux.get_domain` docs say so too.
* A `default_domain`/`default_mux_server_domain`/`--domain` that names a
  not-yet-discovered domain waits unbounded: a timeout there would treat
  "discovery is slow" as "the domain doesn't exist" and fail startup
  outright, which is exactly what happens on the slow-`wsl.exe` machines
  this targets. This is also the pre-existing behaviour, since discovery
  used to be synchronous. The wait re-checks periodically and logs while it
  is still waiting, so a slow or stuck startup is diagnosable from the log
  rather than being silent.
* Async callers (`split_pane`, `spawn_tab_or_window`, the GUI's `--domain`
  handling, the mux server's startup) `await` a `DomainDiscoveryWaitFuture`
  and then resolve wait-free, rather than parking the main-thread executor
  on a condvar. Blocking resolution from the GUI thread would freeze
  repaints and input for the whole wait -- every synchronous resolver call
  site left in the GUI thread (including `DetachDomain`) was audited and
  switched to the wait-free variant.
* A config reload that switches to an explicit `wsl_domains` list cancels
  the in-flight run and applies the whole of the rest of that reload -- the
  explicit WSL domains, the exec domains, the serial ports and the new
  default -- while still holding the discovery lock, so no waiter can
  observe a half-applied configuration. Publishing after only the WSL part
  would let a woken waiter be told an exec or serial domain the new config
  does define is invalid, or read the previous default; since `Mux` has no
  WSL-specific knowledge, a waiter parked on discovery may well be waiting
  for a name that has nothing to do with WSL, which is what makes that
  observable. Cancelling resets to
  `NotStarted`, never `Done`: conflating "cancelled without discovering
  anything" with "completed" would make a later reload back to
  auto-discovery skip discovery entirely for the rest of the process's
  life. That atomic cancel-and-replace is the only cancellation entry
  point exposed; the two-phase split it is built on stays crate-private,
  since a caller that drops the drained `Waker`s, returns early, or panics
  between the two halves parks every async waiter forever, and
  `#[must_use]` only warns.
* `wsl.exe` failures stay failures. They finish the run as `NotStarted`
  (retryable on the next reload) and are logged, rather than being cached
  as a successful empty list -- otherwise one transient failure means "you
  have no WSL domains" until restart. A `pending_default` name a completed
  run already failed to find is remembered, so a stale/typo'd
  `default_domain` doesn't spawn a fresh discovery thread on every config
  reload just to re-derive the same "not found" answer.
* Concurrent `wsl.exe` invocations *on this path* are capped at one via a
  claimed-timestamp guard (`wezterm.default_wsl_domains()` deliberately
  bypasses it, since it is documented to always produce a live answer);
  a caller that finds a run already in flight waits for it (bounded)
  and adopts its result instead of starting a second one, but a genuinely
  abandoned (permanently hung) claim is detected by age and treated as
  stale rather than blocking every later caller for the same timeout
  forever. A run still in flight past `DOMAIN_DISCOVERY_STUCK_THRESHOLD` is
  likewise superseded by the next `begin_domain_discovery` call rather than
  making it a no-op, so a permanently hung `wsl.exe` doesn't pin the state
  machine in `InProgress` for the life of the process.

`wezterm.mux.get_domain()`/`wezterm.mux.all_domains()` stay synchronous
snapshot lookups: both are reachable from synchronous lua callbacks
(`format-tab-title` and friends), where an mlua async function would fail
with "attempt to yield from outside a coroutine". This means a
`gui-startup`/`mux-startup` handler that runs before discovery finishes may
transiently not see an auto-discovered WSL domain by name; a config that
needs one deterministically available at startup should set `wsl_domains`
explicitly. `wezterm.default_wsl_domains()` keeps its documented
always-live semantics and is deliberately not served from the cache; only
wezterm's own internal default-list computation is cached. Because that
cache lives as long as the process does, a long-lived one (notably
`wezterm-mux-server`) won't notice a distribution installed or removed
after it started, even across a reload; the changelog and the docs for
`default_wsl_domains()` spell that out, along with the always-live escape
hatch and its own limit -- it picks up additions, but domains already
registered with the mux are never unregistered.

The concurrency details (lock ordering, panic-safety via
`FinishDomainDiscoveryOnDrop`, `catch_unwind` around the caller-supplied
`accepts_as_default` predicate, generation-based staleness, wake-after-
unlock, and how a superseded run's late success gets adopted -- including
applying the *current* default request rather than the one it was spawned
with) are covered by doc comments at each site and by named regression
tests in `mux/src/lib.rs` and `config/src/wsl.rs`; not restated here.

Known limitation: `WslDistro::load_distro_list` still invokes `wsl.exe`
without a deadline, so a permanently hung one is worked around (startup no
longer blocks on it, and stuck runs are superseded) rather than resolved --
each supersede leaves the blocked thread and its child process behind.
Giving that call a timeout with a kill and reap would address the root
cause, but that is pre-existing behaviour unrelated to the latency this
change targets, so it is left for a follow-up.

## Test plan

- [x] `cargo test -p mux -p config --lib` (61 tests, new regression tests for
      each of: non-blocking startup, bounded vs. unbounded waits, the
      construction-time fallback while the cache is unpopulated, applying
      `pending_default` at all three sites it can be applied, cancel/replace
      semantics, zero-distribution discovery, stuck-run supersession)
- [x] `cargo check` across `wezterm-gui`, `wezterm-mux-server`,
      `wezterm-mux-server-impl`, `mux-lua`, `wezterm`
- [x] `cargo +nightly fmt --all -- --check`
- [x] Built and ran both debug and release `wezterm-gui`/`wezterm` on
      Windows with a real (stopped) WSL distribution installed; GUI startup
      no longer blocks on `wsl.exe`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3550,6 +3550,7 @@ name = "mux"
 version = "0.1.0"
 dependencies = [
  "anyhow",
+ "async-io",
  "async-trait",
  "base64",
  "bintree",
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -943,11 +943,35 @@ impl Config {
         }
     }
 
+    /// Returns the WSL domain list for this config.
+    ///
+    /// **This can block the calling thread**, potentially for tens of seconds, if `wsl_domains`
+    /// isn't configured explicitly: it falls through to `WslDomain::default_domains()`, which shells
+    /// out to `wsl.exe -l -v` (or waits for another in-flight invocation of it) on a cache miss.
+    /// Nothing in this crate calls this method anymore -- prefer `wsl_domains_cached()`, the
+    /// non-blocking accessor used everywhere latency matters, and treat a `None` result as "not
+    /// discovered yet" rather than triggering discovery yourself.
     pub fn wsl_domains(&self) -> Vec<WslDomain> {
         if let Some(domains) = &self.wsl_domains {
             domains.clone()
         } else {
             WslDomain::default_domains()
+        }
+    }
+
+    /// Like `wsl_domains()`, but returns a non-blocking result for the
+    /// auto-discovery case: if `wsl_domains` is not configured, returns
+    /// `Some(cached list)` once discovery has populated it, or `None` while
+    /// discovery hasn't completed yet. Callers must treat `None` differently
+    /// from `Some(vec![])` ("discovery finished and found nothing") -- see
+    /// `LocalDomain::resolve_wsl_domain` for why. Never shells out to
+    /// `wsl.exe`. Use this on latency-sensitive paths (like pane spawn) to
+    /// avoid blocking.
+    pub fn wsl_domains_cached(&self) -> Option<Vec<WslDomain>> {
+        if let Some(domains) = &self.wsl_domains {
+            Some(domains.clone())
+        } else {
+            WslDomain::cached_default_domains()
         }
     }
 
--- a/config/src/lua.rs
+++ b/config/src/lua.rs
@@ -333,7 +333,13 @@ end
 
         wezterm_mod.set(
             "default_wsl_domains",
-            lua.create_function(|_, ()| Ok(crate::WslDomain::default_domains()))?,
+            // Deliberately calls the uncached compute_default_domains(),
+            // not default_domains() (used internally by discovery/spawn
+            // paths that shouldn't repeatedly shell out): this function
+            // is documented to reflect what's currently installed, and a
+            // caller invoking it directly is opting into paying the
+            // `wsl.exe` cost for a live answer.
+            lua.create_function(|_, ()| Ok(crate::WslDomain::compute_default_domains()))?,
         )?;
 
         wezterm_mod.set("font", lua.create_function(font)?)?;
--- a/config/src/wsl.rs
+++ b/config/src/wsl.rs
@@ -2,9 +2,11 @@ use crate::config::validate_domain_name;
 use crate::*;
 use luahelper::impl_lua_conversion_dynamic;
 use std::collections::HashMap;
+use std::sync::{Condvar, Mutex};
+use std::time::{Duration, Instant};
 use wezterm_dynamic::{FromDynamic, ToDynamic};
 
-#[derive(Default, Debug, Clone, FromDynamic, ToDynamic)]
+#[derive(Default, Debug, Clone, PartialEq, FromDynamic, ToDynamic)]
 pub struct WslDomain {
     #[dynamic(validate = "validate_domain_name")]
     pub name: String,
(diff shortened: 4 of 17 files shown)


PHPCraftdream:
Context: you suggested opening this
[here](https://github.com/PHPCraftdream/wezterm/commit/9c7bb15081a83fdbee8b45271fb3a5d2c8218b94#commitcomment-196046168),
after pointing at Wez's reply on #7782. This PR is the answer to that reply — the
state machine is the "synchronization for discovery" he said was missing, and the
`default_domain`/`--domain` path waits unbounded so starting into a WSL domain
doesn't regress.

Same disclosure as on #8060: the engineering is Claude's, not mine — my part was
direction and verification. For calibration rather than as an argument, the fork you
were looking at is built the same way, as is everything on my account this year: 584
commits in the active fork, no human-written line in any of them.

I said in that thread I'd send this upstream as a PR and address the gap Wez pointed
out — I consider that delivered. Thanks for the upstream worth building on.

PR #8060, opened 2026-08-17
promise: fix two lost wakeups and a panic in spawn_into_new_thread

PHPCraftdream:
Three independent problems in the same future.

The worker thread does `tx.send(res).unwrap()`, which panics if the
receiving future has already been dropped -- a cancelled or timed-out
caller takes down a worker thread rather than being ignored. Nothing
reads the result at that point, so dropping it is the correct response.

More seriously, `poll` had a lost-wakeup window: it did a single
`try_recv`, and only on `Empty` went on to store the waker. A worker that
sent its result between those two steps found no waker to wake, and the
future stayed `Pending` with nobody left to poll it again -- a hang, not a
delay. Register the waker first and then re-check, the standard ordering
for this pattern.

The third is the same failure from the other direction: if `f()` panics,
the unwind skips both the send and the wake, so a waker registered by an
earlier poll is never woken and the future parks forever. `poll` already
knows what to do about a worker that produced no result -- it reports the
disconnected channel as an error -- it just never gets asked. Waking from
a drop guard covers the unwind path too.

Order matters there, and not in the obvious direction: a closure's
captured variables outlive the locals declared in its body, so a guard
holding only the waker would wake the poller while the sender was still
alive, and the poll it triggers would find the channel empty but still
connected, re-register, and sleep again -- the original bug, reintroduced
on the panic path. The sender is moved into an inner scope so it is
dropped first and the wake finds the channel disconnected. The guard also
tolerates a poisoned waker lock rather than unwrapping it, since it can
run while already unwinding, where a second panic aborts the process.
The `wake()` call itself is left unguarded: `Waker`'s contract says it
does not panic, and the only route to one here is a poisoned scheduler
mutex, which has already broken every other spawn in the process.

`spawn_into_new_thread` backs several paths: `termwiztermtab`,
`wezterm-client`'s domain attach, the overlay helpers.

The crate had no test module before this; added two, both deterministic.
One is a plain sanity check on the happy path, to catch a future
regression that breaks the channel or the wake outright (eg. removing
`tx.send`). The other pins down the panic-in-`f` hang this fixes: on the
old code it never returns, and shows up as a hung test rather than a
clean failure -- verified by temporarily removing the guard and running
just that test under an external timeout. The other lost-wakeup (the
`poll`-side race between `try_recv` and registering the waker) genuinely
needs the worker to land inside a window measured in the gap between two
adjacent statements, which isn't something these tests attempt to pin
without adding injection points to the production path.

## Test plan

- [x] `cargo test -p promise --lib` (2 new tests: happy path, and a
      regression test for the panic-in-`f` hang, mutation-tested by
      temporarily removing the `WakeOnExit` guard and confirming the test
      hangs under an external timeout instead of failing cleanly)
- [x] `cargo check -p promise -p mux -p wezterm-client -p wezterm-gui`
      (the three real consumers of `spawn_into_new_thread`)
- [x] `cargo +nightly fmt --all -- --check`

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -152,6 +152,11 @@ As features stabilize some brief notes about them will accumulate here.
   search matching. Thanks to @mrdziuban! #7385
 
 #### Fixed
+* `wezterm connect` and other operations backed by `spawn_into_new_thread`
+  (overlays such as the launcher and debug overlay, `wezterm-client`'s
+  domain attach) could hang forever instead of completing or reporting an
+  error, if the background thread finished its work in a narrow window
+  while the caller was going to sleep, or if it panicked.
 * `ResetTerminal` (RIS) did not reset the `modifyOtherKeys` state. A program
   that left it enabled and exited uncleanly could leave ctrl keys emitting
   escape sequences that the shell doesn't expect. RIS now also resets the
--- a/promise/src/spawn.rs
+++ b/promise/src/spawn.rs
@@ -74,17 +74,51 @@ where
 
     let thread_waker = Arc::clone(&holder);
     std::thread::spawn(move || {
-        // Run the thread
-        let res = f();
-        // Pass the result back
-        tx.send(res).unwrap();
-        // If someone polled the thread before we got here,
-        // they will have populated the waker; extract it
-        // and wake up the scheduler so that it will poll
-        // the result again.
-        let mut waker = thread_waker.waker.lock().unwrap();
-        if let Some(waker) = waker.take() {
-            waker.wake();
+        // Wake the poller from a drop guard rather than inline at the end of
+        // this closure, so that it happens on every exit from this thread --
+        // including `f()` panicking and unwinding straight past here. `poll`
+        // below already knows how to report a thread that produced no result
+        // (it turns the disconnected channel into an error), but it only gets
+        // the chance if something wakes it up to look; without this the future
+        // would stay parked for the rest of the process's life.
+        struct WakeOnExit(Arc<WakerHolder>);
+
+        impl Drop for WakeOnExit {
+            fn drop(&mut self) {
+                // Tolerate a poisoned lock rather than unwrapping: this may
+                // run while already unwinding, where panicking a second time
+                // aborts the process. `wake()` below is left unguarded --
+                // `Waker`'s contract says it doesn't panic, and the only way
+                // it could here is a poisoned scheduler mutex, which has
+                // already broken every other spawn in the process.
+                let mut waker = match self.0.waker.lock() {
+                    Ok(waker) => waker,
+                    Err(err) => err.into_inner(),
+                };
+                // If someone polled the thread before we got here,
+                // they will have populated the waker; extract it
+                // and wake up the scheduler so that it will poll
+                // the result again.
+                if let Some(waker) = waker.take() {
+                    waker.wake();
+                }
+            }
+        }
+
+        let _wake_on_exit = WakeOnExit(thread_waker);
+
+        {
+            // Rebinding moves the sender into this inner scope, so that it is
+            // dropped before the guard above wakes anyone. The order matters:
+            // a poll racing in between a wake and the sender's drop would find
+            // the channel empty but still connected, re-register its waker,
+            // and go back to sleep with nothing left to wake it.
+            let tx = tx;
+            // Run the thread
+            let res = f();
+            // Pass the result back, but don't panic if the receiving future
+            // was already dropped/cancelled (channel disconnected).
+            let _ = tx.send(res);
         }
     });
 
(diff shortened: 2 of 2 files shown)


bew:
Hello, how did you find out about this issue?
If there is no related issue, could you create a new one for this following the bug template? With proper environment info & reproduction steps that I can test out, and additionally the root-cause you identified & its effects..

This looks mostly AI-generated, but for transparency: how much AI was used for this contribution? What for? (research/implementation/testing/PRwriting)

PHPCraftdream:
Thanks for looking. Answers below.

## How I found it

There's no existing issue — this came out of #8059 (WSL domain discovery latency,
#7782).

An early version of that change implemented its async waits on top of
`spawn_into_new_thread`. `split_pane` and `spawn_tab_or_window` are `async fn` on
the single-threaded main-thread executor, so they needed to wait for discovery
without blocking it, and this helper looked like the ready answer. Reviewing that
design turned up the lost-wakeup window in `poll`, and it mattered: layered under
the WSL wait, a perfectly *successful* discovery could still leave a pane spawn
`Pending` forever. That made it a blocker for #8059.

Discovery later moved onto its own future (`DomainDiscoveryWaitFuture`), which is
why #8059 as it stands doesn't touch this path any more. The defects in the helper
remain for its existing users, hence a separate PR rather than folding them into
that one.

## The three defects

1. `tx.send(res).unwrap()` panics the worker thread if the receiving future was
   already dropped. A cancelled or timed-out caller takes down a thread instead of
   being ignored; nothing reads the result at that point, so dropping it is the
   correct response.

2. `poll` did a single `try_recv` and only stored the waker on `Empty`. A worker
   that sends between those two steps takes the waker mutex, finds nothing, and
   doesn't wake; `poll` then stores its waker and returns `Pending` with nobody
   left to poll it again. Fixed by registering first, then re-checking.

3. A panic in `f()` unwinds past both the send and the wake, so a waker registered
   by an earlier poll is never woken and the future parks for the life of the
   process. `poll` already knows how to report a worker that produced no result —
   it turns the disconnected channel into an error — it just never gets asked.
   Waking from a drop guard covers the unwind path.

The drop order in (3) isn't the obvious one: a closure's captured variables outlive
the locals declared in its body, so a guard holding only the waker would wake the
poller while the sender was still alive. The poll it triggers would find the
channel empty but still connected, re-register, and sleep again — defect (2),
reintroduced on the panic path. The sender is moved into an inner scope so it drops
first and the wake finds a disconnected channel.

## Reproduction

**(3)** — on this branch, delete the line
`let _wake_on_exit = WakeOnExit(thread_waker);` from `promise/src/spawn.rs` and run:

```
cargo test -p promise --lib spawn_into_new_thread_panic_reports_an_error_instead_of_hanging
```

It stops finishing and has to be killed. Restore the line and it passes
immediately. Not strictly deterministic: if the worker finishes unwinding before
the first poll, the receiver sees a disconnected channel and reports an error
instead of hanging. In practice the poll gets there first.

**(1)** — deterministic in a test: hold `f` on a channel, drop the `Task`, release
`f`, and the worker panics on `unwrap`. In the tree, `attach()` in
`wezterm-client/src/domain.rs` awaits such a future directly, so a cancelled
connection attempt looks reachable — I didn't trace whether the cancellation
actually drops it. The overlay call sites in `wezterm-gui/src/overlay/mod.rs` go
through `spawn(future).detach()` and shouldn't be able to hit it.

**(2)** — no manual reproduction. The window is a few instructions wide; the
argument is structural.

No field occurrence for any of the three. The tests came afterwards — they pin the
behaviour down, they didn't discover it.

## Testing

The crate had no test module; this adds two, both deterministic. One is a sanity
check on the happy path, to catch a future regression that breaks the channel or
the wake outright. The other is the panic test above.

Both were mutation-tested: the fix was temporarily disabled and the test confirmed
to fail — or, for the panic case, to hang under an external timeout — before being
restored. That's what shows the tests aren't vacuous.

Manual testing was building and running both debug and release on Windows. That
shows the change doesn't break the working paths; it isn't evidence the bugs were
being hit.

## AI usage

The engineering here is Claude's, not mine. It found the defect, wrote the fix,
wrote the tests, wrote the commit message, and wrote this comment. My part was
direction: setting the constraints, making the calls at each fork (keeping the
existing timeout semantics rather than changing them, splitting this out of #8059,
deciding which edge cases to fix versus document), throwing out the attempts that
didn't survive inspection, and verifying the result on hardware. Weigh it as
machine-produced work with a human editor, not the other way round.

The review was machine work too: 21 written review passes across the whole series,
about eleven of them touching this PR, deliberately spread across different models
rather than asking the same one twice, including one non-Anthropic model as a
cross-check. Each produced a written report whose surviving findings were fixed and
re-reviewed. Several caught real defects in the AI's own earlier fixes — one "fix"
turned out to do nothing at all, and one change introduced a regression a later
pass found. Volume isn't independence, though: none of those passes was a human
reading the diff, except me.

## Where that leaves it

I can file an issue if you want one, though it would be a code-audit report rather
than a bug report with reproduction steps. If anything here doesn't suit you, for
whatever reason, say so directly and close the PR.

bew:
> I can file an issue if you want one, though it would be a code-audit report rather
> than a bug report with reproduction steps.

A code-audit report wouldn't be of much use though.. 
The reason I would prefer to have a proper bug report with repro steps rather than a code-audit report, is that this allows to think about the problem first, rather that being biased by an implementation, allowing for judgment about the PR direction & implementation.

This is not specific to AI-assisted PRs, it's the bar for any claimed concurrency bug.
But it matters more here: we don't want to vibe this project, every claimed issue needs human-verified evidence it's real, not just a model's structural argument for it.

By your own account (1) isn't traced end-to-end and (2) has no reproduction at all, just theory. A repro forces that check against the actual code.
(3) already has one via the test, that's fine. An issue with repro steps for (1) and (2) would help.

PHPCraftdream:
Filed: #8062, with repro steps rather than a code audit.

Branch to run it from: https://github.com/PHPCraftdream/wezterm/tree/repro-8060
It carries upstream's unfixed `spawn.rs` plus four tests, and the fix as a
separate patch instead of as code, so exactly one thing changes between the two
states. `bash repro-8060/verify.sh` runs the whole before/after matrix.

Both windows are a handful of instructions wide, so I made them reachable
rather than lucky: two `cfg(test)` hooks turn the window into a rendezvous
where `poll` waits until the worker has finished completely. On the unfixed
code the lost-wakeup and panic cases hang every time and have to be killed.

On your point that (1) wasn't traced end to end -- I traced it, and it doesn't
hold up. All four call sites either detach or are driven by something that
does, nothing races or times them out, and neither production user of
`ScopedExecutor` reaches this helper. The only drop today is process teardown,
where the panic is invisible. So (1) goes in the issue as hardening, not as a
bug; I'd rather correct that myself than have you find it.

(2) and (3) stand, and (3) has a user-visible end: `start_overlay` calls
`schedule_cancel_overlay` after `func`, so a panic in `func` leaves the overlay
pane on screen with nothing left to close it.

One question in the issue is yours to decide: the hooks are what make these
deterministic, and they sit in production source under `cfg(test)`. They're
deliberately not in this PR. Say the word if you'd rather have the
deterministic tests in-tree and carry them.

Issue #6344, opened 2024-11-01
Crash on zero-height kitty graphic


joouha:
### What Operating System(s) are you seeing this problem on?

Linux Wayland

### Which Wayland compositor or X11 Window manager(s) are you using?

KWin

### WezTerm version

20241031-235112-0983ae90

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

Displaying a kitty graphic with the parameters `c=0,h=0` causes a divide-by-zero error, causes the terminal to crash, and all is lost.

### To Reproduce

Display a kitty graphic with `c=0,h=0`, e.g.:

```bash
echo -e '\e_Gr=0,h=0,a=T,t=d,f=100;iVBORw0KGgoAAAANSUhEUgAAAAsAAAALCAYAAACprHcmAAAACXBIWXMAAAGKAAABigEzlzBYAAAAOUlEQVQYlZXOwQ0AMAzCQEdi7yaT0xWAN7JuDCac2PQKYxflycOoICOKtPIuqFCg4/LzKxiz6xjyAYh9DR1sLUN1AAAAAElFTkSuQmCC\e\'
```

### Configuration

Default

### Expected Behavior

I'd probably expect wezterm to not display an image and not crash

### Logs

```
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.54s
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.54s
     Running `target/debug/wezterm-gui`
     Running `target/debug/wezterm-gui`
10:06:48.539  ERROR  env_bootstrap > panic at term/src/terminalstate/image.rs:114:21 - attempt to divide by zero
   0: env_bootstrap::register_panic_hook::{{closure}}
             at /home/user/data/projects/_forks/wezterm/env-bootstrap/src/lib.rs:176:18
   1: std::panicking::rust_panic_with_hook
   2: std::panicking::begin_panic_handler::{{closure}}
   3: std::sys::backtrace::__rust_end_short_backtrace
   4: rust_begin_unwind
   5: core::panicking::panic_fmt
   6: core::panicking::panic_const::panic_const_div_by_zero
   7: wezterm_term::terminalstate::image::<impl wezterm_term::terminalstate::TerminalState>::assign_image_to_cells::{{closure}}
             at /home/user/data/projects/_forks/wezterm/term/src/terminalstate/image.rs:114:21
[...]

PR #8064, opened 2026-08-18
fix(kitty): nine protocol key and default-value corrections

i-am-logger:
Nine places where the kitty parser disagrees with the protocol. One commit each,
every one with a test that fails without it.

Four are the same bug: a key whose protocol default is `0` was read as a literal
`0` rather than as absent.

| | wezterm read/wrote | should be | backed by |
| --- | --- | --- | --- |
| `S=0` | a request for zero bytes | the same command as omitting `S` | default + impl |
| raw `f=24`/`f=32` | refused unless the length matched exactly | pixel count from `f`, `s`, `v` | spec + impl |
| placement `w` `h` `c` `r` `=0` | a rectangle of nothing | not given | spec (`w`/`h`), impl (`c`/`r`) |
| compose `w=0` `h=0` | copied nothing | the whole frame | spec |
| `p=0` | a placement filed under id 0 | no placement id | spec |
| `O` (offset) | written under `S`, overwriting the size | `O` | spec |
| frame gap | read and written as `Z` | `z` | spec |
| transmit-and-display | encoded `a=Q` | `a=T` | spec |
| delete at z-index | encoded `d=p`, a broader delete | `d=q` | spec |

The placement row is what divided by zero in #6344. That fix guarded the division
by refusing the placement, which stopped the crash but left a conformant image
undrawn; this removes the case at its source and keeps the guard for a placement
that really has nothing to draw. The last three rows change only `to_keys`, so
they reach consumers of `termwiz::escape` that re-encode.

### References

Spec quotes are from
[graphics-protocol.rst](https://raw.githubusercontent.com/kovidgoyal/kitty/master/docs/graphics-protocol.rst);
the reference implementation is
[kitty/graphics.c](https://github.com/kovidgoyal/kitty/blob/master/kitty/graphics.c).
I have marked what the spec **states** separately from what only the
implementation decides, because two of these are not spec-stated and I would
rather say so than have you find it.

- **`S=0`** — the [control data reference](https://sw.kovidgoyal.net/kitty/graphics-protocol/#control-data-reference)
  gives `S` default `0`, and defines Default as "the value they take when
  missing", so a missing `S` and `S=0` are the same value. That `0` must not mean
  a zero-length read is the *implementation's* decision, not the spec's:
  `mmap_img_file` treats a zero size as "map the whole file" via `fstat`. kitty
  also cannot distinguish the two — `GraphicsCommand g = {0}` with no presence
  flag (`parse-graphics-command.h`).
- **raw length** — spec, "RGB and RGBA data": *"Since `f=24` there are three
  bytes per pixel and therefore the pixel data must be `3 * 10 * 20 = 600`
  bytes."* That the source may be **longer** is the implementation's call:
  `process_image_data` errors only when the source is *short*
  (`mapped_file_sz < data_sz` → `ENODATA`), never when it is long. Note the spec
  shows `t=s` both with `S` (`s=10,v=2,t=s,S=80,O=10`, a partial read) and
  without (`i=31,s=10,v=2,t=s`) — so `S` is optional for raw formats, not
  confined to PNG.
- **placement `w`/`h`** — spec states it: *"The width (in pixels) of the image
  area to display. By default, the entire width is used."* For **`c`/`r`** the
  spec gives the same `0` default but never says what a zero means, so reading it
  as absent is wezterm's inference — a defensible one, since zero cells is not a
  placement.
- **compose `w`/`h`** — spec states it, [animation composition](https://sw.kovidgoyal.net/kitty/graphics-protocol/#composing-animation-frames):
  *"By default, the entire width is used."* kitty agrees: `width = g->width ?
  g->width : img->width`.
- **`p=0`** — spec: *"add the `p` key with a number between `1` and
  `4294967295`"*, and *"Not specifying a placement id or using `p=0` … results in
  multiple placements"*. kitty treats it as match-all: `!g->placement_id ||
  ref->client_id == g->placement_id`.
- **`d=q` vs `d=p`** — spec, [deleting images](https://sw.kovidgoyal.net/kitty/graphics-protocol/#deleting-images):
  `q` is *"Delete all placements that intersect a specific cell having a specific
  z-index"*. That `p` ignores z-index is inferred from its row omitting it, and
  confirmed by `point_filter_func` having no z term. Neither is cursor-relative —
  both take the cell from `x`/`y`; `d=c` is the cursor one.
- **`O`/`S`, `z`, `a=T`** — all in the [control data reference](https://sw.kovidgoyal.net/kitty/graphics-protocol/#control-data-reference)
  key tables.

### How to test

```
cargo test --all
```

Eight tests added. Each fails on `main`; the parser ones are
struct-literal round trips in `wezterm-escape-parser/src/apc.rs`, so a reviewer
can read the expected value directly.

`kitty_zero_dimension_image_does_not_panic` asserted that `r=0,h=0` is refused,
which is what the placement fix corrects. It is now
`kitty_zero_dimension_image_is_displayed`, with a separate test covering the
placement that genuinely has nothing to draw — a source origin outside the image,
which the #6344 guard still refuses.

Known divergence, stated rather than hidden: on `t=d` kitty caps the slack at
`data_sz + 10` bytes and returns `EFBIG` beyond it, while this accepts any
over-long source and truncates. The motivating case is `t=f`/`t=t`/`t=s`, where
kitty is permissive too.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -156,35 +156,16 @@ As features stabilize some brief notes about them will accumulate here.
 #### Fixed
 * perf: Terminal images were hashed three times each on the transmit path; the sha256
   an RGBA image already carries is now reused instead. Thanks to @i-am-logger! #8065
-* Fixed raw format kitty images being refused when the file or shared memory
-  object holding them is longer than the image itself, which a shared memory
-  object always is once it is rounded up to a page. The pixel count follows from
-  `f`, `s` and `v`, so only that much is read.
-* Fixed a kitty image data size of zero being read as a request for no bytes.
-  `S` defaults to zero, so that is a size the client did not give.
-* Fixed the kitty image protocol data offset being re-encoded under the `S` key,
-  which overwrote the size and dropped `O` entirely. This affects anything that
-  parses an image escape sequence and writes it back out through `Display`.
-* Fixed the kitty animation frame gap being read and written as `Z` rather than
-  the `z` the protocol uses, so a client's frame gap was ignored and every
-  animated frame fell back to the 40ms default.
-* Fixed transmit-and-display re-encoding as `a=Q`, which is not an action the
-  protocol defines; it is `a=T`. This affects anything that parses an image
-  escape sequence and writes it back out through `Display`.
-* Fixed a delete that names a z-index re-encoding as `d=p`, which deletes every
-  placement in the cell whatever its z-index; it is `d=q`. This affects anything
-  that parses an image escape sequence and writes it back out through `Display`.
-* Fixed kitty images being refused when the placement gives `w`, `h`, `c` or `r`
-  as zero. All four default to zero, and zero is how the protocol spells "not
-  given": for `w` and `h` it says the entire width and height are used, and it
-  gives `c` and `r` the same default without defining a zero. Reading those
-  zeros as dimensions is what divided by zero in #6344. A placement with nothing
-  to draw, where the source origin lies outside the image, is still refused.
-* Fixed a kitty animation frame composition with `w=0` or `h=0` composing
-  nothing. Both default to zero, which the protocol reads as the whole frame.
-* Fixed a kitty delete with `p=0` matching only placements recorded under a
-  placement id of zero. Placement ids run from 1, so `p=0` is the same command
-  as omitting `p`: it names every placement of the image.
+* Fixed nine places where the kitty image protocol implementation disagreed with
+  the spec. Keys whose protocol default is `0` were read as a literal zero
+  rather than as absent, and raw images were refused when their source held more
+  bytes than their dimensions imply, as a shared memory object always does.
+  Between them these refused conformant images, and were the root of the divide
+  by zero in #6344. #8064
+* Fixed the kitty image data offset, the animation frame gap, transmit-and-display
+  and delete-at-a-z-index being re-encoded under the wrong keys. This affects
+  anything that parses an image escape sequence and writes it back out through
+  `Display`. #8064
 * `ResetTerminal` (RIS) did not reset the `modifyOtherKeys` state. A program
   that left it enabled and exited uncleanly could leave ctrl keys emitting
   escape sequences that the shell doesn't expect. RIS now also resets the
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -156,6 +156,35 @@ As features stabilize some brief notes about them will accumulate here.
 #### Fixed
 * perf: Terminal images were hashed three times each on the transmit path; the sha256
   an RGBA image already carries is now reused instead. Thanks to @i-am-logger! #8065
+* Fixed raw format kitty images being refused when the file or shared memory
+  object holding them is longer than the image itself, which a shared memory
+  object always is once it is rounded up to a page. The pixel count follows from
+  `f`, `s` and `v`, so only that much is read.
+* Fixed a kitty image data size of zero being read as a request for no bytes.
+  `S` defaults to zero, so that is a size the client did not give.
+* Fixed the kitty image protocol data offset being re-encoded under the `S` key,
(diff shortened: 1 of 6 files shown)


wezterm-escape-parser/src/apc.rs
```diff
@@ -949,6 +994,7 @@ pub struct KittyImageFrame {
 
     /// Gap in milliseconds of this frame from the next one.
     /// Zero or omitted values are interpreted as 40ms.
+    /// A negative gap asks for a gapless frame, which is not supported.
     /// z=...
     pub duration_ms: Option<u32>,
```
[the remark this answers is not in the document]
i-am-logger: Fixed. It now says what actually happens rather than "not supported": a negative gap does not parse as `u32`, so it arrives as `None` and takes the same 40ms as an omitted gap.

docs/changelog.md
```diff
@@ -152,6 +152,35 @@ As features stabilize some brief notes about them will accumulate here.
   search matching. Thanks to @mrdziuban! #7385
 
 #### Fixed
+* Fixed raw format kitty images being refused when the file or shared memory
```
bew: I think this is way too many details to be in the user changelog, this should be simplified to a few lines about improving the kitty impl to better match the spec, and mention the PR like other changelog entries.

term/src/test/image.rs
```diff
@@ -70,3 +85,54 @@ fn kitty_image_with_zero_pixel_dimensions_does_not_panic() {
...
+            }
+        }
+    }
+    panic!("no image was attached to the screen");
+}
+
+/// For the raw formats the protocol derives the pixel count from `f`, `s` and
+/// `v`, so a source holding more than that is read up to the length the image
+/// needs. A shared memory object reporting more than was staged in it is the
+/// case that matters.
```
bew: about that last sentence, is `A shared memory object` referring to the shm memory way of transmitting the image (I don't think it's this one), or just referring to the image data?
please rephrase to make this less ambiguous.

wezterm-escape-parser/src/apc.rs
```diff
@@ -111,17 +111,26 @@ impl KittyImageData {
             "d" => Some(Self::Direct(String::from_utf8(payload.to_vec()).ok()?)),
             "f" => Some(Self::File {
                 path: String::from_utf8(base64_decode(payload.to_vec()).ok()?).ok()?,
-                data_size: geti(keys, "S"),
+                data_size: match geti(keys, "S") {
+                    None | Some(0) => None,
+                    n => n,
+                },
```
bew: instead of duplicating this match many times, could you refactor this into a named helper or a variant of `geti`?
This should make the code more obvious & easier to read 🤔

bew (commented):
I didn't review the details, but here are a few comments on the general structure of this changeset.

wezterm-escape-parser/src/apc.rs
i-am-logger: Done — `geti_zero_is_absent`, in `75ec0ea`.

It was seventeen sites rather than a few, so thanks for pushing on it. The two arms that read `None | Some(0)` as `AlphaBlending` are a different rule and are left as they are.

term/src/test/image.rs
i-am-logger: It meant the `t=s` transport, and you are right that nothing there said so — the test itself sends `t=d`, so the sentence had nothing to attach to.

Rephrased in `fcb2e9b`: it now names `t=s`, says why an over-long source is the normal case there (a POSIX shared memory object reports its size rounded up to a page), and says the test reaches the same length check inline instead.

docs/changelog.md
i-am-logger: Agreed, it was out of keeping with the entries around it. Cut to two lines in `75ec0ea`: one for what was misread, one for what was re-encoded under the wrong key, both citing the PR. #6344 stays where the divide by zero belongs.

PR #8065, opened 2026-08-18
perf(image): hash each transmitted image once

i-am-logger:
An `Rgba8` image is hashed three times on the transmit path. `compute_hash`
recomputed the sha256 that `Rgba8` already stores — once for the cache key in
`raw_image_to_image_data` (`term/src/terminalstate/image.rs:290`), once in
`ImageData::with_data` (`wezterm-cell/src/image.rs:564`), on top of the one
`new_single_frame` computed and kept (`image.rs:271`). It now returns the stored
hash, so three passes become one.

One pass over a 13.18MiB buffer (2400x1440 RGBA) is about 22ms on an Apple
M-series; `main` does three of them per image.

The value is unchanged — what `Rgba8` stores is the same sha256 `compute_hash`
was recomputing. `wezterm-gui`'s frame cache already keyed on that stored field
(`glyphcache.rs:918`), including for images arriving over the mux, so this makes
`compute_hash` agree with what the renderer already trusted.

The other variants are untouched: `EncodedFile` still hashes on demand,
`EncodedLease` already returned its content id without hashing, and `AnimRgba8`
keeps hashing because its `compute_hash` also covers frame durations.

Same redundancy `1c9bd347` fixed in `ImageData::hash`, in a path that commit did
not reach.

### Testing

```
cargo test --all
```

One test covers the invariant this relies on: that `compute_hash()` on an
`Rgba8` still describes its pixels after a kitty `a=f` in-place frame edit.
Break the refresh and it fails; break the blit and it fails on a different
assertion, so the two can be told apart.

It cannot detect a revert to recomputing, and no test can — the stored and the
recomputed value are equal by construction. What it guards is the invariant that
makes returning the stored one safe.

*AI-disclosure: drafted by AI, manually reviewed and iterated on by me.*

### DIFF
--- a/wezterm-cell/src/image.rs
+++ b/wezterm-cell/src/image.rs
@@ -216,6 +216,8 @@ pub enum ImageDataType {
         data: Vec<u8>,
         width: u32,
         height: u32,
+        /// sha256 of data, which is what compute_hash reports. Update it
+        /// whenever data changes.
         hash: [u8; 32],
     },
     /// Data is an animated sequence
@@ -304,14 +306,19 @@ impl ImageDataType {
 
     pub fn compute_hash(&self) -> [u8; 32] {
         use sha2::Digest;
-        let mut hasher = sha2::Sha256::new();
         match self {
-            ImageDataType::EncodedFile(data) => hasher.update(data),
-            ImageDataType::EncodedLease(lease) => return lease.content_id().as_hash_bytes(),
-            ImageDataType::Rgba8 { data, .. } => hasher.update(data),
+            // Both of these already hold the hash, so neither builds a hasher.
+            ImageDataType::EncodedLease(lease) => lease.content_id().as_hash_bytes(),
+            ImageDataType::Rgba8 { hash, .. } => *hash,
+            ImageDataType::EncodedFile(data) => {
+                let mut hasher = sha2::Sha256::new();
+                hasher.update(data);
+                hasher.finalize().into()
+            }
             ImageDataType::AnimRgba8 {
                 frames, durations, ..
             } => {
+                let mut hasher = sha2::Sha256::new();
                 for data in frames {
                     hasher.update(data);
                 }
@@ -320,9 +327,9 @@ impl ImageDataType {
                     let b = d.to_ne_bytes();
                     hasher.update(b);
                 }
+                hasher.finalize().into()
             }
-        };
-        hasher.finalize().into()
+        }
     }
 
     /// Divides the animation frame durations by the provided
@@ -584,5 +591,30 @@ impl ImageData {
 
     pub fn hash(&self) -> [u8; 32] {
         self.hash
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    /// The hash stored with an Rgba8 is the hash of its data.
+    #[test]
+    fn rgba8_compute_hash_matches_its_data() {
+        let data: Vec<u8> = (0..=255u8).cycle().take(16 * 16 * 4).collect();
+        let expected = ImageDataType::hash_bytes(&data);
+
+        let img = ImageDataType::new_single_frame(16, 16, data);
+
+        assert_eq!(img.compute_hash(), expected);
+    }
+
+    /// Distinct pixels must produce distinct hashes.
+    #[test]
+    fn rgba8_compute_hash_follows_the_data() {
+        let a = ImageDataType::new_single_frame(2, 2, vec![0u8; 16]);
+        let b = ImageDataType::new_single_frame(2, 2, vec![1u8; 16]);
+
+        assert_ne!(a.compute_hash(), b.compute_hash());
     }
 }
--- a/wezterm-cell/src/image.rs
+++ b/wezterm-cell/src/image.rs
@@ -310,11 +310,7 @@ impl ImageDataType {
             // Both of these already hold the hash, so neither builds a hasher.
             ImageDataType::EncodedLease(lease) => lease.content_id().as_hash_bytes(),
             ImageDataType::Rgba8 { hash, .. } => *hash,
-            ImageDataType::EncodedFile(data) => {
-                let mut hasher = sha2::Sha256::new();
-                hasher.update(data);
-                hasher.finalize().into()
-            }
+            ImageDataType::EncodedFile(data) => ImageDataType::hash_bytes(data),
             ImageDataType::AnimRgba8 {
                 frames, durations, ..
             } => {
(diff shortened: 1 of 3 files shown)


wezterm-cell/src/image.rs
```diff
@@ -308,7 +310,7 @@ impl ImageDataType {
         match self {
             ImageDataType::EncodedFile(data) => hasher.update(data),
```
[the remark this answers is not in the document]
i-am-logger: Fixed — each arm now builds its own hasher, so the two that already hold a hash never construct one.

Worth being accurate about the cost though: `Sha256::new()` is a stack struct init of a few nanoseconds, against ~4.12ms for the hash itself, so this is readability rather than measurable overhead. Constructing something and discarding it in a commit about removing wasted hashing was still the wrong look.

wezterm-cell/src/image.rs
```diff
@@ -304,14 +306,19 @@ impl ImageDataType {
...
-            ImageDataType::EncodedFile(data) => hasher.update(data),
-            ImageDataType::EncodedLease(lease) => return lease.content_id().as_hash_bytes(),
-            ImageDataType::Rgba8 { data, .. } => hasher.update(data),
+            // Both of these already hold the hash, so neither builds a hasher.
+            ImageDataType::EncodedLease(lease) => lease.content_id().as_hash_bytes(),
+            ImageDataType::Rgba8 { hash, .. } => *hash,
+            ImageDataType::EncodedFile(data) => {
+                let mut hasher = sha2::Sha256::new();
+                hasher.update(data);
+                hasher.finalize().into()
```
bew: is there any reason not to use `hash_bytes` here for `EncodedFile`?

```suggestion
                ImageDataType::hash_bytes(data)
```

wezterm-cell/src/image.rs
```diff
@@ -586,3 +593,28 @@ impl ImageData {
...
+        let expected = ImageDataType::hash_bytes(&data);
+
+        let img = ImageDataType::new_single_frame(16, 16, data);
+
+        assert_eq!(img.compute_hash(), expected);
+    }
+
+    /// Distinct pixels must produce distinct hashes.
+    #[test]
+    fn rgba8_compute_hash_follows_the_data() {
```
bew: I'm honestly not convinced these tests are very useful 🤔
How do you feel about those?

wezterm-cell/Cargo.toml
```diff
@@ -23,6 +23,13 @@ wezterm-color-types.workspace = true
 wezterm-escape-parser.workspace = true
 wezterm-dynamic.workspace = true
 
+# sha2 compiles its aarch64 sha256 intrinsics only under the asm feature; x86
+# and x86_64 select their backend by architecture and do not need it. The
+# feature also pulls in sha2-asm, which refuses to build for Windows targets,
+# hence the target gate.
```
bew: nit: such small reformatting makes the flow of this comment easier to read:
_(I think it's something you should teach your LLM how to do properly..)_

```suggestion
# sha2 compiles its aarch64 sha256 intrinsics only under the `asm` feature;
# x86 and x86_64 already select their backend by architecture and do not need it.
#
# The `asm` feature also pulls in `sha2-asm`, which refuses to build for Windows targets,
# hence the target gate.
```

bew (commented):
Thanks, changes seems good.
Small comments only 👇

bew:
Looking around the `sha2` crate, I saw that we're using version 0.10 which has the `asm` feature you're adding, but this was removed in sha2 version `0.11` ([ref](https://github.com/RustCrypto/hashes/blob/master/sha2/CHANGELOG.md#0110-2026-03-25)).

`uv` ~recently did this upgrade at https://github.com/astral-sh/uv/pull/20831, and notes:
> Upgrade to sha2 0.11, which selects its ARM backend automatically when CPU feature detection is available. This preserves accelerated distribution hashing on Apple Silicon and Linux/Android ARM while removing sha2-asm and its target-specific configuration.

So I'm thinking it might be worth upgrading `sha2` to `0.11`, could you try this and see what your benchmark says about it?
(maybe in a separate PR to avoid blocking this one in case we decide to stay on 0.10 🤔)

WDYT?

i-am-logger:
Measured, and it changed what this PR should be — so I have force-pushed and
dropped the `asm` commit. The force-push entry above has the before-and-after if
you want to look. If you would rather stay on 0.10 and take the target gate now,
say so and I will put that commit back on top.

Same buffer as before, 13.18MiB (2400x1440 RGBA) in a scratch crate, 3 warmups
and 20 timed iterations, one sha256 pass:

| build | per pass |
|---|---|
| 0.10, no features | ~22ms |
| 0.10 with `asm` | ~4ms |
| **0.11, nothing declared** | **~4ms** |
| 0.11, `--cfg sha2_256_backend="soft"` | ~20ms |

Run-to-run spread is a few percent, so the two hardware rows are not
distinguishable - 0.11 gives up nothing against `asm`.

On 0.10 the aarch64 branch is `#[cfg(all(feature = "asm", target_arch =
"aarch64"))]`, so a default build falls through to the portable backend. On 0.11
there is no such feature: both are compiled and `cpufeatures` chooses at
runtime. The last row is the control. x86 is unaffected either way — it already
selected SHA-NI at runtime on 0.10.

So the `asm` feature was the wrong mechanism, and adding a target gate here to
remove it later seemed worse than not adding it.

**This PR is now the deduplication alone** — `compute_hash` returns the sha256
that `Rgba8` already carries, three passes to one. It no longer touches
`Cargo.toml` or `Cargo.lock`: two source files and the changelog. That also
takes away the comment block you commented on.

The backend is now #8068, a one-line bump of the workspace
`sha2`. One thing to know before you look: `digest` 0.10 stays in the tree
alongside 0.11 for `sha1`, which arrives through `zbus`, so Linux and BSD builds
carry both and macOS and Windows do not.

Two unrelated notes. #8063 and #8064 have no check runs at all - their check
suites sit at `action_required`, which is where a fork PR's workflows wait to be
approved; #8065's ran, so it was approved. And `first_image` still collides with
#8064, as flagged on #8061 - so does the changelog block, which I found by trying
the merge.

*AI-disclosure: drafted by AI, manually reviewed and iterated on by me.*

wezterm-cell/src/image.rs
i-am-logger: No reason at all — applied. `hash_bytes` is the same three lines, so the arm was
just an inlined copy of it.

wezterm-cell/Cargo.toml
i-am-logger: Force-pushed since you looked: this PR no longer touches `Cargo.toml`, so the
block went with it. Formatting taken for the next comment that needs it.

wezterm-cell/src/image.rs
i-am-logger: You're right, both unit tests are gone. They only ever see a freshly built
image, which is the one state where a stored hash cannot be stale. I broke
`compute_hash`, then broke the `a=f` hash refresh in turn: they caught the first
and missed the second - they live in `wezterm-cell` and the refresh is in `term`,
so they could not have seen it.

The term test stays, rewritten. Its assertions read the `hash` field rather than
going through `compute_hash`, so a regression in that function slipped past it -
with the `Rgba8` arm returning a constant the old test still passed and the
rewritten one fails. It now takes its values from `compute_hash()` and asserts
the edited pixel:

```rust
k9::assert_equal!(&data[0..4], &[0xff, 0x00, 0x00, 0xff][..]);
k9::assert_equal!(edited, ImageDataType::hash_bytes(data));
```

The pixel line is what separates a blit that silently did nothing from a hash
that went stale — under the old `assert_ne!(*hash, transmitted)` both failed on
the same line. Net two tests fewer than the PR had.

`kitty_frame_compose` (`a=c`) edits the same pixels and has no coverage. Worth
saying that this PR raises the stakes on it: a stale hash used to reach only the
gui frame cache, and now it is what `compute_hash` returns. Happy to cover it
here or as a follow-up.

PR #8067, opened 2026-08-18
Missing verb in docs/config/keyboard-concepts.md

dstork:
"will" (or something equivalent) seems to be missing from the documentation on dead keys.


bew (approved):
Seems right, thank you!

PR #8073, opened 2026-08-20
pty: Replace winapi with windows-sys

xtqqczze:
Contributes to https://github.com/wezterm/wezterm/issues/7797.

### DIFF
--- a/pty/src/lib.rs
+++ b/pty/src/lib.rs
@@ -46,6 +46,8 @@ use serde::{Deserialize, Serialize};
 use std::io::Result as IoResult;
 #[cfg(windows)]
 use std::os::windows::prelude::{AsRawHandle, RawHandle};
+#[cfg(windows)]
+use windows_sys::Win32::System::Threading::TerminateProcess;
 
 pub mod cmdbuilder;
 pub use cmdbuilder::CommandBuilder;
@@ -303,9 +305,7 @@ impl ChildKiller for ProcessSignaller {
     fn kill(&mut self) -> IoResult<()> {
         if let Some(handle) = &self.handle {
             unsafe {
-                if winapi::um::processthreadsapi::TerminateProcess(handle.as_raw_handle() as _, 127)
-                    == 0
-                {
+                if TerminateProcess(handle.as_raw_handle(), 127) == 0 {
                     return Err(std::io::Error::last_os_error());
                 }
             }
--- a/pty/src/win/mod.rs
+++ b/pty/src/win/mod.rs
@@ -5,11 +5,10 @@ use std::os::windows::io::{AsRawHandle, RawHandle};
 use std::pin::Pin;
 use std::sync::Mutex;
 use std::task::{Context, Poll};
-use winapi::shared::minwindef::DWORD;
-use winapi::um::minwinbase::STILL_ACTIVE;
-use winapi::um::processthreadsapi::*;
-use winapi::um::synchapi::WaitForSingleObject;
-use winapi::um::winbase::INFINITE;
+use windows_sys::Win32::Foundation::STILL_ACTIVE;
+use windows_sys::Win32::System::Threading::{
+    GetExitCodeProcess, GetProcessId, TerminateProcess, WaitForSingleObject, INFINITE,
+};
 
 pub mod conpty;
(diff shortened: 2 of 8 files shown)


bew:
Please edit PR description to at least link to the tracking issue

In general: never open a PR without description 🙏

pty/src/win/mod.rs
```diff
@@ -24,11 +23,11 @@ pub struct WinChild {
 
 impl WinChild {
     fn is_complete(&mut self) -> IoResult<Option<ExitStatus>> {
-        let mut status: DWORD = 0;
+        let mut status: u32 = 0;
```
bew, line 26: I'd prefer to use the more specific type from winsys: `NTSTATUS`

e.g. https://docs.rs/windows-sys/latest/windows_sys/Win32/Foundation/constant.STILL_ACTIVE.html

pty/src/win/psuedocon.rs
```diff
@@ -12,30 +12,28 @@ use std::os::windows::io::{AsRawHandle, FromRawHandle};
...
 };
-use winapi::um::wincon::COORD;
-use winapi::um::winnt::HANDLE;
 
 pub type HPCON = HANDLE;
 
-pub const PSUEDOCONSOLE_INHERIT_CURSOR: DWORD = 0x1;
-pub const PSEUDOCONSOLE_RESIZE_QUIRK: DWORD = 0x2;
-pub const PSEUDOCONSOLE_WIN32_INPUT_MODE: DWORD = 0x4;
+pub const PSUEDOCONSOLE_INHERIT_CURSOR: u32 = 0x1;
```
bew, line 25: Similarly for these, can you explain why you changed them to u32 instead of using the more 'windowsy' DWORD type?

pty/src/lib.rs
```diff
@@ -303,9 +305,7 @@ impl ChildKiller for ProcessSignaller {
     fn kill(&mut self) -> IoResult<()> {
         if let Some(handle) = &self.handle {
             unsafe {
-                if winapi::um::processthreadsapi::TerminateProcess(handle.as_raw_handle() as _, 127)
-                    == 0
-                {
+                if TerminateProcess(handle.as_raw_handle() as _, 127) == 0 {
```
bew: I don't think you need the `as _` anymore 🤔

Please check other similar simplifications

bew (changes requested):
As mentioned below, can you explain each different non-import-changes you did in this PR?

pty/src/win/mod.rs
xtqqczze: The signature of [`GetExitCodeProcess`](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-getexitcodeprocess) is:

```c++
BOOL GetExitCodeProcess(
  [in]  HANDLE  hProcess,
  [out] LPDWORD lpExitCode
);
```

Since `DWORD` is `u32` while `NTSTATUS` is `i32`, I’d rather keep the output parameter as `u32` to match the API signature and cast the constant instead.

pty/src/win/psuedocon.rs
xtqqczze: `DWORD` isn’t defined in the `windows-sys` crate; since it’s just an alias for `u32`, the expected approach is to use `u32` directly.

pty/src/lib.rs
xtqqczze: Target type is `HANDLE = *mut core::ffi::c_void`, but `RawHandle = raw::HANDLE = *mut std::os::raw::c_void`, so it's arguable.

xtqqczze: `std::os::raw::c_void` maps to `core::ffi::c_void` anyway, removed.

pty/src/win/mod.rs
bew: Alright

pty/src/win/psuedocon.rs
bew: alright, I went to search the reasons, posted at https://github.com/wezterm/wezterm/issues/7797#issuecomment-5413245095

bew:
Thanks for the PR! 🚀

PR #8075, opened 2026-08-20
filedescriptor: prepare for new release

xtqqczze:


xtqqczze:
Depends on https://github.com/wezterm/wezterm/pull/7973

Cargo.toml
```diff
@@ -266,7 +266,7 @@ winapi = "0.3.9"
 window = {path="window"}
 window-funcs = { path = "lua-api-crates/window-funcs" }
 windows = "0.33.0"
-windows-sys = { version = "0.61.2", default-features = false }
+windows-sys = { version = ">=0.60.2, <0.62", default-features = false }
```
bew, line 269: hmm why the range?

bew:
As I mentioned in #8026, I don't have control over crate releases.
I think the crates will all get released in bulk when we do the next Wezterm release, so I don't think this PR is of much use right now.. I'd vote to close it

Cargo.toml
xtqqczze: some dependencies might not be on the latest version, so this helps reduce duplicate `windows-sys` crates.

xtqqczze:
The next wezterm release might not be for a while, and I'd like to use the `filedescriptor` crate in [`coreutils`](https://github.com/uutils/coreutils) with the changes I've contributed.

Cargo.toml
mati865: This is a common practice for libraries. Windows crates are quite a mess.

PR #8081, opened 2026-08-21
procinfo: Replace ntapi with windows-sys

xtqqczze:
Contributes to https://github.com/wezterm/wezterm/issues/7797.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3570,7 +3570,6 @@ dependencies = [
  "mlua",
  "names",
  "nix",
- "ntapi",
  "parking_lot",
  "percent-encoding",
  "portable-pty",
@@ -4514,6 +4513,7 @@ dependencies = [
  "ntapi",
  "wezterm-dynamic",
  "winapi",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


xtqqczze:
I think we need to keep using `ntapi`, since we’re doing something undocumented with `ProcessInformation`. The documentation describes it as a boolean indicating whether a process is running under WOW64, but we actually treat it as a pointer to the 32-bit Process Environment Block (PEB).

bew:
> I think we need to keep using `ntapi`

I saw that https://github.com/microsoft/windows-rs/issues/4844 has been closed, next could be to try to ask `ntapi` to switch from `winapi` to `windows-rs` 🤔

xtqqczze:
> next could be to try to ask `ntapi` to switch from `winapi` to `windows-rs` 🤔

This has already been requested, doesn't look like it will happen anytime soon: https://github.com/MSxDOS/ntapi/issues/9, https://github.com/MSxDOS/ntapi/pull/17

xtqqczze:
@bew How about we define `RTL_USER_PROCESS_PARAMETERS` and `RTL_USER_PROCESS_PARAMETERS32` and `UNICODE_STRING32` ourselves in order to remove the `ntapi` crate?

bew:
> @bew How about we define `RTL_USER_PROCESS_PARAMETERS` and `RTL_USER_PROCESS_PARAMETERS32` and `UNICODE_STRING32` ourselves in order to remove the `ntapi` crate?

yeah good idea, make sure to explain why we do this & link to impl in ntapi 🙏

PR #8096, opened 2026-08-25
Add command_palette_line_height config option

rafaelsteil:
## Summary
- Add `command_palette_line_height` (default `1.0`) so command palette row spacing can be scaled independently of global `line_height`.
- Apply the scale in both palette layout paths so row height and `max_rows_on_screen` stay consistent.
- Related: https://github.com/wezterm/wezterm/discussions/6615

#### Example of height `2.0`
image

## Test plan
- [ ] Open the command palette with default config and confirm spacing is unchanged
- [ ] Set `config.command_palette_line_height = 1.3` and confirm taller rows
- [ ] Confirm global `line_height` does not affect the palette unless this option is set

## Disclosure of AI usage
Planning and implementation made with the help of AI. Code reviewed by me, and UI manually tested and validated.

### DIFF
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -164,7 +164,7 @@ pub struct Config {
 
     #[dynamic(
         default = "default_one_point_oh_f64",
-        validate = "validate_command_palette_line_height"
+        validate = "validate_line_height"
     )]
     pub command_palette_line_height: f64,
 
@@ -2183,17 +2183,7 @@ fn validate_row_or_col(value: &u16) -> Result<(), String> {
 fn validate_line_height(value: &f64) -> Result<(), String> {
     if *value <= 0.0 {
         Err(format!(
-            "Illegal value {value} for line_height; it must be positive and greater than zero!"
-        ))
-    } else {
-        Ok(())
-    }
-}
-
-fn validate_command_palette_line_height(value: &f64) -> Result<(), String> {
-    if *value <= 0.0 {
-        Err(format!(
-            "Illegal value {value} for command_palette_line_height; it must be positive and greater than zero!"
+            "Illegal value {value}; it must be positive and greater than zero!"
         ))
     } else {
         Ok(())
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -162,6 +162,12 @@ pub struct Config {
     #[dynamic(default = "default_command_palette_font_size")]
     pub command_palette_font_size: f64,
 
+    #[dynamic(
+        default = "default_one_point_oh_f64",
+        validate = "validate_command_palette_line_height"
+    )]
+    pub command_palette_line_height: f64,
+
     pub command_palette_rows: Option<usize>,
     #[dynamic(default = "default_command_palette_fg_color")]
     pub command_palette_fg_color: RgbaColor,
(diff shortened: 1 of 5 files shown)


config/src/config.rs
```diff
@@ -2184,6 +2190,16 @@ fn validate_line_height(value: &f64) -> Result<(), String> {
     }
 }
 
+fn validate_command_palette_line_height(value: &f64) -> Result<(), String> {
+    if *value <= 0.0 {
+        Err(format!(
+            "Illegal value {value} for command_palette_line_height; it must be positive and greater than zero!"
```
bew: (to be verified) I think the config name is already present in any error you get, no need to repeat it in the error message I think 🤔

If so, could you refactor and use the `validate_line_height` for this config as well? no need to duplicate code for this

docs/config/lua/config/command_palette_line_height.md
```diff
@@ -0,0 +1,24 @@
...
+[ActivateCommandPalette](../keyassignment/ActivateCommandPalette.md).
+The default is `1.0`, which uses the font-specified metrics.
+
+If the command palette feels too vertically cramped then you can set
+`command_palette_line_height = 1.2` to increase the vertical spacing by 20%.
+Conversely, setting `command_palette_line_height = 0.9` will decrease the
+vertical spacing by 10%.
+
+This option is independent of the global
+[line_height](line_height.md) setting, which applies only to terminal cells.
```
bew: Move this to the end of the page, and rewrite it as:

```suggestion
See [line_height](line_height.md), which only applies to terminal cells.
```

(with a quick read I read `dependent` instead of `INdependent` 😅, so this edit removes the ambiguity)

docs/changelog.md
```diff
@@ -75,6 +75,10 @@ As features stabilize some brief notes about them will accumulate here.
   easier to spot the remaining candidates. Thanks to @mr-felixoid and @bew! #7752
 
 #### New
+* [command_palette_line_height](config/lua/config/command_palette_line_height.md)
+  option to scale the vertical spacing of rows in the command palette,
+  independently of the global [line_height](config/lua/config/line_height.md)
+  setting.
```
bew:
```suggestion
  independently of [line_height](config/lua/config/line_height.md) which is for terminal cells only.
  Thanks to @rafaelsteil! #8096
```

bew (commented):
Thanks for the PR, looks pretty straight forward ;)

Just a few comments

config/src/config.rs
rafaelsteil: Confirmed that the error message already includes the field. Pushed a change as you suggested. Btw, I validated it with a test in `config.rs` but didn't include it as there was no other tests there. 

```rust
#[cfg(test)]
mod tests {
    use super::*;
    use wezterm_dynamic::{FromDynamic, Object, ToDynamic, Value};

    #[test]
    fn command_palette_line_height_rejects_non_positive() {
        let mut obj = Object::default();
        obj.insert("command_palette_line_height".to_dynamic(), Value::I64(0));
        let err = Config::from_dynamic(&Value::Object(obj), Default::default()).unwrap_err();
        assert_eq!(
            err.to_string(),
            "Error processing Config::command_palette_line_height: Illegal value 0; it must be positive and greater than zero!"
        );
    }
}
```

rafaelsteil:
PR updated with the requested changes!

docs/config/lua/config/command_palette_line_height.md
```diff
@@ -0,0 +1,23 @@
...
+If the command palette feels too vertically cramped then you can set
+`command_palette_line_height = 1.2` to increase the vertical spacing by 20%.
+Conversely, setting `command_palette_line_height = 0.9` will decrease the
+vertical spacing by 10%.
+
+```lua
+config.command_palette_line_height = 1.2
+```
+
+See [line_height](line_height.md), which only applies to terminal cells.
```
bew:
```suggestion
See also [line_height](line_height.md), which only applies to terminal cells.
```

bew:
Works perfectly well, thank you!

Issue #8049, opened 2026-08-12
Improve default font weight changes for Normal vs Dim/Bold


bfoersterling:
### What Operating System(s) are you seeing this problem on?

Linux X11

### Which Wayland compositor or X11 Window manager(s) are you using?

X11 / `i3`.

### WezTerm version

20260812-070121-fe3006ae

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

This command will not print in bold/bright color:
```
printf "\e[1mfoo\e[0m"
```
This problem also translates to the program `aerc` which does not display bold colors when started inside `wezterm`.

### To Reproduce

```
printf "\e[1mfoo\e[0m"
```
and compare with
```
printf "foo"
```

### Configuration

I tried many different color schemes and even custom colors.\
These are the same colors I use in `terminator`:
```
config.colors = {
    foreground = '#dcdfe4',

        ansi = {
    '#282c34',
    '#e06c75',
    '#98c379',
    '#e5c07b',
    '#61afef',
    '#c678dd',
    '#56b6c2',
    '#dcdfe4',
  },
  brights = {
    '#282c34',
    '#e06c75',
    '#98c379',
    '#e5c07b',
    '#61afef',
    '#c678dd',
    '#56b6c2',
    '#dcdfe4',
  },
}
```

I tried setting `bold_brightens_ansi_colors` but it did not have any effect.

### Expected Behavior

```
printf "\e[1mfoo\e[0m"
```
should print bolder/brighter than
```
printf "foo"
```

### Logs

Debug Overlay
wezterm version: 20260812-070121-fe3006ae x86_64-unknown-linux-gnu
Window Environment: X11 i3
Lua Version: Lua 5.4
OpenGL: Mesa Intel(R) UHD Graphics 630 (CML GT2) 4.6 (Compatibility Profile) Mesa 26.1.5-arch1.1
Enter lua statements or expressions and hit Enter.
Press ESC or CTRL-D to exit

### Anything else?

In `terminator` this works out of the box.

bew:
What font are you using?
Have you tried with the default config?

please do a screenshot of what you see for `printf "foo\e[1mfoo\e[0mfoo\n"`

----

With default configs on MacOS I see:
<img width="356" height="57" alt="Image" src="https://github.com/user-attachments/assets/87f82baf-931a-47b7-918a-7e7e97239317" />
[...]

PR #8097, opened 2026-08-25
fonts: Improve default weights: normal→dim/bold is more visible

hphng:
we have a discussion why need to change at #8049 

As the title say, change the increasing bolder value from 200 to 400

### DIFF
--- a/config/src/font.rs
+++ b/config/src/font.rs
@@ -201,6 +201,12 @@ impl Default for FontWeight {
 }
 
 impl FontWeight {
+    /// Amount by which `lighter()` / `bolder()` shift the weight when
+    /// synthesizing dim / bold variants. Chosen large enough that bold
+    /// is clearly distinct from regular by default.
+    /// See issue #8049 for examples.
+    const WEIGHT_STEP: u16 = 300;
+
     pub const fn from_opentype_weight(w: u16) -> Self {
         Self(w)
     }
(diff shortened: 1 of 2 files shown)


bew:
As mentioned in https://github.com/wezterm/wezterm/issues/8049#issuecomment-5414601679 😅:
> mention #8049 issue for reference in a comment around that change 🙏

(to explain why 400 was chosen)

config/src/font.rs
bew: looking around the change, I see this function also uses 200, it would be interesting to see (try different fonts) to see if 400 subtraction would be better here as well 🤔

Would be worth asking on #8049 to try lower weights as well for the fonts the guy tested 🤔

config/src/font.rs
```diff
@@ -214,7 +214,8 @@ impl FontWeight {
     }
 
     pub fn bolder(self) -> Self {
-        Self::from_opentype_weight(self.to_opentype_weight() + 200)
+        // +200 was too subtle to distinguish from normal; see issue #8049
```
bew: Actually looking at the `FontWeight` type where this function is defined in, it represents a given weight, not necessarily 'Normal'.

To make it more generic we could rephrase this like:
```suggestion
        // +200 used initially was too subtle to distinguish any difference from e.g. normal & bold
        // See issue #8049 for examples.
```
(note: I've split the comment on 2 lines to fit ~100 chars line length)

config/src/font.rs
hphng: yeah I am about to talk about this because `lighter()` is opposite with `bolder()`

PR #8098, opened 2026-08-25
config: Replace winapi with windows-sys

xtqqczze:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -983,7 +983,7 @@ dependencies = [
  "wezterm-input-types",
  "wezterm-ssh",
  "wezterm-term",
- "winapi",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


Cargo.lock
```diff
@@ -983,7 +983,7 @@ dependencies = [
  "wezterm-input-types",
  "wezterm-ssh",
  "wezterm-term",
- "winapi",
+ "windows-sys 0.61.2",
```
bew, line 986: do you know why most other deps don't have a version mentioned like this in the cargo lock, but windows-sys always do?

xtqqczze: It's because different dependencies in your tree require incompatible versions of it.

```console
$ cargo tree --invert windows-sys
error: specification `windows-sys` is ambiguous
help: re-run this command with one of the following specifications
  windows-sys@0.45.0
  windows-sys@0.48.0
  windows-sys@0.52.0
  windows-sys@0.59.0
  windows-sys@0.60.2
  windows-sys@0.61.2
  ```
  
This is why it might be helpful to specify the version as a range.

PR #8092, opened 2026-08-25
feat: add support for Kitty Pointer Shapes Protocol

NSPC911:
Adds support for the [Mouse Pointer Shapes](https://sw.kovidgoyal.net/kitty/pointer-shapes/) protocol from Windows.

This is the first terminal to support the protocol on Windows; this highlights certain problems with Windows' cursor handling. Certain cursors don't work as mentioned by spec (specifically, cursors for alias, copy, zoom-in and zoom-out use the default cursor, while the grab and grabbing versions use the pointer version. crosshair and cell use the same icon).

I don't have a MacOS machine to test this PR on it would be nice if someone could help me with that.

resolves #7907 

Draft PR for now until I'm signaled that this PR will be merged (I want to squash the commits later)

PR #8100, opened 2026-08-26
window: Replace MouseCursor with cursor-icon from winit

NSPC911:
Refactored based on the suggestion in https://github.com/wezterm/wezterm/pull/8092#pullrequestreview-5021264959


window/src/os/macos/window.rs
```diff
@@ -1225,20 +1269,21 @@ impl WindowInner {
...
                 // that we don't know about.
-                let instance: id = match cursor {
-                    MouseCursor::Arrow => msg_send![ns_cursor_cls, arrowCursor],
-                    MouseCursor::Text => msg_send![ns_cursor_cls, IBeamCursor],
-                    MouseCursor::Hand => msg_send![ns_cursor_cls, pointingHandCursor],
-                    MouseCursor::SizeUpDown => msg_send![ns_cursor_cls, resizeUpDownCursor],
-                    MouseCursor::SizeLeftRight => msg_send![ns_cursor_cls, resizeLeftRightCursor],
+                let selector = mouse_cursor_selector(cursor);
+                let responds: BOOL = msg_send![ns_cursor_cls, respondsToSelector: selector];
+                let selector = if from_yes_no(responds) {
```
bew: let's convert early, so we don't have to deal with with MacOS stuff at all in normal code after getting the value.
And let's make that var name explicit!
```suggestion
                let cursor_is_supported = from_yes_no(msg_send![ns_cursor_cls, respondsToSelector: selector]);
                let selector = if cursor_is_supported {
```

window/src/os/x11/cursor.rs
```diff
@@ -367,15 +363,21 @@ impl CursorInfo {
...
+            CursorIcon::Pointer | CursorIcon::Grab | CursorIcon::Grabbing => HAND1,
+            CursorIcon::Text | CursorIcon::VerticalText => XTERM,
+            CursorIcon::NResize
+            | CursorIcon::SResize
+            | CursorIcon::NsResize
+            | CursorIcon::RowResize => SB_V_DOUBLE_ARROW,
+            CursorIcon::EResize
+            | CursorIcon::WResize
+            | CursorIcon::EwResize
+            | CursorIcon::ColResize => SB_H_DOUBLE_ARROW,
```
bew: WDYT about handling more cursors types, using more constants from https://docs.rs/xcb-util/0.3.0/src/xcb_util/cursor.rs.html ?

window/src/os/macos/window.rs
```diff
@@ -983,6 +983,50 @@ fn screen_point_to_cartesian(point: ScreenPoint) -> NSPoint {
     }
 }
 
+fn mouse_cursor_selector(cursor: CursorIcon) -> Sel {
```
bew: let's also rename as `mouse_icon_selector`

window/src/os/macos/window.rs
```diff
@@ -983,6 +983,50 @@ fn screen_point_to_cartesian(point: ScreenPoint) -> NSPoint {
...
+    match cursor {
```
bew, line 989: let's document where to find mouse selectors:
https://developer.apple.com/documentation/appkit/nscursor#overview (official icons)
https://github.com/mstg/OSX-Runtime-Headers/blob/9e1686ba1c48e0ca17f6baf7cfb209d4f7cfa4f2/AppKit/NSCursor.h (all)

window/src/os/macos/window.rs
```diff
@@ -983,6 +983,50 @@ fn screen_point_to_cartesian(point: ScreenPoint) -> NSPoint {
...
+        CursorIcon::Cell | CursorIcon::Crosshair => sel!(crosshairCursor),
+        CursorIcon::Copy => sel!(dragCopyCursor),
+        CursorIcon::EResize => sel!(resizeRightCursor),
+        CursorIcon::EwResize | CursorIcon::ColResize => sel!(resizeLeftRightCursor),
+        CursorIcon::Grab | CursorIcon::Move | CursorIcon::AllScroll => sel!(openHandCursor),
+        CursorIcon::Grabbing => sel!(closedHandCursor),
+        CursorIcon::Help | CursorIcon::ContextMenu => sel!(contextualMenuCursor),
+        CursorIcon::NResize => sel!(resizeUpCursor),
+        CursorIcon::NeResize | CursorIcon::NeswResize | CursorIcon::SwResize => {
+            sel!(_windowResizeNorthEastSouthWestCursor)
```
bew: where did you find this one? 🤔
I'd like to also link to where to find all non-documented selectors like this one

bew (commented):
Changes look good, did you test it on Linux & Windows? (I'll test MacOS)

NSPC911:
By working, if you meant whether the old and new behaviours are the same, yeah, they have the same behaviour, the text cursor appears when mouse reporting is disabled, and the normal cursor appears when mouse reporting is enabled

window/src/os/x11/cursor.rs
```diff
@@ -364,21 +412,9 @@ impl CursorInfo {
...
-            | CursorIcon::NsResize
-            | CursorIcon::RowResize => SB_V_DOUBLE_ARROW,
-            CursorIcon::EResize
-            | CursorIcon::WResize
-            | CursorIcon::EwResize
-            | CursorIcon::ColResize => SB_H_DOUBLE_ARROW,
-            _ => TOP_LEFT_ARROW,
-        };
+        // `/usr/include/X11/cursorfont.h`
+        // <https://docs.rs/xcb-util/0.3.0/src/xcb_util/cursor.rs.html>
```
bew: Move that comment above the list of constants, more appropriate there I think

bew (approved):
Tested on MacOS, works fine 👍

bew:
> here i think?

Yes 😄

bew:
Thank you!

PR #8099, opened 2026-08-26
termwiz: Replace winapi with windows-sys

xtqqczze:
Contributes to https://github.com/wezterm/wezterm/issues/7797.

Issue #8051, opened 2026-08-13
Soundness: Unsound Enum Transmutation on Deserialized CellAttributes Bitfields


Manishearth:
> [!NOTE]
> This finding was identified during an agentic unsafe Rust code review performed by Gemini AI, followed by human review and verification.

## The Issue

In `src/cell.rs`, `CellAttributes` derives Serde `Deserialize` (when the `use_serde` feature is enabled) and stores terminal styling flags inside a raw integer field `attributes: u32`: 

https://github.com/wezterm/wezterm/blob/fe3006aefcdc4c22924e7bce966b2c430dade4f1/wezterm-cell/src/lib.rs#L54-L57

Untrusted input (such as JSON or bincode payloads) can deserialize arbitrary integer bit patterns directly into `attributes` without validation. The `bitfield!` macro defines public getter methods (`intensity()`, `underline()`, `blink()`, `semantic_type()`, `vertical_align()`) that extract bit slices and execute `mem::transmute` into target `#[repr(u8)]` enums

https://github.com/wezterm/wezterm/blob/fe3006aefcdc4c22924e7bce966b2c430dade4f1/wezterm-cell/src/lib.rs#L126-L161

Several target `#[repr(u8)]` enums have fewer valid discriminants than allowed by their extracted bitmasks. For example, [`Intensity`](https://github.com/wezterm/wezterm/blob/fe3006aefcdc4c22924e7bce966b2c430dade4f1/wezterm-escape-parser/src/csi.rs#L43-L47) occupies 2 bits (mask `0b11`, values 0 through 3), but defines only 3 valid variants (`Normal = 0`, `Bold = 1`, `Half = 2`). When any getter method is invoked on deserialized data containing an out-of-range bit pattern (including automatically during `Debug::fmt`), `mem::transmute` constructs an invalid enum discriminant, which is UB.

https://github.com/wezterm/wezterm/blob/fe3006aefcdc4c22924e7bce966b2c430dade4f1/wezterm-cell/src/lib.rs#L201-L202

<details><summary>Minimal Reproduction (Miri)</summary>

```rust
use termwiz::cell::CellAttributes;

fn main() {
    // Obtain a default CellAttributes instance and serialize it to a serde_json::Value
    let mut val = serde_json::to_value(&CellAttributes::default()).unwrap();

    // Inject an invalid bit pattern into the internal `attributes` u32 bitfield.
[...]

PR #8107, opened 2026-08-28
wezterm-escape-parser: Replace winapi with windows-sys

xtqqczze:

### DIFF
--- a/wezterm-escape-parser/src/apc.rs
+++ b/wezterm-escape-parser/src/apc.rs
@@ -319,11 +319,11 @@ fn read_shared_memory_data(
 
 #[cfg(all(feature = "kitty-shm", windows))]
 mod win {
-    use winapi::um::handleapi::CloseHandle;
-    use winapi::um::memoryapi::{
-        FILE_MAP_ALL_ACCESS, MapViewOfFile, OpenFileMappingW, UnmapViewOfFile, VirtualQuery,
+    use windows_sys::Win32::Foundation::{CloseHandle, HANDLE};
+    use windows_sys::Win32::System::Memory::{
+        FILE_MAP_ALL_ACCESS, MEMORY_BASIC_INFORMATION, MEMORY_MAPPED_VIEW_ADDRESS, MapViewOfFile,
+        OpenFileMappingW, UnmapViewOfFile, VirtualQuery,
     };
-    use winapi::um::winnt::{HANDLE, MEMORY_BASIC_INFORMATION};
 
     struct HandleWrapper {
         handle: HANDLE,
@@ -331,7 +331,7 @@ mod win {
 
     struct SharedMemObject {
         _handle: HandleWrapper,
-        buf: *mut u8,
+        buf: MEMORY_MAPPED_VIEW_ADDRESS,
     }
 
     impl Drop for HandleWrapper {
@@ -345,7 +345,7 @@ mod win {
     impl Drop for SharedMemObject {
         fn drop(&mut self) {
             unsafe {
-                UnmapViewOfFile(self.buf as _);
+                UnmapViewOfFile(self.buf);
             }
         }
     }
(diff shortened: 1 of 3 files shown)


wezterm-escape-parser/src/apc.rs
```diff
@@ -423,8 +423,13 @@ mod win {
...
             size = size.min(val as usize);
         }
-        let buf_slice = unsafe { std::slice::from_raw_parts(shm.buf.add(offset), size) };
-        let data = buf_slice.to_vec();
+        let data = unsafe {
+            let src = shm.buf.Value.cast::<u8>().add(offset);
+            let mut data = Vec::with_capacity(size);
+            std::ptr::copy_nonoverlapping(src, data.as_mut_ptr(), size);
+            data.set_len(size);
+            data
```
bew: what is this change about? why is it needed?

xtqqczze: > ### 2. Aliasing and Immutability Violation on Windows Shared Memory Slices (`src/escape/apc.rs:428`) 🔴 🚨
> 
> -   **Priority**: 🔴 High
> - **Threat Vector**: 🚨 Untrusted Input
> -   **Bug Type**: Aliasing Violation
> 
> In `read_shared_memory_data`, Windows shared memory for Kitty Image Protocol IPC is opened via `OpenFileMappingW` and mapped into memory via `MapViewOfFile` with `FILE_MAP_ALL_ACCESS`. The code constructs a shared reference slice `&[u8]` over this memory via `std::slice::from_raw_parts` and copies it:
> 
> ```rust
> 
> let buf_slice = unsafe { std::slice::from_raw_parts(shm.buf.add(offset), size) };
> let data = buf_slice.to_vec();
> 
> ```
> 
> Because the shared memory mapping is accessible by external processes or concurrent threads with write permissions, external modification of the shared memory during the lifetime of the `&[u8]` slice violates Rust's aliasing rules (shared references `&T` require that the underlying memory is not mutated for their duration). Furthermore, if the mapped memory pages were allocated by an untrusted external process and left uninitialized, reading them through `&[u8]` causes immediate Undefined Behavior due to reading uninitialized memory.

https://github.com/wezterm/wezterm/issues/8051#issue-5145217015

xtqqczze: Since I touched this code I thought I'd fix it, unless you'd prefer a separate PR?

bew: yeah let's keep it separate please, the concerns are different here

winapi removal can be merged quickly and the security can get reviewed independently

bew:
Yeah sorry for the delay here, thanks for the changes!

wezterm-escape-parser/src/apc.rs
xtqqczze: Addressed in https://github.com/wezterm/wezterm/pull/8129.

xtqqczze:
This merge probably broke `main`:

```log
error[E0432]: unresolved import `winapi::um::synchapi`
   --> wezterm-client\src\discovery.rs:27:21
    |
 27 |     use winapi::um::synchapi::{CreateMutexW, ReleaseMutex, WaitForSingleObject};
    |                     ^^^^^^^^ could not find `synchapi` in `um`
    |
note: found an item that was configured out
   --> D:\a\wezterm\wezterm\vendor\winapi-0.3.9\src\um\mod.rs:235:38
    |
235 | #[cfg(feature = "synchapi")] pub mod synchapi;
    |       --------------------           ^^^^^^^^
    |       |
    |       the item is gated behind the `synchapi` feature
```

xtqqczze:
Most likely this conflicted with https://github.com/wezterm/wezterm/pull/8099. The lesson here is probably to only have one `winapi` PR in flight at a time.

bew:
> This merge probably broke main

Oops yes 😬

> The lesson here is probably to only have one `winapi` PR in flight at a time.

Or I should have required a `main` merge in PR with full CI checks before merge!
Multiple PRs in-flights shouldn't be an issue, it's just ready-to-merge checks that needs to be better here IMO

xtqqczze:
The best solution is to configure a [merge queue](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue).

mati865:
> Most likely this conflicted with #8099. The lesson here is probably to only have one `winapi` PR in flight at a time.

Not using winapi crate is your best bet in the current Rust ecosystem 😉.

xtqqczze:
@mati865 Contributions toward the remaining work in https://github.com/wezterm/wezterm/issues/7797 would be welcome.

Issue #2405, opened 2022-08-14
Ctrl + L completely removes the previous output


chenzhiwei:
### What Operating System(s) are you seeing this problem on?

Linux Wayland

### Which Wayland compositor or X11 Window manager(s) are you using?

Archlinux + Wayland + KDE + KWin

### WezTerm version

20220807-113146-c2fee766

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

No, and I'll explain why below

### Describe the bug

When I pressed ctrl+l on gnome-terminal, the buffer was shifted upwards outside my view, but I could still scroll back and find the stuff there...

In wez-term, when I press ctrl+l, what's currently in the screen gets cleared and lost forever... while the buffer outside the screen view remains there, unchanged.

### Configuration

I installed with flatpak, nothing configured.

### Expected Behavior

After run `Ctrl + L`, the previous output should be found by scrolling back.

wez:
This isn't a bug; gnome-terminal behaves differently from other terminals in how it handles `CTRL-L`. wezterm behaves the same way as xterm, as do most other terminals.

bew:
(probably too much info in one go if you've never heard of these things, but at least now it's written somewhere)

---

To give a bit more info and propose a basic solution:
Usually with `Ctrl-L` the shell (bash, zsh, ...) (or the program `clear`) tell the terminal to clear the terminal in the following way (using [CSI ANSI sequences](https://en.wikipedia.org/wiki/ANSI_escape_code#CSI_(Control_Sequence_Introducer)_sequences)): (for example)

1. `CSI H` Put cursor at the top left corner of the terminal
2. `CSI 2 J` Clear what's currently displayed on the terminal

With this scenario, existing things displayed on the terminal are simply cleared, only lines that were not visible anymore are still available in the scrollback buffer (when you scroll up).

A solution that should work on any terminal, is to change what `Ctrl-L` do to first save existing  content in the scrollback buffer:
1. `CSI n S` Scroll the visible content of the terminal by `n` lines (needs to be computed), to move all useful terminal output to the scrollback buffer
2. then do the same operations as before: put cursor to top left corner, then clear the terminal

To know by how much the terminal needs to be scrolled by (find `n`), it's possible to ask the terminal where the cursor is currently (using `CSI 6n`) and read the reply, then scroll the terminal by the cursor row.

---
If you're using zsh as your shell, you could get inspired by what an implementation of the ideal behavior I did in my config:
[...]

PR #8116, opened 2026-08-30
term: add erase_display_scrolls_into_scrollback option

### LINKED ISSUE #8105: clear -x truncates the last output — scrollback content is incomplete
## Describe the bug

After running `clear -x`, scrolling back up does not show the complete previous output. The last command's output is truncated — only a partial portion of what was visible on screen before `clear -x` is preserved in the scrollback buffer.

## To Reproduce

1. Run a command that produces more than a screenful of output (e.g. a table with many rows, or `ls -la` in a large directory).
2. Run `clear -x` (which sends `\e[H\e[2J` — clears screen without `\e[3J` scrollback erase).
3. Scroll up with the mouse to look for the previous output.
4. Observe that the output is incomplete — only a portion of what was on screen is in the scrollback.

## Expected Behavior

`clear -x` only clears the visible screen (`\e[H\e[2J`) without erasing scrollback (`\e[3J`). The content that was visible on screen before `clear -x` should be fully preserved in the scrollback buffer, so scrolling up reveals all of it.

Other terminals (Alacritty, gnome-terminal, macOS Terminal.app) preserve the full visible content in scrollback after `clear -x`.

## Environment

- **WezTerm version**: `20260826-113711-78cd82db`
- **OS**: Linux x86_64 (kernel 6.6.0)
- **Display**: X11
- **Shell**: zsh 5.9
- **TERM**: xterm-256color

## Additional context

`clear -x` sends `\e[H\e[2J` (cursor home + erase entire display). WezTerm implements `\e[2J` per the ANSI spec: it erases the screen content without first moving the visible lines into the scrollback. This means any lines that were on screen but not yet scrolled off naturally are lost — only lines that already scrolled into scrollback before `clear -x` was run are preserved.

In contrast, Alacritty and gnome-terminal scroll the visible lines into scrollback before erasing, so the full screen content is recoverable by scrolling up.

I understand this may be the intended spec-conformant behavior (see #2405 and #6620), but the truncation of the last output makes `clear -x` much less useful. Would it be possible to add a configuration option to enable scrollback-preserving behavior for `\e[2J`, or to have `clear -x`-style clears preserve visible content in scrollback?

dylanpulver:
Adds an opt-in `erase_display_scrolls_into_scrollback` (default `false`). With it
set, `CSI 2 J` on the primary screen scrolls the screen into the scrollback and
then erases it, instead of erasing in place. Nothing changes for anyone who
doesn't set it, so the conformance position from #2405 still holds by default.

The scroll only covers rows up to the last row that holds something, and that is
the part worth reviewing. Scrolling all `physical_rows` unconditionally is the
obvious implementation, but then every `Ctrl-L` at an empty prompt pushes a
screenful of blanks in and evicts real history, so the option would eat the
scrollback it exists to protect. `erase_display_scroll_does_not_pad_scrollback`
and `..._of_empty_screen_is_a_no_op` both fail against that version.

The scroll happens before the existing erase rather than instead of it, so
`CSI 2 J` still paints the whole screen with the current SGR background. Doing it
as an early return leaves rows that were already blank carrying their old
attributes; `erase_display_scroll_still_erases_to_current_background` covers that.

Seven tests in `term/src/test/csi.rs`. Reverting the source fails four of them;
the other three pin behaviour this must not change (`CSI 0 J`, an already-empty
screen, and the default-off path). `cargo test --all` and `cargo fmt --all` are
clean locally.

Known limits:

- The alternate screen is skipped explicitly. It has no scrollback, so the
  outcome is the same either way and that guard is defensive rather than covered
  by a test.
- "Holds something" means text, so a row of spaces carrying only a background
  colour counts as empty and gets trimmed.
- Plain `clear` also sends `CSI 3 J`, which erases the scrollback, so this is
  visible with `clear -x` and the shell's `Ctrl-L` rather than with `clear`.

Happy to rename the option or move the trimming rule if you'd rather it matched
something else.

AI disclosure, since you ask this on PRs: the implementation, the tests, the docs
page and this description were drafted with Claude Opus 5 via Claude Code.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -77,6 +77,11 @@ As features stabilize some brief notes about them will accumulate here.
   easier to spot the remaining candidates. Thanks to @mr-felixoid and @bew! #7752
 
 #### New
+* [erase_display_scrolls_into_scrollback](config/lua/config/erase_display_scrolls_into_scrollback.md)
+  option to make `CSI 2 J` (as sent by `clear -x` and the shell's `Ctrl-L`)
+  scroll the screen into the scrollback instead of erasing it in place. Off by
+  default, which leaves the spec behavior unchanged. Thanks to @dylanpulver!
+  #8105
 * [command_palette_line_height](config/lua/config/command_palette_line_height.md)
   option to scale the vertical spacing of rows in the command palette,
   independently of [line_height](config/lua/config/line_height.md) which is for terminal cells only.
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -272,6 +272,14 @@ pub struct Config {
     #[dynamic(default)]
     pub enable_checksum_rectangular_area: bool,
 
+    /// Whether `CSI 2 J` (erase the whole display, as sent by eg: `clear`)
+    /// should scroll the screen contents into the scrollback instead of
+    /// erasing them in place.
+    /// Disabled by default because erasing in place is what the spec
+    /// describes and what xterm does.
+    #[dynamic(default)]
+    pub erase_display_scrolls_into_scrollback: bool,
+
     /// Specifies the width of a new window, expressed in character cells
     #[dynamic(default = "default_initial_cols", validate = "validate_row_or_col")]
     pub initial_cols: u16,
--- a/config/src/terminal.rs
+++ b/config/src/terminal.rs
@@ -86,6 +86,10 @@ impl wezterm_term::TerminalConfiguration for TermConfig {
         self.configuration().enable_checksum_rectangular_area
     }
 
+    fn erase_display_scrolls_into_scrollback(&self) -> bool {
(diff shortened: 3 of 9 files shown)


docs/changelog.md
```diff
@@ -77,6 +77,11 @@ As features stabilize some brief notes about them will accumulate here.
   easier to spot the remaining candidates. Thanks to @mr-felixoid and @bew! #7752
 
 #### New
+* [erase_display_scrolls_into_scrollback](config/lua/config/erase_display_scrolls_into_scrollback.md)
+  option to make `CSI 2 J` (as sent by `clear -x` and the shell's `Ctrl-L`)
+  scroll the screen into the scrollback instead of erasing it in place. Off by
+  default, which leaves the spec behavior unchanged. Thanks to @dylanpulver!
+  #8105
```
bew: Link to PR, not to issue

Issue #8017, opened 2026-08-05
[DRAFT] macOS: migrate from legacy `objc`/`cocoa`/`core-*` crates to `objc2*`


bew:
> [!WARNING]
> **AI-disclosure**:
> This plan & issue was **drafted** on Perplexity using `GPT-5.6 Terra Thinking`

---

The native macOS window implementation currently depends on the legacy `objc` crate and `cocoa`.

`objc` has not received releases for 7 years at time of writing, and contains obsolete `cfg(cargo_clippy)` conditionals ([removed in 2024](https://blog.rust-lang.org/2024/02/28/Clippy-deprecating-feature-cargo-clippy/)).
Current Rust/Cargo check-cfg validation reports these as unexpected configuration values, producing **184 warnings** like this during project build.

<img width="1518" height="226" alt="Image" src="https://github.com/user-attachments/assets/b2f9f119-028a-4153-86cc-68c9230ad7fe" />

The macOS backend in `window/src/os/macos/` is a substantial native implementation. It manages the AppKit application/window/view lifecycle, custom Objective-C view and delegate classes, input and IME handling, window state, display integration, and rendering-context lifecycle.

👉 The goal should therefore be an incremental bindings migration, not a rewrite of the backend.

## Timing

This should be planned for **after the next WezTerm release**.
The work touches central macOS windowing and input paths, so it is better started at the beginning of a development cycle, with enough time for nightly testing and regression fixes before the following release.

Until then, I think we can live with the warnings, or suppress them temporarily 🤔

## Proposed direction

Migrate to the maintained `objc2` ecosystem (full docs at: https://docs.rs/objc2):

- `objc2` for Objective-C runtime interaction, selectors, message sends, ownership, and class definitions.
- `objc2-app-kit` for AppKit APIs such as `NSApplication`, `NSWindow`, `NSView`, menus, events, and screens.
- `objc2-foundation` for Foundation types and APIs.
- `objc2-core-foundation`, `objc2-core-graphics`, `objc2-core-text`, and `objc2-quartz-core` for existing framework usage as applicable.
- `block2` and `dispatch2` where Objective-C blocks or GCD integration are needed.
- `objc2-metal` only if required by direct Metal-specific code.

## Suggested _incremental_ plan

> [!WARNING]
> I didn't review this carefully, do not take it as the actual plan.
> I think however that it's a good draft for the overall direction.

[...]

PR #8117, opened 2026-08-31
wezterm-font: migrate core_text.rs off objc/cocoa to objc2

jason-mac:
## Summary

Related to #8017. Migrates `wezterm-font/src/locator/core_text.rs` off the legacy `objc`/`cocoa` crates onto `objc2`.

This is a small effort for migration towards `objc2`. One function, (`build_fallback_list_impl`) needed `cocoa`/`objc`, to read the user's `AppleLanguages` system preference used for building the font-fallback cascade.

## Changes

- Replaced the `msg_send![class!(NSUserDefaults), standardUserDefaults]` / `msg_send![user_defaults, stringArrayForKey:apple_lang]` calls with typed `objc2_foundation::NSUserDefaults::standardUserDefaults()` / `.stringArrayForKey(&key)`.
- The rest of the file (CoreText/Core Foundation font matching) is unchanged and still uses `core-foundation`'s `CFArray<CFString>` type, so the resulting `NSArray<NSString>` is bridged into a `CFArray<CFString>` using `CFArray::wrap_under_get_rule`.
- Removed the now-unused `cocoa`/`objc` dependencies from `wezterm-font/Cargo.toml`.
- No behavior changes, only bindings changes.

## AI Usage
AI was used for documentation lookup into `objc2`, `objc` and `cocoa` for understanding and verification of API usage between them. 
AI was also used for implementation sanity checks and as a tool for understanding relevant functions' control flow.

## Testing

- [x] `cargo check -p wezterm-font`
- [x] `cargo test -p wezterm-font`
- [x] `cargo fmt --all -- --check`


wezterm-font/src/locator/core_text.rs
```diff
@@ -3,15 +3,15 @@
...
 use crate::locator::{FontDataSource, FontLocator, FontOrigin};
 use crate::parser::ParsedFont;
-use cocoa::base::id;
 use config::{FontAttributes, FontStretch, FontStyle, FontWeight};
 use core_foundation::array::CFArray;
 use core_foundation::base::{CFRange, TCFType};
 use core_foundation::dictionary::CFDictionary;
 use core_foundation::string::{CFString, CFStringRef};
 use core_text::font::*;
 use core_text::font_descriptor::*;
```
bew: you should also change `core_*` crates to their `objc2-core-*` counterpart, so we don't have to fiddle with pointers manually anymore (hopefully)

jason-mac: Thanks for review, I took time to make some adjustments.

bew:
Hello! Thanks for starting this work!

What's your experience with MacOS development ? (Without LLMs)

For transparency, what / how much manual (non-LLM) review/iteration/testing did you do? 🤔
How did you test these changes? (cargo test is not enough for testing OS integrations like this)

I'll have to get familiar with these new interface and functions to get a feel for whether this is the right migration path, but it's great to see the code being less obscure!

Could you tell me why we still have a few casts (I saw a few `cast_unchecked`s & a `downcast`), is there a reasonable way to avoid them or what makes them necessary ?

Also please update the title and PR description to be up-to-date and less specific to the code.
(I believe this was LLM generated? I'd prefer simpler but targeted description, mentioning your choices rather than describing what I can already see in the set of changes)

Thank you for your help !

jason-mac:
Hello, sorry for the late reply. Thanks for taking a look! I dont want to leave this PR stale as I would like to make this contribution at some point. I wanted to take on this issue since, at first glance, it seemed like simple binding changes. 

Claude was used in the assistance of this conversion. This is my first pass at MacOS development (if you can even call it that) and if relevant this is also my thirdish pass with Rust (coming from mostly C++ background).

Implementation literally involved going line by line and trying a direct conversion from the old crates -> `objc2` while trying to maintain logical equivalence as much as possible. I thought this strategy would minimize any possible regressions and its impact if there were going to be any.

I think I am going to take a little step back and try to understand the interfaces (some actual intersting history in NS and CF lol) a bit more to ensure that the conversion makes sense and that the API swap is equivalent as much as it can be. I will go through another pass sometime this week or weekend and make another update after doing a bit more reading.

Feel free to make any changes you see fit or make further suggestions. As a long time wezterm user, I thank you for being an active maintainer!

Issue #8126, opened 2026-09-04
Add a `workspace-changed` event


jakealti:
**Is your feature request related to a problem? Please describe.**

There is no way for a config to learn that the active workspace changed.

**Describe the solution you'd like**

A window event, fired on every change of the active workspace:

```lua
wezterm.on('workspace-changed', function(window, pane, workspace, prior)
  -- prior is the workspace that was active before, or nil
end)
```

The previous name is already computed and discarded today, in
`Mux::set_active_workspace_for_client`:

```rust
info.active_workspace.replace(workspace.to_string());  // returns the old name
```

**Describe alternatives you've considered**

- Polling `update-status`, which is what configs do today. Unreliable when wezterm switches for you, and it means every module that cares keeps its own copy of "which workspace was I on".
- #4938 asks for a `SwitchToLastWorkspace` action, which would cover one use of this. The reporter there suggests the same thing: "Another option might be an event when the workspace is switched." An event also covers falling back to a sensible workspace when one is destroyed, and repainting a status bar on change rather than once a second.

**Additional context**

None.

bew:
> The reporter there suggests the same thing

It's also you the reporter there 🤣 I guess you used AI here and didn't tell it who you are..

I agree the event is a good idea, I'll checkout your PR 😉

jakealti:
Yes, I used AI to write the text. I read it myself so it wouldn't be obvious, but that phrase seems to have slipped through :)

The PRs are not ready yet, so I opened them as drafts. I will review them and make sure everything works as expected by testing the changes manually.

cenkalti:
I've implemented this and would like to check the shape before taking it further.

I proposed:

```lua
wezterm.on('workspace-changed', function(window, pane, workspace, prior)
```

Having built it, I don't think that shape works.

**It fires once per gui window, for a change that happened once.** The active workspace is a mux-wide value, so every copy carries the same `workspace` and the same `prior`. With three windows open, one switch calls the handler three times with identical arguments, and anything doing whole-switch work has to hand-roll a dedupe guard. I wrote the same guard twice in my own config and then documented it on the event's own page, which felt like a sign the shape was wrong.

[...]

PR #8128, opened 2026-09-04
gui: add workspace-changed event

jakealti:
Adds a `workspace-changed` event, so a config can react when the active workspace changes. Motivation and detail in #8126.

```lua
wezterm.on('workspace-changed', function(window, pane, workspace, prior)
  -- prior is the workspace that was active before, or nil
end)
```

`prior` is already computed and thrown away today in `Mux::set_active_workspace_for_client`, and is unrecoverable once the switch has happened.

Additive: a new event, nothing renamed or removed, no config or protocol changes. The one exception worth a look is that the mux no longer notifies subscribers when the active workspace is set to the value it already has, without which the event would fire for switches that did not happen.

Docs page and changelog entry included. `cargo test --all` passes. Verified by hand that the event fires on a deliberate switch, fires when the last window of the active workspace is closed, and does not fire on a no-op set.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -77,6 +77,9 @@ As features stabilize some brief notes about them will accumulate here.
   easier to spot the remaining candidates. Thanks to @mr-felixoid and @bew! #7752
 
 #### New
+* [workspace-changed](config/lua/window-events/workspace-changed.md) event, emitted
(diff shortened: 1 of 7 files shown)


jakealti:
Closing in favour of https://github.com/wezterm/wezterm/pull/8137

PR #8130, opened 2026-09-04
style: Add newline after inner attribute for readability

xtqqczze:
Add a newline after inner attributes to make it clear that the attribute applies to the enclosing item rather than the following item.

### DIFF
--- a/termwiz/src/widgets/mod.rs
+++ b/termwiz/src/widgets/mod.rs
@@ -1,6 +1,3 @@
-// Ideally this would be scoped to WidgetId, but I can't seem to find the
-// right place for it to take effect
-#![allow(clippy::new_without_default)]
 use crate::color::ColorAttribute;
 use crate::input::InputEvent;
 use crate::surface::{Change, CursorShape, CursorVisibility, Position, SequenceNo, Surface};
--- a/wezterm-cell/src/lib.rs
+++ b/wezterm-cell/src/lib.rs
@@ -1,5 +1,7 @@
-#![cfg_attr(not(feature = "std"), no_std)]
 //! Model a cell in the terminal display
+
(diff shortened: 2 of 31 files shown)


termwiz/src/widgets/mod.rs
```diff
@@ -1,6 +1,7 @@
 // Ideally this would be scoped to WidgetId, but I can't seem to find the
 // right place for it to take effect
 #![allow(clippy::new_without_default)]
```
bew: could we resolve this comment now with newer Rust version?
(compared to when that comment was added)

wezterm-cell/src/lib.rs
```diff
@@ -1,4 +1,5 @@
 #![cfg_attr(not(feature = "std"), no_std)]
+
 //! Model a cell in the terminal display
```
bew: could you also move this module-wide doc comment at the very top?

wezterm-escape-parser/src/lib.rs
bew: similarly here

bew (commented):
Good idea, thank you for this!

Do you know if there's a way to enforce this at rustfmt level? or could we add a CI step in the formatting job for this? 🤔

xtqqczze:
I'll push changes addressing review comments once https://github.com/wezterm/wezterm/pull/8132 is merged.

xtqqczze:
> Do you know if there's a way to enforce this at rustfmt level?

No there isn't, the closest thing is the [`clippy::empty_line_after_outer_attr`](https://rust-lang.github.io/rust-clippy/master/#empty_line_after_outer_attr) lint. This does suggest a new `no_empty_line_after_inner_attr` lint (https://github.com/rust-lang/rust-clippy/issues/17687).

bew:
Thanks!

PR #8137, opened 2026-09-05
gui: add workspace-changed event

cenkalti:
Adds a `workspace-changed` event, so a config can react when the active workspace changes. Motivation in #8126.

```lua
wezterm.on('workspace-changed', function(workspace, prior)
  -- prior is the workspace that was active before
end)
```

It is a gui event emitted once per switch rather than a window event.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -77,6 +77,8 @@ As features stabilize some brief notes about them will accumulate here.
   easier to spot the remaining candidates. Thanks to @mr-felixoid and @bew! #7752
 
 #### New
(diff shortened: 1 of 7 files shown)


bew:
Hey, I saw your PR, I'm working on getting more understanding/intuition of the role, responsibilities & quirks of the big `wezterm-gui` crate and I started documenting its content incrementally!
Also I never used workspaces so I have to get up to speed ^^'

I'm trying to get a better overview first instead of 'blindly' review your PR, so please wait a bit more I'm getting there 🙏

PR #8141, opened 2026-09-06
Fix Linux compose locale initialization

### LINKED ISSUE #8119: Linux GUI never starts: locale query returns "C" (setlocale never initialized) makes compose-table acquisition fatal
### What Operating System(s) are you seeing the problem on?

Linux

### Which Wayland compositor / X11 WM (if applicable)

COSMIC (Pop!_OS 24.04) — reproduces on both Wayland and X11 paths

### WezTerm version

wezterm 20260829-194257-08e5e0af (built from source, HEAD 08e5e0a)

### Did you try this in a new, clean WezTerm config file?

- [x] Yes, I tried a new, clean config file

### Describe the bug

On a fresh session the GUI never starts — both display paths die during keyboard init:

**Wayland:**
```
xkbcommon: ERROR: couldn't find a Compose file for locale "C" (mapped to "C")
ERROR window::os::wayland::keyboard > Error processing keymap change: Failed to acquire compose table from locale
ERROR env_bootstrap > panic at window/src/os/wayland/keyboard.rs:113:38 - !?
thread 'main' panicked ... no keymap
```

**X11:**
```
xkbcommon: ERROR: couldn't find a Compose file for locale "C" (mapped to "C")
ERROR wezterm_gui > Failed to acquire compose table from locale; terminating
```

### Root cause

`query_lc_ctype()` in `window/src/os/x11/keyboard.rs` calls `libc::setlocale(libc::LC_CTYPE, NULL)` — a **query-only** call. Rust programs never call `setlocale(..., "")` at startup, so the C runtime locale is still the default `"C"`, and the query always returns `"C"` regardless of `$LANG`/`$LC_CTYPE`. `Table::new_from_locale` then fails for `"C"`, which is fatal (wayland: keymap mapper `None` → `expect("no keymap")` panic; X11: `terminating`).

The macOS-only `env-bootstrap` locale probe exists, but the Linux path has no equivalent initialization.

Two-layer reproduction note (Pop!_OS 24.04): even with the locale query fixed, `xkbcommon`'s `compose.dir` lookup for `en_US.UTF-8` fails on this system (the Compose file exists at `/usr/share/X11/locale/en_US.UTF-8/Compose`), so a fallback to a direct file load is also needed.

### Fix (both applied locally, builds and runs green on Wayland/WebGpu)

**1. Initialize the locale from the environment before querying** (`window/src/os/x11/keyboard.rs`):

```rust
fn query_lc_ctype() -> anyhow::Result<&'static OsStr> {
    // Rust never calls setlocale; a NULL query always returned "C".
    unsafe { libc::setlocale(libc::LC_CTYPE, b"\0".as_ptr().cast()) };
    let ptr = unsafe { libc::setlocale(libc::LC_CTYPE, std::ptr::null()) };
    ensure!(!ptr.is_null(), "failed to query locale");
    let cstr = unsafe { CStr::from_ptr(ptr) };
    Ok(OsStr::from_bytes(cstr.to_bytes()))
}
```

**2. Compose-table fallback chain** at the three `new_from_locale` sites in the same file: on failure, retry via `Table::new_from_buffer` with the direct `/usr/share/X11/locale/en_US.UTF-8/Compose` bytes, then `/usr/share/X11/locale/C/Compose` (empty-but-valid on some distros) before giving up:

```rust
let table = xkb::compose::Table::new_from_locale(&context, locale, xkb::compose::COMPILE_NO_FLAGS)
    .or_else(|_| {
        let bytes = std::fs::read("/usr/share/X11/locale/en_US.UTF-8/Compose")
            .map_err(|e| anyhow!("compose file read failed: {e}"))?;
        xkb::compose::Table::new_from_buffer(&context, bytes, "en_US.UTF-8",
            xkb::compose::FORMAT_TEXT_V1, xkb::compose::COMPILE_NO_FLAGS)
            .map_err(|_| anyhow!("direct Compose file parse failed"))
    })
    .or_else(|_| {
        let bytes = std::fs::read("/usr/share/X11/locale/C/Compose")
            .map_err(|e| anyhow!("C compose read failed: {e}"))?;
        xkb::compose::Table::new_from_buffer(&context, bytes, "C",
            xkb::compose::FORMAT_TEXT_V1, xkb::compose::COMPILE_NO_FLAGS)
            .map_err(|_| anyhow!("C Compose parse failed"))
    })
    .map_err(|_| anyhow!("Failed to acquire compose table from locale"))?;
```

After both patches the GUI starts cleanly under Wayland (WebGpu) and X11 on this box.

Related: #6888 (same terminal state on Fedora; the `dead_hamza` Compose errors there are the second layer of this).

be-student:
Initialize `LC_CTYPE` from the environment before creating compose tables. All three keyboard constructors use a shared loader that delegates Compose-file discovery to xkbcommon. If loading the selected locale fails, log the fallback at INFO and retry with the `C` locale. The helper documentation explains locale initialization and fallback behavior.

Validation at `c44c96af13141701980f6c648123f7d840028ac0`:
- `cargo fmt --check`, `cargo check -p wezterm-gui`, and `git diff --check` pass on macOS. The macOS build does not compile the Linux X11 keyboard module.
- A standalone harness using the exact `query_lc_ctype` helper passes environment-driven UTF-8 initialization, `LC_ALL` precedence, and invalid-locale error checks on macOS. Replacing initialization with the original query-only call fails the UTF-8 expectation.
- No live Linux GUI/keyboard test is claimed. GitHub currently reports no hosted checks for this branch.

AI disclosure: OpenAI Codex assisted with analysis, implementation, automated validation, and PR wording. This update and its audit were also AI-assisted; no separate human review or manual GUI testing is claimed.

Closes #8119

### DIFF
--- a/window/src/os/x11/keyboard.rs
+++ b/window/src/os/x11/keyboard.rs
@@ -569,11 +569,7 @@ impl Keyboard {
             .ok_or_else(|| anyhow!("Failed to load system default keymap"))?;
 
         let state = xkb::State::new(&keymap);
-        let locale = query_lc_ctype()?;
-
-        let table =
-            xkb::compose::Table::new_from_locale(&context, locale, xkb::compose::COMPILE_NO_FLAGS)
-                .map_err(|_| anyhow!("Failed to acquire compose table from locale"))?;
+        let table = new_compose_table(&context)?;
         let compose_state = xkb::compose::State::new(&table, xkb::compose::STATE_NO_FLAGS);
 
         let phys_code_map = build_physkeycode_map(&keymap);
@@ -607,11 +603,7 @@ impl Keyboard {
         .ok_or_else(|| anyhow!("Failed to parse keymap state from file"))?;
 
         let state = xkb::State::new(&keymap);
-        let locale = query_lc_ctype()?;
-
-        let table =
-            xkb::compose::Table::new_from_locale(&context, locale, xkb::compose::COMPILE_NO_FLAGS)
-                .map_err(|_| anyhow!("Failed to acquire compose table from locale"))?;
+        let table = new_compose_table(&context)?;
         let compose_state = xkb::compose::State::new(&table, xkb::compose::STATE_NO_FLAGS);
 
         let phys_code_map = build_physkeycode_map(&keymap);
@@ -652,11 +644,7 @@ impl Keyboard {
 
         let state = xkb::x11::state_new_from_device(&keymap, connection, device_id);
 
-        let locale = query_lc_ctype()?;
-
-        let table =
-            xkb::compose::Table::new_from_locale(&context, locale, xkb::compose::COMPILE_NO_FLAGS)
-                .map_err(|_| anyhow!("Failed to acquire compose table from locale"))?;
+        let table = new_compose_table(&context)?;
         let compose_state = xkb::compose::State::new(&table, xkb::compose::STATE_NO_FLAGS);
 
         {
@@ -820,9 +808,45 @@ impl Keyboard {
     }
 }
 
+fn new_compose_table(context: &xkb::Context) -> anyhow::Result<xkb::compose::Table> {
+    let locale = query_lc_ctype()?;
+    xkb::compose::Table::new_from_locale(context, locale, xkb::compose::COMPILE_NO_FLAGS)
+        .or_else(|_| new_compose_table_from_file(context, locale))
+        .map_err(|_| anyhow!("Failed to acquire compose table for locale {locale:?}"))
+}
+
+fn new_compose_table_from_file(
+    context: &xkb::Context,
+    locale: &OsStr,
+) -> Result<xkb::compose::Table, ()> {
+    let locale_dir = std::path::Path::new("/usr/share/X11/locale");
+    for (candidate, candidate_locale) in [
+        (locale_dir.join(locale).join("Compose"), locale),
+        (locale_dir.join("C/Compose"), OsStr::new("C")),
+    ] {
+        let Ok(bytes) = std::fs::read(candidate) else {
+            continue;
+        };
+        if let Ok(table) = xkb::compose::Table::new_from_buffer(
+            context,
+            bytes,
+            candidate_locale,
+            xkb::compose::FORMAT_TEXT_V1,
+            xkb::compose::COMPILE_NO_FLAGS,
+        ) {
+            return Ok(table);
+        }
+    }
+    Err(())
+}
+
 fn query_lc_ctype() -> anyhow::Result<&'static OsStr> {
-    let ptr = unsafe { libc::setlocale(libc::LC_CTYPE, std::ptr::null()) };
-    ensure!(!ptr.is_null(), "failed to query locale");
+    // Rust does not initialize the C runtime locale from the environment.
+    let ptr = unsafe { libc::setlocale(libc::LC_CTYPE, b"\0".as_ptr().cast()) };
+    ensure!(
+        !ptr.is_null(),
+        "failed to initialize locale from the environment"
+    );
     let cstr = unsafe { CStr::from_ptr(ptr) };
     Ok(OsStr::from_bytes(cstr.to_bytes()))
 }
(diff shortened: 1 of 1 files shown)


window/src/os/x11/keyboard.rs
```diff
@@ -820,9 +808,45 @@ impl Keyboard {
...
 }
 
+fn new_compose_table(context: &xkb::Context) -> anyhow::Result<xkb::compose::Table> {
+    let locale = query_lc_ctype()?;
+    xkb::compose::Table::new_from_locale(context, locale, xkb::compose::COMPILE_NO_FLAGS)
+        .or_else(|_| new_compose_table_from_file(context, locale))
+        .map_err(|_| anyhow!("Failed to acquire compose table for locale {locale:?}"))
+}
+
+fn new_compose_table_from_file(
```
bew, line 821: could you add a doc comment for the main functions (new/existing) you're touching with this PR?
Focus on the function's role and specific behaviors if any

window/src/os/x11/keyboard.rs
```diff
@@ -820,9 +808,45 @@ impl Keyboard {
...
+            return Ok(table);
+        }
+    }
+    Err(())
+}
+
 fn query_lc_ctype() -> anyhow::Result<&'static OsStr> {
-    let ptr = unsafe { libc::setlocale(libc::LC_CTYPE, std::ptr::null()) };
-    ensure!(!ptr.is_null(), "failed to query locale");
+    // Rust does not initialize the C runtime locale from the environment.
```
bew: let's mention what we're doing exactly
```suggestion
    // Rust does not initialize the C runtime locale from the environment.
    // so we pass "" below to force initialize the locale from the environment.
```

bew (commented):
Hello! Welcome to Wezterm ;)

be-student:
Thanks for asking. This contribution used OpenAI Codex for the issue analysis, implementation, testing, and PR wording. There was no separate manual implementation pass before the PR was opened. In this review iteration, I checked the relevant `setlocale` and xkb compose-table paths, applied the requested doc comments and exact initialization explanation, ran `cargo fmt --check`, and ran `cargo check -p wezterm-gui` successfully.

window/src/os/x11/keyboard.rs
```diff
@@ -820,9 +808,50 @@ impl Keyboard {
...
+}
+
+/// Load a locale-specific Compose file, then fall back to the portable C locale.
+fn new_compose_table_from_file(
+    context: &xkb::Context,
+    locale: &OsStr,
+) -> Result<xkb::compose::Table, ()> {
+    let locale_dir = std::path::Path::new("/usr/share/X11/locale");
+    for (candidate, candidate_locale) in [
+        (locale_dir.join(locale).join("Compose"), locale),
```
bew, line 827: Hmm this is already what `xkb::compose::Table::new_from_locale` does, why reimplement it ?

Also for C we could simply call that function again, passing `"C"` as the locale, I don't think this function is useful as is 🤔

WDYT?

window/src/os/x11/keyboard.rs
```diff
@@ -820,9 +808,30 @@ impl Keyboard {
     }
 }
 
+/// Build the compose table for the process locale, retrying with the C locale
+/// when xkbcommon cannot resolve it. xkbcommon owns Compose-file discovery.
```
bew:
```suggestion
/// if xkbcommon cannot resolve it.
```

window/src/os/x11/keyboard.rs
```diff
@@ -820,9 +808,30 @@ impl Keyboard {
...
+fn new_compose_table(context: &xkb::Context) -> anyhow::Result<xkb::compose::Table> {
+    let locale = query_lc_ctype()?;
+    xkb::compose::Table::new_from_locale(context, locale, xkb::compose::COMPILE_NO_FLAGS)
+        .or_else(|_| {
+            xkb::compose::Table::new_from_locale(
+                context,
+                OsStr::new("C"),
+                xkb::compose::COMPILE_NO_FLAGS,
+            )
```
bew, line 822: TODO: add info log when that happens!

PR #8143, opened 2026-09-07
mux: enforce search result limit within logical lines

cuishuang:
`LocalPane::search(limit)` could return more results than requested when a single logical line contained multiple matches.

The limit was checked only before processing each logical line, so one line could add several matches at once. The existing `==` check also failed to stop the search after the result count had exceeded the limit.


mux/src/localpane.rs
```diff
@@ -679,168 +686,179 @@ impl Pane for LocalPane {
...
-            };
-            let stable_idx = sr.start;
+        Ok(results)
+    }
+}
 
-            if haystack.is_empty() {
-                return true;
-            }
+fn search_logical_line(
```
bew, line 718: Could you add doc comments on new / existing functions/structs that you are touching in this PR ?
Focus on the role/responsibility of each, and mentioning its quirks if relevant.
_Thank you!_

bew (commented):
Hello! Thanks for the PR, welcome to Wezterm!

For transparency, how much AI was used for this contribution?
.. What for? (research/implementation/testing/PRwriting)
.. And what / how much manual review/iteration/testing did you do? 🤔

mux/src/localpane.rs
cuishuang: I use CodeX helping identify the result-limit bug, refactor the logical-line search helper, add regression tests, and add  documentation comments and the mux test suite.


And I manually reviewed the implementation and the final diff, including the responsibilities and quirks described in the new doc comments. I also iterated on the tests after catching an incorrect assertion in the cross-line test.

mux/src/localpane.rs
```diff
@@ -679,168 +696,190 @@ impl Pane for LocalPane {
...
+    lines: &[&Line],
+    stable_idx: StableRowIndex,
+    uniq_matches: &mut HashMap<String, usize>,
+    coords: &mut Option<Vec<SearchCoord>>,
+    results: &mut Vec<SearchResult>,
+) {
+    if coords.is_none() {
+        coords.replace(make_coords(lines, stable_idx));
+    }
+    let coords = coords.as_ref().unwrap();
```
bew: FIXME: this whole thing could be better written to avoid the unwrap 🤔

mux/src/localpane.rs
```diff
@@ -679,168 +696,190 @@ impl Pane for LocalPane {
...
+
+    if haystack.is_empty() {
+        return true;
+    }
+
+    let haystack = match pattern {
+        CompiledPattern::CaseInSensitiveString(_) => Cow::Owned(haystack.to_lowercase()),
+        _ => haystack,
+    };
+    let mut coords = None;
```
bew: TODO: doc this, I'm not sure what this is for exactly / what it holds

mux/src/localpane.rs
```diff
@@ -679,168 +696,190 @@ impl Pane for LocalPane {
...
-            let (start_x, start_y) = haystack_idx_to_coord(byte_idx, coords);
-            let (end_x, end_y) = haystack_idx_to_coord(byte_idx + text.len(), coords);
-            results.push(SearchResult {
-                start_x,
-                start_y,
-                end_x,
-                end_y,
-                match_id,
+/// Build byte-offset-to-cell mappings for the physical lines in one logical line.
+fn make_coords(lines: &[&Line], stable_row: StableRowIndex) -> Vec<SearchCoord> {
```
bew: FIXME: why do we need to iterate all cells of all lines of the searched range to make those coords?

this fn is called once for each found match, in `found_match` fn 😬

mux/src/localpane.rs
```diff
@@ -679,168 +696,190 @@ impl Pane for LocalPane {
...
+/// Append one match after converting its byte range into terminal coordinates.
+/// Match IDs are assigned by distinct matched text and are shared across lines.
+fn found_match(
+    text: &str,
+    byte_idx: usize,
+    lines: &[&Line],
+    stable_idx: StableRowIndex,
+    uniq_matches: &mut HashMap<String, usize>,
+    coords: &mut Option<Vec<SearchCoord>>,
+    results: &mut Vec<SearchResult>,
```
bew: `uniq_matches`, `coords` & `results` could go in a `ctx: &mut SearchContext` param 🤔
Could hold `limit` as well? or not?

mux/src/localpane.rs
```diff
@@ -679,168 +696,190 @@ impl Pane for LocalPane {
...
+                found_match(
+                    s,
+                    idx,
+                    lines,
+                    stable_idx,
+                    uniq_matches,
+                    &mut coords,
+                    results,
+                );
+                if limit.is_some_and(|limit| results.len() >= limit as usize) {
```
bew: if we put `limit` in `SearchContext` we could move this condition in that context type impl 🤔

PR #8147, opened 2026-09-09
gui: add native SelectAll action

alb3rtg:
Adds an opt-in `wezterm.action.SelectAll` key assignment for selecting all text in the active pane, including scrollback and the visible viewport.

The action operates on the regular terminal selection; it does not activate Copy Mode or Search Mode. This supports a normal macOS `CMD-A` binding without Lua callbacks or overlay state.

Example:
```lua
{ key = "a", mods = "SUPER", action = wezterm.action.SelectAll }
```

Tests:
- `cargo fmt --all -- --check`
- `cargo check -p wezterm-gui`
- `cargo test -p config` (8 passed)

The full documentation-container build was not run because neither Docker nor Podman is installed locally.


docs/config/lua/keyassignment/SelectAll.md
```diff
@@ -0,0 +1,22 @@
...
+config.keys = {
+  {
+    key = 'a',
+    mods = 'SUPER',
+    action = act.SelectAll,
+  },
+}
+```
+
+{{since('nightly')}}
```
bew: move this directive at the start of the page

wezterm-gui/src/termwindow/selection.rs
```diff
@@ -116,6 +116,21 @@ impl super::TermWindow {
         self.window.as_ref().unwrap().invalidate();
     }
 
+    pub fn select_all(&mut self, pane: &Arc<dyn Pane>) {
+        let dims = pane.get_dimensions();
+        let start = SelectionCoordinate::x_y(0, dims.scrollback_top);
+        let end = SelectionCoordinate::x_y(
+            usize::MAX,
+            dims.physical_top + dims.viewport_rows.saturating_sub(1) as StableRowIndex,
```
bew: FIXME: is this the right way to get the last row?

wezterm-gui/src/termwindow/selection.rs
```diff
@@ -116,6 +116,21 @@ impl super::TermWindow {
...
+        );
+        let mut selection = self.selection(pane.pane_id());
+        selection.origin = Some(start);
```
bew: FIXME: do we need to re-set all of those fields? doesn't `self.selection` already does some of that?
(what does that function do really? might need a rename otherwise)

wezterm-gui/src/commands.rs
```diff
@@ -1745,6 +1745,14 @@ pub fn derive_command_from_key_assignment(action: &KeyAssignment) -> Option<Comm
             icon: None,
         },
         DisableDefaultAssignment => return None,
+        SelectAll => CommandDef {
+            brief: "Select all text in the current pane".into(),
+            doc: "Selects all text in the current pane, including scrollback, without entering Copy Mode or Search Mode".into(),
+            keys: vec![],
+            args: &[ArgType::ActivePane],
```
bew: FIXME: what's that?

PR #8152, opened 2026-09-12
Add opt-in Windows UI Automation input support

### LINKED ISSUE #7791: Windows: Win+H voice typing produces no input (no TSF text store)

Windows

### WezTerm version

No — but I confirmed by reading `window/src/os/windows/window.rs` on `master` (HEAD as of filing) that the relevant code path does not exist on either branch (see Describe the bug).

### Describe the bug

Pressing `Win+H` opens the Windows voice-typing flyout, the mic activates, and recognised words appear in the flyout, but nothing is inserted into WezTerm. The same chord works in Notepad, Microsoft Terminal, and the VS Code integrated terminal.

Windows voice typing delivers recognised text via TSF — `ITextStoreACP::InsertTextAtSelection` on the focused window's registered text store. `window/src/os/windows/window.rs` uses IMM32 exclusively (`ImmGetContext`, `WM_IME_COMPOSITION`, etc.); no `ITextStoreACP` / `ITfThreadMgr` / `ITfDocumentMgr` is registered with the thread manager, so Windows has nowhere to deliver the recognised text and drops it silently. The IMM↔TSF compatibility bridge bridges composition messages but does not bridge voice-typing insertions.

This is the Windows analog of #4592, which was resolved on macOS by #7536 (returning a valid `selectedRange` from `NSTextInputClient`). On Windows the structural cost is higher because there is no existing TSF surface to fix — it has to be added.

### To Reproduce

1. Launch WezTerm on Windows 11 with default config.
2. Focus WezTerm; place the cursor at a shell prompt.
3. Press `Win+H`; speak a short phrase ("hello world").
4. Flyout shows the recognised text; nothing arrives in WezTerm.

Cross-check: same steps with Notepad or Microsoft Terminal focused — text is inserted.

### Configuration

no config

### Expected Behavior

Words recognised by Win+H voice typing are delivered to WezTerm as if typed.

### Logs

No relevant logs — the failure is silent at WezTerm's level because no message arrives. Confirmable externally with `Inspect.exe` / `AccEvent.exe`: WezTerm's top-level window exposes no `ITextStoreACP` and is not registered as a TSF focus target.

### Anything else?

Related (context, not duplicates):

- #7536 (merged) — macOS dictation fix, in-repo precedent for the cross-platform pattern.
- #4592, #5180 — macOS-side history.
- #7738, #7529, #7157, #7700, #7721 — recent panics and reentrancy issues in the existing IMM32 path. A TSF stub would live alongside IMM32 behind a config flag (default off), so it does not intersect with this cluster.

Happy to put together a PR. Proposed shape: register an `ITextStoreACP2` with the thread manager that reports selection `(0,0)` / length `0` and forwards `InsertTextAtSelection` through the existing keystroke path. No composition tracking, no ACP↔screen mapping in this first cut — same minimal-disclosure pattern as #7536. Would follow `cargo test --all` / `cargo fmt --all` and update the IME docs per CONTRIBUTING. If you'd rather discuss approach first on Matrix or in a Discussion before I open anything, happy to do that — just let me know.

imoldmao:
On Windows, Typeless can paste into WezTerm yet show its fallback copy card because the focused terminal does not expose an editable UI Automation element. This patch adds an opt-in input bridge at the terminal cursor using AccessKit.

Set `enable_win32_accessibility_input = true` before creating a window. The default remains `false`.

The bridge exposes a writable `MultilineTextInput` with Value/Text patterns and routes submitted text through the active pane's existing `send_paste` path. Requests are checked against the current pane ID and window keyboard focus; a GUI modal removes the input target. Window lifetime tokens prevent stale callbacks from reaching a reused handle, and accessibility notifications are raised after window state borrows have been released.

The element is an immediately consumed pending-input buffer: its value and text range remain empty. It does not represent the screen, scrollback, shell command line, or terminal selection, and SetValue cannot replace text already sent to an application. No Enter is synthesized; multiline text follows the receiving application's normal paste semantics. Full screen-reader support and TSF/Win+H are outside this patch. In particular, this does not claim to fix #7791.

Validation:
- [Windows release build, tests and package](https://github.com/imoldmao/wezterm/actions/runs/34608295373) pass for source `55ec393`: 333 passed, 39 skipped across 72 test binaries, including all three new input model tests. Formatting and documentation checks also pass.
- The portable build was run on Windows. Native probes found a writable Edit element, empty Value/Text document, one degenerate selection, and the system's focused element resolving to `Terminal input`.
- Chinese multiline SetValue input reached an inert receiver exactly once with the existing bracketed-paste delimiters and no extra Enter.
- With Typeless 2.5.0, a captured Chinese dictation trial arrived exactly once. The user confirmed no fallback copy card in an additional dictation test, then also verified normal dictation without the card in a regular terminal window using their existing personal configuration plus the opt-in flag. This is a local compatibility result, not a claim about all dictation applications or all workflows.
- Native pane switching invalidated the old input provider; the command palette removed and then restored the input target; destroying one window invalidated its provider while another window in the same process remained usable. Ctrl+C still reached the test receiver as byte `03` and exited it.
- A background-target .NET ValuePattern call brought its target to the foreground and changed native keyboard focus before the observed insertion. That test is not counted as proof of rejection while a window remains in the background. The dispatch-time keyboard-focus guard remains in place.

Image-paste workflows and a complete screen-reader session were not exercised. The included `ci/windows-accessibility-smoke.ps1` can inspect native patterns and optionally insert into a dedicated test receiver; the configuration reference documents the opt-in behavior and buffer limitations.


window/src/os/windows/accessibility.rs
```diff
@@ -0,0 +1,224 @@
...
+            let caret = TextPosition {
+                node: text_id,
+                character_index: 0,
+            };
+            input.set_text_selection(TextSelection {
+                anchor: caret,
+                focus: caret,
+            });
+
+            // 输入立即交给终端；这里仅表示待提交的空缓冲区，不冒充屏幕文本。
```
bew: this is an international project, so could you review all comments in this PR and translate them to English?

window/src/os/windows/accessibility.rs
```diff
@@ -0,0 +1,224 @@
...
+const ROOT: NodeId = NodeId(0);
+
+#[derive(Default)]
+pub(super) struct InputState {
+    target: Option<(usize, Rect)>,
+}
+
+impl InputState {
+    fn input_id(pane: usize) -> NodeId {
+        NodeId(1 + 2 * pane as u64)
```
bew: how does this _id_ work? why this value?

window/src/os/windows/window.rs
```diff
@@ -100,7 +101,7 @@ lazy_static! {
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Ord, PartialOrd)]
-pub(crate) struct HWindow(HWND);
+pub(crate) struct HWindow(pub(super) HWND);
```
bew: Hmm why make it public / available in the module?

docs/config/lua/config/enable_win32_accessibility_input.md
```diff
@@ -0,0 +1,30 @@
...
+text range are empty because committed text is sent to the terminal through
+the normal paste path. It does not expose the terminal screen, scrollback,
+shell command line, or terminal selection. Setting its value inserts text;
+it cannot replace text previously sent to an application. No Enter key is
+added, and the receiving application's normal paste behavior applies.
+
+Input requests are accepted only for the focused window and active pane.
+This is experimental input assistance, not full screen-reader support or
+Windows Text Services Framework support. Windows voice typing (Win+H) is
+not covered by this option.
```
bew: Do you know what would be needed to support Windows voice typing?
Why is Typeless different in its requirement from the apps?

ci/windows-accessibility-smoke.ps1
```diff
@@ -0,0 +1,76 @@
+﻿param(
```
bew: Could you add a top doc comment explaining what this script is for, when it should be used?

config/src/config.rs
```diff
@@ -723,6 +723,9 @@ pub struct Config {
 
     #[dynamic(default = "default_true")]
     pub use_ime: bool,
+    /// 为 Windows 辅助输入工具提供立即提交的终端输入缓冲区。
+    #[dynamic(default)]
+    pub enable_win32_accessibility_input: bool,
```
bew: Do you know of any downside of having it enabled all the time (no config option) ?

docs/config/lua/config/enable_win32_accessibility_input.md
```diff
@@ -0,0 +1,30 @@
+---
+tags:
+  - keys
+---
+# `enable_win32_accessibility_input`
+
+{{since('nightly')}}
+
+Windows only. Defaults to `false`. Set this option before creating a window:
```
bew: why BEFORE ?

I think that ~all options for Wezterm are dynamically reloadable, and should be able to apply on an existing window, are you saying that this one can't? Why is that?

PR #7536, opened 2026-01-29
fix(macos): enable dictation by returning valid selectedRange

planteman:
## Summary
This PR enables macOS dictation/voice input support by returning a valid `NSRange(0, 0)` from `selectedRange` instead of `NSNotFound`.

## Problem
macOS dictation (voice input) doesn't work in WezTerm because the `selectedRange` method in the `NSTextInputClient` protocol implementation returns `NSNotFound`, which tells macOS there is no valid text cursor position.

## Solution
Return `NSRange::new(0, 0)` (cursor at position 0, no selection) instead of `NSNotFound`. This signals to macOS that there is a valid cursor position where dictated text can be inserted.

## References
- This is the same fix that resolved dictation in Emacs: https://xenodium.com/macos-dictation-returns-to-emacs-fix-merged/
- Tested on macOS - dictation now works correctly

Fixes #4592

PR #8160, opened 2026-09-14
fix(macos): return a valid selectedRange to enable dictation

### LINKED ISSUE #4592: Support voice input on macOS

Zeraphim:
## Problem

`selected_range` in the `NSTextInputClient` implementation returns `{NSNotFound, 0}`. macOS reads that as "this view has no insertion point", so voice input silently bails out: the system dictation UI never appears, and third-party dictation apps that probe the focused element for a selected text range find nothing and abort before recording starts.

Fixes #4592, #5180.

## Fix

Return an empty selection at index 0 instead. `NSTextInputClient` expects character indices into a text storage buffer and a terminal emulator doesn't maintain one, but an empty range at 0 satisfies the protocol and is enough for the dictation machinery to treat the view as a valid target. Text still arrives through the existing `insert_text` / `set_marked_text` callbacks, so nothing else changes. kitty and Emacs use the same workaround.

Also adds `NSSpeechRecognitionUsageDescription` to the macOS bundle so the speech recognition permission prompt has a usage string.

Two files, +11/−1. No config surface, no render changes.

## Verification

Tested on macOS 26.5 (Darwin 25.5.0, Apple Silicon) with MacWhisper 14.8.1, which inserts dictated text at the cursor of the focused element.

MacWhisper logs the failure explicitly. Before the patch, on WezTerm `20240203-110809`, every dictation attempt aborts within 2ms:

```
[dictationController] -------------- 🎙️ STARTING DICTATION --------------
[dictationSession]    🗣️ (EB2E) Start was called
[dictationSession]    🗣️ (EB2E) noActiveTextFieldFound
[dictationController] handleNonRecoverableError (id: EB2E). Error: 'No Active Textfield'
```

38 consecutive failures, no recording, no text. The same app dictates into iTerm2 and Terminal without issue on the same machine, so the difference is the terminal, not the dictation app or its permissions.

After the patch, built from `main` @ `699fd77b4`, the same three keypresses produce:

```
[dictationController] -------------- 🎙️ STARTING DICTATION --------------
[dictationSession]    🗣️ (3F43) Model load finished; continuing as .transcribing
[dictationController] OVERLAY: state became .finished, hiding overlay (id: 3F43)
```

3 dictation sessions, 0 `noActiveTextFieldFound`, transcript inserted at the shell prompt. MacWhisper's own history table records `com.github.wez.wezterm` as the target app for the first time.

## Relationship to existing PRs

This is deliberately a subset of #7453, which is already approved and also fixes this bug. #7453 bundles the `selected_range` change with new `compose_fg` / `compose_bg` palette options and render changes in `wezterm-gui`.

This PR carries only the dictation fix so there is a minimal-diff option on the table. **If maintainers prefer #7453, please close this one in its favour** — I'd rather the fix land in either form. #7536 was an earlier standalone attempt, closed as a duplicate of #7453.

I'm aware review throughput is limited right now; no urgency implied. Adding independent third-party-dictation confirmation to the record in case it helps whoever picks this up.


bew:
Let's use #7435 instead, thanks for the ping ;)

Issue #7980, opened 2026-07-26
shell-integration: unquoted base64 substitution word-splits user vars >57 bytes (WEZTERM_PROG truncated + spurious vars, Linux)


soren-achebe:
### What Operating System(s) are you seeing this problem on?

Linux

### Which Wayland compositor or X11 Window manager(s) are you using?

_Not applicable — the bug is in `assets/shell-integration/wezterm.sh` itself and reproduces in any terminal (verified via raw PTY capture)._

### WezTerm version

`wezterm.sh` as of current main (`76b606ec597a3c0263fa60321548637451c0a547`); the affected line is unchanged in the latest release.

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes — the script is identical on main today, so nightly behaves the same.

### Describe the bug

`__wezterm_set_user_var` encodes values with an **unquoted** backtick substitution:

```sh
printf "\033]1337;SetUserVar=%s=%s\007" "$1" `echo -n "$2" | base64`
```

GNU coreutils `base64` (the default on Linux) wraps its output at 76 columns. Because the substitution is unquoted, the shell **word-splits the wrapped lines into separate arguments**, and `printf` then re-uses its format string for the leftovers. So any value whose base64 encoding exceeds 76 chars (i.e. any value longer than **57 bytes**) is emitted as:

1. `SetUserVar=<name>=<first 76 base64 chars>` — the var is silently **truncated**, and
2. garbage sequences `SetUserVar=<line N>=<line N+1>` pairing up the remaining lines — i.e. **spurious user vars whose *names* are base64 fragments**, each firing a `user-var-changed` event.

Practical impact: `WEZTERM_PROG` is truncated to 57 chars for any longer command line (a `cargo test --workspace ...` easily exceeds that), and `WEZTERM_HOST` breaks on long FQDNs. Anyone consuming these from `user-var-changed` or `pane:get_user_vars()` on Linux gets wrong values plus noise events. macOS/BSD `base64` doesn't wrap, which is presumably why this hasn't been noticed.

### To Reproduce

With the function extracted verbatim from `wezterm.sh` (no GUI needed):

```console
$ __wezterm_set_user_var WEZTERM_PROG "echo $(printf 'x%.0s' {1..120})" | cat -v
[...]

PR #8180, opened 2026-09-17
Fixes https://github.com/wezterm/wezterm/issues/7980

jessp01:
Also fixes:

In assets/shell-integration/wezterm.sh line 428:
```
if [[ ! -n "${BLE_VERSION-}" ]]; then
```
^-- SC2236 (style): Use -z instead of ! -n.


assets/shell-integration/wezterm.sh
```diff
@@ -35,7 +35,6 @@ case "$TERM" in
     return 0
   ;;
 esac
-
```
bew, line 38: beware of useless line changes! here and near end of diff

assets/shell-integration/wezterm.sh
```diff
@@ -434,16 +433,17 @@ fi
 # It requires the `base64` utility to be available in the path.
 __wezterm_set_user_var() {
   if hash base64 2>/dev/null ; then
-    if [[ -z "${TMUX-}" ]] ; then
-      printf "\033]1337;SetUserVar=%s=%s\007" "$1" `echo -n "$2" | base64`
+    if [[ -z "${TMUX}" ]] ; then
```
bew: why this change for TMUX var?

assets/shell-integration/wezterm.sh
jessp01: Fixed.

assets/shell-integration/wezterm.sh
jessp01: Because the whole point is to check whether `$TMUX` is unset (using `-z`), and so, there's no need for a default value, in the event it is not:)

taminomara: It's a defense against `set -u` though.

jessp01: Sure, but `assets/shell-integration/wezterm.sh` does not set -u and there's no real issue if `TMUX` is unset (as per my last comment). Still, it's not really worth a debate, I've reinstated it now.

Issue #7959, opened 2026-07-21
Unix-domain `proxy_command` via `wsl.exe` hangs at "Checking server version" when spawned by the GUI


jamesbraza:
### What Operating System(s) are you seeing this problem on?

Windows

### WezTerm version

20240203-110809-5046fc22

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

No, and I'll explain why below

### Describe the bug

I set up a mux domain into WSL2 to bypass https://github.com/microsoft/terminal/issues/9232. Since WSL2 lacks AF_UNIX interop (the [docs' `socket_path` recipe][mux-docs] notes it "only works with WSL 1"), I configured a unix domain that bridges the mux protocol over `wsl.exe` stdio pipes with `proxy_command`. Connecting hangs forever at:

```text
Connect to Proxy(["wsl.exe", "--exec", "/usr/bin/wezterm", "cli",
                  "--no-auto-start", "proxy"])
Connected!
Checking server version
```

Evidence collected while the window hangs:

- The `wezterm-mux-server` inside WSL is healthy: local `wezterm cli --no-auto-start list` (inside WSL) answers normally throughout.
- The Linux-side proxy process (`wezterm cli --no-auto-start proxy`) is alive and connected to the socket during the hang; the version-check PDU exchange simply never completes across the `wsl.exe` stdio bridge.
- The identical PDU round-trip through the identical bridge works when `wsl.exe` is spawned from a console-attached process: running `wsl.exe -- env WEZTERM_UNIX_SOCKET=/run/user/1000/wezterm/sock wezterm cli --no-auto-start list` from PowerShell prints the server's pane list, i.e. requests and responses cross the pipe fine.
- When wezterm-gui spawns the proxy, a spurious **console window appears** (titled `C:\Program Files\WezTerm\wezterm.exe`), suggesting `wsl.exe` allocates/attaches a console when its parent is a GUI-subsystem process, plausibly interfering with the binary stdio streams.
- Reproduced with both `{ 'wsl.exe', 'wezterm', 'cli', '--no-auto-start', 'proxy' }` and the `--exec` form; with the server pre-started (no `serve_command` race); and launched both from Win+R and from a shell.

Workaround that avoids the bug entirely: an `ssh_domains` entry into WSL2 (`127.0.0.1:2222`, `multiplexing = 'WezTerm'`) works, since the GUI then never spawns `wsl.exe` in the data path.

[...]

PR #8182, opened 2026-09-18
fix: hang on ProxyCommand in Windows

gwangyi:
In Windows, HANDLE/SOCKETs from a single source by DuplicateHandle share a single kernel backing object FILE_OBJECT. It's protected by mutex and all ReadFile/WriteFile operations are serialized unless it is marked as non-blocking. That means WriteFile won't finish until simultaneous ReadFile call returns.

Marking it non-blocking won't solve the issue because calling Read/Write ops on an OVERLAPPED handle without OVERLAPPED argument is undefined. Because of these problems, making two pairs of pipes and detached copying loop is a standard way to redirect stdin/out to a socket.

This pull request contains a commit that introduces "as_stdio_pair()" that returns two separated Stdio objects bound to each pipe in Windows with detached copying loops. For non-Windows OSes, each socket has separate queues of read and write, which means no hang on simultaneous I/O on a single socket. Therefore, "as_stdio_pair()" is basically the same as calling "as_stdio()" two times.


sebastianstudniczek:
Curious whether this might fix this issue?
https://github.com/wezterm/wezterm/issues/7959

bew:
> Also, adds emulated socketpair() because
> filedescriptor::socketpair() doesn't seem to work in Windows correctly.

Note that `filedescriptor` is a crate in this repo as well, shouldn't this go in it to fix socketpair at the source?

---

gwangyi:
For the AI usage, I used AI just once to figure out what is the difference between `if cfg() { ... }` and `#[cfg(...)]`, which is very dumb question from the first contribution in Rust.

I'm using wezterm with this fix in daily tasks since several months ago. I haven't got any issue related this. Since it is a change with quite small changing radius, I thought having daily tasks is enough.

My thought about having change in this way instead of putting it in `filedescriptor` crate is that it is related to a fundamental difference about implementing socket between UNIX-like OSes and Windows, so the user of filedescriptor should handle the difference to keep filedescriptor thin enough. It's just a humble opinion from a hobby-rust-coder.

bew:
> My thought about having change in this way instead of putting it in `filedescriptor` crate is that it is related to a fundamental difference about implementing socket between UNIX-like OSes and Windows, so the user of filedescriptor should handle the difference to keep filedescriptor thin enough. It's just a humble opinion from a hobby-rust-coder.

Well, `filedescriptor` IS the crate that should own & manage this OS-specific distinction, consumer crates shouldn't have to care about that, so changing that crate should be the right way to go here.

gwangyi:
I rewrote the commit; detached copy looping is moved to filedescriptor crate.

I think this is the story behind: ReadFile/WriteFile on SOCKET holds a mutex in the kernel object FILE_OBJECT. DuplicateHandle doesn't clone this data structure, so multiple handles from a single SOCKET shares the mutex. Unless the SOCKET is set as non-blocking, all the read/write ops are serialized; write won't be complete until preceding read op is done, which means hang. Marking non-blocking makes child processes call ReadFile/WriteFile without OVERLAPPED data structure on non-blocking socket, which is undefined behavior. We should use 2 pipe pairs for this purpose.

Since this commit doesn't let the child inherit SOCKETs, I rewrote socketpair_impl to more standard way. I couldn't figure out on this issue yet, but it is being stuck as well.

Issue #4826, opened 2024-01-21
Wezterm wgpu VRAM memory leak


takase1121:
### What Operating System(s) are you seeing this problem on?

Linux Wayland

### Which Wayland compositor or X11 Window manager(s) are you using?

Operating System: Arch Linux 
KDE Plasma Version: 5.27.10
KDE Frameworks Version: 5.114.0
Qt Version: 5.15.12
Kernel Version: 6.7.0-zen3-1-zen (64-bit)
Graphics Platform: Wayland
Processors: 12 × AMD Ryzen 5 5500U with Radeon Graphics
Memory: 5.6 GiB of RAM
Graphics Processor: AMD Radeon Graphics
Manufacturer: LENOVO
Product Name: 82LM
System Version: IdeaPad 5 14ALC05

### WezTerm version

20240120-211320-b4c5dc74

### Did you try the latest nightly build to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version of the nightly that I tried

### Describe the bug

Wezterm keeps on expanding VRAM and GTT (shared memory between CPU and GPU) until it runs out of memory.
This happens especially with interactive terminal applications, e.g. htop, nano.
After wezterm is closed, the memory is freed properly.

### To Reproduce

1. Use wgpu as the backend (`config.front_end = "WebGpu"` which is the default)
2. Open htop
3. Wait patiently or scroll really fast to update the screen

### Configuration

```lua
local wezterm = require 'wezterm'
local config = {}

config.color_scheme = 'kanagawabones'
config.hide_tab_bar_if_only_one_tab = true
config.use_ime = true
config.visual_bell = {
  fade_in_duration_ms = 75,
  fade_out_duration_ms = 75,
  target = 'CursorColor',
}
config.warn_about_missing_glyphs = false
config.default_cursor_style = 'SteadyBar'
config.front_end = "WebGpu"

return config
```

### Expected Behavior

To have it not explode on my face and OOM everything.

### Logs

```
warning: queue 0x5761ee5c4380 destroyed while proxies still attached:
  zwp_primary_selection_offer_v1@4278190086 still attached
  wl_data_offer@4278190085 still attached
  wl_data_offer@4278190084 still attached
  wl_buffer@47 still attached
  wl_buffer@48 still attached
  wl_shm_pool@50 still attached
  wl_buffer@52 still attached
  wl_shm_pool@54 still attached
  wl_data_offer@4278190081 still attached
[...]

PR #8184, opened 2026-09-18
webgpu: poll the device after present to fix a leak on vulkan

Zakkaus:
wgpu only frees dropped resources when the device is polled, and we never
polled it, so on the vulkan backend (nvidia) the per-frame buffers and bind
groups pile up until wezterm-gui hits ~10GB and gets OOM killed. OpenGL is
not affected.

With this it stays flat at ~440MB.

refs: #4826


Zakkaus:
For a long time ,NVIDIA has been using 10GB of RAM, causing an OOM. 

I tested it locally for a today and it was stable with no OOM.

wezterm-gui/src/termwindow/render/draw.rs
```diff
@@ -146,6 +146,9 @@ impl crate::TermWindow {
         webgpu.queue.submit(std::iter::once(encoder.finish()));
         output.present();
 
+        // let wgpu free the per-frame buffers and bind groups we just dropped
```
bew, line 149: Let's explain a bit more about what this is about 🤔
```suggestion
        // Reclaim wgpu resources dropped last frame: wgpu only frees them once
        // it confirms (via fence) the GPU submission using them has completed,
        // and that check only runs inside poll().
        // Without polling, freed buffers/bind groups pile up forever on fence-gated
        // backends (Vulkan), leaking memory until OOM.
        // Poll is non-blocking, so this costs nothing. #4826
```

bew (commented):
Thank you for this, #4826 was a pretty old issue!

I've asked people there to try it out, to verify it's not a "if fixes it for me only" type of fix, and is actually correct in practice.

---

Zakkaus:
> Thank you for this, #4826 was a pretty old issue!
> 
> I've asked people there to try it out, to verify it's not a "if fixes it for me only" type of fix, and is actually correct in practice.
> 
> ---
> For transparency, how much AI was used for this contribution? What for? (research/implementation/testing/PRwriting)
> And what / how much manual review/iteration/testing did you do? 🤔


Sorry, my earlier comparison was flawed, please hold off on testing. I'm testing it by hand now to find the real cause, it may take a while. Without a fix wezterm is basically unusable on my desktop.

